PR #44, opened 2018-06-06
Expand to us-east-1, use latest AMIs

errordeveloper:


errordeveloper:
@bboreham I've made it simpler for now and will open WIP PR for param validation.

PR #48, opened 2018-06-06
WIP: Add config parameter validation

### LINKED ISSUE #28: review parameter validation
Right now it's a bit ad-hoc, especially ASG size, but also instance types and regions are not validated at all.

https://github.com/weaveworks/eksctl/blob/0a39d1a2c1d107e1da90c27889d43bddba42ab3a/cmd/eksctl/create.go#L70

errordeveloper:
Closes #28.

### DIFF
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -29,15 +29,6 @@ func createCmd() *cobra.Command {
 	return cmd
 }
 
-const (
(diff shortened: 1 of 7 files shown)


errordeveloper:
I'd like to consider using https://github.com/octago/sflags or something similar instead of the hand-rolled version I have here.

Issue #57, opened 2018-06-09
Unable to create cluster using profile


richardcase:
I'm trying to create a cluster with using a specific profile. I've set the **AWS_PROFILE** environment variable (as per [this](https://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html)) but its still creating the cluster using the default cli profile.

We use specific profiles to distinguish between our environments (dev, non-prod, prod) and we're also using MFA.

Looking at #56 it would appear that using profiles can work?

richardcase:
@errordeveloper - thanks, i haven't had time to try but will try today and let you know. I'm sure its a case of PEBKAC.

errordeveloper:
@richardcase see https://github.com/weaveworks/eksctl/issues/56#issuecomment-396064846, and please let us know if that would be helpful.

richardcase:
Yes that would be useful.

I thought i'd check the AWS cli with the AWS_PROFILE environment variable set and it correctly uses the profile and i get the output from the correct role/account.

But if i then use eksctl with the same environment variables set i now get errors:

```
keksctl\2018-06-11T08:01:38+01:00 [✖]  checking AWS STS access – cannot get role ARN for current session: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, XXX-dev.
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get http://169.254.169.254/latest/meta-data/iam/security-credentials: dial tcp 169.254.169.254:80: connect: host is down
```

I'm guessing that the AWS cli does something extra when running or some subtle differneces between the Python SDK and Go SDK. 

I will keep on looking,

richardcase:
@errordeveloper - i managed to get a cluster created. I had to use the following first:
```bash
aws sts assume-role ....
```
And then use the output to set the AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN. After which i could create the cluster using `eksctl create cluster`.

If i use the[ aws cli](https://github.com/aws/aws-cli) i only need to set AWS_PROFILE (or --profile) and not the other environment variables. I guess the CLI is doing something extra (doing the assume role). Is this something eksctl could also do?

vkhatri:
I had to create a default profile (~/.aws/credentials) to get `eksctl` working. 

[...]

Issue #29, opened 2018-06-05
kubeconfig writer should merge with existing files and respect KUBECONFIG


errordeveloper:
https://github.com/weaveworks/eksctl/blob/0a39d1a2c1d107e1da90c27889d43bddba42ab3a/cmd/eksctl/create.go#L125

richardcase:
@errordeveloper - i can look at this. If thats ok?

richardcase:
Are you thinking something  along these lines:
- If --kubeconfig, --auto-kubeconfig or KUBECONF env var are NOT set then the default kube config location should be use (~/.kube/conf).
- If KUBECONF is set then we use this (assuming --kubeconfig and --auto-kubeconfig aren't specified)
- if KUBECONF is set BUT --kubeconfig or --auto-kubeconfig are specified then KUBECONF is ignored
- if a kubeconfig already exists then we merge in the cluster details with the existing file and create  the relevant cluster/context/user entries.

richardcase:
It would also be worth re-visiting the **write-kubeconfig** and **auto-kubeconfig** flags. 

I wonder if auto-kubeconfig should be set to false as a default so that if no flags are supplied it picks up the standard kubeconf location.

errordeveloper:
@richardcase I agree, we should make `--auto-kubeconfig=flase` the default, it's a bug that it's the opposite right now.

errordeveloper:
@richardcase I've opened #63 to fix the default `--auto-kubeconfig`.

errordeveloper:
> If --kubeconfig, --auto-kubeconfig or KUBECONF env var are NOT set then the default kube config location should be use (~/.kube/conf).

Yes, indeed. (nit: `~/.kube/config`).

> If KUBECONF is set then we use this (assuming --kubeconfig and --auto-kubeconfig aren't specified)

Yes.

> if KUBECONF is set BUT --kubeconfig or --auto-kubeconfig are specified then KUBECONF is ignored

Yes.

> if a kubeconfig already exists then we merge in the cluster details with the existing file and create the relevant cluster/context/user entries.

Yes, and I think the user should be able to control weather `current-context` gets be updated.

richardcase:
I've got pretty much got all of the above working. 

Good idea with the `current-context`.  Do you think if there is no existing config to set the context authomatically but if there is an existing config then you need to specify something like `--set-context`?

errordeveloper:
> I've got pretty much got all of the above working.

[...]

PR #59, opened 2018-06-12
AWS Profile Changes

richardcase:
Changes to ensure that AWS profiles are supported. This involved making
sure that the AWS config file is loaded (SharedConfigEnabled) and
also making sure we have a TokenProvider set.

Added an explicit --profile flag that can be used to explicity specify
which AWS profile you would like to use. This will override any profile
that you have specified via AWS_PROFILE.

If endpoints are being overriden then the credentials from the initial
session creation are shared with any subsequent session creation to
ensure that the tokens are shared (otherwise you may get multiple MFA
prompts).

Issue #57

### DIFF
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -64,6 +64,7 @@ func createClusterCmd() *cobra.Command {
 
 	fs.StringVarP(&cfg.ClusterName, "cluster-name", "n", "", fmt.Sprintf("EKS cluster name (generated if unspecified, e.g. %q)", exampleClusterName))
 	fs.StringVarP(&cfg.Region, "region", "r", DEFAULT_EKS_REGION, "AWS region")
+	fs.StringVarP(&cfg.Profile, "profile", "p", "", "AWS Profile to use. This overrides the AWS_PROFILE environment variable if its set")
 
 	fs.StringVarP(&cfg.NodeType, "node-type", "t", DEFAULT_NODE_TYPE, "node instance type")
 	fs.IntVarP(&cfg.Nodes, "nodes", "N", DEFAULT_NODE_COUNT, "total number of nodes (for a static ASG)")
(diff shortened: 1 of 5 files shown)


cmd/eksctl/create.go
```diff
@@ -64,6 +64,7 @@ func createClusterCmd() *cobra.Command {
 
 	fs.StringVarP(&cfg.ClusterName, "cluster-name", "n", "", fmt.Sprintf("EKS cluster name (generated if unspecified, e.g. %q)", exampleClusterName))
 	fs.StringVarP(&cfg.Region, "region", "r", DEFAULT_EKS_REGION, "AWS region")
+	fs.StringVarP(&cfg.Profile, "profile", "p", "", "AWS Profile to use. This overrides the AWS_PROFILE environment variable if its set")
```
marccarre: Minor/Grammar: `if its set`. What about something like the below?
`AWS profile to use. If provided, this overrides the AWS_PROFILE environment variable`
N.B.: this comment applies to `delete.go`, `get.go` and `utils.go` as well.

marccarre:
N.B.:
1. I tried setting a valid profile with `AWS_PROFILE` and it worked as expected.
2. I tried setting a valid profile via `--profile` and it worked as expected.
3. I tried setting an invalid `AWS_PROFILE` and a valid `--profile` and it did override the env. var. and worked as expected.
4. I tried setting an invalid `AWS_PROFILE` and it failed as expected.
5. I tried setting an invalid `--profile` and it failed as expected.
6. I tried not to set `AWS_PROFILE` nor `--profile` and it failed as expected.

cmd/eksctl/create.go
richardcase: Yes i think that would be better. I'll make the change now.

richardcase:
@marccarre - small change with the text you suggested for the description of the --profile flag.

cmd/eksctl/create.go
marccarre: Thank you so much! :+1:

errordeveloper:
@richardcase thanks a lot!

richardcase:
@marccarre @errordeveloper - i think i've run into an issue with this with 'create cluster' and specifically `CreateDefaultNodeGroupAuthConfigMap'.  This has come to light as part of my testing for #29.  I will let you know what i find.

PR #63, opened 2018-06-15
Make `--auto-kubeconfig=false` the default

errordeveloper:


errordeveloper:
@richardcase feel free to incorporate with other changes we discussed in #29.

PR #66, opened 2018-06-17
Support merging of config files and KUBECONFIG

richardcase:
When writing the kubeconfig for the new EKS cluster then the following
will happen:
- If `--kubeconfig`, `--auto-kubeconfig` or `KUBECONFING` env var are NOT set
  then the default kube config location will be used (~/.kube/conf).
- If `KUBECONFIG` is set then we use this (assuming `--kubeconfig` and
  `--auto-kubeconfig` aren't specified).
- if `KUBECONFIG` is set **BUT** `--kubeconfig` or `--auto-kubeconfig` are
  specified then `KUBECONFIG` will be ignored.
- if a kubeconfig file already exists then we merge in the new EKS
  cluster details with the existing file.
- If `set-context` is true then the `current-context` will be set to that
  of the new EKS cluster.
- If `set-context` is true (default) then if we are merging with an
  existing configuration file with a `current-context` set then this won't
  be updated.

Issue #29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,11 @@ gitTag := $(shell git describe --tags --abbrev=0 --always)
 build:
 	go build -ldflags "-X main.gitCommit=$(gitCommit) -X main.builtAt=$(builtAt)" ./cmd/eksctl
 
+.PHONY: test
+test:
+	go test $(shell go list ./... | grep -v "^github.com/weaveworks/eksctl/vendor" | sort -u)
+
+
 .PHONY: update-bindata
 update-bindata:
 	go generate ./pkg/eks
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/weaveworks/eksctl/pkg/eks"
 	"github.com/weaveworks/eksctl/pkg/utils"
+	"github.com/weaveworks/eksctl/pkg/utils/kubeconf"
 )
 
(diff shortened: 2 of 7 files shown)


richardcase:
Errors on further testing, closing.

errordeveloper:
@richardcase did you have any chance to resolve the issues? I'd love to get this in before beta.1 (aiming to cut it before end of the week or early next week), otherwise we might need to leave this until after 0.1.0.

richardcase:
@errordeveloper  The merging of the config file and looking at the KUBECONFIG file all works. But when i was testing it i've discovered an issue with `create cluster`.  Looking into it this is actually a problem related to the changes i made for #57. Its looking like a problem with using the `--profile` flag and the profile not passed to heptio authenticator when we generate a token so that it can then create the config map.

errordeveloper:
@richardcase ah, I see, that sounds like a bug and doesn't surprise me, as authenticator creates a separate client. I previously use `AWS_PROFILE` in all of my tests, and that obviously worked. I think we should re-open and review this, and fix the authenticator bug separately.

richardcase:
@errordeveloper - sounds like a plan. I have a few changes that i need to include in this. I'll get them committed and pushed.

errordeveloper:
Great, thanks!

richardcase:
@errordeveloper - it should be good for testing.

errordeveloper:
@richardcase I've added some more changes to this, and figured it'll be easier to review as one PR – please see #80.

Issue #78, opened 2018-06-22
Support debug logging from AWS SDK


richardcase:
Add another debug level (perhaps v=5) where we enable debug logging of the AWS Go SDK (with request tracing). 

This will help with issue investigation but also with implementing mocking of services for tests.

PR #80, opened 2018-06-22
Extend kubeconfig writer to support global path with merging

errordeveloper:
Incorporates @richardcase's changes from #78.


errordeveloper:
@richardcase so I couldn't stop wondering if [`k8s.io/client-go/tools/clientcmd`](https://godoc.org/k8s.io/client-go/tools/clientcmd) should be able to do the merging for us, but all references to "merge" in the doc seem to refer to merging of configs on load... So I looked at [`github.com/kubicorn/kubicorn/pkg/kubeconfig`](https://github.com/kubicorn/kubicorn/blob/5914f8b1ee8beb76b1a64a28a44203f5b1d4becd/pkg/kubeconfig/kubeconfig.go) and that seems to be very specific to kubicorn and mostly implements whatever is needed to fetch the config via SSH. I also looked at [`github.com/kubernetes/kops/pkg/kubeconfig`](https://github.com/kubernetes/kops/blob/3489ea317e7fe705ccb7ebf55fe18c873fe9754e/pkg/kubeconfig/kubecfg_builder.go) and discovered that it is also very specific to kops, but it uses **`clientcmd.ModifyConfig`** which looks intriguing.

errordeveloper:
I've reviewed the `kubeconfig` package in kops, I think we should be able to use it actually, but it might need some refactoring (it uses `glog` and `fmt`, but that should be easy to fix upstream). It seems like the package does what we want, I'll need to take another read to confirm that 100%. I'm also very happy to merge this to start with.

richardcase:
@errordeveloper - you are right `KubeconfigBuilder` from kops would work. I like the fact that it uses [ModifyConfig](https://github.com/kubernetes/client-go/blob/484f27892430b961df38fe6715cc396409207d9f/tools/clientcmd/config.go#L158) to do the read of existing and saving of the modified config. As you say very intriguing.

We could refactor to use it directly or change our code to do something similar and call ModifyConfig. I'm happy to do either of these 2 if you want.

richardcase:
As you pointed out kops implementation use `glog` and `fmt'` and it doesn't support `ExecConfig` that we require for Heptio Authenticator (unless i've missed this).

I wonder if we should refactor our implementation to use a similar approach to kops and take advantage of [clientcmd.NewDefaultPathOptions](https://github.com/kubernetes/client-go/blob/484f27892430b961df38fe6715cc396409207d9f/tools/clientcmd/config.go#L137),  [clientcmd.ModifyConfig](https://github.com/kubernetes/client-go/blob/484f27892430b961df38fe6715cc396409207d9f/tools/clientcmd/config.go#L158) and also [clientcmd.GetStartingConfig](https://github.com/kubernetes/client-go/blob/484f27892430b961df38fe6715cc396409207d9f/tools/clientcmd/config.go#L84).

I also like that they do a delete of the cluster, context and authinfos. Our delete logic is a bit crude at the moment (i.e. only delete an autogenerated config).

What do you think?

errordeveloper:
So I've missed `ExecConfig` part. I think refactoring to make use of `clientcmd.NewDefaultPathOptions`, `clientcmd.ModifyConfig` and `clientcmd.GetStartingConfig`, as you said, is a good idea. Perhaps we can get that in, and open an issue to coverage with kops later.

richardcase:
I will refactor based on the above against the `issue-29-merge-kubeconfig` branch.

richardcase:
As discussed I've updated our implementation to use `clientcmd.NewDefaultPathOptions`, `clientcmd.ModifyConfig` and  `clientcmd.GetStartingConfig`.

I will raise an issue with kops to support `ExecConfig` and link it to an issue here to track and change our implementation.

errordeveloper:
@richardcase I'll test this tomorrow and merge

pkg/utils/kubeconfig/kubeconfig_test.go
```diff
@@ -23,7 +23,7 @@ func TestCreateNewKubeConfig(t *testing.T) {
 			"test-context": {AuthInfo: "test-user", Cluster: "test-cluster", Namespace: "test-ns"}},
 	}
 
-	err := kubeconfig.WriteToFile(configFile.Name(), &testConfig, false)
+	err := kubeconfig.WriteKubeCfg(configFile.Name(), &testConfig, false)
```
errordeveloper: Actually, I think we should just call this`kubeconfig.Write` – wdyt?

richardcase: Yes i think you are right. That would be simpler.  I'll make the change.

errordeveloper:
I've spotted a couple of issues, turns out `NewClientSetWithEmbeddedToken` expects `current-context` to be set, also `utils.CheckAllCommands` will break... I'm testing a fix a the moment, will push once I confirm it works.

errordeveloper:
@richardcase please take a look

@jstrachan FYI

cmd/eksctl/create.go
```diff
@@ -79,7 +79,7 @@ func createClusterCmd() *cobra.Command {
 
 	fs.BoolVar(&writeKubeconfig, "write-kubeconfig", true, "toggle writing of kubeconfig")
 	fs.BoolVar(&autoKubeconfigPath, "auto-kubeconfig", false, fmt.Sprintf("save kubconfig file by cluster name, e.g. %q", kubeconfig.AutoPath(exampleClusterName)))
-	fs.StringVar(&kubeconfigPath, "kubeconfig", "", "path to write kubeconfig (incompatible with --auto-kubeconfig)")
+	fs.StringVar(&kubeconfigPath, "kubeconfig", kubeconfig.DefaultPath, "path to write kubeconfig (incompatible with --auto-kubeconfig)")
```
richardcase: Its good to restore the DefaultPath. I was undecided about this.

pkg/utils/kubeconfig/kubeconfig.go
```diff
@@ -15,36 +15,38 @@ import (
...
 )
 
+var DefaultPath = clientcmd.RecommendedHomeFile
+
 // Write will write Kubernetes client configuration to a file.
 // If path isn't specified then the path will be determined by client-go.
 // If file pointed to by path doesn't exist it will be created.
 // If the file already exists then the configuration will be merged with the existing file.
-func Write(path string, newConfig *api.Config, setContext bool) error {
+func Write(path string, newConfig *api.Config, setContext bool) (string, error) {
```
richardcase: I like the addition of returning the path to the config file.

errordeveloper:
I spotted a bug since last review, here is how I fixed it:
```diff
diff --git a/cmd/eksctl/create.go b/cmd/eksctl/create.go
index b6557a4..04d4e93 100644
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -144,12 +144,12 @@ func doCreateCluster(cfg *eks.ClusterConfig, name string) error {

                if writeKubeconfig {
                        config := clientConfigBase.WithExecHeptioAuthenticator()
-                       filename, err := kubeconfig.Write(kubeconfigPath, config.Client, setContext)
+                       kubeconfigPath, err = kubeconfig.Write(kubeconfigPath, config.Client, setContext)
                        if err != nil {
                                return errors.Wrap(err, "writing kubeconfig")
                        }

-                       logger.Success("wrote %q", filename)
+                       logger.Success("wrote %q", kubeconfigPath)
                } else {
                        kubeconfigPath = ""
                }
```
Essentially, without this we wouldn't see effective path of the `kubeconfig` file when it was set by `KUBECONFIG` env var.

PR #83, opened 2018-06-25
Make --name argument parsing stricter

errordeveloper:
Hey @christopherhein, I managed to incorporate some of my improvements on top of #81.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -14,7 +14,20 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-func ClusterName() string {
+// ClusterName generates a neme string when a and b are empty strings.
+// If either a or b are non-empty, it returns whichever is non-empty.
+// If neither a nor b are empty, it returns empty name, to indicate
+// ambigous usage.
+func ClusterName(a, b string) string {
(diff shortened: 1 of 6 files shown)


errordeveloper:
I've overlooked what `fs.MarkDeprecated` does, and it's really unnecessary anyway, anyone using `--cluster-name` will just see an error and  cobra prints usage info right away, so no need to make it all redundant for little benefit. I removed that and re-tested – all works nicely now. Please review once again so we get green tick and I'll merge. Perhaps we should cut another alpha after this.

errordeveloper:
@christopherhein would you mind to review it once again? Thanks!

pkg/utils/utils.go
```diff
@@ -14,7 +14,16 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-func ClusterName() string {
+func ClusterName(a, b string) string {
```
stefanprodan: @errordeveloper could you add a comment describing the logic in here, why you return string empty when both values are not nil?

errordeveloper: 👍

errordeveloper:
@stefanprodan I've documented the behaviour of this function – PTAL.

Issue #23, opened 2018-06-05
polling for stack and cluster status should timeout


errordeveloper:
It can stall forever right now, at least in theory.

https://github.com/weaveworks/eksctl/blob/0a39d1a2c1d107e1da90c27889d43bddba42ab3a/pkg/eks/cfn.go#L56

https://github.com/weaveworks/eksctl/blob/0a39d1a2c1d107e1da90c27889d43bddba42ab3a/pkg/eks/eks.go#L76

PR #84, opened 2018-06-27
Add AWS API call timeouts

archisgore:
This commit fixes issue: https://github.com/weaveworks/eksctl/issues/23

By accepting a timeout parameter, in number of seconds, it allows
the polling operations to exit after a default of five minutes (300 seconds),
but one that is modifiable.

This is immensely useful in automation and scripted scenarios.

Also ran 'go fmt' which may have caused some whitespace changes in
other files. Will undo them if they are undesired.


pkg/eks/api.go
```diff
@@ -52,6 +52,8 @@ type ClusterConfig struct {
 	SSHPublicKeyPath string
 	SSHPublicKey     []byte
 
+	AwsOperationTimeoutSeconds int
```
errordeveloper: Please use `AWSOperationTimeoutSeconds`.

archisgore:
Thanks for the prompt review! I renamed the field. Let me know if this looks good.

Issue #75, opened 2018-06-20
Hung on  creating control plane


samirspatel:
```
eksctl create cluster --cluster-name dev --region us-east-1
```

This has been hung on `2018-06-20T14:16:22-05:00 [ℹ]  creating control plane "dev"` for over thirty minutes now. If it does take this long, it would be nice to show some additional output periodically to let me know it's doing something

```
2018-06-20T14:15:01-05:00 [ℹ]  importing SSH public key "/Users/samirpatel/.ssh/id_rsa.pub" as "EKS-dev"
2018-06-20T14:15:01-05:00 [ℹ]  creating EKS cluster "dev" in "us-east-1" region
2018-06-20T14:15:01-05:00 [ℹ]  creating ServiceRole stack "EKS-dev-ServiceRole"
2018-06-20T14:15:01-05:00 [ℹ]  creating VPC stack "EKS-dev-VPC"
2018-06-20T14:15:21-05:00 [✔]  created ServiceRole stack "EKS-dev-ServiceRole"
2018-06-20T14:16:22-05:00 [✔]  created VPC stack "EKS-dev-VPC"
2018-06-20T14:16:22-05:00 [ℹ]  creating control plane "dev"
```

yamaszone:
Seeing the same issue. Hung for more than 45+ min:
```
eksctl create cluster --cluster-name cluster-eks --nodes 2 --kubeconfig ./kubeconfig.cluster-eks.yaml
2018-06-21T06:15:51Z [ℹ]  importing SSH public key "/home/core/.ssh/id_rsa.pub" as "EKS-cluster-eks"
2018-06-21T06:15:51Z [ℹ]  creating EKS cluster "cluster-eks" in "us-west-2" region
2018-06-21T06:15:51Z [ℹ]  creating VPC stack "EKS-cluster-eks-VPC"
2018-06-21T06:15:51Z [ℹ]  creating ServiceRole stack "EKS-cluster-eks-ServiceRole"
2018-06-21T06:16:12Z [✔]  created ServiceRole stack "EKS-cluster-eks-ServiceRole"
```
Environment information:
- CoreOS (Kernel 4.14.19-coreos)

grdnrio:
I have the same issue. Both the VPC and ServiceRole stacks create successfully, but then it hangs on creating the control plane.

`2018-06-21T16:50:19+01:00 [✔]  created VPC stack "EKS-joeg-VPC"
[...]

PR #85, opened 2018-06-27
Ensure taskErr channel is buffered

archisgore:
The Problem:

Specifically under ClusterProvider.CreateCluster, there are 4
tasks that need to complete.

The first two tasks run in parallel. The next 2 run serially,
but are not gated against success of the first two.

This means that if more than one task fails, the cluster creation
process can get stuck attempting to write to the unbuffered
taskErr channel.

The operational fix:

Buffer the channel with 5 slots for errors, ensuring there
are no surprises.

eksctl is not some high-performance app, so having a buffer
is useful.

The systematic fix:

When writing to the channel, provide a default escape clause
that allows a failed write to be noticed, and warned of. This
will prevent silent hangings.


rade:
Hmm. Perhaps `CreateCluster` should return `[]error`. That would make for a cleaner API and avoid the channel sizing problems, though it does mean errors cannot be reported incrementally. The code currently doesn't do the latter; in fact another way to address the problem would be to implement that functionality - reading errors from `errChan` from a separate go-routine and reporting them would avoid the blocking.

archisgore:
Can't return []error because the polling for stack-completion happens in a goroutine.

I think having another go-routine to read errors off the channel as they happen is the other viable option. Let me know if that's what you want. I can make the changes.

rade:
> Can't return []error because the polling for stack-completion happens in a goroutine.

I don't follow. `runCreateTask` could return an error slice. Yes, it would use a channel internally, but it knows the bound, so can ensure it is not blocking.

Mind you, the whole error handling is a little weird - why do task functions take a `chan error` _and_ return `error`?

errordeveloper:
I believe this code can be simplified a lot.

@archisgore so this will fix silence that we are seeing e.g. in #75? If so, I'd be happy to merge this with an outlook that we will simplify this code later on.

archisgore:
Yes, this change will fix that too. And I agree, I'll clean up the code. I've got a lot of PRs coming out. I'm using this as a production scripting tool, so I'll be around for any technical debt payback for sure.

archisgore:
Already began work for a revamp: https://github.com/weaveworks/eksctl/pull/87

rade:
> @rade what do you think?

I am not all happy with a) hard-coded '5', since that requires knowledge of what happens under the covers, and b) the non-blocking sends since these will produce log messages for what is essentially a bug in the code.

The quickest way to address these, which doesn't require a revamping/cleaning-up of the whole error handling, would be to read from `errChan` in a separate go-routine, thus avoiding the blocking.

archisgore:
You got it. :-) No more buffering, and separate go-routine to read the error channel.

pkg/eks/api.go
```diff
@@ -137,10 +137,12 @@ func (c *ClusterProvider) runCreateTask(tasks map[string]func(chan error) error,
 			logger.Debug("task %q started", tn)
 			errs := make(chan error)
 			if err := task(errs); err != nil {
+				logger.Debug("task %q failed to start due to error %#v", tn, err)
```
rade: Do we need these log statements? Won't the errors be logged at the higher level anyway.

archisgore: Not needed. I'll take them out. I just really wanted to know when debugging, the difference between a returned error, vs a channel-read error. Sorry about this.

cmd/eksctl/create.go
```diff
@@ -124,10 +124,15 @@ func doCreateCluster(cfg *eks.ClusterConfig) error {
...
+
 		// create each of the core cloudformation stacks
-		ctl.CreateCluster(taskErr)
+		go ctl.CreateCluster(taskErr)
+
 		// read any errors (it only gets non-nil errors)
 		for err := range taskErr {
+			if err == nil {
+				break
+			}
```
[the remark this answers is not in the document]
archisgore: That was my misunderstanding on reading from channels. Sorry. Will remove the `nil` check.

This routine still returns an `error`. What would you prefer by way of consolidating multiple errors and returning them? Do you want to chain them? Return an array of errors? Return the last one?

archisgore:
All changes you asked for implemented. I only need your preference on what to do with multiple errors scenario, and I'll take care of it.

rade:
> I only need your preference on what to do with multiple errors scenario, and I'll take care of it.

We could use https://github.com/hashicorp/go-multierror, perhaps.

Also, as I mentioned in my review, this code does need some comment as to what is going on, and why.

archisgore:
Yes absolutely. I'll add comments and everything. I just needed to know what you wanted, so that I can document it correctly. I'll use go-multierror as you requested.

rade:
> I'll use go-multierror as you requested.

_suggested_ ;) I haven't used it myself before, so there may be gotchas.

Btw, please run some before/after manual tests for the 'multiple errors' case and post the output here - i.e. it should block in 'before' (i.e. on master) and print out multiple errors in 'after' (i.e. on this branch).

rade:
btw, ...

> This means that if more than one task fails, the cluster creation process can get stuck attempting to write to the unbuffered `taskErr` channel.

Given that the channel is entirely unbuffered, won't the cluster creation get stuck even when just a _single_ task fails?

errordeveloper:
@archisgore looks like `go ctl.CreateCluster(taskErr)` is the indeed the simplest fix for this. I am probably misunderstanding something about channels, but what I do understand very well is that this code is unnecessarily complicated anyway, which is my fault! I'll apply this simple fix along with a few other changes, then cut a release and consider simplifying the code after that (we should be able to use a single nested CloudFormation stack, and thereby reduce the need for most of the Go codes we have around CloudFormation and EKS API). Thanks for your contributions so far, despite getting anything merged, you helped a lot!

PR #86, opened 2018-06-27
Make cluster creation and deletion idempotent, repeatable and resumable from partial state

archisgore:
Depends on: https://github.com/weaveworks/eksctl/pull/84 and https://github.com/weaveworks/eksctl/pull/85

I made some extensive changes to make `eksctl` be able to pick up partially-created or existing resources in AWS whenever they are partially created.

This is intended to allow resumption of a partially executed create run, and it also sets the stage for launching clusters in existing VPCs.


archisgore:
Moving out of WIP status. This is now ready for review. I was able to ctrl+c the script partially, and it successfully picks up and continues. This is crucial once the timeouts kick in, because an EKS cluster can take a lot more than 5 minutes to become ready.

archisgore:
Yes, that makes sense. I'll re-submit once a testing structure/framework/guidelines are posted, with proper tests. I agree that there's likely to be someone else who's wanted to test against AWS SDK stubs.

PR #87, opened 2018-06-27
Error-handling revamp

archisgore:
This'll be a stream of commits. The PR will let you see what I'm doing, and comment. I am not shy about people looking over my shoulder, so you'll see all the dirty code, reverts, etc. Once we all like it, I can rewrite history so it looks good.

Prompted by: https://github.com/weaveworks/eksctl/pull/85


archisgore:
This is out of WIP status. No more buffered channels. No more duplicate returns of errors. No more confusing return-nil-to-indicate-success ad-hoc protocols.

Channels are created and cleaned up next to each other. All calls are non-blocking. All callers must wait on returning error channels for an error or nil (which automatically happens when the channel is closed.)

archisgore:
This is an abandoned idea.

Issue #46, opened 2018-06-06
Testing


errordeveloper:
We could probably add some unit tests, but the critical code path is reliant on AWS APIs and mocking CloudFormation is going to involve a lot of effort due to its complex nature (unless there is something ready-made we can use). Also, we would also need to mock Kubernetes somehow, there is probably prior art, but that could be very specialised. Personally, I feel that integration test is what we need the most.

errordeveloper:
We might want to start on this before 0.1.0, but I wouldn't block 0.1.0 on lack of automated testing, unless others strongly object.

richardcase:
@errordeveloper - happy to look at this if you want?

Raffo:
Happy to help with the review @richardcase 😉

errordeveloper:
Hey folks, please go ahead! Would be good to outline a short plan and
discuss it before anyone spends a lot of time on this.

Raffo:
Sounds good to me. 

@richardcase what's your plan for that?

richardcase:
@Raffo - i will have a look and come up with a plan.  

There are things like [localstack](https://github.com/localstack/localstack) and also the Go SDK has interfaces for the services to help build mocks, [like](https://github.com/aws/aws-sdk-go/blob/master/service/cloudformation/cloudformationiface/interface.go) this for CloudFormation (i haven't checked interfaces for the other dependent services like STS).

Raffo:
I think the approach with interfaces would be good and that's what I've used in the past. We've used a similar approach while implementing testing for external DNS here: https://github.com/kubernetes-incubator/external-dns/blob/master/provider/aws.go#L79

richardcase:
@Raffo @errordeveloper - what about the following initially:

- Add tests to each of the packages (like we've done with the kubeconfig change). The decision here is whether the tests for a package are contained within the package and have access to all the non-exported functions/structs/etc or if instead the tests site in a test package and test only via the exported functions/structs/etc (seeing only what the user of the package would see). I used the later for the kubeconfig tests.
- We can mock the AWS services using the interfaces and inject these mocks where required for the unit tests (this may require some refactoring).
- We create a separate integration test suite that tests against AWS (when we've worked out who can pay/sponsor). I quite like how minikube have done [this](https://github.com/kubernetes/minikube/tree/master/test/integration). We'd have to perform operations with eksctl and then check its done what we expected using the AWS sdk directly (or something else).
- Start with the `get` command
- Add code coverage output

[...]

PR #88, opened 2018-07-01
Use AWS Service interfaces

richardcase:
Changed `providerServices` to use the AWS service interfaces instead
of the service structs directly. This is an initial change to help with
implementing testing in the future.

Issue #46


Raffo:
Sweet! I'm looking forward for the rest of the plan, I can help, just ping me 😃

PR #93, opened 2018-07-04
Optionally enable full access to ECR

### LINKED ISSUE #92: add a CLI option to enable policies on the worker nodes (e.g. for ECR)
we're using eksctl in `jx create cluster eks` and its working well - many thanks!. 

However it'd be even more awesome if we could add a CLI option to be able to enable the `arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPowerUser` policy on the worker nodes; so that nodes are able to push to ECR. Then we can use ECR OOTB in Jenkins X on clusters created via `eksctl`

errordeveloper:
- this is primarily for use with Jenkins X (fix #92)
- also provide foundation to expose similar features for other addons


errordeveloper:
So heres what we get from this: 
```
15:40:48 UTC+0100 | ROLLBACK_IN_PROGRESS | AWS::CloudFormation::Stack | EKS-scrumptious-monster-1530714596-DefaultNodeGroup | The following resource(s) failed to create: [NodeInstanceProfile]. . Rollback requested by user.
15:40:47 UTC+0100 | CREATE_FAILED | AWS::IAM::InstanceProfile | NodeInstanceProfile | Roles has too many elements. The limit is 1.
15:40:46 UTC+0100 | CREATE_COMPLETE | AWS::IAM::Role | NodeInstanceRole |  
15:40:37 UTC+0100 | CREATE_COMPLETE | AWS::IAM::Role | NodeInstanceRoleExtra
```
Will need to rethink how we handle this.

jstrachan:
I wonder if we need a single Role and we make the `ExtraManagedPolicyArns` a parameter and we default to passing in these as values:
```
```
then adding the extra role if `--full-ecr-access` is specified?

pkg/eks/api.go
```diff
@@ -69,8 +70,28 @@ type ClusterConfig struct {
...
+
+type ClusterAddons struct {
+	WithIAM AddonIAM
 }
 
+type AddonIAM struct {
+	AmazonEC2ContainerRegistryPowerUser bool
+}
+
+// func NewClusterConfig() *ClusterConfig {
```
errordeveloper: This is not needed

errordeveloper:
I'll cleanup and squash.

Issue #105, opened 2018-07-09
Delete all resources on error


errordeveloper:
We should delete all resources we create in case of an error (optionally to start with, as they maybe be handy for debugging), and provide a tool a way to delete any partially-created clusters.

michaelbeaumont:
Duplicate of #2172

Issue #91, opened 2018-07-03
detect existing KeyPair to avoid failure on second cluster creation


jstrachan:
I created a cluster and it worked great. I zapped it via the CloudFormation console.

Then the next time I tried to create a cluster I get an error:
```
2018-07-03T12:07:39+01:00 [ℹ]  importing SSH public key "/Users/jstrachan/.ssh/id_jstrachan_testing_rsa.pub" as "EKS-eks1"
2018-07-03T12:07:40+01:00 [✖]  importing SSH public key: InvalidKeyPair.Duplicate: The keypair 'EKS-eks1' already exists.
```
I had to manually work around this via:
```
aws ec2 delete-key-pair --key-name EKS-eks1 --region us-west-2
```
It'd be nice to detect this error message and log as a warning and continue; or better use the `aws` CLI to query if the KeyPair exists for the given name first?

errordeveloper:
> It'd be nice to detect this error message and log as a warning and continue; or better use the aws CLI to query if the KeyPair exists for the given name first?

Something like that... but probably better to add unique name suffix based on checksum, and when such name exists, double check fingerprint, and error if it's absolutely not the same thing, but continue otherwise.

Also, if you want to make use of existing key, you can just pass `--ssh-public-key EKS-eks1`.

https://github.com/weaveworks/eksctl/blob/a71ba696af68b6656fccaa8e49e8f7c15715cee1/pkg/eks/auth.go#L26-L63

errordeveloper:
I think checking the fingerprint should be enough really, will take a stab at it.

errordeveloper:
Arguably https://github.com/weaveworks/eksctl/issues/105 is another way to fix this, but #107 is a plausible start (and needed anyway).

errordeveloper:
Arguably #105 is another way to fix this, but #106 is a plausible start (and needed anyway).

PR #106, opened 2018-07-09
Add fingerprint suffix to EC2 key pair

errordeveloper:
The purpose of this is to:
- ensure that for a given key and a cluster name the key name will always be unique
- avoid having to manually delete the key when cluster creation fails (as discussed in #91) (see also #105)


errordeveloper:
@jstrachan thanks for the review! 🍰

PR #117, opened 2018-07-12
Update authenticator name when checking its existence

### LINKED ISSUE #112: heptio-authenticator-aws binary changed
https://github.com/weaveworks/eksctl/blob/1c76d1b52ff8f6111092b1632d08680aa2a21f02/pkg/utils/kubectl.go#L36

Heptio Authenticator repo has moved to https://github.com/kubernetes-sigs/aws-iam-authenticator, thus returns an error for those who use the new repo

gchaincl:
Also added some cosmetic changes:
Moved the function from kubectl.go to utils.go given that the binary
checking is not related to kubectl
And update the function name from CheckHeptioAuthenticatorAWS to a more
generic one.

Closes #112

### DIFF
--- a/pkg/utils/kubectl.go
+++ b/pkg/utils/kubectl.go
@@ -46,22 +46,12 @@ func CheckKubectlVersion(env []string) error {
 	return nil
 }
 
-func CheckHeptioAuthenticatorAWS() error {
-	path, err := exec.LookPath("heptio-authenticator-aws")
-	if err == nil {
-		logger.Debug("heptio-authenticator-aws: %q", path)
-	} else {
-		return fmt.Errorf("heptio-authenticator-aws not installed")
-	}
-	return nil
-}
-
 func CheckAllCommands(kubeconfigPath string, isContextSet bool, contextName string, env []string) error {
 	if err := CheckKubectlVersion(env); err != nil {
 		return err
 	}
 
-	if err := CheckHeptioAuthenticatorAWS(); err != nil {
+	if err := checkAuthenticator(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -38,3 +40,23 @@ func FileExists(path string) (bool, error) {
...
 }
+
+// CheckHeptioAuthenticatorAWS checks for the authenticator binary existence.
+// DEPRECATED: Use CheckAuthenticatorBinary
+func CheckHeptioAuthenticatorAWS() error {
+	return CheckAuthenticatorBinary()
+}
+
+// CheckAuthenticatorBinary checks for the authenticator binary existence.
+func CheckAuthenticatorBinary() error {
```
errordeveloper: Simply calling it `CheckAuthenticator` (or indeed `checkAuthenticator`) would be fine. It's not a binary we are looking for, we want to see if command is in the path or not, binary maybe present anywhere else despite the fact...

errordeveloper:
Did you think about plug this to  https://github.com/weaveworks/eksctl/blob/master/pkg/eks/auth.go#L189?

pkg/utils/utils.go
```diff
@@ -38,3 +40,23 @@ func FileExists(path string) (bool, error) {
...
+	binaries := []string{"heptio-authenticator-aws", "aws-iam-authenticator"}
+	for _, bin := range binaries {
+		path, err := exec.LookPath(bin)
+		if err == nil {
+			// binary was found
+			logger.Debug("%s: %q", bin, path)
+			return nil
+		}
+	}
+	return fmt.Errorf("aws-iam-authenticator not installed")
```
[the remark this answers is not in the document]
errordeveloper: The thing is, released binaries are still caller 'heptio-authenticator-aws', as there hasn't been a release since repo was moved.

pkg/utils/kubectl.go
```diff
@@ -98,3 +88,17 @@ func CheckAllCommands(kubeconfigPath string, isContextSet bool, contextName stri
...
+	binaries := []string{"heptio-authenticator-aws", "aws-iam-authenticator"}
+	for _, bin := range binaries {
+		path, err := exec.LookPath(bin)
+		if err == nil {
+			// binary was found
+			logger.Debug("%s: %q", bin, path)
+			return nil
+		}
+	}
+	return fmt.Errorf("Neither aws-iam-authenticator nor heptio-authenticator-aws are installed")
```
errordeveloper: Error message should begin with lowercase

gchaincl:
Let's do it in a separate PR

PR #121, opened 2018-07-13
Add some tests

richardcase:
Show the work-in-progress for adding tests to eksctl. Its currently pretty rough.

- Added an initial couple of tests to see if mocking of the AWS services is realistic.

Issue #46


errordeveloper:
@richardcase I would be curious to see how we would test https://github.com/weaveworks/eksctl/commit/3435db8cfd7e75fca4c2811dd906a6d33a465a02, I think it maybe an easier place to start (maybe along with SSH key importer), as these are simple and well isolated APIs. Also we will probably stop using EKS API directly sooner or later in favour of CloudFormation.

errordeveloper:
@richardcase please review, and merge if you are happy with my changes, I think this is all in a good shape now and we can iterate on actual tests in separate PRs.

richardcase:
@errordeveloper - i think it would be fairly easy to create tests around [SelectAvailabilityZones](https://github.com/weaveworks/eksctl/commit/3435db8cfd7e75fca4c2811dd906a6d33a465a02). I will give it a go.

pkg/eks/mocks/mocks.go
```diff
@@ -0,0 +1,8 @@
+package mocks
+
+import _ "github.com/vektra/mockery"
+
+//go:generate mockery -dir=../../../vendor/github.com/aws/aws-sdk-go/service/cloudformation/cloudformationiface -name=CloudFormationAPI -output=./
```
richardcase: I like this. Better than the shell script i was using.

pkg/eks/api_test.go
```diff
@@ -0,0 +1,28 @@
...
+	sts *mocks.STSAPI
+}
+
+func (m MockProvider) CloudFormation() cloudformationiface.CloudFormationAPI { return m.cfn }
+func (m MockProvider) mockCloudFormation() *mocks.CloudFormationAPI {
+	return m.CloudFormation().(*mocks.CloudFormationAPI)
+}
+
+func (m MockProvider) EKS() eksiface.EKSAPI   { return m.eks }
+func (m MockProvider) mockEKS() *mocks.EKSAPI { return m.EKS().(*mocks.EKSAPI) }
```
richardcase: I like this. I had initially used a struct but ripped it out. Having the function that does the type cast will make the test code neater.

PR #125, opened 2018-07-18
Fix two typos

errordeveloper:
@bowlesns thanks for spotting these!


bowlesns:
No problem, thank you for the tool!

Issue #76, opened 2018-06-21
Create EKS cluster using Cloudformation instead of raw API calls


christopherhein:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-cluster.html

PR #126, opened 2018-07-18
Create control plane with cloudformation

nckturner:
- retains old behavior with flag but defaults to new
- copied assets into a new asset dir 1.10.3/2018-07-18
- Fixes https://github.com/weaveworks/eksctl/issues/76


errordeveloper:
This is great, thanks a lot! I was just thinking of looking into it myself. I will review later on.

errordeveloper:
@nckturner actually, I'm not sure we need the flag, because it's a change in implementation details, and users should rely on such a thing, unless you thinks it fundamentally different?

pkg/eks/cfn.go
```diff
@@ -289,6 +290,81 @@ func (c *ClusterProvider) DeleteStackServiceRole() error {
 	return nil
 }
 
+func (c *ClusterProvider) stackNameDefaultControlPlane() string {
+	return "EKS-" + c.cfg.ClusterName + "-DefaultControlPlane"
```
errordeveloper: just `-ControlPlane`, I think...

nckturner:
@errordeveloper yeah I would have left the flag out had it been my project, but I wasn't sure if backwards compatibility was required, so I threw it in for good measure.  Would you like me to leave it in to be removed in the next release or so, or just remove it now?  I'm not sure why someone *wouldn't* want a cfn resource for their cluster, but perhaps there could be scenarios.  It does (should) make deleting the cluster different, as you should do it through the stack.

errordeveloper:
I am with you. Don't think there is a need to keep clusters creation
backwards compatible, but let's make sure deletions are backwards
compatible. We don't need a flag for deletions, we can just attempt
deleting the cluster if is present, otherwise delete the CFN stack. And we
can break this compatibility in the next release for sure.

errordeveloper:
I think the refactoring PR will take a while. We can land this right after 0.1.0, so hopefully before the end of this week! Do you want to try rebasing?

errordeveloper:
And of course, now we have deletion also.

nckturner:
Ah thanks, I was just working on this last night, but you beat me to it!  Sounds good to me!

PR #131, opened 2018-07-20
Disable zone filter, add flag to set zones manually

### LINKED ISSUE #118: UnsupportedAvailabilityZoneException: lack of capacity to support cluster
unable to create cluster control plane: UnsupportedAvailabilityZoneException: Cannot create cluster '****-**' because us-east-1c, the targeted availability zone, does not currently have sufficient capacity to support the cluster. Retry and choose from these availability zones: us-east-1a, us-east-1b, us-east-1d

errordeveloper:
This the only sane way to fix UnsupportedAvailabilityZoneException,
espcially because it turns out that this is somehow different in every
AWS account (at least that what seems to be the case from the docs.

> You may receive an error that one of the Availability Zones in your
> request does not have sufficient capacity to create an Amazon EKS
> cluster. If this happens, the error output contains the Availability
> Zones that can support a new cluster. Retry creating your cluster
> with at least two subnets that are located in the supported
> Availability Zones for your account.

In the future we may try parsing the error message and retrying
automatically, but it doesn't seem feasible right now as we would
need to improve some of the internals and provide stack update
functionality.

Close #118. Finally.

### DIFF
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -42,6 +42,7 @@ var (
 	kubeconfigPath     string
 	autoKubeconfigPath bool
 	setContext         bool
+	availabilityZones  []string
 )
 
 func createClusterCmd() *cobra.Command {
@@ -73,6 +74,8 @@ func createClusterCmd() *cobra.Command {
 	// TODO: https://github.com/weaveworks/eksctl/issues/28
 	fs.IntVarP(&cfg.MinNodes, "nodes-min", "m", 0, "minimum nodes in ASG")
 	fs.IntVarP(&cfg.MaxNodes, "nodes-max", "M", 0, "maximum nodes in ASG")
+
+	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
 
 	fs.StringVar(&cfg.SSHPublicKeyPath, "ssh-public-key", DEFAULT_SSH_PUBLIC_KEY, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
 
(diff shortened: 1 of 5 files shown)


pkg/az/az.go
```diff
@@ -84,8 +84,8 @@ type AvailabilityZoneSelector struct {
 func NewSelectorWithDefaults(ec2api ec2iface.EC2API) *AvailabilityZoneSelector {
 	avoidZones := map[string]bool{
 		// well-known over-populated zones
-		"us-east-1a": true,
-		"us-east-1b": true,
+		// "us-east-1a": true,
```
errordeveloper: I am not quite convinced that we have to take this out completely... WDYT?

richardcase: We shouldn't need this in an ideal world and I suppose it is helpful to avoid known over populated zones.

If we automatically retried when getting `UnsupportedAvailabilityZoneException` then this is not needed.

cmd/eksctl/create.go
```diff
@@ -74,6 +75,8 @@ func createClusterCmd() *cobra.Command {
 	fs.IntVarP(&cfg.MinNodes, "nodes-min", "m", 0, "minimum nodes in ASG")
 	fs.IntVarP(&cfg.MaxNodes, "nodes-max", "M", 0, "maximum nodes in ASG")
 
+	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
```
richardcase: I like the addition of being able to explicitly set the zones. I know this is the approacvh we'd take.

PR #132, opened 2018-07-22
Make CloudFormation easier to compose

### LINKED ISSUE #127: Make CloudFormation easier to compose
At the moment we have a number of issues where compassable CloudFormation would help (#42, #69, #120, #122).

Managing templates as YAML and embedding them in Go using `bindata` is not particularly delightful, but not an issue per-se. What's hard is piecing together Go code and any logic that is implemented with CloudFormation  intrinsic functions, stack parameters and conditions.

Additionally, it would be much more convenient if we have a single stack composed of optional parts (VPC, IAM role, control plane), and separate (nested?) stacks for nodegroups. This would simplify much of our code.

So we could either invent something ourselves, if we wanted to, but I found that [goformation](https://github.com/awslabs/goformation#marshalling-cloudformationsam-described-with-go-structs-into-yamljson) already does what's needed (and I made an improvement (~https://github.com/awslabs/goformation/pull/95~, https://github.com/awslabs/goformation/pull/103) to it that allows to use `Ref`s and other intrinsics). 

The idea would be to compose the stack using Go structs. First of all we would use Go instead of CloudFormation for conditionals, as it's much easier to grasp all the parameters. We would leverage some CloudFormation runtime functionality to the minimum extended needed, but anything that would take more then one nested intrinsic function would be done in Go, unless information is only available inside of CloudFormation runtime (e.g. resource IDs for VPCs, subnets etc).

errordeveloper:
This PR introduces a big change. We remove checked-in CloudFormation templates, in favour of Go code. It fixes #127.

The use-facing changes are:

- addition of `--ssh-access` flag, which if turned off by default, as it turned out we didn't fully enable SSH access and SG changes were required to open port 22 (this was added mostly because it shows nicely how new model allows us modify templates easily without having to have to instances of LC wrapped with a conditional.
- addition of `--max-pods-per-node` flag, which is also a good example of what was hard to expose before
- number of stacks reduced to two
- CloudFormation stacks have no useful parameters and are no longer directly manageable for the user (we never promised this, it was accidental)

Bug fixes:
- `--nodes=0` will no longer hang waiting for actual nodes

# TODOs:
- [x] add logic to wait for nodegroup to be deleted
- [x] merge #126 and rebase
- [x] squash commits
- [ ] review the code and discuss test coverage (help needed)
- [x] more testing (help needed)
- [x] get clarity on https://github.com/awslabs/goformation/pull/103
- [ ] document new packages
- [x] update sample output in the docs
- [x] fix `kubectl exec` and `kubectl logs`
- [x] remove TODOs from code
- [ ] ~review any improvements from https://github.com/awslabs/amazon-eks-ami/pull/16~ (will be incorporated in follow-up PR)


errordeveloper:
We are pretty close to having this ready now, just a few bugs to fix before it's ready for someone to review.

errordeveloper:
The code is pretty much ready now, it's certainly functional.
```
 [0] >> ./eksctl create cluster
2018-08-02T08:24:17+01:00 [ℹ]  setting availability zones to [us-west-2b us-west-2c us-west-2a]
2018-08-02T08:24:17+01:00 [ℹ]  creating EKS cluster "adorable-outfit-1533194656" in "us-west-2" region
2018-08-02T08:24:17+01:00 [ℹ]  creating cluster stack "eksctl-adorable-outfit-1533194656-cluster"
2018-08-02T08:35:58+01:00 [✔]  created cluster stack "eksctl-adorable-outfit-1533194656-cluster"
2018-08-02T08:35:58+01:00 [ℹ]  creating nodegroup stack "eksctl-adorable-outfit-1533194656-nodegroup-0"
2018-08-02T08:39:40+01:00 [✔]  created nodegroup stack "eksctl-adorable-outfit-1533194656-nodegroup-0"
2018-08-02T08:39:40+01:00 [✔]  all EKS cluster "adorable-outfit-1533194656" resources has been created
2018-08-02T08:39:40+01:00 [✔]  saved kubeconfig as "/Users/ilya/.kube/config"
2018-08-02T08:39:42+01:00 [ℹ]  the cluster has 0 nodes
2018-08-02T08:39:42+01:00 [ℹ]  waiting for at least 2 nodes to become ready
2018-08-02T08:40:15+01:00 [ℹ]  the cluster has 2 nodes
2018-08-02T08:40:15+01:00 [ℹ]  node "ip-192-168-130-34.us-west-2.compute.internal" is ready
2018-08-02T08:40:15+01:00 [ℹ]  node "ip-192-168-222-80.us-west-2.compute.internal" is ready
2018-08-02T08:40:17+01:00 [ℹ]  kubectl command should work with "/Users/ilya/.kube/config", try 'kubectl get nodes'
2018-08-02T08:40:17+01:00 [✔]  EKS cluster "adorable-outfit-1533194656" in "us-west-2" region is ready
 [0] >> kubectl get nodes
NAME                                           STATUS    ROLES     AGE       VERSION
ip-192-168-130-34.us-west-2.compute.internal   Ready     <none>    45s       v1.10.3
ip-192-168-222-80.us-west-2.compute.internal   Ready     <none>    52s       v1.10.3
 [0] >> kubectl get pods
No resources found.
 [0] >> kubectl get pods --all-namespaces
NAMESPACE     NAME                       READY     STATUS    RESTARTS   AGE
kube-system   aws-node-j59b8             1/1       Running   1          1m
kube-system   aws-node-xfpdp             1/1       Running   1          56s
kube-system   kube-dns-7cc87d595-7nmpc   3/3       Running   0          7m
kube-system   kube-proxy-rf6jj           1/1       Running   0          56s
kube-system   kube-proxy-txmx7           1/1       Running   0          1m
 [0] >>
```
I haven't tested how it different combinations of flags, but I cannot foresee any issue there (most of the logic hasn't changed).

errordeveloper:
I've cleaned up the history now, it should be reviewable. It'd be great if a few of you could take a look at it. I'm happy to split some packages as separate PRs, if that helps...

cmd/eksctl/create.go
```diff
@@ -203,3 +207,38 @@ func doCreateCluster(cfg *eks.ClusterConfig, name string) error {
 
 	return nil
 }
+
+func runCreateTask(tasks map[string]func(chan error) error, taskErrs chan error) {
```
errordeveloper: We might be able to simplify things and get rid of this now actually...

cmd/eksctl/create.go
```diff
@@ -203,3 +207,38 @@ func doCreateCluster(cfg *eks.ClusterConfig, name string) error {
...
+				return
+			}
+			logger.Debug("task %q returned without errors", tn)
+		}(taskName)
+	}
+	logger.Debug("waiting for %d tasks to complete", len(tasks))
+	wg.Wait()
+}
+
+func createCluster(ctl *eks.ClusterProvider, taskErrs chan error) {
```
errordeveloper: As we only have two tasks, which are serial.

pkg/cfn/builder/api.go
```diff
@@ -0,0 +1,197 @@
...
+func getOutput(stack *cfn.Stack, key string) *string {
+	for _, x := range stack.Outputs {
+		if *x.OutputKey == key {
+			return x.OutputValue
+		}
+	}
+	return nil
+}
+
+func setOutput(obj interface{}, key, value string) error {
```
errordeveloper: This probably needs some comments, reflect can get so hairy...

pkg/cfn/builder/api_test.go
```diff
@@ -0,0 +1,229 @@
...
+				&cfn.Output{
+					OutputKey:   &k,
+					OutputValue: &v,
+				})
+		}(k, v)
+	}
+	return s
+}
+
+var _ = Describe("CloudFormation template builder API", func() {
```
errordeveloper: For now I've written test just for the most complicated and deep parts of this, we can probably add a few broader tests that – not sure yet what those should be.

pkg/cfn/builder/api_test.go
```diff
@@ -0,0 +1,229 @@
...
+			Expect(obj.Resources["NodeGroup"].Properties.Tags[0].PropagateAtLaunch).To(Equal("true"))
+			Expect(obj.Resources["NodeGroup"].Properties.Tags[1].Key).To(Equal(map[string]interface{}{
+				"Fn::Sub": "kubernetes.io/cluster/${ClusterName}",
+			}))
+			Expect(obj.Resources["NodeGroup"].Properties.Tags[1].Value).To(Equal("owned"))
+			Expect(obj.Resources["NodeGroup"].Properties.Tags[1].PropagateAtLaunch).To(Equal("true"))
+		})
+	})
+
+	Describe("UserData", func() {
```
errordeveloper: This mostly tests cloudconfig package, but in the context of how we actually use it.

pkg/cfn/builder/cluster.go
```diff
@@ -0,0 +1,78 @@
...
+		spec: spec,
+	}
+}
+
+func (c *clusterResourceSet) AddAllResources() {
+	c.rs.template.Description = clusterTemplateDescription
+	c.rs.template.Description += clusterTemplateDescriptionDefaultFeatures
+	c.rs.template.Description += templateDescriptionSuffix
+
+	_, globalCIDR, _ := net.ParseCIDR("192.168.0.0/16")
```
errordeveloper: These CIDR are hardcoded, as that's what we had before, except that it was in YAML...

pkg/cfn/builder/cluster.go
```diff
@@ -0,0 +1,78 @@
...
+
+	subnets := map[string]*net.IPNet{}
+	_, subnets[c.spec.AvailabilityZones[0]], _ = net.ParseCIDR("192.168.64.0/18")
+	_, subnets[c.spec.AvailabilityZones[1]], _ = net.ParseCIDR("192.168.128.0/18")
+	_, subnets[c.spec.AvailabilityZones[2]], _ = net.ParseCIDR("192.168.192.0/18")
+
+	c.addResourcesForVPC(globalCIDR, subnets)
+	c.addResourcesForIAM()
+	c.addResourcesForControlPlane("1.10")
```
errordeveloper: Same goes for the version, also there is only one available right now.

pkg/cfn/builder/iam.go
```diff
@@ -0,0 +1,122 @@
+package builder
```
errordeveloper: The idea is to keep same types of resources grouped together, as we might compose IAM-only stack in the future

pkg/cfn/builder/vpc.go
```diff
@@ -0,0 +1,136 @@
+package builder
```
errordeveloper: Here it's the same as with `iam.go` – we want these grouped together.

pkg/cfn/manager/nodegroup.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (c *StackCollection) makeNodeGroupParams(sequence int) map[string]string {
+	return map[string]string{
+		builder.ParamClusterName:      c.spec.ClusterName,
+		builder.ParamClusterStackName: c.spec.ClusterStackName,
+		builder.ParamNodeGroupID:      fmt.Sprintf("%d", sequence),
+	}
+}
+
+func (c *StackCollection) CreateNodeGroup(errs chan error) error {
```
errordeveloper: There is still some duplication between here and `CreateCluster`, but I'd rather address that separately.

pkg/cfn/builder/api.go
```diff
@@ -0,0 +1,197 @@
...
+	return makeRef(name)
+}
+
+func (r *resourceSet) newStringParameter(name, defaultValue string) *gfn.StringIntrinsic {
+	return r.newParameter(name, "String", defaultValue)
+}
+
+func (r *resourceSet) newResource(name string, resource interface{}) *gfn.StringIntrinsic {
+	r.template.Resources[name] = resource
+	e := reflect.ValueOf(resource).Elem()
```
errordeveloper: Needs a comment – it for setting `Name` tag on all resources that have tags. It also appends resource type to the value (which is what CloudFormation cannot do for us, and it seems like the kind of things we everyone does and wants to see).

pkg/cfn/builder/iam.go
```diff
@@ -0,0 +1,122 @@
...
+	})
+
+	n.instanceProfile = n.newResource("NodeInstanceProfile", &gfn.AWSIAMInstanceProfile{
+		Path:  gfn.NewString("/"),
+		Roles: makeSlice(refIR),
+	})
+	n.rs.attachAllowPolicy("PolicyTagDiscovery", refIR, "*", []string{
+		"ec2:DescribeTags",
+	})
+	n.rs.attachAllowPolicy("PolicyStackSignal", refIR,
```
errordeveloper: This looks a little bit complicated, maybe there is a better way to do this, I am not sure right now.

pkg/cfn/builder/nodegroup.go
```diff
@@ -0,0 +1,144 @@
...
+		InstanceType: gfn.NewString(n.spec.NodeType),
+		UserData:     n.userData,
+	}
+	if n.spec.NodeSSH {
+		lc.KeyName = gfn.NewString(n.spec.SSHPublicKeyName)
+	}
+	refLC := n.newResource("NodeLaunchConfig", lc)
+	// currently goformation type system doesn't allow specifying `VPCZoneIdentifier: { "Fn::ImportValue": ... }`,
+	// and tags don't have `PropagateAtLaunch` field, so we have a custom method here until this gets resolved
+	n.newResource("NodeGroup", &awsCloudFormationResource{
```
errordeveloper: Again, a special case – see comment above.

pkg/cloudconfig/cloudconfig.go
```diff
@@ -0,0 +1,132 @@
+package cloudconfig
```
errordeveloper: This one is quite simple, I've tried one from Juju project and it was too complicated and had messy deps.

pkg/nodebootstrap/userdata.go
```diff
@@ -0,0 +1,112 @@
+package nodebootstrap
```
errordeveloper: This one should be quite self-explanatory...

pkg/cfn/manager/waiters.go
```diff
@@ -0,0 +1,182 @@
...
+	cfn "github.com/aws/aws-sdk-go/service/cloudformation"
+
+	"github.com/kubicorn/kubicorn/pkg/logger"
+
+	"github.com/weaveworks/eksctl/pkg/cfn/builder"
+)
+
+// cloudformation.WaitUntilStackCreateComplete doesn't detect in-progress status early enough,
+// so this is custom version that is more suitable for our use, as there is no way to add any
+// custom acceptors
```
errordeveloper: Arguably, this new code is probably as many lines (or more), however it's really not as complicated as what we've had, we mostly define structs here, and call the AWS SDK to do polling (only if we didn't have to poll...).

Fun fact: CloudFormation API is actually XML... why cannot we write templates in XML with XSLT and all that? 🤔 😉

pkg/cfn/manager/waiters.go
```diff
@@ -0,0 +1,182 @@
...
+	logger.Debug("done %s", msg)
+
+	return nil
+}
+
+func (c *StackCollection) doWaitUntilStackIsCreated(name string) error {
+	return c.waitWithAcceptors(name,
+		makeAcceptors(
+			cfn.StackStatusCreateComplete,
+			[]string{
```
errordeveloper: It unfortunate we have to list every stack status we want to error on, but I'd rather keep this code simple and not try to make it do anything too clever.

pkg/cfn/builder/iam.go
richardcase: I like this

pkg/cfn/builder/cluster.go
richardcase: Do you think this will always be hardcoded?

pkg/cfn/builder/cluster.go
```diff
@@ -0,0 +1,80 @@
...
+	c.rs.template.Description += templateDescriptionSuffix
+
+	_, globalCIDR, _ := net.ParseCIDR("192.168.0.0/16")
+
+	subnets := map[string]*net.IPNet{}
+	_, subnets[c.spec.AvailabilityZones[0]], _ = net.ParseCIDR("192.168.64.0/18")
+	_, subnets[c.spec.AvailabilityZones[1]], _ = net.ParseCIDR("192.168.128.0/18")
+	_, subnets[c.spec.AvailabilityZones[2]], _ = net.ParseCIDR("192.168.192.0/18")
+
+	c.addResourcesForVPC(globalCIDR, subnets)
```
richardcase: Should `subnets` and `globalCIDR` live inside `addResourcesForVPC`?

pkg/cfn/manager/tasks.go
```diff
@@ -0,0 +1,54 @@
...
+		}(t)
+	}
+	logger.Debug("waiting for %d tasks to complete", len(tasks))
+	wg.Wait()
+}
+
+// CreateClusterWithInitialNodeGroup runs two tasks to create
+// the stacks for use with CLI; any errors will be returned
+// as a slice on completion of one of the two taks
+func (s *StackCollection) CreateClusterWithInitialNodeGroup() []error {
```
richardcase: Would be good to have this visualised / documented in the wiki for when someone new to eksctl comes along. They'd know that that there are 2 stacks and the resources in each are x, y, z.

cmd/eksctl/create.go
```diff
@@ -75,25 +76,27 @@ func createClusterCmd() *cobra.Command {
...
 	fs.StringVar(&kubeconfigPath, "kubeconfig", kubeconfig.DefaultPath, "path to write kubeconfig (incompatible with --auto-kubeconfig)")
 	fs.BoolVar(&setContext, "set-kubeconfig-context", true, "if true then current-context will be set in kubeconfig; if a context is already set then it will be overwritten")
 
-	fs.DurationVar(&cfg.WaitTimeout, "aws-api-timeout", eks.DefaultWaitTimeout, "")
+	fs.DurationVar(&cfg.WaitTimeout, "aws-api-timeout", api.DefaultWaitTimeout, "")
 	fs.MarkHidden("aws-api-timeout") // TODO deprecate in 0.2.0
-	fs.DurationVar(&cfg.WaitTimeout, "timeout", eks.DefaultWaitTimeout, "max wait time in any polling operations")
+	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
```
richardcase: Do we need `dry-run` functionality?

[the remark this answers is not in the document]
errordeveloper: I think that flag would be mostly useful for debugging, or if you really have to hand-craft some feature that `eksctl` doesn't support yet. The main issue is that this code (by design) avoided use of parameters in CloudFormation as much as possible (e.g. subnets, node AMIs, SSH key name), and it's not unlikely even remaining parameters might go away, if there is a good reason for that. Any human readable template should have parameters, otherwise the human using it will try to add it in, and that would be a sad thing.

cmd/eksctl/utils.go
```diff
@@ -110,7 +114,7 @@ func writeKubeconfigCmd() *cobra.Command {
 	return cmd
 }
 
-func doWriteKubeconfigCmd(cfg *eks.ClusterConfig, name string) error {
+func doWriteKubeconfigCmd(cfg *api.ClusterConfig, name string) error {
```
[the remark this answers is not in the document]
errordeveloper: Yes, I'm happy with it too. But I was more of necessity, it was a circular dependency.

cmd/eksctl/create.go
errordeveloper: Not sure what it would do, CloudFormation API doesn't have this when you want to create a stack, but it there is something for updating one... Maybe we could discuss it separately.

cmd/eksctl/utils.go
```diff
@@ -55,12 +57,14 @@ func waitNodesCmd() *cobra.Command {
 
 	fs.StringVar(&utilsKubeconfigInputPath, "kubeconfig", "kubeconfig", "path to read kubeconfig")
 	fs.IntVarP(&cfg.MinNodes, "nodes-min", "m", DEFAULT_NODE_COUNT, "minimum number of nodes to wait for")
-	fs.DurationVar(&cfg.WaitTimeout, "timeout", eks.DefaultWaitTimeout, "how long to wait")
+	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "how long to wait")
```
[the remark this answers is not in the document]
errordeveloper: It's actually for all the wait loops, CloudFormation ones, and node join... I agree the description is unhelpful.

pkg/cfn/builder/api.go
errordeveloper: I was tempted to put all set/get output functions in `outputs.go`, but otherwise these are very tied in to what we need here.

pkg/cfn/builder/cluster.go
errordeveloper: No, not at all. These will ultimately get set from a config file (when we have that). I'm less keen on having flags for each of these. They are here for now, I just didn't want to do anything clever about these as part of this PR.

pkg/cfn/builder/cluster.go
errordeveloper: These ultimately come from above, TBD where exactly, but in the long term – Cluster API config file is the most likely source for these.

pkg/cfn/manager/tasks.go
errordeveloper: Yes, indeed. I also think there are quite a few things that need to documented very specifically, i.e. why this is being done (even if it's just a pointer to the discussion in one the issues).

cmd/eksctl/utils.go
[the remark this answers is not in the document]
richardcase: I agree i can't imagine why someone would set it to anything different. But yes an indication to that effect would be good.

cmd/eksctl/create.go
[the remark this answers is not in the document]
richardcase: I was thinking for the scenario where someone would want to use `eksctl` to generate the templates and use that template in a wider infrastructure provisioning script. But if this was the case then they could just exec eksctl directky (like we do with kops & terraform).

Perhaps for debugging we could have a hidden flag or even a log level that dumps the template (if thats not already done)?

pkg/cfn/builder/api.go
```diff
@@ -0,0 +1,210 @@
...
+
+func setOutput(obj interface{}, key, value string) error {
+	e := reflect.ValueOf(obj).Elem()
+	if e.Kind() == reflect.Struct {
+		f := e.FieldByName(key)
+		if f.IsValid() && f.CanSet() {
+			switch f.Kind() {
+			case reflect.String:
+				f.SetString(value)
+			case reflect.Slice:
```
errordeveloper: It feels like there are few too many switches. Perhaps I should think about how to make it more readable also.

pkg/cfn/manager/nodegroup.go
```diff
@@ -0,0 +1,46 @@
...
+		builder.ParamClusterStackName: c.spec.ClusterStackName,
+		builder.ParamNodeGroupID:      fmt.Sprintf("%d", sequence),
+	}
+}
+
+func (c *StackCollection) CreateInitialNodeGroup(errs chan error) error {
+	return c.CreateNodeGroup(0, errs)
+}
+
+func (c *StackCollection) CreateNodeGroup(seq int, errs chan error) error {
```
richardcase: Currently `0` is harded to be passed in as seq. What are the plans with this? Are we going to allow multiple node groups?

cmd/eksctl/delete.go
```diff
@@ -71,34 +74,51 @@ func doDeleteCluster(cfg *eks.ClusterConfig, name string) error {
...
+		return false
 	}
 
-	if err := ctl.DeleteControlPlane(); err != nil {
-		debugError(err)
+	// We can remove all 'DeprecatedDelete*' calls in 0.2.0
+
+	stackManager := ctl.NewStackManager()
+
+	if err := stackManager.WaitDeleteNodeGroup(); err != nil {
```
richardcase: There is a change in behaviour here. In previous versions when deleting a cluster it would kick of the stack deletion but not wait. Now we have the following behaviour:
- Wait for the deletion of the node group stack then
- Start the deletion of the cluster stack BUT don't wait

We should probably be consistent with the deletion of both stacks (i.e. not wait or wait). The behavaior before was to not wait. But i think waiting for both is good.

cmd/eksctl/create.go
```diff
@@ -133,18 +136,22 @@ func doCreateCluster(cfg *eks.ClusterConfig, name string) error {
 	logger.Info("creating EKS cluster %q in %q region", cfg.ClusterName, cfg.Region)
 
 	{ // core action
-		taskErr := make(chan error)
-		// create each of the core cloudformation stacks
-		go ctl.CreateCluster(taskErr)
+		stackManager := ctl.NewStackManager()
+		logger.Info("will create 2 separate CloudFormation stacks for cluster itself and the initial nodegroup")
+		logger.Info("if you encounter any issues, check CloudFormation console first")
```
richardcase: > if you encounter any issues, check CloudFormation console first

Is there anything we could do here to help? I'm thinking the "verification" functionality would help.

pkg/cfn/manager/nodegroup.go
errordeveloper: Yes. I think numeric IDs make sense, but we can review that when we get to it. At least calling initial one `nodegroup-0` makes sense, but if user may need to give their extra groups more descriptive names, we can probably cater for it.
Right now I fell like numeric ones could work nicely if we want to say "I want an extra group with different types of instances and some custom IAM roles", e.g. without user having to specify a name... but really it's mostly a placeholder for now.

cmd/eksctl/delete.go
errordeveloper: We have to wait because nodegroup stack import values from the cluster stack, so cluster stack is actually protected until nodegroup stack is gone. Waiting for cluster to delete will take much longer, it takes basically at least as long as it take to create it (perhaps they have some kind of lazy deletion, like the EC2 instances which sit in terminated state for long time).

cmd/eksctl/create.go
errordeveloper: Yes, for sure. But here I just wanted to make it clear to the user where to go. It's based on feedback from @alexellis, who wasn't aware of CloudFormation, so didn't know that it's the go-to place, he basically ended-up looking all over AWS console and deleting resources one-by-one...

cmd/eksctl/delete.go
errordeveloper: We need to review deletions in general. Right the behaviour is a little weird, the output always says that your cluster will be delete even if the cluster doesn't exists. I can assure you this doesn't actually cause any distraction! :)

richardcase: Ah, that makes sense.

cmd/eksctl/create.go
richardcase: We can add an issue to give that extra feedback from cloudformation in the future.

richardcase:
@errordeveloper  - this all looking good to me so far. Really like the change.

cmd/eksctl/create.go
errordeveloper: #155

pkg/cfn/builder/nodegroup.go
```diff
@@ -0,0 +1,144 @@
...
+
+	n.vpc = makeImportValue(ParamClusterStackName, cfnOutputClusterVPC)
+
+	userData, err := nodebootstrap.NewUserDataForAmazonLinux2(n.spec)
+	if err != nil {
+		return err
+	}
+	n.userData = userData
+
+	n.rs.newStringParameter(ParamClusterName, "")
```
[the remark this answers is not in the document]
errordeveloper: Is this a suggestion or just an observation? ;)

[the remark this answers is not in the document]
richardcase: Good question!! It started out as a suggestion that was going to be "should we set this here as its in the spec and get rid of makeNodeGroupParams". But that would mean adding nodegroupid to the spec. So really its an observation and can be ignored.

errordeveloper:
I'm wrestling with awslabs/goformation#103, I might put it on hold and merge this first, but would like to get somewhere close to a more rounded PR with tests passing etc.

cmd/eksctl/create.go
christopherhein: In theory we could use the validate api when you ran dry run after outputting the templates to STDOUT - https://docs.aws.amazon.com/cli/latest/reference/cloudformation/validate-template.html

Note sure if it would really be useful…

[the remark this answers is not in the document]
mikemorris: Providing (and documenting) an integration with Terraform would be 😍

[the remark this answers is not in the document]
errordeveloper: @mikemorris we aren't planning to integrate with Terraform, but we are keen to hear about your use-case – please open an issue and tell us about it 😄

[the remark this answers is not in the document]
mikemorris: @errordeveloper Basically the same as https://github.com/kubernetes/kops/blob/master/docs/terraform.md - we're currently using kops to generate cluster spec and terraform to manage state. Considering migrating EKS and would love to stick to a similar workflow.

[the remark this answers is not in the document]
richardcase: @mikemorris - we use the same kops/terraform method and it would be good to do the same with EKS/terraform. If no one else looks at it then i will definetly have a look.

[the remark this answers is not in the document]
errordeveloper: Let's fork this thread into an issue :)

cmd/eksctl/create.go
```diff
@@ -75,25 +76,27 @@ func createClusterCmd() *cobra.Command {
 	fs.IntVarP(&cfg.MinNodes, "nodes-min", "m", 0, "minimum nodes in ASG")
 	fs.IntVarP(&cfg.MaxNodes, "nodes-max", "M", 0, "maximum nodes in ASG")
 
+	fs.IntVar(&cfg.MaxPodsPerNode, "max-pods-per-node", 0, "maximum number of pods per node (set automatically if unspecified)")
 	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
 
+	fs.BoolVar(&cfg.NodeSSH, "ssh-access", false, "control SSH access for nodes")
```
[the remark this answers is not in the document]
errordeveloper: I'd like to review flags later, would be nice to see grouping... as part of that we can see which help messages are still unclear.

cmd/eksctl/create.go
errordeveloper: I think validation is already done, cause I've seen validation errors...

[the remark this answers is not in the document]
mikemorris: #181

PR #137, opened 2018-07-25
Implement gready deletion

### LINKED ISSUE #136: ignore errors during deletion
Right most delete calls have to succeed, and that's not useful for most people (especially those who are not familiar with CloudFormation).

errordeveloper:
Fixes #136.


richardcase:
I like the idea of ignoring errors on deletion so that it continues if a resource was never created (or has been manually deleted). I think this is good for now.

In the future we should try to be a bit more intelligent with handling errors, so #138 has been raised to improve this in a later release.

errordeveloper:
I should add a note on failed creation, so that users can see what to do to cleanup...

PR #139, opened 2018-07-25
fix: add roles for NLB support on EKS

### LINKED ISSUE #133: can we use NLB with EKS clusters created via eksctl?
I've been trying out NLB instead of ELB as NLBs have stable static IP addresses; so you can use the nip.io trick of setting up Ingress without mandating a custom DNS name being used to setup wildcard CND via a CNAME to kick the tyres on EKS.

Things work great on kops but don't seem to work on EKS which am guessing is related to roles. It looks like these kinds of roles are required for setting up NLB on AWS with kops
https://gist.github.com/micahhausler/4f3a2ee540f5714e6dd91b4bacace3ae#file-create-cluster-sh-L30

so am guessing something similar is required on the `NodeInstanceRole`. Will try figure out whats really required...

errordeveloper:
fixes #133


errordeveloper:
@jstrachan in my test I can see latest version of this results in LB of type 'network' being created in my account, I suppose that means it worked? ;)

jstrachan:
@errordeveloper that sounds like it worked! 

I've just tested it on Jenkins X now and its working perfectly! Huge thanks!!!

errordeveloper:
@jstrachan my pleasure!

PR #140, opened 2018-07-25
ListClusters Improvements

richardcase:
Started to add output printers, currently printers for:
- YAML
- JSON
- Textual table layout (very rudimentry).

These still need to be hooked into list custers (as does the paging), this is very much WIP.


Gopkg.toml
```diff
@@ -72,3 +72,7 @@ required = [
 [[constraint]]
   name = "github.com/onsi/ginkgo"
   version = "1.5.0"
+
+[[constraint]]
+  name = "gopkg.in/yaml.v2"
```
errordeveloper: This one was unfavoured for a long time, but perhaps v2 has no major issue... however, I think it's better to stick to `"github.com/ghodss/yaml"`, which is what Kubernetes uses.

cmd/eksctl/get.go
```diff
@@ -43,6 +51,7 @@ func getClusterCmd() *cobra.Command {
 	fs := cmd.Flags()
 
 	fs.StringVarP(&cfg.ClusterName, "name", "n", "", "EKS cluster name")
+	fs.IntVar(&pageSize, "page-size", DEFAULT_PAGE_SIZE, "The number of results to return in single query")
```
errordeveloper: perhaps `--max-results` would be better here...

errordeveloper:
@richardcase have you looked at [k8s.io/kubernetes/pkg/printers](https://godoc.org/k8s.io/kubernetes/pkg/printers) at all? That's what kubectl uses, it may or may not be useful to us here...

errordeveloper:
I've peeked at [`clusterctl`](https://github.com/kubernetes-sigs/cluster-api/tree/master/clusterctl), and there is no get command as far as I can see (it only has create, validate & delete right now), which makes sense as there is no where really to get a list from...

errordeveloper:
I suppose [`kops get clusters`](https://github.com/kubernetes/kops/blob/master/cmd/kops/get.go) is pretty close to what we will likely end up with, it's probable worse having a look at what goes on there, even just copying table format maybe something to consider.

Gopkg.toml
richardcase: I'll change this.

cmd/eksctl/get.go
richardcase: Not sure about renaming to `--max-results` as this implies that it will only return results up to this maximum. 

Kubectl uses `--chunk-size`. How about that instead?

richardcase:
I did look at https://godoc.org/k8s.io/kubernetes/pkg/printers. We can't re-use the printers but it did provide some inspiration. The printers in kubectl assume that all objects to print implement the `Object` interface.

richardcase:
> I suppose kops get clusters is pretty close to what we will likely end up with, it's probable worse having a look at what goes on there, even just copying table format maybe something to consider.

I think this depends how generic we make the printers. My current table printer is pretty bad so i may look at including this.

I'll make some changes today.

cmd/eksctl/get.go
errordeveloper: Ah, my bad... Sure, let's call it `--chunk-size`.

errordeveloper:
> whats your preference with referencing the package or creating a version of it in our codebase?

We already use kops packages, no need to copy code if we don't have to! (just double check if it doesn't call glog, kops code does that all the time and it can be a little nasty).

richardcase:
> just double check if it doesn't call glog, kops code does that all the time and it can be a little nasty

It does use glog, does that mean we need to copy the code then?

errordeveloper:
@richardcase hm, lets see what we can do about that...

I'd like to make sure we don't enable glog globally, and don't actually write any output with it.
That is because it has a very weird format, and it can (at least used to) register CLI flags without you asking for it.

I'd consider forking the code, including that as dependency and making a PR to somehow eliminate glog from the given code path. Copying seems like the last resort, in case PR gets turned down...

pkg/printers/printers_test.go
```diff
@@ -26,9 +26,18 @@ func TestTablePrinter(t *testing.T) {
 		},
 	}
 	printer := NewTablePrinter()
-	printer.PrintObj(test, os.Stdout)
+
+	printer.(*TablePrinter).AddColumn("CLUSTERNAME", func(c *eks.Cluster) string {
```
errordeveloper: Perhaps just `NAME`, as argument name is `--name`.

cmd/eksctl/get.go
```diff
@@ -3,14 +3,16 @@ package main
...
 	"os"
+	"strings"
+	"time"
 
+	"github.com/weaveworks/eksctl/pkg/eks"
 	"github.com/weaveworks/eksctl/pkg/printers"
 
-	"github.com/spf13/cobra"
-
+	awseks "github.com/aws/aws-sdk-go/service/eks"
```
errordeveloper: good idea to call this differently, I just got a little confused reading the other file... would you mind to rename this everywhere else?

errordeveloper: Maybe in a separate PR :)

.vscode/launch.json
```diff
@@ -0,0 +1,46 @@
+{
```
errordeveloper: I am not sure if we should add this here.. perhaps sample configs in another directory or `HACKING.md` along with other notes/pointers?

cmd/eksctl/get.go
```diff
@@ -43,10 +56,12 @@ func getClusterCmd() *cobra.Command {
 	fs := cmd.Flags()
 
 	fs.StringVarP(&cfg.ClusterName, "name", "n", "", "EKS cluster name")
+	fs.IntVar(&chunkSize, "chunk-size", DEFAULT_CHUNK_SIZE, "Return large lists in chunks rather than all at once. Pass 0 to disable.")
 
 	fs.StringVarP(&cfg.Region, "region", "r", DEFAULT_EKS_REGION, "AWS region")
 	fs.StringVarP(&cfg.Profile, "profile", "p", "", "AWS creditials profile to use (overrides the AWS_PROFILE environment variable)")
 
+	fs.StringVarP(&output, "output", "o", "log", "Specifies the output printer to use")
```
errordeveloper: I think `table` should be the default.

pkg/printers/table.go
```diff
@@ -0,0 +1,30 @@
+package printers
+
+import (
+	"io"
+
+	"k8s.io/kops/util/pkg/tables"
```
errordeveloper: https://github.com/kubernetes/kops/issues/5551

Gopkg.toml
```diff
@@ -72,3 +76,7 @@ required = [
 [[constraint]]
   name = "github.com/onsi/ginkgo"
   version = "1.5.0"
+
+[[override]]
+  name = "github.com/gophercloud/gophercloud"
```
errordeveloper: 🤦‍♂️ https://github.com/kubernetes/kops/issues/5551

pkg/printers/printers_test.go
richardcase: I agree. It was calld `NAME` when i initially made the change. So i'll change it back.

cmd/eksctl/get.go
richardcase: Done, #142

.vscode/launch.json
richardcase: Ah, this was a mistake in all honesty. But yes i think having sample configs and a HACKING.md would be good. I'll remove this from the PR and have a separate PR.

cmd/eksctl/get.go
richardcase: I wasn't sure which one to make default so i went to "as-is" functionality. I'll change to table.

Gopkg.toml
richardcase: Great stuff. I hadn't got around to making kops changes eyt (just using as is). The kops PR makes sense. Perhaps i'll have a look after i get teh final changes on this done.

errordeveloper:
@richardcase how much work there is left to do on this?

richardcase:
> how much work there is left to do on this?

There are are only 2 things left to do:
1. Add additional test for the table printer to test the scenario where we have multiple clusters. (this test is currently commented out).
2. Remove the `glog` calls from the error paths of the kops code we import.

errordeveloper:
> Remove the glog calls from the error paths of the kops code we import.

Yes, I think we should try to do it, but I don't think we have to block the release. If it's easy enough (i.e. simply convert to errors) then we can use a fork of kops for now, otherwise just leaving it and fixing out-of-band would probably be the right thing. 

I checked now, and it looks like glog doesn't add user-facing impact, i.e. it doesn't register flags (it used to), and doesn't affect our logger in any way. These were things I've seen in the past, when I used a dependency that had glog.

errordeveloper:
Not so sure we need to include ARN column, it makes the output very very wide and there is little use to it, don't you think?

richardcase:
> Not so sure we need to include ARN column, it makes the output very very wide and there is little use to it, don't you think?

It is pretty long, i'll remove ARN. I'll add kubernetes version into the table instead. 

Would we want a list of the security group ids included?

errordeveloper:
> I'll add kubernetes version into the table instead.

> Would we want a list of the security group ids included?

There would be one for the cluster, but also one per nodegroup actually... May be we could add just the one for the cluster, and cover nodegroups when we have a `eksctl get nodegroups` command.

richardcase:
Changes have been made:

screen shot 2018-08-01 at 09 54 04

What do you think? Still too long?

errordeveloper:
@richardcase I think this is better, thanks a lot! We can always tweak it later.

richardcase:
@errordeveloper - If we leave the glog stuff as is and work on those separately then theres no more changes on my side.

errordeveloper:
@richardcase thanks a lot, I'll review today! could you please rebase?

pkg/printers/json_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+		BeforeEach(func() {
+			printer = NewJSONPrinter()
+		})
+
+		It("should not be nil", func() {
+			Expect(printer).ShouldNot(BeNil())
+		})
+
+		//TODO - finish
```
errordeveloper: Does this still need finishing up, or it's rather good enough?

pkg/printers/table_test.go
```diff
@@ -0,0 +1,162 @@
...
+			printer.(*TablePrinter).AddColumn("ARN", func(c *awseks.Cluster) string {
+				return *c.Arn
+			})
+		})
+
+		It("should not be nil", func() {
+			Expect(printer).ShouldNot(BeNil())
+		})
+
+		//TODO - finish
```
errordeveloper: ditto...

pkg/printers/yaml_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+		BeforeEach(func() {
+			printer = NewYAMLPrinter()
+		})
+
+		It("should not be nil", func() {
+			Expect(printer).ShouldNot(BeNil())
+		})
+
+		//TODO - finish
```
errordeveloper: and here..

errordeveloper:
Also, seems like logger is not working as intended:

```
./eksctl get clusters --output=log
2018-08-02T09:09:24+01:00 [ℹ]  []*string{(*string)(0xc42052c490), (*string)(0xc42052c4a0), (*string)(0xc42052c4b0), (*string)(0xc42052c4c0)}
```

Seems like it would be an easy fix, but I wonder if we need it at all?

pkg/eks/eks.go
```diff
@@ -143,27 +147,65 @@ func (c *ClusterProvider) GetCredentials(cluster awseks.Cluster) error {
...
 		}
+		allClusterNames = append(allClusterNames, clusters...)
+		if nextToken != nil && *nextToken != "" {
+			token = *nextToken
+		} else {
+			break
+		}
 	}
+
+	logger.Debug("clusters = %#v", allClusterNames)
```
errordeveloper: This seems to print only the pointers, I'm not sure if we need it at all...
```
2018-08-02T09:10:12+01:00 [▶]  clusters = []*string{(*string)(0xc4205307e0), (*string)(0xc4205307f0), (*string)(0xc420530820), (*string)(0xc420530830)}
```

pkg/printers/json_test.go
richardcase: Fixed.

pkg/printers/table_test.go
richardcase: Fixed.

pkg/printers/yaml_test.go
richardcase: Fixed.

pkg/eks/eks.go
richardcase: I removed the `log' printer. It was always an after thought.

PR #141, opened 2018-07-26
license: document files under MPL-2 and CDDL

dholbach:


errordeveloper:
I believe there are license files already in `vendor/`, maybe you can simply point at those instead of adding new copies? Also, I don't think we actually use the code as such.

dholbach:
I'm trying to follow the instructions on https://www.apache.org/legal/resolved.html#category-b - sorry for not pointing this out in the PR description.

LICENSE.MPL-2.0
```diff
@@ -0,0 +1,385 @@
+eksctl makes use of the following pieces of code which are licensed under
+the MPL-2.0:
+
+  https://github.com/go-sql-driver/mysql
```
errordeveloper: Actually, we don't vendor this one. It maybe a sub-dependency, but we don't drag it along.

LICENSE.MPL-2.0
```diff
@@ -0,0 +1,385 @@
...
+
+You can find our copy of the code in ./vendor/.
+
+At the point of writing they also pull in
+
+  https://github.com/hashicorp/hcl
```
errordeveloper: Yes, we have this vendored...

LICENSE.CDDL
```diff
@@ -0,0 +1,387 @@
+eksctl makes use of code which in turn pull in code under the CDDL.
+At the point of writing this is:
+
+  https://github.com/howeyc/gopass
```
errordeveloper: This one is vendored.

errordeveloper:
So you are saying that APL guideline dictate this convention of having `LICENSE.<TYPE>` in top-level directory?

dholbach:
I updated the PR and it reflects current vendoring reality.

If you prefer, I could re-do the PR to add a brief section to `README.md` about licenses. It's what is suggested on the [Apache Legal site](https://www.apache.org/legal/resolved.html#category-b).

`hcl` includes a copy of the MPL-2.0 license, so just pointing to that should be fine AFAIUI.

`gopass` includes `terminal_solaris.go` which is under CDDL-1.0, but it just contains a brief bit in the header, not the full license text. §3.1 of [the license](https://opensource.org/licenses/CDDL-1.0) seems to require this though.

LMK what you think.

dholbach:
Sorry, please review again - basically I just "updated the branch" locally. Did another review of the source, it's fine right now (7701fdc1074d86c5223726334146a5c1d6a12343).

PR #153, opened 2018-08-09
Get cluster fixes

### LINKED ISSUE #145: eksctl get clusters is not clear about cluster status
I have a cluster that hasn't been created yet, it's called `floral-mushroom-1533218176`.

It appears in regular list:
```
> eksctl get clusters
NAME
beautiful-gopher-1528350240
floral-mushroom-1533218176
>
```

But trying to get more info returns nothing:
```
> eksctl get clusters floral-mushroom-1533218176
>
```

I think we should at status column, or always filter ready clusters, and provide a flag to show non-ready ones.

### LINKED ISSUE #146: eksctl get clusters output is confusing when no clusters a present
```
> eksctl get clusters
NAME
>
```

I think it should say "No clusters found" instead, which would be more like what `kubectl` does.

richardcase:
Changed the 'list clusters' functionality:

- If there are no clusters and you are using the table printer then `No clusters found` will be output. If you are using the JSON or YAML printers then an empty array is returned. Fixes #146.
- If specifying a clustername and the cluster doesn't have a status of `ClusterStatusActive` then it will still print its details. Before this it was ignored. Fixes #145.
- Added additional tests to support the above and started to refactor tests so common 'fake' creation is separated out.


.vscode/launch.json
```diff
@@ -0,0 +1,44 @@
+{
```
errordeveloper: do you want to add this to your `~/.config/git/ignore`?

pkg/testutils/cluster.go
```diff
@@ -0,0 +1,34 @@
...
+		},
+	}
+
+	return cluster
+}
+
+// ValidClusterStatus checks that the provided status is a valid
+// status of a cluster according to AWS SDK
+func ValidClusterStatus(status string) bool {
+	valid := map[string]bool{awseks.ClusterStatusActive: true, awseks.ClusterStatusCreating: true, awseks.ClusterStatusDeleting: true, awseks.ClusterStatusFailed: true}
```
errordeveloper: I'd put each key on a separate line here..

errordeveloper: Actually, is this used anywhere yet?
I'm not entirely sure about having it here, I though we already have some filtering functionality in `pkg/eks/api.go`.. I'd try to fix whatever needs fixing there, e.g. pass filter option or something like that...

errordeveloper: Actually, it'd be good to leave any major improvements to how we get out cluster until after #132, as after that we would be looking at only CloudFormation-managed clusters. When we do that, we might want to let anyone list all EKS clusters with e.g. something like `--include-unmanaged` or `--incude-all`.

.vscode/launch.json
richardcase: Added...thanks.

pkg/testutils/cluster.go
richardcase: I removed this function. I was going to call it from `NewFakeCluster` to valid the status being passed it. But as its a test i decided not to and left this in. Its been removed.

PR #160, opened 2018-08-09
Fix code coverage reporting

richardcase:
This is to fix code coverage. 

- Changed coverage to run it circle golang container.


errordeveloper:
@richardcase is this still WIP?

richardcase:
> is this still WIP?

Yes, but worked out the problem (i think).

PR #168, opened 2018-08-20
Use aws-iam-authenticator when it's installed

### LINKED ISSUE #147: detect which authenticator command to use in kubeconfig
We've merged #117, which recognises new name of the authenticator command, but we still write kubecofig file for `heptio-authenticator-aws`. Ideally, we should keep the default behaviour as is, mostly because there are still no officially released binaries (https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/119).

Simplest way to do this would be by calling `exec.LookupPath("aws-iam-authenticator")` from `WithExecHeptioAuthenticator` (which we should probably rename). We shouldn't have to do anything more sophisticated, as we won't need to supporting both authenticator versions for too long.

errordeveloper:
### Description

Fixes #147.

### Checklist
- [x] All tests passing (i.e. `make tests`)
- [ ] Added yourself to the humans.txt file

### DIFF
--- a/pkg/eks/auth.go
+++ b/pkg/eks/auth.go
@@ -22,7 +22,8 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"k8s.io/kops/pkg/pki"
-	"k8s.io/kops/upup/pkg/fi/utils"
+
+	"github.com/weaveworks/eksctl/pkg/utils"
 )
 
 func (c *ClusterProvider) getKeyPairName(fingerprint *string) string {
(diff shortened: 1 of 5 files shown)


errordeveloper:
@richardcase once we merge this, let's tag 0.1.1 🚢!

pkg/eks/auth.go
```diff
@@ -186,13 +187,13 @@ func (c *ClusterProvider) NewClientConfig() (*ClientConfig, error) {
...
 
-func (c *ClientConfig) WithExecHeptioAuthenticator() *ClientConfig {
+func (c *ClientConfig) WithExecAuthenticator() *ClientConfig {
 	clientConfigCopy := *c
 
 	x := clientConfigCopy.Client.AuthInfos[c.ContextName]
 	x.Exec = &clientcmdapi.ExecConfig{
 		APIVersion: "client.authentication.k8s.io/v1alpha1",
-		Command:    "heptio-authenticator-aws",
+		Command:    utils.DetectAuthenticator(),
```
richardcase: Nice, its great kops has this already.

pkg/utils/kubectl.go
```diff
@@ -89,16 +93,39 @@ func CheckAllCommands(kubeconfigPath string, isContextSet bool, contextName stri
...
+		path, err := exec.LookPath(cmd)
 		if err == nil {
-			// binary was found
-			logger.Debug("%s: %q", bin, path)
+			// command was found
+			logger.Debug("%s: %q", cmd, path)
 			return nil
 		}
 	}
 	return fmt.Errorf("neither aws-iam-authenticator nor heptio-authenticator-aws are installed")
```
richardcase: Feel free to ignore this one! Would be good if the error message used the constants defined so there is less chance of deviation of error message from the commands.

errordeveloper: Yes, I thought about this, but we will soon care less about the legacy command, so I decided not too worry.

pkg/eks/auth.go
errordeveloper: This is in our utils package, which now wraps the single kops utils function we use.

PR #173, opened 2018-08-27
trivial: Update .gitignore

philoserf:
fix binary ignore, as it was it also caused folders named 'eksctl' to be ignored, for example, `/cmd/eksctl`

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/eksctl/173)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 .goreleaser.brew.combined.yml
 
 # Locally built binary
-eksctl
+/eksctl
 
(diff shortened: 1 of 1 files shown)


philoserf:
test failures are unrelated to the change I have proposed here

errordeveloper:
> test failures are unrelated to the change I have proposed here

@philoserf I think there maybe a bug in how coveralls is called, I will take a look later on.

PR #179, opened 2018-09-04
Update kops to alleviate incidental dependencies

errordeveloper:
### Description

We got a bag of deps from kops (https://github.com/kubernetes/kops/issues/5551), but it's been fixed now! 🎉 

### Checklist


errordeveloper:
@richardcase per https://github.com/kubernetes/kops/issues/5551#issuecomment-409965074:

> I refactored the package for cleaner package dependencies, primarily by allowing for you to register pretty-print functions.

Should we update our printers to take advantage of the changes in kops?

Gopkg.toml
```diff
@@ -33,7 +33,7 @@ required = [
 
 [[constraint]]
   name = "k8s.io/kops"
-  version = "1.9.1"
+  branch = "master"
```
richardcase: Would the plan then be to change this to a specific version again once there is a tag that includes the change?

errordeveloper: Yes. There are multiple other packages in very much the same situation, but I wouldn't make it a high priority just yet...

errordeveloper:
@richardcase any thoughts on whether printers package needs an update to make better use of changes in https://github.com/kubernetes/kops/pull/5565?

errordeveloper:
>@richardcase any thoughts on whether printers package needs an update to make better use of changes in kubernetes/kops#5565?

I've looked at it, and don't think those changes affect our usage.

richardcase:
@errordeveloper - like you i suspect not but i will take a proper look.

PR #180, opened 2018-09-04
WIP: Static bootstrap

errordeveloper:
### Description

Make node bootstrap more static, use less shell scripts

### Checklist
- [ ] Code compiles correctly (i.e `make build`)
- [ ] Added tests that cover your change (if possible)
- [ ] All tests passing (i.e. `make test`)
- [ ] Added/modified documentation as required (such as the README)
- [ ] Added yourself to the `humans.txt` file

PR #184, opened 2018-09-05
Make node bootstrap (mostly) static, update goformation

errordeveloper:
This PR carries changes to support upstreaming of gofromation changes, along with node bootstrap simplification. It replaces #180 (at least for now, I might separate it out later).


richardcase:
Reviewing this now

errordeveloper:
Thank you for reviewing and merging! Let's look at #210 when I'm back from
holidays.

PR #187, opened 2018-09-06
Fix nil pointer error on create timeout.

jmcarp:
Fix a nil pointer error on stack create timeout. Looks like we mixed up two lines here.


errordeveloper:
Thanks a lot @jmcarp! Well spotted :)

Issue #183, opened 2018-09-04
Support for create a GPU-accelerated cluster


arun-gupta:
**Why do you want this feature?**

EKS cluster can be created with a GPU-accelerated AMI as explained at https://aws.amazon.com/blogs/compute/running-gpu-accelerated-kubernetes-workloads-on-p3-and-p2-ec2-instances-with-amazon-eks/. `eksctl create cluster` API should provide support for that.

**What feature/behavior/change do you want?**

Update `eksctl create cluster` to pick the right AMI based upon instance type or allow the AMI id to be explicitly specified.

arun-gupta:
// CC @countspongebob

richardcase:
We'll have to do this i 2 parts, using the correct AMI for the instance type is the first part. And then when we have the addon framework we can add the NVIDIA device framework.

Until we get the addon framework we can add supply instructions about manually running the device framework.

errordeveloper:
@richardcase sounds like a good plan!

So to clarify, after 1st part is done, users would be able to run `eksctl create cluster --node-type=<gpuInstanceType> [<flags>]` followed by `kubectl apply -f <gpuDriverManifest> [<flags>]`  to get going. And right now, one cannot do it, as AMI we use doesn't have GPU support. Is that correct?

And once we have add-ons (0.2.0), we will be able to cater for `eksctl create cluster  --addons=nvidia-gpu` to do the right thing.

@arun-gupta how does that sound?

arun-gupta:
Sounds like a great plan to me, can't wait to try it out.

arun-gupta:
Automatic installation of addon would be super!

errordeveloper:
I think we can close this now, as further work will be tracked in #205. The @richardcase's changes are in master now and will be available in 0.1.2 binaries, which we are planning to cut today.

To use create a cluster with GPU nodes, you will need to do the following:
1. get `eksctl` 0.1.2
2. subscribe to GPU AMIs
3. run `eksctl create cluster --node-type=p3.8xlarge --timeout=40m`

I don't know if timeout doesn't have to be that long, but @richardcase should be able to confirm what value worked. The thing is that GPU nodes take longer too boot.

Docs PRs are welcome, but I'd like to get the 0.1.2 out as soon as possible, as #190 is pretty urgent.

richardcase:
The main thing is you need to subscribe to the GPU Supported AMI:

https://aws.amazon.com/marketplace/pp/B07GRHFXGM

[...]

Issue #49, opened 2018-06-06
automate AMI updates


errordeveloper:
Currently, AMI IDs have to copied from [the docs](https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html), we need to make it easier to automate (e.g. via a script).

JordanFaust:
Commenting here rather then creating a new issue. As part of this the cli should expose an option to create a cluster with a custom AMI for the worker nodes. Thoughts on rolling these two things together?

errordeveloper:
Custom AMI will be supported soon, we just need to expose a flag, I'd like to do that as part of #48 (there is code already, but I'd like to simplify the code by using a library, if I can).

errordeveloper:
```
aws ec2 describe-images \
    --filters \
        Name=owner-id,Values=602401143452 \
        Name=virtualization-type,Values=hvm \
        Name=root-device-type,Values=ebs \
        Name=is-public,Values=true \
        Name=name,Values=amazon-eks-node-\* \
    --output text \
    --query 'max_by(Images[], &CreationDate).ImageId'
```

richardcase:
I'm making this change as part of #192 as you suggested.

richardcase:
@errordeveloper - do you think that this can be closed now (after #221)? Or would you like to wait until we get an automated test that checks for every evening for new image versions?

errordeveloper:
@richardcase yes :)

PR #192, opened 2018-09-06
WIP: AMI resolution and overriding (allowing GPU support)

richardcase:
### Description
If an AMI isn't specified then it will be resolved based
on region and instance type. It uses "resolvers" to do this.
Currebntly there are 2 resolvers:
- Default which uses the default EKS AMIs for a region
- GPU which for GPU instance types (p2 or p3) will use the
specific EKS GPU optimized image for a region.

This change also allows you to specify an AMI to use instead of
using resolution.

If using the GPU isntances then the NVIDIA device plugin will
still need to be installed manually:
https://github.com/NVIDIA/k8s-device-plugin
This will be an 'addon' in the future.

Issue #183

### Checklist

### DIFF
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -93,6 +93,8 @@ func createClusterCmd() *cobra.Command {
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
 
+	fs.StringVar(&cfg.NodeAMI, "node-ami", "", "if not supplied then eksctl will automatically set the AMI based on region/instance type; if supplied it will override the AMI to use for the nodes")
+
 	return cmd
 }
 
--- a/pkg/ami/ami.go
+++ b/pkg/ami/ami.go
@@ -1,0 +1,70 @@
+package ami
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ResolveAMI will resolve an AMI from the supplied region
+// and instance type. It will invoke a specific resolver
+// to do the actual detrminng of AMI.
+func ResolveAMI(region string, instanceType string) (string, error) {
+	var resolver Resolver
+
+	if strings.HasPrefix(instanceType, "p2") ||
+		strings.HasPrefix(instanceType, "p3") {
+		resolver = &GpuResolver{}
+	} else {
+		resolver = &DefaultResolver{}
+	}
+
+	return resolver.Resolve(region, instanceType)
+}
+
+// Resolver provides an interface to enable implementing multiple
+// ways to determine which AMI to use from the region/instance type.
+type Resolver interface {
+	Resolve(region string, instanceType string) (string, error)
+}
+
+// DefaultResolver resolves the AMi to the defaults for the region
+type DefaultResolver struct {
+}
+
+// Resolve will return an AMI to use based on the default AMI for each region
+// TODO: https://github.com/weaveworks/eksctl/issues/49
+// currently source of truth for these is here:
+// https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html
+func (r *DefaultResolver) Resolve(region string, instanceType string) (string, error) {
+	switch region {
+	case "us-west-2":
+		return "ami-73a6e20b", nil
+	case "us-east-1":
+		return "ami-dea4d5a1", nil
+	default:
+		return "", fmt.Errorf("Unable to resolve AMI for region %s and instance type %s", region, instanceType)
+	}
+}
+
+// GpuResolver resolves the AMI for GPU instances types.
+type GpuResolver struct {
+}
+
+// Resolve will return an AMI based on the region for GPU instance types
+func (r *GpuResolver) Resolve(region string, instanceType string) (string, error) {
+	if !strings.HasPrefix(instanceType, "p2") &&
+		!strings.HasPrefix(instanceType, "p3") {
+		return "", fmt.Errorf("Cannot resolve AMI as the instance type isn'y GPU optimized")
+	}
+
+	switch region {
+	case "us-west-2":
+		return "ami-0d20f2404b9a1c4d1", nil
+	case "us-east-1":
+		return "ami-09fe6fc9106bda972", nil
+	case "eu-west-1":
+		return "ami-09e0c6b3d3cf906f1", nil
+	default:
+		return "", fmt.Errorf("Unable to resolve AMI for region %s and instance type %s", region, instanceType)
+	}
+}
--- a/pkg/ami/ami_test.go
+++ b/pkg/ami/ami_test.go
@@ -1,0 +1,72 @@
+package ami_test
+
+import (
+	"github.com/weaveworks/eksctl/pkg/ami"
+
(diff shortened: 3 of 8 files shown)


pkg/ami/ami.go
```diff
@@ -0,0 +1,95 @@
...
+		return "ami-0b2ae3c6bda8b5c06"
+	case "eu-west-1":
+		return "ami-066110c1a7466949e"
+	default:
+		return ""
+	}
+}
+
+// GpuResolver resolves the AMI for GPU instances types.
+type GpuResolver struct {
```
errordeveloper: nit: GPUResolver

pkg/ami/ami_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Entry("with gpu (p2) instance and us-west-2", ResolveCase{
+			Region:       "us-west-2",
+			InstanceType: "p2.xlarge",
+			ExpectedAMI:  "ami-0d20f2404b9a1c4d1",
+			ExpectError:  false,
+		}),
+		Entry("with gpu (p3) instance and us-east-1", ResolveCase{
+			Region:       "us-east-1",
+			InstanceType: "p3.2xlarge",
+			ExpectedAMI:  "ami-09fe6fc9106bda972",
```
errordeveloper: It seems like we'd need to update this every time AMI changes... If we solve #49 by code generation, would you generate code for this also? I seems like a map would be easier to generate then switch+tests, or we could use a static JSON asset... I'm happy to land this as is, but just saying that we might need to change it soon enough.

richardcase: Yeah i completely agree. I'll take a look at doing this and then squash/rebase

richardcase: I'll work on #49 after this.

richardcase:
@errordeveloper - might be worth merging this in first so people can pick up the new AMIs. I'll start working on #49 now but as a separate PR.

richardcase:
@errordeveloper - changed back to WIP as i'm changing the code to resolve the AMI dynamically at runtime and never have hardcoded values in it. This will satisfy #49. 

The GPU images are an interesting one as the nodegroup stack gets stuck in creating for a very long time and then fails. It appears that the GPU images can only be used if you 'subscribe; through the marketplace. I'll continue testing this.

errordeveloper:
I am rather torn about that, in a way I would prefer more deterministic
option here, perhaps we could flag auto-resolution for the time being.

I am not sure what the default behaviour should be. Perhaps we can discuss
this at the end.

errordeveloper:
One thought that cross my mind is to have '--node-ami' flag that would take any AMI or one of special values, e.g. 'latest' and 'fixed'. Where 'latest' would use auto-resolution, while 'fixed' could be hardcoded, and eventually we could make it settable by the user via a config file. Not suggesting the default yet.

richardcase:
I've incorporated the feedback around `fixed` and `latest`.

richardcase:
When using p2/p3 instances you need to subscribe to the EKS GPU Enabled node:
https://aws.amazon.com/marketplace/pp/B07GRHFXGM

If you don't then the nodegroup stack will eventually fail saying you need to subscribe to the image.

cmd/eksctl/create.go
```diff
@@ -119,14 +122,18 @@ func doCreateCluster(cfg *api.ClusterConfig, name string) error {
...
-		return fmt.Errorf("--region=%s is not supported only %s and %s are supported", cfg.Region, EKS_REGION_US_WEST_2, EKS_REGION_US_EAST_1)
+	if cfg.Region != EKS_REGION_US_WEST_2 && cfg.Region != EKS_REGION_US_EAST_1 && cfg.Region != EKS_REGION_EU_WEST_1 {
+		return fmt.Errorf("--region=%s is not supported only %s, %s and %s are supported", cfg.Region, EKS_REGION_US_WEST_2, EKS_REGION_US_EAST_1, EKS_REGION_EU_WEST_1)
 	}
 
 	if err := ctl.SetAvailabilityZones(availabilityZones); err != nil {
 		return err
 	}
 
+	if err := ctl.EnsureAmi(); err != nil {
```
errordeveloper: nit: AMI

pkg/ami/aws.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/pkg/errors"
+)
+
+// IsAmiAvailableInAWS checks if a given ami is available in AWS
+func IsAmiAvailableInAWS(api ec2iface.EC2API, ami string) (bool, error) {
```
errordeveloper: nit: shouldn't need to use `AMI`  in functions names in package called `ami`..

pkg/ami/aws.go
```diff
@@ -0,0 +1,66 @@
...
+	if len(output.Images) < 1 {
+		return false, nil
+	}
+
+	return *output.Images[0].State == "available", nil
+}
+
+// QueryAWSForEksAmi will get the AMI is for the EKS nodesby querying
+// AWS for theimage to use. You need to supply a name pattern to use.
+func QueryAWSForEksAmi(api ec2iface.EC2API, namePattern string) (string, error) {
```
errordeveloper: how about `FindImageForEKS`?

richardcase:
Thanks for the review @errordeveloper. Changes have been made.

errordeveloper:
@richardcase I've looked over it once again and added a few changes to accommodate for supporting multiple OS families in the future, please merge whenever you are ready tomorrow and let's try to get 0.1.2 out of the door tomorrow :)

errordeveloper:
I have a few notes, I don't think we have to address these now, but perhaps things to discuss in the future:

- static resolver consists of two resolvers, while auto one has only one
- could this code be simpler, e.g. will chaining resolvers prove useful?
- should we propagate use of error types through out the code? are there benefits besides matching? (they seem to add lines of code that don't quite do anything)
- some error messages begin with capital letter, I've been told it's bad for some reason (maybe it was a linter rule?)

I am very happy with this PR, just wanted to mention things that raised some vague doubts, which seem like great topics to discuss! :)

richardcase:
Thanks for the approval @errordeveloper - squashed and merged.

Good comments and some points to ponder and discuss.

> static resolver consists of two resolvers, while auto one has only one could this code be simpler, e.g. will chaining resolvers prove useful?

I struggled with this very question. At the moment there are only 2 cases (non-GPU and GPU) and its probably overkill having 2 separate static resolvers chained together. I went with future extensibility in mind but perhaps it would've been better to stick to YAGNI. Perhaps we give it a while and if there are no new cases of different AMIs to use we simplify down to 1 static resolver. 

> should we propagate use of error types through out the code? are there benefits besides matching? (they seem to add lines of code that don't quite do anything)

For me the biggest benefit is from a testing perspective. Its cleaner to test for a specific error type than it is for a error message. This way you can change the error message without having to change the test.

> some error messages begin with capital letter, I've been told it's bad for some reason (maybe it was a linter rule?)

Ah, not heard of that. I will take a look and she what the deal is. I think we should run linting over the code base soon and get this as part of the CI build so it becomes part of our development workflow.

PR #193, opened 2018-09-06
Add linting target

richardcase:
### Description
A new make target has been added to run gometalinter.
There are a lot of linting errors!!

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,7 @@ test: generate
 
 .PHONY: lint
 lint:
-	gometalinter --exclude=^vendor\/  ./...
-
-.PHONY: install-lint-deps
-install-lint-deps:
-	curl https://git.io/vp6lP | sh
-	gometalinter --install
+	@gometalinter ./...
 
 .PHONY: ci
 ci: test lint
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -19,6 +19,19 @@ test: generate
 	@go test -v -covermode=count -coverprofile=coverage.out ./pkg/... ./cmd/...
 	@test -z $(COVERALLS_TOKEN) || goveralls -coverprofile=coverage.out -service=circle-ci
 
+.PHONY: lint
+lint:
+	gometalinter --exclude=^vendor\/  ./...
+
+.PHONY: install-lint-deps
+install-lint-deps:
+	curl https://git.io/vp6lP | sh
```
errordeveloper: Could you please add `github.com/alecthomas/gometalinter` to https://github.com/weaveworks/eksctl/blob/master/build/Gopkg.toml and https://github.com/weaveworks/eksctl/blob/master/build/install.sh instead? That way all the dependencies will be included in the build image, or installed locally via `make install-build-deps`.

richardcase: Seems like and easy thing to do but been struggling getting this working. I must be doing something stupid!

errordeveloper: happy to pick this up, would be nice to include in 0.1.2!

errordeveloper:
I made it fail only on errors, we should probably change it long-term, for the purpose of this PR, it seems right to not let it make all our builds fail.

richardcase:
Great @errordeveloper  - thanks for fixing this.

PR #207, opened 2018-09-13
Docs updated with GPU and --node-ami

richardcase:
### Description
The [README](README.md) has been updated with details of using GPU
instances types. Also a section on using --node-ami has been added.

### Checklist


README.md
```diff
@@ -40,8 +40,9 @@ A cluster will be created with default parameters
 - default EKS AMI
 - `us-west-2` region
 - dedicated VPC (check your quotas)
+- using static AMI resolver
```
errordeveloper: I'd avoid introducing 'resolver' here, saying 'use official EKS AMI' seems sufficient here...

README.md
```diff
@@ -142,6 +143,41 @@ To delete a cluster, run:
...
 eksctl delete cluster --name=<name> [--region=<region>]
 ```
+### GPU Support
+
+If you'd like to use GPU instance types (i.e. [p2](https://aws.amazon.com/ec2/instance-types/p2/) or [p3](https://aws.amazon.com/ec2/instance-types/p3/) ) then the first thing you need to do is subscribe to the [EKS-optimized AMI with GPU Support](https://aws.amazon.com/marketplace/pp/B07GRHFXGM). If you don't do this then node creation will fail.
+
+After subscribing to the AMI you can create a cluster specifying the GPU instance type you'd like to use for the nodes. For example:
+
+```
+eksctl create cluster -t p2.xlarge
```
errordeveloper: `--node-type=p2.xlarge` is preferred, so it's consistent and descriptive

README.md
```diff
@@ -142,6 +143,41 @@ To delete a cluster, run:
...
+
+| Keyword | Description |
+| ------------ | -------------- |
+| static       | Indicates that the AMI images ids embedded into eksctl should be used. This relates to the static resolvers. |
+| auto        | Indicates that the AMI to use for the nodes should be found by querying AWS. This relates to the auto resolver. | 
+
+If, for example, AWS release a new version of the EKS node AMIs and a new version of eksctl hasn't been released you can use the latest AMI by doing the following:
+
+```
+eksctl create cluster --node-ami auto
```
errordeveloper: nit: `--node-ami=auto`

PR #221, opened 2018-09-20
Generation of static AMI list

richardcase:
### Description
The list of AMIs to use for the static resolver is now generated
via go:generate. The list is queried from AWS and output to a code
file.

A separate makefile target called 'generate-ami' has been added that
runs go generate. The reason for a separate target is the fact that
AWS credentials are required for the code generation and this would
break the current workflow for the 'generate' target.

Another makefile target called 'ami-check' has been added. The
intended purpose of this by an automated scheduled process that
will run to check if new images are available.

Issue #49 

### Checklist


richardcase:
Ah, rebased now. Thanks @stefanprodan

Issue #175, opened 2018-08-27
Add Support for Auto-Deploying a Default Storage Class


christopherhein:
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
To ease the setup of EKS cluster for production usecases

**What feature/behavior/change do you want?**
When #132  gets merged if we have the possibility it would be nice to add an option to auto deploy a storageclass and make it the default.

The manifest that would need to be applied would be:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: gp2
  annotations:
    "storageclass.kubernetes.io/is-default-class": "true"
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
reclaimPolicy: Retain
mountOptions:
  - debug
```

errordeveloper:
@christopherhein yes, more then happy to do it! Could you please clarify the `#321` reference, which repo is it for?

christopherhein:
LOL, sorry about that I flipped that entirely, I meant #132 updating the above comment to reflect. Sorry about that.

karinnainiguez:
Hey @christopherhein !  
I'm hoping to be a new collaborator from the EKS team and help out with this.  But I brought it up to my team, and was wondering if you could clarify what the use case for this would be.  

As I understand it, this would allow an end user to create a cluster, and not have to run the `kubectl create -f <storage-class.yaml>` command.  Is that correct?   The question that came up was regarding the responsibility of deciding what type of storage class to make default.  Would allowing the user to skip this command make it more difficult for any user that may not want to use the chosen default storage class for their respective applications?  

My apologies if I'm misunderstanding the concept.  We can take this offline and discuss if you'd like. 
Thanks for your help!

christopherhein:
Hey @karinnainiguez, it shouldn't be set as a default with all `eksctl` clusters, what I'm hoping for is similar to what we have with `--full-ecr-access` have `--gp2-storage-class` or have `--storage-class=gp2` the latter only if we think that we'd ever have a way to set anything other than `gp2` by default.

The real benefit is not having to configure your storage class if you are trying to get up and running quickly and be able to use PVs without having to write or go find this manifest file.

karinnainiguez:
Thank you so much @christopherhein!  I really appreciate the clarification.  

[...]

PR #224, opened 2018-09-21
Storage Class flag

karinnainiguez:
### Description
The optional flag of `--storage-class` is set to true unless the user decides to exclude it with the flag `eksctl create cluster --storage-class=false`.  The new cluster will contain a default StorageClass of gp2 provisioned by EBS. That can be confirmed using the `kubectl get storageclass` command. 

Issue #175 
Note - Discussed making the flag a string with StorageClass type, but decided to make it a bool flag instead that creates a gp2 type and sets it as default to better capture the typical use case, without putting more burden on the user.

### Checklist
- [X] All tests passing (i.e. `make test`)
- [X] Added/modified documentation as required (such as the README)
- [X] Added yourself to the `humans.txt` file (added myself in previous PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ To create the same kind of basic cluster, but with a different name, run:
 eksctl create cluster --name=cluster-1 --nodes=4
 ```
 
+To prevent a default StorageClass of gp2 provisioned by EBS:
+
+```
+eksctl create cluster --storage-class=false
+```
+
 To write cluster credentials to a file other than default, run:
 
 ```
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ To create the same kind of basic cluster, but with a different name, run:
 eksctl create cluster --name=cluster-1 --nodes=4
 ```
 
-To prevent a default StorageClass of gp2 provisioned by EBS:
+A default [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) (gp2 volume type provisioned by EBS) will be added automatically when creating a cluster.  If you want to prevent this, use the `--storage-class` flag.  For example:
 
 ```
 eksctl create cluster --storage-class=false
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -91,7 +91,7 @@ func createClusterCmd() *cobra.Command {
 	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
-	fs.BoolVar(&cfg.Addons.Storage, "storage-class", true, "if true then a default StorageClass of type gp2 provisioned by EBS will be created")
+	fs.BoolVar(&cfg.Addons.Storage, "storage-class", true, "if true (default) then a default StorageClass of type gp2 provisioned by EBS will be created")
 
 	fs.StringVar(&cfg.NodeAMI, "node-ami", ami.ResolverStatic, "Advanced use cases only. If 'static' is supplied (default) then eksctl will use static AMIs; if 'auto' is supplied then eksctl will automatically set the AMI based on region/instance type; if any other value is supplied it will override the AMI to use for the nodes. Use with extreme care.")
 
--- a/cmd/eksctl/create.go
+++ b/cmd/eksctl/create.go
@@ -91,6 +91,7 @@ func createClusterCmd() *cobra.Command {
 	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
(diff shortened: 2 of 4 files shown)


cmd/eksctl/create.go
```diff
@@ -188,6 +189,13 @@ func doCreateCluster(cfg *api.ClusterConfig, name string) error {
 			return err
 		}
 
+		// add default storage class
+		if cfg.Addons.Storage == true {
+			if err := ctl.AddDefaultStorageClass(clientSet); err != nil {
+				return err
+			}
+		}
+
```
christopherhein: I might recommend moving this after the `NodeAuthConfigMap` creation. Even though this should work without any issues it feels a little odd to create this before the worker nodes are able to connect to the control plane… @errordeveloper @richardcase thoughts?

christopherhein: "this" representing the full `AddDefaultStorageClass` block.

karinnainiguez: Thanks @christopherhein!  I can definitely do that if y'all decide that's best.  Let me know @richardcase / @errordeveloper 

:)

richardcase: @christopherhein - yes i agree with you. I think it makes more logical sense.

cmd/eksctl/create.go
```diff
@@ -91,6 +91,7 @@ func createClusterCmd() *cobra.Command {
 	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
+	fs.BoolVar(&cfg.Addons.Storage, "storage-class", false, "Add a default StorageClass of type gp2 provisioned by EBS")
```
richardcase: From the discussion in #175, it sounds like the flag should be a string which defaults to gp2.  So, something similar to:

`fs.StringVar(&cfg.Addons.Storage, "storage-class", "gp2, "Adds a StorageClass. Defaults to a type gp2 provisioned by EBS")`

So running `eksctl create cluster` would create a default storage class of gp2 but this can be overridden.

@christopherhein @karinnainiguez  - what do you think?

christopherhein: Sorry this conversation should have been external. We had a short discussion about the implementation and had recommended following the path that `kops` uses rather than implement the full gamut of storage classes just set a default so out of the box PVs/PVCs work and let users that understand and want to change it do so after the fact with their own settings. `kops` just auto deploys the `gp2` class and calls it a day.

Do think it would be necessary to have all classes configurable at cluster creation?

richardcase: @christopherhein - the makes sense to me and i guess if we where to use another volume type (such as io1) then we'd may have to  add additional parameters to cover things like iopspergb.  Flipping to `true` as default would be good.

karinnainiguez: Awesome!  Thank you @christopherhein and @richardcase for all your help with this! 

I just submitted the change to make the flag true as default.  Let me know if you need me to make any other changes.

cmd/eksctl/create.go
karinnainiguez: Awesome!  I just moved that new function after the nodes join.  Thank you so much!

christopherhein:
Awesome, thanks @karinnainiguez I think we might need to rebase `master` back in then lgtm!

README.md
```diff
@@ -79,6 +79,12 @@ To create the same kind of basic cluster, but with a different name, run:
 eksctl create cluster --name=cluster-1 --nodes=4
 ```
 
+To prevent a default StorageClass of gp2 provisioned by EBS:
```
richardcase: Perhaps a slight change here. How about something like:

A default [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) (gp2 volume type provisioned by EBS) will be added automatically when creating a cluster. If you want to prevent this use the  `--storage-class` flag. For example:
```
eksctl create cluster --storage-class=false
```

cmd/eksctl/create.go
```diff
@@ -91,6 +91,7 @@ func createClusterCmd() *cobra.Command {
 	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
 
 	fs.BoolVar(&cfg.Addons.WithIAM.PolicyAmazonEC2ContainerRegistryPowerUser, "full-ecr-access", false, "enable full access to ECR")
+	fs.BoolVar(&cfg.Addons.Storage, "storage-class", true, "if true then a default StorageClass of type gp2 provisioned by EBS will be created")
```
richardcase: A very minor suggestion

"if true (default) then a default StorageClass of type gp2 provisioned by EBS will be created"

pkg/eks/storageclass.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	storage "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+func (c *ClusterProvider) AddDefaultStorageClass(clientSet *clientset.Clientset) error {
```
richardcase: Perhaps adding a logger.Debug somewhere to output the fact that we are creating a storage class. Really just to help with any debugging of issues.

karinnainiguez: Thank you so much for reviewing the changes @richardcase! I really appreciate the help.  Feel free to let me know if you find anything else that can be improved.  I'd be happy to incorporate it.  :)

PR #229, opened 2018-09-26
Configure node volume size and encryption.

### LINKED ISSUE #214: --node-volume-size flag for create cluster
It is currently not possible to set the volume size used for the nodes.  It defaults to 20GB and it is only possible to modify the volume sizes after creation using AWS tools.

A --node-volume-size flag that can be used to set larger node volume size on cluster creation.

jmcarp:
[Fixes #214]

We should probably look up the root device name from the AMI instead of hard-coding `/dev/xvda`. Happy to add that here if that makes sense.

### Description

### Checklist


jmcarp:
Actually, noticed that encryption doesn't work here, since the ebs snapshots associated with the base AMIs aren't encrypted. Rolling back to just configuring volume size.

cmd/eksctl/create.go
```diff
@@ -66,6 +66,7 @@ func createClusterCmd() *cobra.Command {
 
 	fs.StringVarP(&cfg.Region, "region", "r", api.DEFAULT_EKS_REGION, "AWS region")
 	fs.StringVarP(&cfg.Profile, "profile", "p", "", "AWS credentials profile to use (overrides the AWS_PROFILE environment variable)")
+	fs.IntVarP(&cfg.NodeVolumeSize, "node-volume-size", "", 0, "Node volume size (in GB)")
```
errordeveloper: It would be good to use a constant, e.g. `api.DEFAULT_NODE_VOLUME_SIZE`.

README.md
```diff
@@ -136,6 +136,12 @@ apply to CloudFormation stacks but not EKS clusters.
 eksctl create cluster --tags environment=staging --region=us-east-1
 ```
 
+To configure node volume size, use the `--node-volume-size` flag.
+
+```
+eksctl create cluster --node-volume-size 50
```
errordeveloper: Please use `--node-volume-size=50`, for consistency with the rest of the readme.

cmd/eksctl/create.go
errordeveloper: Also, please move it just above `cfg.MaxPodsPerNode`, to match the order in which it's defined in the struct.

jmcarp: The default behavior here is to use the volume size from the AMI. I don't know if we need a named constant for the empty value.

jmcarp:
Thanks, updated.

jmcarp:
@errordeveloper: anything else I should update here?

errordeveloper:
@jmcarp thanks, this great!

Issue #222, opened 2018-09-20
Command to list supported regions


richardcase:
**Why do you want this feature?**
It would be good to have an eksctl command that lists the EKS supported regions. Something like `eksctk utils list-available-regions`. 

**What feature/behavior/change do you want?**
This should list out the regions where eks is cupported

This is a result of discussion in #217

bowlesns:
I'll take this. Thanks.

richardcase:
Great, thanks @bowlesns

philoserf:
There isn't an API from AWS to provide this information today. The project would need to maintain a list of regions until an API is available. It looks like today AWS keeps a list of available regions in two separate documents in the user guide at [awsdocs/amazon-eks-user-guide](https://github.com/awsdocs/amazon-eks-user-guide/tree/master/doc_source)

Is this a responsibility the project team wants to take on?

redborian:
There is already a hard-coded list of supported regions in `eksctl`. The team already has the responsibility of keeping that hard-coded list up-to-date.
It would be awesome if you can somehow expose them to the user.

richardcase:
> It would be awesome if you can somehow expose them to the user. 

I know that @bowlesns has been working on this.

errordeveloper:
I think this was important when only a few regions were supported, it seems like we don't really need it any more as EKS is in most regions now.

PR #231, opened 2018-09-27
WIP: Added eksctl get supported-regions command

bowlesns:
In response to issue #222, I added a command to allow a user
to get the regions where EKS is currently supported. After
some discussion with another contributor, thought it would
be better under the "get" command.

The command is simple, and prints out a few lines showing
which regions EKS is currently supported in. I also added
the url where the most up to date information is on AWS.

Issue #222

### Description

### Checklist


bowlesns:
Adding as a WIP. Wanted to ask if it's worth adding in the "printers" functionality that was just added to output multiple formats, or if the command will suffice as-is.

richardcase:
Great, thanks @bowlesns.

I like the idea of having the list of regions outputted as a table. But not sure if its actually needed.

@errordeveloper  - Do you think it would be good if we used the table printer or just leave it using formatted text?

redborian:
@bowlesns Just a thought: Instead of hard-coding the command output, can we get the list of regions from `eksctl/cmd/eksctl/create.go`?

errordeveloper:
Let's refactor the code a little bit and add `[]string` to hold supported regions.

richardcase:
@errordeveloper @bowlesns  - apologies as this was initially under utils and then i thought it flowed better being under `get`. But if by having something under `get` this also implies there should be `delete` and `create` then i was wrong in my advice (sorry @bowlesns)

errordeveloper:
I've happened to add `api.SupportedRegions` [in #273](https://github.com/weaveworks/eksctl/pull/273/files#diff-9c8e1b263f86d71b19e53c9ce3853aa1R30), which we will probably merge pretty soon.

errordeveloper:
Hey @bowlesns I'm gonna have to close this, but feel free to re-open if you want to get back to it!

Issue #235, opened 2018-10-04
Default region setting not respected


jpds:
**What happened?**

In `~/.aws/config`, I have this set:

```
[default]
region = eu-west-1
```

When I  ran `eksctl create cluster`, it started creating one in `us-west-2`.

richardcase:
@jpds  - interesting issue, thanks. We don't take into consideration at all the region specified in the default or named profile. If a region isn't specified on the command line (`--region/-r`) then we default to `us-west-2`. To get some more clarification, what do you think to these scenarios:

**No region in AWS config AND no region flag supplied**
Then we can default to `us-west-2` like we currently do

**EKS enabled region in AWS config AND no region flag supplied**
Then we default to the region specified in the AWS config. So in your example `eu-west-1`. We need to take into consideration default and named profiles here.

**Non EKS enabled region in AWS config AND no region flag supplied**
Then we can default to `us-west-2` like we currently do.

**Region in AWS config AND region flag supplied**
Then we use the region specified in the flag (printing error if the region supplied isn't EKS enabled)

PR #241, opened 2018-10-05
Changed region handling

richardcase:
### Description
The region handling for the commands has changed. Previously, if
you didn't specify the region flag (-r/--region) then we would
default to us-west-2 even if you had a different region specified
in the AWS environment variables or shared config. With this change
we only default to us-west-2 if there is no region specified using the
flay AND there is no region coming from the AWS shared config /
environment variables.

Issue #235

### Checklist


errordeveloper:
```
 [0] >> ./eksctl create cluster
2018-10-10T16:49:14+01:00 [✖]  checking AWS CloudFormation access – cannot list stacks: MissingRegion: could not find region configuration
 [1] >> ./eksctl get cluster
2018-10-10T16:49:27+01:00 [✖]  checking AWS CloudFormation access – cannot list stacks: MissingRegion: could not find region configuration
 [1] >>
```

Issue #206, opened 2018-09-12
Stricter linting


errordeveloper:
As of #193, we have `make lint`, but it's in errors-only mode and fairly light. Would be good to review and make it stricter.

https://github.com/weaveworks/eksctl/blob/584e85ad3bec1daf73165ed748639fb6f8023a29/.gometalinter.json#L1-L6

PR #246, opened 2018-10-10
Changed linter and lint fixes

richardcase:
### Description
Changed the linter from gometalinter to [GolangCI-Lint](
https://github.com/golangci/golangci-lint) as it is faster
and the configuration and output are better.

Fixed issues highlighted by the new linter. The following
linters have been disabled and should be enabled in the future:

- gocyclo
- lll
- gosec

This is further to work already carried out in #206

### Checklist


errordeveloper:
CI seems unhappy - any ideas? IIRC, we have CGO disabled, so is that why we are getting some CGO weirdness there?

richardcase:
@errordeveloper - yeah i thought from the message it may bbe because CGO is disabbled. I'm going to have a look this morning.

errordeveloper:
We have `CGO_ENABLED=0`, but only when build releases:
```
.goreleaser.yml:      - CGO_ENABLED=0
```

errordeveloper:
I've enabled GolangCI, let's see how it works. But for now, I'd just stick to what we have, I don't think gometalinter is that slow. If we could just make GolangCI do its job, it'd be fine by me to use it along the side with current CircleCI setup.

errordeveloper:
Here is an idea: let's merge this changes, but without enabling new linter in CircleCI =)

errordeveloper:
I like this, now we have some CI redundancy and GolangCI reported before CircleCI jobs finished, because it doesn't run unit tests.

richardcase:
Nice, thats an even better idea

PR #191, opened 2018-09-06
WIP: Nodegroup Scaling

richardcase:
### Description
This is currently work in progress and is dependent on a [change to goformation](https://github.com/awslabs/goformation/pull/103) so that ASG can be correctly unmarshalled.

It builds succesfully but it does not work yet.

Issue #116 

### Checklist
- [x] Code compiles correctly (i.e `make build`)
- [ ] Added tests that cover your change (if possible)
- [ ] All tests passing (i.e. `make test`)
- [ ] Added/modified documentation as required (such as the README)
- [ ] Added yourself to the `humans.txt` file

Issue #116, opened 2018-07-12
Allow ASG size modification from eksctl


stevesloka:
It would be helpful to have `eksctl` allow for updating the desired node count for a cluster. I have some dev workflows where I want to scale up my node counts and then scale back.

Command example: 
```
$ eksctl edit cluster --nodes=5
```
Some tricky things that could happen are if the user wants to scale beyond the `max` defined in the ASG. Should `eksctl` return error or update the ASG to allow more instances? 

I'm happy to work through a design and help implement some features if desired.

errordeveloper:
Thanks, this is definetly very useful feature we should add soon! I was thinking we need to add subcommand for nodegroups anyway, and this would fit nicely as `eksctl scale nodegroup --cluster=<name> --nodes=5 [<nodeGroupName>]` (inspired by `kubectl scale deployment`, of course).

richardcase:
Looking at this now.

errordeveloper:
Closed via #254.

PR #254, opened 2018-10-13
Nodegroup scaling 

richardcase:
### Description
Initial version of nodegroup scaling has been added. This scales
by modifying the CloudFormation template for the nodegroup. The
modified template is used to create a changeset that is then
executed.

When scaling down/in (i.e. reducing the number of nodes) we rely
solely on the the resulting change to the ASG. This means that
the node(s) that are to be terminated aren't drained and so pods
running on the terminating nodes may cause errors. In the future
we may consider picking the EC2 instances to be terminated and then
drain the nodes and then create a termination policy to ensure those
nodes are killed.

This supercedes #191 and it relates to issue #116.

Todo:
- [x] Cloudformation template modification
- [x] Changeset implementation
- [x] Workload testing whilst scaling up
- [x] Workload testing whilst scaling down  

### Checklist


cmd/eksctl/scale.go
```diff
@@ -0,0 +1,78 @@
...
+	cmd.AddCommand(scaleNodeGroupCmd())
+
+	return cmd
+}
+
+func scaleNodeGroupCmd() *cobra.Command {
+	cfg := &api.ClusterConfig{}
+
+	cmd := &cobra.Command{
+		Use:   "nodegroup",
```
errordeveloper: @richardcase would you be interested to add get/create/delete commands also #257? Right now, get will be most important, because user may wish to know how many nodes there are in a given cluster before they scale. They could use kubectl, but we don't have an obvious link from the cluster that kubectl is pointed at, to eksclt, perhaps something to think about (I've opened #258 to continue that discussion).

pkg/cfn/manager/waiters.go
```diff
@@ -107,6 +115,43 @@ func (c *StackCollection) waitWithAcceptors(i *Stack, acceptors []request.Waiter
...
+		},
+	}
+	logger.Debug("start %s", msg)
+	if waitErr := w.WaitWithContext(ctx); waitErr != nil {
+		s, err := c.describeStackChangeset(i, changesetName)
+		if err != nil {
+			logger.Debug("describeChangesetErr=%v", err)
+		} else {
+			logger.Critical("unexpected status %q while %s", *s.Status, msg)
+			c.troubleshootStackFailureCause(i, desiredStatus)
```
errordeveloper: Ok, so this is being used in wait-for-update code path, but there is still a chance of error when creating a changeset that we don't have a way to troubleshoot like this just yet?

cmd/eksctl/scale.go
```diff
@@ -0,0 +1,78 @@
...
+				logger.Critical("%s\n", err.Error())
+				os.Exit(1)
+			}
+		},
+	}
+
+	fs := cmd.Flags()
+
+	fs.StringVarP(&cfg.ClusterName, "name", "n", "", "EKS cluster name")
+	fs.IntVarP(&cfg.Nodes, "nodes", "N", 0, "total number of nodes (scale to this number)")
```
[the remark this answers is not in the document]
richardcase: We do have the following test later to cover this:
```
if cfg.Nodes < 1 {
		return fmt.Errorf("number of nodes must be greater than 0. Use the --nodes/-N flag")
	}
```

cmd/eksctl/scale.go
richardcase: Sure, i'd be happy to do #257 as well. Would you want that before we ship scale?

pkg/cfn/manager/waiters.go
richardcase: Yes there is a chance that changeset cannot be created. The changeset doesn't have events as far as i can see. We could take out the `troubleshootStackFailureCause` and actually replace with this from describing the changeset:
```
	// A description of the change set's status. For example, if your attempt to
	// create a change set failed, AWS CloudFormation shows the error message.
	StatusReason *string `type:"string"`
```
Do you think that would be better?

pkg/cfn/manager/api.go
```diff
@@ -104,6 +108,30 @@ func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, pa
...
+	if err != nil {
+		return err
+	}
+	changeset, err := c.describeStackChangeset(i, &changesetName)
+	if err != nil {
+		return err
+	}
+	logger.Debug("changes = %#v", changeset.Changes)
+	if err := c.doExecuteChangeset(stackName, changesetName); err != nil {
+		logger.Warning("error executing Cloudformation changeset %s in stack %s. Check the Cloudformation console for further details", changesetName, stackName)
```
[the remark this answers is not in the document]
richardcase: 'existing event tracer' - is this in relation to the addition of `waitWithAcceptorsChangeSet` and instead could we re-use `waitWithAcceptors`? If thats the case then we could generalise these down into 1 function but i felt that this would then make the the `doWaitUntil*` functions more complicated. Happy to generlise this doen to 1 function if you think its best?

errordeveloper:
No, I am happy to ship scale first, we will provide docs and it should be
already pretty useful to many people!

errordeveloper:
Scaling to zero is legit use-case actually (basically a way to save money),
we shouldn't prevent it as such. I just think there should be no default
value, that's all.

richardcase:
> Scaling to zero is legit use-case actually (basically a way to save money), we shouldn't prevent it as such. I just think there should be no default value, that's all.

Good point. I'll change that.

cmd/eksctl/scale.go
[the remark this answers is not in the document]
richardcase: I'm changign this to default to -1 and then updating the test to check for 0 or greater.

errordeveloper:
@richardcase so annoying GitHub doesn't attach email reply to the thread... sorry about that, I was hoping they fixed it.

cmd/eksctl/scale.go
errordeveloper: No, I am happy to ship scale first, we will provide docs and it should be
already pretty useful to many people!

pkg/cfn/manager/api.go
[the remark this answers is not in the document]
errordeveloper: I am thinking of code that was added in #202. Namely, can we do anything before suggesting users to go to CloudFormation console (as you know, it doesn't exactly tell you what the issue is at glance, so it's not the most helpful). I'll read through the code once again, perhaps it's already hits `c.troubleshootStackFailureCause` by this point...

pkg/cfn/manager/waiters.go
richardcase: I think `troubleshootStackFailureCause` makes sense for the execution of the changeset, as this is basically like `create cluster` except that its success events are different.

But for the creation of the changeset then it doesn't fit really, so i've added code to log the reason returned

pkg/cfn/manager/api.go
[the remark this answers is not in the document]
richardcase: The following `doWaitUntilStackIsUpdated` will hit `c.troubleshootStackFailureCause` but `c.doExecuteChangeset` won't. This error handler should only be hit if we couldn't create the changeset (whether thats a validation error with the request or with communicating with the AWS endpoint).

cmd/eksctl/scale.go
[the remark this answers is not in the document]
errordeveloper: I thought there was a way to have flags without default values, but seems like I was wrong in thinking that...

pkg/cfn/manager/waiters.go
errordeveloper: sgtm

pkg/cfn/manager/waiters.go
```diff
@@ -143,8 +143,7 @@ func (c *StackCollection) waitWithAcceptorsChangeSet(i *Stack, changesetName *st
 		if err != nil {
 			logger.Debug("describeChangesetErr=%v", err)
 		} else {
-			logger.Critical("unexpected status %q while %s", *s.Status, msg)
-			c.troubleshootStackFailureCause(i, desiredStatus)
```
errordeveloper: So you removed `troubleshootStackFailureCause` after all, or was it unintentional?

richardcase: That was intentional for just the changeset creation, but i added logging the reason. Its still in there for actually eexecuting the changeset (which does the actual stack update).

errordeveloper: 👍

pkg/cfn/manager/nodegroup.go
```diff
@@ -1,11 +1,23 @@
...
 	"github.com/weaveworks/eksctl/pkg/cfn/builder"
 
+	cfn "github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/kubicorn/kubicorn/pkg/logger"
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
+)
+
+const (
+	desirecCapacityPath = "Resources.NodeGroup.Properties.DesiredCapacity"
```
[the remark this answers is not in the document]
richardcase: Good spot, thanks. Just corrected it.

pkg/cfn/manager/api.go
```diff
@@ -104,6 +108,31 @@ func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, pa
...
+		return err
+	}
+	return c.doWaitUntilStackIsUpdated(i)
```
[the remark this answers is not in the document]
errordeveloper: This should be pretty quick normally, as far CloudFormation update is concerned.

I just ran a test and this is the sequence:

```
2018-10-16T16:28:32+01:00 [▶]  start waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:28:32+01:00 [▶]  waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:28:50+01:00 [▶]  waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:29:10+01:00 [▶]  waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:29:29+01:00 [▶]  waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:29:45+01:00 [▶]  waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
2018-10-16T16:29:46+01:00 [▶]  done after 1m13.220525803s of waiting for CloudFormation stack "eksctl-attractive-wardrobe-1539702643-nodegroup-0" to reach "UPDATE_COMPLETE" status
```

So it's around 1m for the actual update, which is okay by me. We may wish to add another stage where we'd wait for nodes to register, which could be a subject to `--wait` flag – what do you think?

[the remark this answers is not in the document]
richardcase: Good call @Lazyshot.....especially in light of your recenet change #262 👍

richardcase:
> We may want to go ahead and add integration tests for this:
> Scale to 2 nodes, check for kubernetes nodes via kube api, scale back down to 1 (no wait?)

That would be very nice. I'll also add this.

pkg/cfn/manager/api.go
```diff
@@ -220,3 +249,61 @@ func (c *StackCollection) DescribeStackEvents(i *Stack) ([]*cloudformation.Stack
 	}
 	return resp.StackEvents, nil
 }
+
+func (c *StackCollection) doCreateChangesetRequest(i *Stack, action string, description string, templateBody []byte,
+	parameters map[string]string, withIAM bool) (string, error) {
+
+	changesetName := fmt.Sprintf("eksctl-%s-%d", action, time.Now().Unix())
```
errordeveloper: Is there max length to this? It doesn't appear to get namespaced:

```
arn:aws:cloudformation:eu-west-1:123456789:changeSet/eksctl-scale-nodegroup-1539703695/0d6c1284-4d0d-44d8-9827-d1eaff527d4a
```

It'd be better if prefixed it with stack name, e.g. `eksctl-attractive-wardrobe-1539702643-nodegroup-0-scale-<timestamp>`.

pkg/cfn/manager/api.go
```diff
@@ -220,3 +249,61 @@ func (c *StackCollection) DescribeStackEvents(i *Stack) ([]*cloudformation.Stack
...
+	}
+	logger.Debug("creating changeset, input = %#v", input)
+	s, err := c.cfn.CreateChangeSet(input)
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("creating changest %q for stack %q", changesetName, *i.StackName))
+	}
+	logger.Debug("changeset = %#v", s)
+	return changesetName, nil
+}
+func (c *StackCollection) doExecuteChangeset(stackName string, changesetName string) error {
```
errordeveloper: nit: I'd keep naming consisten with the SDK, i.e. `ChangeSet` and `changeSet`

pkg/cfn/manager/api.go
```diff
@@ -220,3 +249,61 @@ func (c *StackCollection) DescribeStackEvents(i *Stack) ([]*cloudformation.Stack
...
+		input.Parameters = append(input.Parameters, p)
+	}
+	logger.Debug("creating changeset, input = %#v", input)
+	s, err := c.cfn.CreateChangeSet(input)
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("creating changest %q for stack %q", changesetName, *i.StackName))
+	}
+	logger.Debug("changeset = %#v", s)
+	return changesetName, nil
+}
```
errordeveloper: nit: new line

pkg/cfn/manager/api.go
richardcase: Sure, we can do that

richardcase: I'll add and push a new change tomorrow

pkg/cfn/manager/waiters.go
```diff
@@ -107,6 +115,42 @@ func (c *StackCollection) waitWithAcceptors(i *Stack, acceptors []request.Waiter
...
+			}
+			logger.Debug(msg)
+			req, _ := c.cfn.DescribeChangeSetRequest(input)
+			req.SetContext(ctx)
+			return req, nil
+		},
+	}
+	logger.Debug("start %s", msg)
+	if waitErr := w.WaitWithContext(ctx); waitErr != nil {
+		s, err := c.describeStackChangeset(i, changesetName)
```
golangcibot: invalid operation: c (variable of type *StackCollection) has no field or method describeStackChangeset

Lazyshot:
We can definitely have the integration tests in another PR

richardcase:
Rebased but build is failing.

errordeveloper:
@richardcase I've re-triggered, looks like it could be a flake...

richardcase:
Thanks @errordeveloper. Coudl you approve again when you get time and i'll merge in.

richardcase:
I've created #267 to make sure we don't forget to add the integration test.

Issue #213, opened 2018-09-15
--watch flag for 'cluster delete'


paulbsch:
**Why do you want this feature?**
It would be helpful to be able see/monitor the results of 'cluster delete' on the command line.

**What feature/behavior/change do you want?**
Currently, 'cluster delete' immediately exits after sending the delete commands upstream.  I'm proposing a --watch flag that could be set to true/false.  Then eksctl could be directed to watch/monitor the cluster deletion until it finishes/fails.

Currently, the default behavior on cluster creation is essentially "--watch=true".  It might also make sense to make "--watch=true" the default on delete?

richardcase:
@paulbsch - sounds like a reasonable idea for delete. I wonder of `--wait` may be better?

paulbsch:
@richardcase I'm indifferent as to what the flag is called.  Whatever works.

Lazyshot:
`--wait` was added in #262

PR #262, opened 2018-10-15
Add wait flag for delete cluster command

Lazyshot:
Adding ability for a user to have the delete cluster command
wait until all resources are removed before exiting.

Issue #213

Note: Reopened due to issue with NodeGroup stacks must be fully deleted before the cluster stack deletion can be requested.

### Checklist

### DIFF
--- a/cmd/eksctl/delete.go
+++ b/cmd/eksctl/delete.go
@@ -14,6 +14,10 @@ import (
 	"github.com/weaveworks/eksctl/pkg/utils/kubeconfig"
 )
 
+var (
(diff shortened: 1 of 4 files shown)


errordeveloper:
> It makes sense enforcing the wait for the node group to delete.

This is required because nodegroup stack imports values from cluster stack, which implies [cluster stack cannot be deleted](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html) while there exists a stack that imports any of the exported values.

Issue #257, opened 2018-10-15
nodegroup as a resource


errordeveloper:
**Why do you want this feature?**

A cluster may have one or more nodegroups, we need a way for user to see nodegroups and basic info about them (size, OS family and AMI ID, creation date, other EC2 info), get list of nodes within a group, create and/or delete a group. This will also pave the way for #100.

Also, as we are adding `eksclt scale nodegroup`, we will need a way for the user to find what's the size of a nodegroup they wish to scale.

**What feature/behavior/change do you want?**

- `eksctl get nodegroup`
- `eksctl create nodegroup`
- `eksctl delete nodegroup`

Flags and features TBD.

richardcase:
@errordeveloper - i'll happily work on this as a follow up to `scale nodegroup`

richardcase:
This will also include getting a list of changes to a node group when doing the `get`

Raffo:
Is this completed by #272?

snstanton:
As part of this, will you be able to add labels to the nodes in each group?  We currently have a cluster with multiple nodegroups and we map work to the appropriate group by assigning different labels to the nodes in each group.  I'm currently editing the cloudformation templates after initial creation, but it would be great to be able to specify a list of labels to apply in each group.

errordeveloper:
Yes, that would be certainly possible. In your current setup, do you set
--node-labels kubelet flag as well as EC2 tags?

richardcase:
Good, idea i will add it to the list for the PR

snstanton:
Yes, we currently edit the UserData to add something like: `--node-labels=cluster=system` to the `ExecStart` command.

kylemclaren:
I know this is being worked on but thought I'd chime in and say it would be a really useful feature for me. So I could, for example, have a nodegroup of memory-optimized instances with a label that allows me to only run rabbitmq pods on those nodes.

mumoshu:
I believe #332 is ready to be reviewed.

errordeveloper:
Not sure why this was still open, closing now.

PR #272, opened 2018-10-17
Initial command restructure

richardcase:
### Description
Prior to adding 'create/get/delete nodegroup' (#257) some refactoring
has been done on the commands. We had a single file for all
the commands for that verb so this has been changed to separate
folders/files.

The version command has been left where it for simplicity with the code gen as its already in main.

The new package name has been called `pkg/ctl' based on feedback.

### Checklist


errordeveloper:
🎉

PR #273, opened 2018-10-18
Improve AMI generator

errordeveloper:
### Description

I've used jenifer in the `--max-pods` code, and it seems to work nicely (although a little less intuitive then templates). One advantage is that it's all Go code, and it does `gofmt` for you also.

### Checklist


pkg/ami/generate/main.go
```diff
@@ -1,131 +0,0 @@
...
-		})
-
-		amiImages.Gpu = append(amiImages.Gpu, imageDetail{
-			ImageId: gpu,
-			Region:  region,
-		})
-	}
-
-	// Sort on region
-	sort.Slice(amiImages.General, func(i, j int) bool {
```
errordeveloper: I forgot, do does `gofmt` sort map keys or not? If it doesn't, it's nice, otherwise we might have to re-introduce slices...

pkg/ami/static_resolver_ami.go
```diff
@@ -1,21 +1,18 @@
+package ami
 
-// Code generated by go generate; DO NOT EDIT.
+// Generated code
```
golangcibot: comment on exported var StaticImages should be of the form "StaticImages ..."

pkg/ami/static_resolver_ami_generate.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	"github.com/weaveworks/eksctl/pkg/ami"
+	"github.com/weaveworks/eksctl/pkg/eks/api"
+
+	. "github.com/dave/jennifer/jen"
```
Lazyshot: Dot imports are pretty anti-pattern and I don't see any obvious reasons for it. The only general exception is in test code.

https://github.com/golang/go/wiki/CodeReviewComments#import-dot

errordeveloper:
Yeah, I just copied it from example code and felt like it seemed prettier
for this rather verbose library. And it is only a generator, so no big
deal...

errordeveloper:
@richardcase as you authored the original version, would you mind to give this a pass also?

richardcase:
@errordeveloper - i'll have a look now

pkg/ami/static_resolver_ami.go
```diff
@@ -1,21 +1,18 @@
+package ami
```
richardcase: I think we should make sure in this generated file that we keep the comment header that indicates its a generated file as per convention:

`// Code generated by go generate; DO NOT EDIT.`

errordeveloper: Yes, good idea, thanks!

errordeveloper:
Rebased and updated.

errordeveloper:
Due to GitHub outage, latest commits are not showing up yet...

Issue #42, opened 2018-06-06
Can use existing vpc and subnets


tigerwings:
need the capability to specify vpc and subnets in create cluster command to reuse the existing direct connect vpc.

errordeveloper:
I think there will be two major modes here:
1. use a preconfigured VPC, but create new subnets (easiest for the user who is simply short of VPCs)
2. use a preconfigured VPC and subnets (someone who wants certain kind of configures, e.g. as  mention above - with direct connect was; or more generally someone who has to control networking separately)

I think it must be possible to go with first option and and instruct the user to configure routes and additional security features separately.

errordeveloper:
@tigerwings I have a specific question to you personally – can you use VPC peering to connect your EKS VPC with an existing VPC and enable direct connect that way?

tigerwings:
@errordeveloper , vpc peering is not a valid solution. The main reason to use DC VPCs is that we need to access our internal service from DC VPCs. On the other hand, we may setup and tear down an EKS cluster often, the peering ops have too much overheads.

archisgore:
I'm working on this issue as we speak (if nobody else already is). This is a scenario I need for myself in prod. :-)

In I only found eksctl because I was about to write a shell script to rinse-repeat launch clusters. So I'll get around to this very soon.

baank:
Any status update on this ? We can look into picking this up also.

If you are wondering of a use case here's one. In many large enterprises e.g. ours we centrally control the core networking e.g. VPCs but then freely allow various teams to pick and choose what services they want to use. We are wanting to use eksctl but are strictly allowed to create VPCs.

errordeveloper:
@baank thanks for describing your use-case, it's very helpful! Could you clarify if you havr to use pre-existing subnets also, or it would be okay for eksctl to create dedicated subnets?

baank:
@errordeveloper .. No we are not allowed to touch the subnets either.

errordeveloper:
@baank so you would need two subnets, in your case, would you be able to use get two subnets that are dedicated for use with EKS or you would have use subnets that are also used for other things? If there would be other things, are those gonna be any AWS workloads or other Kubernetes clusters?

What size of subnets would you be able to provide? Here is what we currently use with the dedicated VPC (based on standard CloudFormation template provided):
```
VpcBlock: 192.168.0.0/16
Subnet01Block: 192.168.64.0/18
Subnet02Block: 192.168.128.0/18
```
[...]

PR #277, opened 2018-10-19
Add ability to create cluster in existing kops-owned VPC

### LINKED ISSUE #50: Scrape and convert existing Kops cluster api for subnets, security groups, instance groups
I'm not sure if this is out of scope of this project, but I'd love to be able to convert our existing kops clusters to EKS without the manual work.

Thanks!

errordeveloper:
### Description

This will allow kops users to create an EKS cluster in
the same VPC where their kops cluster is. The cluster must
be in one of EKS regions, and it must have 3 subnets in
different AZs.

This fixes #50 and paves the way for #42 as well.

### Checklist


pkg/kops/kops.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/weaveworks/eksctl/pkg/eks/api"
+
+	"k8s.io/kops/pkg/resources/aws"
+	"k8s.io/kops/upup/pkg/fi/cloudup/awsup"
+
+	"github.com/kubicorn/kubicorn/pkg/logger"
+)
+
+type KopsWrapper struct {
```
golangcibot: type name will be used as kops.KopsWrapper by other packages, and that stutters; consider calling this Wrapper

errordeveloper:
@StevenACoffman please take a look and let us know if this solves your needs!

StevenACoffman:
I'm very excited! It looks like `--vpc-from-kops-cluster` argument specifies a `kopsClusterName` to lookup all the vpcs and subnets from an account, find those existing that match the tags that identify that Kops cluster, and re-use that vpc and subnets.

Nice!

errordeveloper:
@StevenACoffman yes, that's exactly what we have here. I'm hoping we can review and merge soon enough.

errordeveloper:
@StevenACoffman yes, that's exactly what we have here. I'm hoping we can review and merge soon enough, and then a release will be due.

errordeveloper:
@StevenACoffman yes, that's exactly what we have here. I'm hoping we can review and merge soon enough, and then a release will be due.

pkg/kops/kops.go
```diff
@@ -0,0 +1,76 @@
+package kops
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+
```
golangcibot: File is not `goimports`-ed

errordeveloper: This is very annoying.

pkg/kops/kops.go
```diff
@@ -0,0 +1,78 @@
+package kops
+
+import (
+	"fmt"
+
+	"github.com/weaveworks/eksctl/pkg/eks/api"
```
golangcibot: File is not `goimports`-ed

errordeveloper: I am rather very confused.

pkg/ctl/create/cluster.go
```diff
@@ -114,8 +119,24 @@ func doCreateCluster(cfg *api.ClusterConfig, name string) error {
...
+	if kopsClusterNameForVPC != "" {
+		if len(availabilityZones) != 0 {
+			return fmt.Errorf("--vpc-from-kops-cluster and --zones cannot be used at the same time")
+		}
+		kw, err := kops.NewWrapper(cfg, kopsClusterNameForVPC)
+		if err != nil {
+			return err
+		}
+
+		if err := kw.UseVPC(); err != nil {
```
Lazyshot: It's not clear that this is altering the cluster config.

I think it would be more clear if we returned the altered version of the ClusterConfig and set it directly to the `ctl.Spec` field.

errordeveloper: I agree it's not explicit enough. It just seemed very convenient to use `cfg` once and create a pointer to it inside the wrapper struct (less code). See my next comment.

pkg/kops/kops.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Wrapper for interacting with a kops cluster
+type Wrapper struct {
+	clusterName string
+	cloud       awsup.AWSCloud
+	spec        *api.ClusterConfig
+}
+
+// NewWrapper constructs a kops wrapper for a given EKS cluster config and kops cluster
+func NewWrapper(cfg *api.ClusterConfig, kopsClusterName string) (*Wrapper, error) {
```
errordeveloper:
```
NewWrapper(region string, kopsClusterName string)
```

pkg/kops/kops.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return &Wrapper{kopsClusterName, cloud, cfg}, nil
+}
+
+func (k *Wrapper) isOwned(t *ec2.Tag) bool {
+	return *t.Key == "kubernetes.io/cluster/"+k.clusterName && *t.Value == "owned"
+}
+
+// UseVPC finds VPC and subnets that give kops cluster uses and add those to EKS cluster config
+func (k *Wrapper) UseVPC() error {
```
errordeveloper:
```
UseVPC(spec api.ClusterConfig)
```

errordeveloper: I guess a nice alternative would be `GetVPC() *api.ClusterVPC`, which would lead to some refactoring, but that maybe a wise idea anyway, we'd probably want to group some related fields inside of `api.ClusterConfig` anyway (most notably nodegroup stuff).

errordeveloper:
@Lazyshot how do you find proposed API changes?

Lazyshot:
Yeah, I think that's better. 

I'm not sure if we want to include the refactoring of the cluster config spec in this PR or not.

errordeveloper:
@Lazyshot I've updated the API, would you mind to take another look please?

errordeveloper:
I've punted discussion to another issue, this PR seems okay to merge as it doesn't get in the way of current critical code path, we can discuss in #282.

Issue #232, opened 2018-09-29
Add option to launch all nodes into the same AZ


armandmcqueen:
**Why do you want this feature?**
We are running ML training workloads on EKS that are very sensitive to network latency. For performance reasons, we need our nodes to exist in a single AZ. Using `--zones` with a single AZ raises an error.

**What feature/behavior/change do you want?**
When launching a new cluster, I would like to be able to pass in a single AZ with the `--zones` option or to have a `--single-az` flag that forces nodes to launch into the same AZ. 

Any workflow that is less unwieldy than manually altering the AutoScaling group's subnet list would be helpful.

gruebel:
Hi,
This is not possible due to API restrictions of AWS. You need to specify at least two Subnets in two different AZs.

errordeveloper:
@armandmcqueen I'm curious to know what happens if you pass e.g. `--zones us-west-2a,us-west-2a,us-west-2a`?

@gruebel is that an EKS API requirement, or something else?

I think in the future, we should be able to allow this to happen for individual node-groups, perhaps `--single-az` would makes sense in that context. It'd be nice to allow users to have single-AZ nodegroups at the same time as nodegroups that span multiple AZs.

errordeveloper:
We will be able to do this very easily soon (namely after #257).

errordeveloper:
@richardcase as you are working on #281, there's now a `NodeGroup.AvailibilityZones` field that can be leveraged for this, and I think it would be sufficient to just expose it as a flag as there is plumbing already in place.

https://github.com/weaveworks/eksctl/blob/99b5d58d340660ba7299ee470888b022754a2d8e/pkg/cfn/builder/nodegroup.go#L113-L122

errordeveloper:
This is definitely done now, and we have ironed out a bug in this week - #432.

dnascimento:
Does it assume only one subnet per AZ? Can we deploy on multiple subnets in the same AZ?

errordeveloper:
No, it is single subnet per AZ. If you have a particular use-case, please
open an issue with more details of that use-case.

PR #281, opened 2018-10-24
WIP:  Nodegroup as a resource

richardcase:
### Description
This is WIP for adding the following:
- eksctl get nodegroup
- eksctl create nodegroup
- eksctl delete nodegroup

### Tasks to complete
- [x] Create `get` command
- [x] Test `get` command
- [ ] Create `create` command
- [ ] Test `create` command
- [ ] Create `delete` command
- [ ] Test `delete` command
- [ ] Update `delete cluster` command
- [ ] Update `scale nodegroup` command
- [ ] Add support for node labels
- [ ] Add integration tests to cover nodegroup lifecycle
- [ ] Update documentation

### Checklist

### DIFF
--- a/pkg/cfn/manager/nodegroup.go
+++ b/pkg/cfn/manager/nodegroup.go
@@ -3,8 +3,9 @@ package manager
 import (
 	"bytes"
 	"fmt"
+	"strconv"
+	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/pkg/errors"
 	"github.com/weaveworks/eksctl/pkg/cfn/builder"
 	"github.com/weaveworks/eksctl/pkg/eks/api"
@@ -19,7 +20,21 @@ const (
 	desiredCapacityPath = "Resources.NodeGroup.Properties.DesiredCapacity"
 	maxSizePath         = "Resources.NodeGroup.Properties.MaxSize"
 	minSizePath         = "Resources.NodeGroup.Properties.MinSize"
+	instanceTypePath    = "Resources.NodeLaunchConfig.Properties.InstanceType"
+	imageIDPath         = "Resources.NodeLaunchConfig.Properties.ImageId"
 )
+
+// NodeGroupSummary represents a summary of a nodegroup stack
+type NodeGroupSummary struct {
+	Seq             int
+	StackName       string
+	MaxSize         int
+	MinSize         int
+	DesiredCapacity int
+	InstanceType    string
+	ImageID         string
+	CreationTime    *time.Time
+}
 
 func (c *StackCollection) makeNodeGroupStackName(id int) string {
 	return fmt.Sprintf("eksctl-%s-nodegroup-%d", c.spec.ClusterName, id)
@@ -89,7 +104,7 @@ func (c *StackCollection) ScaleNodeGroup(id int) error {
 	logger.Info("scaling nodegroup stack %q in cluster %s", name, clusterName)
 
 	// Get current stack
-	template, err := c.getStackTemplate(name)
+	template, err := c.GetStackTemplate(name)
 	if err != nil {
 		return errors.Wrapf(err, "error getting stack template %s", name)
 	}
@@ -137,15 +152,70 @@ func (c *StackCollection) ScaleNodeGroup(id int) error {
 	return c.UpdateStack(name, "scale-nodegroup", descriptionBuffer.String(), []byte(template), nil)
 }
 
-func (c *StackCollection) getStackTemplate(stackName string) (string, error) {
-	input := &cfn.GetTemplateInput{
-		StackName: aws.String(stackName),
-	}
-
-	output, err := c.cfn.GetTemplate(input)
+// GetNodeGroupSummaries returns a list of summaries for the nodegroups of a cluster
+func (c *StackCollection) GetNodeGroupSummaries() ([]*NodeGroupSummary, error) {
+	stacks, err := c.ListStacks(fmt.Sprintf("^(eksctl|EKS)-%s-nodegroup-\\d+$", c.spec.ClusterName), cfn.StackStatusCreateComplete)
 	if err != nil {
-		return "", err
+		return nil, errors.Wrap(err, "getting nodegroup stacks")
 	}
 
-	return *output.TemplateBody, nil
+	summaries := []*NodeGroupSummary{}
+	for _, stack := range stacks {
+		logger.Info("stack %s\n", *stack.StackName)
+		logger.Debug("stack = %#v", stack)
+
+		template, err := c.GetStackTemplate(*stack.StackName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error getting Cloudformation template for stack %s", *stack.StackName)
+		}
+
+		//TODO: create a map function
+		seq := getNodeGroupID(stack.Tags)
+		maxSize := gjson.Get(template, maxSizePath)
+		minSize := gjson.Get(template, minSizePath)
+		desired := gjson.Get(template, desiredCapacityPath)
+		instanceType := gjson.Get(template, instanceTypePath)
+		imageID := gjson.Get(template, imageIDPath)
+
+		summary := &NodeGroupSummary{
+			Seq:             seq,
+			StackName:       *stack.StackName,
+			MaxSize:         int(maxSize.Int()),
+			MinSize:         int(minSize.Int()),
+			DesiredCapacity: int(desired.Int()),
+			InstanceType:    instanceType.String(),
+			ImageID:         imageID.String(),
+			CreationTime:    stack.CreationTime,
+		}
+
+		summaries = append(summaries, summary)
+	}
+
+	return summaries, nil
+}
+
+// GetMaxNodeGroupSeq returns the sequence number og the highest node group
+func (c *StackCollection) GetMaxNodeGroupSeq() (int, error) {
+	stacks, err := c.ListStacks(fmt.Sprintf("^(eksctl|EKS)-%s-nodegroup-\\d+$", c.spec.ClusterName))
+	if err != nil {
+		return -1, errors.Wrap(err, "getting nodegroup stacks")
+	}
+	seq := -1
+	for _, stack := range stacks {
+		stackSeq := getNodeGroupID(stack.Tags)
+		if stackSeq > seq {
+			seq = stackSeq
+		}
+	}
+	return seq, nil
+}
+
+func getNodeGroupID(tags []*cfn.Tag) int {
+	for _, tag := range tags {
+		if *tag.Key == NodeGroupIDTag {
+			i, _ := strconv.Atoi(*tag.Value)
+			return i
+		}
+	}
+	return -1
 }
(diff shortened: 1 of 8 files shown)


pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,89 @@
...
+	"os"
+
+	"github.com/kubicorn/kubicorn/pkg/logger"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/eksctl/pkg/ctl"
+	"github.com/weaveworks/eksctl/pkg/eks"
+	"github.com/weaveworks/eksctl/pkg/eks/api"
+	"github.com/weaveworks/eksctl/pkg/utils"
+
+	awseks "github.com/aws/aws-sdk-go/service/eks"
```
golangcibot: File is not `goimports`-ed

pkg/ctl/get/nodegroup.go
```diff
@@ -0,0 +1,108 @@
+package get
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/weaveworks/eksctl/pkg/cfn/manager"
```
golangcibot: File is not `goimports`-ed

pkg/ctl/create/nodegroup.go
errordeveloper: This one is very frustrating, I wasn't able to reproduce it locally. Perhaps we could just disable it?

Lazyshot: According to the documentation it's disabled by default.

pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,100 @@
+package create
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/weaveworks/eksctl/pkg/cfn/manager"
```
golangcibot: File is not `goimports`-ed

pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,100 @@
...
+	logger.Info("found cluster %s", eksCluster.Name)
+	logger.Debug("cluster = %#v", eksCluster)
+
+	errs := []error{}
+	appendErr := func(err error) {
+		errs = append(errs, err)
+	}
+
+	stackManager := ctl.NewStackManager()
+	if manager.Run(appendErr, stackManager.CreateNextNodeGroup); len(errs) > 0 {
```
golangcibot: empty branch

pkg/ctl/create/nodegroup.go
errordeveloper: Yes, but we have `linters: {enable-all: true}`, so I've disabled it (875e50daa8ec8f3229e63e285c86dd06b554c3ec).

errordeveloper:
@richardcase #293 is probably relevant for creation, especially in the context of #232; but neither have to be required for initial implementation (we can fix such limitations separately, if solution is not obvious).

pkg/ctl/create/cluster.go
```diff
@@ -49,45 +43,22 @@ func createClusterCmd() *cobra.Command {
...
 	fs.BoolVar(&setContext, "set-kubeconfig-context", true, "if true then current-context will be set in kubeconfig; if a context is already set then it will be overwritten")
 
-	fs.DurationVar(&cfg.WaitTimeout, "aws-api-timeout", api.DefaultWaitTimeout, "")
-	// TODO deprecate in 0.2.0
-	if err := fs.MarkHidden("aws-api-timeout"); err != nil {
-		logger.Debug("ignoring error %q", err.Error())
-	}
-	fs.DurationVar(&cfg.WaitTimeout, "timeout", api.DefaultWaitTimeout, "max wait time in any polling operations")
+	//TODO: is this needed for nodegroup creation????
+	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
```
errordeveloper: Yes, this would be needed (see #232). When unspecified it'd be inherited from the cluster stack.

pkg/ctl/create/cluster.go
```diff
@@ -49,45 +43,22 @@ func createClusterCmd() *cobra.Command {
...
-	fs.IntVarP(&ng.MinSize, "nodes-min", "m", 0, "minimum nodes in ASG")
-	fs.IntVarP(&ng.MaxSize, "nodes-max", "M", 0, "maximum nodes in ASG")
-
-	fs.IntVarP(&ng.VolumeSize, "node-volume-size", "", 0, "Node volume size (in GB)")
-	fs.IntVar(&ng.MaxPodsPerNode, "max-pods-per-node", 0, "maximum number of pods per node (set automatically if unspecified)")
-	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
-
-	fs.BoolVar(&ng.AllowSSH, "ssh-access", false, "control SSH access for nodes")
-	fs.StringVar(&ng.SSHPublicKeyPath, "ssh-public-key", defaultSSHPublicKey, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
+	addCommonCreateFlags(fs, cfg, ng)
```
errordeveloper: We might benefit from a separate PR that does some refactoring and cleanup to fix #298. I think that `--region`, `--profile` are somewhat common globally, only a few commands don't inherit those.

richardcase:
@errordeveloper - i'm still working on this. Taking me a little longer than i ever anticipated due to starting a new job.

mumoshu:
@richardcase Hey! Can I take this?

// Btw, thanks for your awesome work ☺️

richardcase:
Hi @mumoshu  - that would be great if you could take it, thanks a lot 👍  I've started a new job and struggling with time.

Do you want to work against this branch/pr or would you like to start clean?

mumoshu:
@richardcase Thanks for the response!

I'd rather keep this PR as-is for reference. But I do want to write a big credit/acknowledgement on your work in my PR.

Would it sounds ok to you?

pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,128 @@
...
+	logger.Debug("cluster = %#v", eksCluster)
+
+	{
+		stackManager := ctl.NewStackManager()
+		maxSeq, err := stackManager.GetMaxNodeGroupSeq()
+		if err != nil {
+			return err
+		}
+		ng.ID = maxSeq + 1
+		logger.Info("will create a Cloudformation stack for nodegroup %d for cluster %s", ng.ID, cfg.ClusterName)
```
mumoshu: @richardcase This seems to result in `CreateNodeGroups` trying to create all the stacks for all the node groups, not only the stack for the specified node gorup.

So, shouldn't we change this to work against single node group per each `create nodegroup` command?

mumoshu: Probably the code would look something like `err := stackManager.CreateNodeGroup(ng)`

pkg/cfn/manager/nodegroup.go
```diff
@@ -137,15 +152,70 @@ func (c *StackCollection) ScaleNodeGroup(id int) error {
...
+		logger.Info("stack %s\n", *stack.StackName)
+		logger.Debug("stack = %#v", stack)
+
+		template, err := c.GetStackTemplate(*stack.StackName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error getting Cloudformation template for stack %s", *stack.StackName)
+		}
+
+		//TODO: create a map function
+		seq := getNodeGroupID(stack.Tags)
```
mumoshu: @richardcase Would you mind clarifying a bit more here?

I have `get nodegroup` working like this in my branch:

```
$ ./eksctl get nodegroup --cluster adorable-sheepdog-1543570021
2018-12-04T16:46:41+09:00 [ℹ]  stack eksctl-adorable-sheepdog-1543570021-nodegroup-0
STACKNAME					SEQ	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
eksctl-adorable-sheepdog-1543570021-nodegroup-0	0	2018-12-04T07:21:34Z	2		2		2			m5.large	ami-0f54a2f7d2e9c88b3
```

But I required no change around this TODO.

pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,128 @@
...
+	logger.Info("found cluster %s", eksCluster.Name)
+	logger.Debug("cluster = %#v", eksCluster)
+
+	{
+		stackManager := ctl.NewStackManager()
+		maxSeq, err := stackManager.GetMaxNodeGroupSeq()
+		if err != nil {
+			return err
+		}
+		ng.ID = maxSeq + 1
```
mumoshu: This breaks after you create a nodegroup and then delete it. We'd better use a semi-random name other than an auto-incremented index. UUID can be used for example, but I'd use the same fancy names used for cluster names.

---

Let's say you had firstly created a nodegroup with ID=0(=maxSeq + 1), the subsequent `create nodegroup` sees maxSeq as 1! Which breaks assumptions in the codebase and results in index out of bound errors.

This happens because CloudFormation DescribeStacks API returns stacks in DELETE_COMPLETE for a while(maybe several to dozen minutes). See the `StackStatus: "DELETE_COMPLETE",` part of the example API response below:

```
2018-12-04T17:05:59+09:00 [▶]  stacks = [{
  Capabilities: ["CAPABILITY_IAM"],
  CreationTime: 2018-12-04 07:21:34.572 +0000 UTC,
  DeletionTime: 2018-12-04 07:49:49.204 +0000 UTC,
  Description: "EKS nodes (AMI family: AmazonLinux2, SSH access: false, subnet topology: Public) [created and managed by eksctl]",
  DisableRollback: false,
  Outputs: [{
      ExportName: "eksctl-adorable-sheepdog-1543570021-nodegroup-0::InstanceRoleARN",
      OutputKey: "InstanceRoleARN",
      OutputValue: "arn:aws:iam::ACCOUNT:role/eksctl-adorable-sheepdog-15435700-NodeInstanceRole-9D8J0JPOZOA"
    }],
  RollbackConfiguration: {

  },
  StackId: "arn:aws:cloudformation:us-west-2:ACCOUNT:stack/eksctl-adorable-sheepdog-1543570021-nodegroup-0/3aed7b30-f795-11e8-99dc-500c337110fd",
  StackName: "eksctl-adorable-sheepdog-1543570021-nodegroup-0",
  StackStatus: "DELETE_COMPLETE",
  Tags: [{
      Key: "eksctl.cluster.k8s.io/v1alpha1/cluster-name",
      Value: "adorable-sheepdog-1543570021"
    },{
      Key: "eksctl.cluster.k8s.io/v1alpha1/nodegroup-id",
      Value: "0"
    }]
}]
```

mumoshu: Update: So I was able to fix the index out of bound errors just by replacing a bunch of `ng.ID` that are used to drill down `ClusterConfig` to a single node group inside it, to use `api.NodeGroup` instead.

I'll keep it going with `ng.ID = maxSeq + 1` for now.

Issue #290, opened 2018-10-30
lower minimum subnets required to 2


errordeveloper:
**Why do you want this feature?**

The EKS docs currently suggest that minimum number of [subnets/AZs required is 2](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html#API_CreateCluster_RequestSyntax). I believe our requirement of 3 subnets/AZs was based on what docs suggested at the time (can check in docs repo).

**What feature/behavior/change do you want?**

I think it may make sense to still default two 3 AZs in all but `us-east-1` AZs. Also kops VPC feature should be easier to use, as requirement would be lowered.

jonk:
I can take a try at this, the requirement for 3 AZs is enforced by the functions at the bottom of https://github.com/weaveworks/eksctl/blob/43771c6a224f0b8942da3bff14e8379bec255832/pkg/eks/api/vpc.go ?

errordeveloper:
Yes, but there maybe other places, but those would be a start.

Issue #252, opened 2018-10-11
handling insufficient EC2 capacity issues


arun-gupta:
Trying to create a GPU-powered cluster and it fails because `us-east-1e` does not have sufficient capacity. The CLI should make that check ahead of time.

```
$ eksctl create cluster eks-gpu --node-type=p3.8xlarge --timeout=40m
2018-10-11T14:41:33-07:00 [ℹ]  setting availability zones to [us-east-1f us-east-1e us-east-1b]
2018-10-11T14:41:33-07:00 [ℹ]  using "ami-058bfb8c236caae89" for nodes
2018-10-11T14:41:33-07:00 [ℹ]  creating EKS cluster "eks-gpu" in "us-east-1" region
2018-10-11T14:41:33-07:00 [ℹ]  will create 2 separate CloudFormation stacks for cluster itself and the initial nodegroup
2018-10-11T14:41:33-07:00 [ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=us-east-1 --name=eks-gpu'
2018-10-11T14:41:33-07:00 [ℹ]  creating cluster stack "eksctl-eks-gpu-cluster"
2018-10-11T14:42:34-07:00 [✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-eks-gpu-cluster" to reach "CREATE_COMPLETE" status
2018-10-11T14:42:34-07:00 [ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
2018-10-11T14:42:34-07:00 [ℹ]  AWS::EC2::SecurityGroup/ControlPlaneSecurityGroup: DELETE_COMPLETE
2018-10-11T14:42:34-07:00 [ℹ]  AWS::IAM::Role/ServiceRole: DELETE_COMPLETE
2018-10-11T14:42:34-07:00 [!]  AWS::IAM::Role/ServiceRole: DELETE_IN_PROGRESS
2018-10-11T14:42:34-07:00 [!]  AWS::EC2::SecurityGroup/ControlPlaneSecurityGroup: DELETE_IN_PROGRESS
2018-10-11T14:42:34-07:00 [ℹ]  AWS::IAM::Policy/PolicyNLB: DELETE_COMPLETE
2018-10-11T14:42:34-07:00 [ℹ]  AWS::EKS::Cluster/ControlPlane: DELETE_COMPLETE
[...]

PR #294, opened 2018-11-01
Lower minimum subnets and AZs to 2

jonk:
### Description
EKS docs suggest that the minimum required subnets/AZs is now 2. Also
updated tests and logging that reference the minimum subnets/AZs.

Fixes Issue #290

### Checklist

### DIFF
--- a/pkg/az/az.go
+++ b/pkg/az/az.go
@@ -8,13 +8,15 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/pkg/errors"
+
+	"github.com/weaveworks/eksctl/pkg/eks/api"
 )
 
 const (
-	// DefaultRequiredAvailabilityZones defines the default number of required availability zones
-	DefaultRequiredAvailabilityZones = 3
+	// RecommendedAvailabilityZones defines the default number of required availability zones
+	RecommendedAvailabilityZones = api.RecommendedSubnets
 	// MinRequiredAvailabilityZones defines the minimum number of required availability zones
-	MinRequiredAvailabilityZones = 2
+	MinRequiredAvailabilityZones = api.MinRequiredSubnets
 )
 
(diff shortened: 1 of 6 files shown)


pkg/az/az.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const (
 	// DefaultRequiredAvailabilityZones defines the number of required availability zones
-	DefaultRequiredAvailabilityZones = 3
+	DefaultRequiredAvailabilityZones = 2
```
errordeveloper: As I mentioned in #290, it would be best to actually keep default to 3 here, but allow 2 in certain cases (namely the `us-east-1` region). So I'd change this to:

```suggestion
	DefaultRequiredAvailabilityZones = 3
	MinRequiredAvailabilityZones = 2
```

pkg/kops/kops.go
```diff
@@ -68,7 +68,7 @@ func (k *Wrapper) UseVPC(spec *api.ClusterConfig) error {
 	}
 	logger.Debug("subnets = %#v", spec.VPC.Subnets)
 	if !spec.HasSufficientPublicSubnets() {
-		return fmt.Errorf("cannot use VPC from kops cluster with less then 3 subnets")
+		return fmt.Errorf("cannot use VPC from kops cluster with less then 2 subnets")
```
errordeveloper: 👍

pkg/az/az_test.go
```diff
@@ -70,7 +70,7 @@ var _ = Describe("AZ", func() {
 				})
 
 				It("should have returned 3 availability zones", func() {
-					Expect(len(selectedZones)).To(Equal(3))
+					Expect(len(selectedZones)).To(Equal(2))
```
errordeveloper: So I think we will basically need to expand the test to cover `DefaultRequiredAvailabilityZones`, and `MinRequiredAvailabilityZones` (as special case for us-east-1).

pkg/az/az.go
errordeveloper: For background on us-east-1, see #118 and #252.

Issue #251, opened 2018-10-11
Refactor region logic


richardcase:
**Why do you want this feature?**
The region setting logic is less than ideal at present as we are forced to re-create the session if there is no region defined by flags/aws config/aws en vars and we want to use the a default region. See #250.

**What feature/behavior/change do you want?**
I would expect us to create the session once and have the region setting logic all in one place. We could provide tests around the region setting logic so we have confidence of no regressions.

To do this in the best way we should try and use the AWS SDK as much as possible and only resort to implementing an AWS config reader ourselves as a last resort. Ideally, we should make a PR to the AWS SDK to make the [loadSharedConfig](https://github.com/aws/aws-sdk-go/blob/f26268b77ea9619d95b1d04b6f3e3c5a03b61110/aws/session/shared_config.go#L80) exported. This function is used when creating a [new session](https://github.com/aws/aws-sdk-go/blob/f26268b77ea9619d95b1d04b6f3e3c5a03b61110/aws/session/session.go#L350)

nikimanoledaki:
Closing this due to inactivity but happy to reopen this if there is still interest in this!

PR #297, opened 2018-11-01
Print out the region when listing clusters

### LINKED ISSUE #256: eksctl get clusters prints log line
As of recent changes, we no see a log line printed above tabular output:

``` $ eksctl get clusters
2018-10-15T11:05:10+01:00 [ℹ]  using region us-west-2
NAME
fabulous-outfit-1538467450
fabulous-painting-1539188916
$
```

I think we could make this prettier, perhaps region could be a column? Bonus point would be to add `--all-regions` also.

gotjosh:
### Description

Adds the region as part of the print output when listing clusters with `eksctl get cluster`. I wanted to try and take a stab at the bonus feature: Add a `--all-regions` option but noticed we have #251 which might make things easier (or harder 😆). Perhaps we should try and lay that one out first?

Happy to create a separate issue for `--all-regions` should we decide to untangle it and close #256 . 

Partially fixes #256

### Checklist

### DIFF
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -18,6 +18,11 @@ import (
 	"github.com/kubicorn/kubicorn/pkg/logger"
 )
 
+type clusterWithRegion struct {
+	Name   string
+	Region string
+}
+
 // DescribeControlPlane describes the cluster control plane
 func (c *ClusterProvider) DescribeControlPlane() (*awseks.Cluster, error) {
 	input := &awseks.DescribeClusterInput{
@@ -84,7 +89,7 @@ func (c *ClusterProvider) ListClusters(chunkSize int, output string) error {
 }
 
 func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.OutputPrinter) error {
-	allClusterNames := []*string{}
+	allClusters := []*clusterWithRegion{}
 
 	getFunc := func(chunkSize int64, nextToken string) ([]*string, *string, error) {
 		input := &awseks.ListClustersInput{MaxResults: &chunkSize}
@@ -104,7 +109,11 @@ func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.Outpu
 		if err != nil {
 			return err
 		}
-		allClusterNames = append(allClusterNames, clusters...)
+
+		for _, clusterName := range clusters {
+			allClusters = append(allClusters, &clusterWithRegion{Name: *clusterName, Region: c.Spec.Region})
+		}
+
 		if nextToken != nil && *nextToken != "" {
 			token = *nextToken
 		} else {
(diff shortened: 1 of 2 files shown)


pkg/eks/eks.go
```diff
@@ -80,11 +80,16 @@ func (c *ClusterProvider) ListClusters(chunkSize int, output string) error {
...
 
-func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.OutputPrinter) error {
-	allClusterNames := []*string{}
+type clusterWithRegion struct {
+	Name   *string
+	Region *string
+}
+
+func (c *ClusterProvider) doListClusters(region string, chunkSize int64, printer printers.OutputPrinter) error {
+	allClusterNames := []*clusterWithRegion{}
```
errordeveloper: could probably use `map[string]string`, but slice of struct is just as fine, as you might want to hang other fields on it eventually...

pkg/eks/eks.go
```diff
@@ -61,7 +61,7 @@ func (c *ClusterProvider) GetCredentials(cluster awseks.Cluster) error {
 }
 
 // ListClusters display details of all the EKS cluster in your account
-func (c *ClusterProvider) ListClusters(chunkSize int, output string) error {
+func (c *ClusterProvider) ListClusters(region string, chunkSize int, output string) error {
```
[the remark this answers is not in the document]
gotjosh: Great shout! 🎖 

Was wondering whenever we had the config within, should have verified the definition of `ClusterProvider` in the first place.

gotjosh:
Hey @errordeveloper, do you mind taking a quick look again? 

Also, two quick things: 

- Is it worth adding tests for the output? (or maybe there're somewhere, I'm just not aware of them) I noticed that our tests at `eks_test.go` are just using spies to assert on the API calls and that no error is thrown but there's nothing for actually asserting the output. I'm happy to add them if you can point me in the right direction. 

- I'll take `--all-regions` into a separate PR if you don't mind.

pkg/eks/eks.go
```diff
@@ -104,7 +104,11 @@ func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.Outpu
 		if err != nil {
 			return err
 		}
-		allClusterNames = append(allClusterNames, clusters...)
+
+		for _, clusterName := range clusters {
+			allClusterNames = append(allClusterNames, &map[string]string{"name": *clusterName, "region": c.Spec.Region})
```
errordeveloper: Oops, I think I've misguided you on this one, I'm sorry... I was assuming something which now appears completely wrong, original version with a struct was actually much better. What I was thinking is that we can have map from names to regions, but that's very very shortsighted, and we's have to have a map of regions to slices of string or something like that. Let's just go back to slice of structs, pardon me that's I've not given it enough thought originally.

gotjosh: That's OK, I'm happy to change it back. Although, I'll admit that I like the way it looks now 😆.

pkg/eks/eks.go
```diff
@@ -18,6 +18,11 @@ import (
 	"github.com/kubicorn/kubicorn/pkg/logger"
 )
 
+type clusterWithRegion struct {
+	Name string
```
golangcibot: File is not `gofmt`-ed with `-s`

gotjosh:
One last look @errordeveloper ? 🙏

Issue #278, opened 2018-10-21
CloudWatch Metrics for AWS CNI plugin


PaulMaddox:
The [AWS CNI plugin](https://github.com/aws/amazon-vpc-cni-k8s) allows automatically exporting the following metrics to CloudWatch:

- The number of errors encountered in ipamd
- The number of ipamd actions inprogress
- The maximum number of ENIs that can be attached to the instance
- The maximum number of IP addresses that can be allocated to the instance
- The number of times ipamD reconciles on ENIs and IP addresses
- The number of add IP address request
- The number of delete IP address request
- The number of ENIs allocated
- The total number of IP addresses
- The number of IP addresses assigned to pods
- AWS API call latency in ms
- The number of times AWS API returns an error
- The number of errors not handled in awsutils library

![example](https://github.com/aws/amazon-vpc-cni-k8s/raw/master/docs/images/cni-metrics-100.png)

These are super useful for EKS cluster administrators, and correct use of dashboards/alarms can save pain.

In order for this to work, the following needs to be added to eksctl:

 - The worker nodes should have `cloudwatch:PutMetricData` permissions added.
 - The following K8s manifest should be deployed: [cni_metrics_helper.yaml].(https://github.com/aws/amazon-vpc-cni-k8s/blob/master/misc/cni_metrics_helper.yaml).
 - And finally, for best experience it would be great to automatically create a CloudWatch Dashboard for each EKS cluster.

PaulMaddox:
Example CloudWatch Dashboard

<img width="1440" alt="example-cni-dashboard" src="https://user-images.githubusercontent.com/4678446/47266964-47686c00-d4f2-11e8-8881-044c0e2adde8.png">

errordeveloper:
I guess we could just enable the permission without a flag. I believe the only implications would be that someone can flood CloydWatch with metrics, which impacts the bill. Creation of dashboard would have to wait for add-ons capability.

errordeveloper:
@PaulMaddox do you agree with my assessment above?

errordeveloper:
I'll open a PR to add IAM role, and we can get it into this week's release.

PaulMaddox:
👍

errordeveloper:
https://github.com/weaveworks/eksctl/pull/296 will add the role, the rest of this feature will have to wait until add-ons are a thing (see #242).

michaelbeaumont:
This is unlikely to be prioritized.
[...]

Issue #109, opened 2018-07-10
Option to use Weave Net CNI


errordeveloper:
I've tested this today, and it worked as intended, but we need more testing before we can declare Weave Net is full supported.

From eksctl perspective, we should make Weave Net an add-on (see #17 and #18), especially because there all the extra steps required.

First, create a cluster with `eksctl create cluster`.

Next, backup and delete `kube-system:DaemonSet/aws-node`:

```
kubectl get ds --namespace=kube-system aws-node --output=yaml --export > aws-node.ds.yaml
kubectl delete ds --namespace=kube-system aws-node 
```

Now, install Weave Net:
```
curl --silent --location \
  "https://cloud.weave.works/k8s/v1.10/net" \
  --output weave-net.yaml
kubectl apply -f weave-net.yaml
```

After `kube-system:DaemonSet/weave-net` is created, all pods will need to be recycled.

The following script can be used to recycle all pods properly:
```
#!/bin/bash

set -o errexit
set -o pipefail
set -o nounset

## Wait until all Weave Net pods are running
weave_net_pods_are_ready() {
  local count
  count="$(
    kubectl get pods \
      --namespace=kube-system \
      --selector='name=weave-net' \
      --output='jsonpath={range .items[*]}{@.metadata.name}:{@.status.phase}{"\n"}{end}' \
      | grep -c -v Running
  )"
  test "${count}" -eq 0
}

## Delete all pods that were created before Weave Net CNI plugin was installed
delete_stale_pods() {
  local namespaces
  namespaces=($(kubectl get namespaces --output='jsonpath={range .items[*]}{@.metadata.name}{"\n"}{end}'))
  for ns in "${namespaces[@]}" ; do
    kubectl delete pods --namespace="${ns}" --selector='name!=weave-net'
  done
  unset ns
}

until weave_net_pods_are_ready ; do sleep 1 ; done

delete_stale_pods
```

NOTE: when we do this, we should be able to bypass `--max-pod` setting based on instance type, as it's based on ENI limits.

errordeveloper:
I was expecting access to `kubernetes` service to break, which is what happen during the tests I did earlier. However, what appears to break now is DNS. Either way, this is most likely due to that ENI network was removed. There must be a way to make this work, potentially without having to use both networks, although we might have to do that for compatibly/stability. Multiple CNI networks is not mainstream, but can be achieved from what I have heard.

micahlmartin:
[...]

PR #303, opened 2018-11-05
Design Proposal #002: Existing VPC

errordeveloper:


docs/proposal-002_existing-vpc.md
```diff
@@ -0,0 +1,67 @@
+# Desing Proposal #002: Existing VPC
+
```
errordeveloper: I should clarify the status. Having it merged doesn't mean it's final, we can refine and augment it as needed in the future.

errordeveloper:
@anodyne4j thanks a lot for your feedback!

> One question - could we run into issues deploying into existing subnets while using EKS-native CNI, with its need for many secondary IPs? What would happen if all the pods in the cluster weren't able to get IP addrs?

I am not deeply familiar with the implementation, but I am pretty certain it's meant to coexist with anything you have in your VPC, from plain EC2 instances, through load balancers to Lambda functions. So I would expect it to try and use what IP addresses are available, which is only a subject to availability of course. You can monitor allocations (and other properties of the CNI driver) via CloudWatch (#278) and/or Prometheus.

However, if you have a short subnet in your VPC, or there are many resources using up the IP address space, you will ultimately run into issue sooner than later. If that's the case, I would consider VPC peering, or something else. Peering should already be possible, and custom CIDR flag is available as of 0.1.8.

And, I must note, that exactly for this and other reasons we have plans to provide Weave Net as an overall option (weaveworks/eksctl#109, weaveworks/weave#3335). By definition overall is exactly for this purposes. I am not up to date on IPv6 support in EKS, I've not heard of it being commonly used in AWS just yet, but I'm keen to hear if anyone used it.

errordeveloper:
I've clarified that this is still a draft and changes are welcome. I'd like to merge it soon, as well as `--vpc-{private,public}-subnets` implementation.

PR #304, opened 2018-11-06
Use error instead of boolean

errordeveloper:


pkg/eks/api/vpc.go
```diff
@@ -80,21 +84,29 @@ func (c *ClusterConfig) HasSufficientPrivateSubnets() bool {
 	return len(c.SubnetIDs(SubnetTopologyPrivate)) >= MinRequiredSubnets
 }
 
+var insufficientSubnetsErr = fmt.Errorf(
```
golangcibot: error var insufficientSubnetsErr should have name of the form errFoo

PR #309, opened 2018-11-07
Clear separation between `api.ClusterConfig` and `api.ClusterProvider`

errordeveloper:
### Description

This refactors how we separate provider configuration and it's recievers
from cluster configuration. We no longer copy cluster configuration in
every places where we need it, instead we pass it explicitly every time.

### Checklist

### DIFF
--- a/pkg/cfn/builder/cluster.go
+++ b/pkg/cfn/builder/cluster.go
@@ -79,7 +79,7 @@ func (c *ClusterResourceSet) addResourcesForControlPlane(version string) {
 	}
 
 	c.newResource("ControlPlane", &gfn.AWSEKSCluster{
-		Name:               gfn.NewString(c.spec.ClusterName),
+		Name:               gfn.NewString(c.spec.Metadata.Name),
 		RoleArn:            gfn.MakeFnGetAttString("ServiceRole.Arn"),
 		Version:            gfn.NewString(version),
 		ResourcesVpcConfig: clusterVPC,
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -36,9 +36,10 @@ type ChangeSet = cloudformation.DescribeChangeSetOutput
 
 // StackCollection stores the CloudFormation stack information
 type StackCollection struct {
-	cfn  cloudformationiface.CloudFormationAPI
-	spec *api.ClusterConfig
-	tags []*cloudformation.Tag
+	cfn         cloudformationiface.CloudFormationAPI
+	waitTimeout time.Duration
+	spec        *api.ClusterConfig
+	tags        []*cloudformation.Tag
 }
 
 func newTag(key, value string) *cloudformation.Tag {
@@ -48,16 +49,17 @@ func newTag(key, value string) *cloudformation.Tag {
 // NewStackCollection create a stack manager for a single cluster
 func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) *StackCollection {
 	tags := []*cloudformation.Tag{
-		newTag(ClusterNameTag, spec.ClusterName),
+		newTag(ClusterNameTag, spec.Metadata.Name),
 	}
-	for key, value := range spec.Tags {
+	for key, value := range spec.Metadata.Tags {
 		tags = append(tags, newTag(key, value))
 	}
 	logger.Debug("tags = %#v", tags)
 	return &StackCollection{
-		cfn:  provider.CloudFormation(),
-		spec: spec,
-		tags: tags,
+		cfn:         provider.CloudFormation(),
+		waitTimeout: provider.WaitTimeout(),
+		spec:        spec,
+		tags:        tags,
 	}
 }
 
@@ -203,7 +205,7 @@ func (c *StackCollection) DeleteStack(name string) (*Stack, error) {
 	}
 	i.StackId = s.StackId
 	for _, tag := range s.Tags {
-		if *tag.Key == ClusterNameTag && *tag.Value == c.spec.ClusterName {
+		if *tag.Key == ClusterNameTag && *tag.Value == c.spec.Metadata.Name {
 			input := &cloudformation.DeleteStackInput{
 				StackName: i.StackId,
 			}
@@ -217,7 +219,7 @@ func (c *StackCollection) DeleteStack(name string) (*Stack, error) {
 	}
 
 	return nil, fmt.Errorf("cannot delete stack %q as it doesn't bare our %q tag", *s.StackName,
-		fmt.Sprintf("%s:%s", ClusterNameTag, c.spec.ClusterName))
+		fmt.Sprintf("%s:%s", ClusterNameTag, c.spec.Metadata.Name))
 }
 
 // WaitDeleteStack kills a stack by name and waits for DELETED status
--- a/pkg/ctl/get/cluster.go
+++ b/pkg/ctl/get/cluster.go
@@ -15,6 +15,7 @@ import (
 var listAllRegions bool
 
 func getClusterCmd() *cobra.Command {
+	p := &api.ProviderConfig{}
 	cfg := api.NewClusterConfig()
 
 	cmd := &cobra.Command{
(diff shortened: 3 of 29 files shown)


pkg/eks/eks.go
```diff
@@ -105,12 +101,12 @@ func (c *ClusterProvider) getClustersRequest(chunkSize int64, nextToken string)
...
 }
 
-func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.OutputPrinter, allClusters *[]*clusterWithRegion, eachRegion bool) error {
+func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.OutputPrinter, allClusters *[]*api.ClusterID, eachRegion bool) error {
 	if eachRegion {
 		// reset region and re-create the client, then make a recursive call
 		for _, region := range api.SupportedRegions() {
-			c.Spec.Region = region
-			if err := New(c.Spec).doListClusters(chunkSize, printer, allClusters, false); err != nil {
+			spec := &api.ProviderConfig{region, c.Provider.Profile(), c.Provider.WaitTimeout()}
```
golangcibot: `api.ProviderConfig` composite literal uses unkeyed fields

pkg/utils/kubeconfig/kubeconfig.go
```diff
@@ -197,9 +193,9 @@ func MaybeDeleteConfig(ctl *api.ClusterConfig) {
 // deleteClusterInfo removes a cluster's information from the kubeconfig if the cluster name
 // provided by ctl matches a eksctl-created cluster in the kubeconfig
 // returns 'true' if the existing config has changes and 'false' otherwise
-func deleteClusterInfo(existing *clientcmdapi.Config, ctl *api.ClusterConfig) bool {
+func deleteClusterInfo(existing *clientcmdapi.Config, id *api.ClusterID) bool {
```
golangcibot: `id` can be `fmt.Stringer`

pkg/eks/api/api.go
```diff
@@ -41,22 +42,39 @@ var DefaultWaitTimeout = 20 * time.Minute
 // DefaultNodeCount defines the default number of nodes to be created
 const DefaultNodeCount = 2
 
+// ClusterID is what identifies the cluster
+type ClusterID struct {
+	Name   string
+	Region string
+}
+
+// String returns cannonical representation of ClusterID
```
golangcibot: `cannonical` is a misspelling of `canonical`

pkg/ctl/get/cluster.go
```diff
@@ -12,17 +12,20 @@ import (
 	"github.com/weaveworks/eksctl/pkg/eks/api"
 )
 
-var listAllRegions bool
+var (
```
golangcibot: `listAllRegions` is a global variable

pkg/ctl/utils/wait_nodes.go
```diff
@@ -12,15 +13,20 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+var (
```
golangcibot: `waitTimeout` is a global variable

pkg/eks/api/api.go
```diff
@@ -41,22 +42,44 @@ var DefaultWaitTimeout = 20 * time.Minute
 // DefaultNodeCount defines the default number of nodes to be created
 const DefaultNodeCount = 2
 
+// ClusterLocator is what identifies the cluster
+type ClusterLocator struct {
+	Name   string
+	Region string
+}
+
+// String returns cannonical representation of ClusterLocator
```
golangcibot: `cannonical` is a misspelling of `canonical`

pkg/utils/kubeconfig/kubeconfig.go
```diff
@@ -197,9 +193,9 @@ func MaybeDeleteConfig(ctl *api.ClusterConfig) {
 // deleteClusterInfo removes a cluster's information from the kubeconfig if the cluster name
 // provided by ctl matches a eksctl-created cluster in the kubeconfig
 // returns 'true' if the existing config has changes and 'false' otherwise
-func deleteClusterInfo(existing *clientcmdapi.Config, ctl *api.ClusterConfig) bool {
+func deleteClusterInfo(existing *clientcmdapi.Config, lc *api.ClusterLocator) bool {
```
golangcibot: `lc` can be `fmt.Stringer`

pkg/eks/eks.go
errordeveloper: I don't think it's such a bad thing really, might disable this check...

pkg/utils/kubeconfig/kubeconfig.go
errordeveloper: WHY?

pkg/cfn/builder/cluster.go
```diff
@@ -79,7 +79,7 @@ func (c *ClusterResourceSet) addResourcesForControlPlane(version string) {
 	}
 
 	c.newResource("ControlPlane", &gfn.AWSEKSCluster{
-		Name:               gfn.NewString(c.spec.ClusterName),
+		Name:               gfn.NewString(c.spec.ID.Name),
```
errordeveloper: `.ID` should be `.Locator`, as struct name is `ClusterLocator`, because it's not an ID or ARN in the sense of AWS API.

pkg/eks/eks.go
```diff
@@ -19,15 +19,10 @@ import (
...
 )
 
-type clusterWithRegion struct {
-	Name   string
-	Region string
-}
-
 // DescribeControlPlane describes the cluster control plane
-func (c *ClusterProvider) DescribeControlPlane() (*awseks.Cluster, error) {
+func (c *ClusterProvider) DescribeControlPlane(lc *api.ClusterLocator) (*awseks.Cluster, error) {
```
errordeveloper: Should be `cl` actually.

pkg/cfn/manager/api.go
```diff
@@ -36,9 +36,10 @@ type ChangeSet = cloudformation.DescribeChangeSetOutput
 
 // StackCollection stores the CloudFormation stack information
 type StackCollection struct {
-	cfn  cloudformationiface.CloudFormationAPI
-	spec *api.ClusterConfig
-	tags []*cloudformation.Tag
+	cfn         cloudformationiface.CloudFormationAPI
+	waitTimeout time.Duration
+	spec        *api.ClusterConfig
```
errordeveloper: We still keep a pointer here, but it's arguably more sensible, as there aren't many different kinds of calls to stack manager from the outside, so it's sensible to set it once.

pkg/eks/api/api.go
```diff
@@ -41,22 +42,45 @@ var DefaultWaitTimeout = 20 * time.Minute
 // DefaultNodeCount defines the default number of nodes to be created
 const DefaultNodeCount = 2
 
-// ClusterProvider provides an interface with the needed AWS APIs
+// ClusterLocator is what identifies a cluster
+type ClusterLocator struct {
```
richardcase: I'm struggling with the terms `ClusterLocator` and `Locator`. I really like the fact we have separated out the fields that uniquely identify a cluster into a separate struct. And from this we can return the string representations etc. 
We are saying that `ClusterLocator` uniquely identifies a cluster and it just so happens that the identity is made up of a location part. I wonder if `ClusterIdentitity` or something similar better describes its purpose.

pkg/eks/api/api.go
```diff
@@ -41,22 +42,45 @@ var DefaultWaitTimeout = 20 * time.Minute
...
+	return fmt.Sprintf("EKS cluster %q in %q region", c.Name, c.Region)
+}
+
+// ClusterProvider is the interface to AWS APIs
 type ClusterProvider interface {
 	CloudFormation() cloudformationiface.CloudFormationAPI
 	EKS() eksiface.EKSAPI
 	EC2() ec2iface.EC2API
 	STS() stsiface.STSAPI
+	Region() string
```
richardcase: We may need some guidance for contributors when to split out / duplicate fields like we have done with `region`.  I get why its done but its probably something we need to keep an eye on just in case there is a proliferation .

errordeveloper: Yes, I agree. I wanted to avoid `Profile()` here, but it's used somewhere and I didn't manage to get rid of it in a sensible manner.

pkg/eks/api/api.go
richardcase: I like `ClusterMeta`. Would it be worth doing that change or just take an action to change `ClusterLocator` in the future?

errordeveloper: Done.

PR #305, opened 2018-11-06
Enable use of existing subnets

errordeveloper:
### Description

This should sufficient to close #40. As discussed in #303, there will be more logical features to add, but this should enable most folks.

This adds `--vpc-{private,public}-subnets` flags. It's entirely up to the user which ones they specify as private or public. This is very clearly defined in #303, but we need to come up with a concise statement to put in the docs. We also have a waning message now, to make sure we don't get too many bug reports because of bad VPC configurations, but we still need to make it crystal clear in the docs.

### Checklist
- [x] Code compiles correctly (i.e `make build`)
- [x] All tests passing (i.e. `make test`)
- [ ] Added/modified documentation as required (such as the README)

PR #310, opened 2018-11-08
Follow-up on #305

### LINKED ISSUE #293: nodegroup stack builder should not assume subnet/AZ mapping based on the order
At the moment we assume order of subnets we get in a list is the same as the order of AZs. This is indeed the case most of the time with `eksctl create cluster`, as the lists are the same.
However, if we want to create a nodegroup separately, that assumption is most likely to break.

I believe we have to options here:

- let cluster stack export the list of AZs in the same order as subnets, which is bound to ordering assumption at the time of creation (which is probably fine)
- when importing subnets from cluster stack, check which AZs they belong to (probably via an API lookup on eksctl side, doing it in CloudFormation would be harder)

Broadly, the boundary is currently unclean between how we use stack exports and how we maintain state on eksctl side, we have a bit of mix going on at the moment and it's most likely to stay that way, as we do what's convenient most of the time, especially because some of CloudFormation facilities are harder to use then others and we only utilise some of them.

This area needs to be reviewed.

### LINKED ISSUE #308: Follow up on #305
When we import subnets we must import VPC CIDR also.

We will also have to fix #293, so we consolidate this code.

errordeveloper:
### Description

- fix #308 
- fix #293

### Checklist


pkg/vpc/vpc.go
```diff
@@ -53,29 +54,64 @@ func (c *ClusterProvider) SetSubnets(spec *api.ClusterConfig) error {
...
+	}
+	output, err := povider.EC2().DescribeVpcs(input)
+	if err != nil {
+		return nil, err
+	}
+	return output.Vpcs[0], nil
+}
 
-	for _, subnet := range output.Subnets {
+// ImportSubnets will update spec with subnets, if VPC ID/CIDR is unkown
```
golangcibot: `unkown` is a misspelling of `unknown`

errordeveloper:
@richardcase @stefanprodan thanks for the review, I'm looking forward to this weeks release with all VPC customisations some users been waiting for!

PR #316, opened 2018-11-09
Create version package, set user agent

errordeveloper:
### Description

### Checklist


errordeveloper:
GolangCI doesn't seem to be very reliable lately, so please ignore it.

pkg/version/version.go
```diff
@@ -0,0 +1,31 @@
...
+//go:generate go run ./release_generate.go
+
+// Info hold version information
+type Info struct {
+	BuiltAt   string
+	GitCommit string `json:",omitempty"`
+	GitTag    string `json:",omitempty"`
+}
+
+var info = Info{
```
golangcibot: `info` is a global variable

pkg/version/norelease.go
```diff
@@ -0,0 +1,7 @@
+// +build !release
+
+package version
+
+var (
```
golangcibot: `builtAt` is a global variable

errordeveloper:
@golangcibot seems to report certain things here, but if I run `golangci-lint` locally I don't get the same, actually I get nothing and just non-zero exit code. I feel that this is an immature tool and we should stick with gometalinter running in CircleCI for now.

errordeveloper:
@christopherhein @nckturner CI is happy now, please review and merge if you are happy.
cc @micahhausler

jirfag:
@errordeveloper it's because you use `enable-all: true` in `.golangci.yml`. We don't recommend users to use this options because when we add a new linter users see unexpected errors.
We're planning to deprecate this option because of that.

christopherhein:
Is there a reason we're not using the standard `version`, `commit`, `date` ldflags that `goreleaser` automatically sets?

```
# Default is `-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}`.
```

From - https://goreleaser.com/build/

errordeveloper:
Didn't know those exist. We would have to set those on the version package
(not main package), will that work?

errordeveloper:
Denis, I see. Perhaps we could update that... But we have also seen a lot
of timeouts, is that related?

jirfag:
> Denis, I see. Perhaps we could update that... But we have also seen a lot
of timeouts, is that related?
Yes, there were timeouts 3 days ago (internal problem), they aren't related.

christopherhein:
> Didn't know those exist. We would have to set those on the version package (not main package), will that work?

I don't think that is possible, what I've been doing is just adding a `version` subcommand like we have and having my `cmd/... as `package main` when go releaser builds it'll automatically add the proper ldflags then I built https://github.com/christopherhein/go-version to present it back to the user with `mybin version [-s]` and if I need the version details in the packages I pass it in through a config.

errordeveloper:
Ah, I see. I think this PR does the job, I don't think there is much that
goreleaser would buy us, in fact passing version in config struct seems a
little awkward...

Issue #299, opened 2018-11-02
Bash completion


braderhart:
**Why do you want this feature?**
Bash completion is standard with kubectl and other CLI tools within the Kubernetes ecosystem

**What feature/behavior/change do you want?**
Add a `eksctl completion` CLI option to generate a Bash completion file, or include the completion file in the repo.

It looks like this should be pretty easy to add:

- https://github.com/weaveworks/eksctl/blob/master/vendor/github.com/spf13/cobra/bash_completions.go

errordeveloper:
I'd be happy to add this, but I do see that #298 is more important right now.

braderhart:
@errordeveloper I don't have a lot of experience with this, but I'll see if I can find some time to figure it out.

errordeveloper:
Closed via #327.

PR #327, opened 2018-11-28
Add shell completion

ozzieba:
Follows https://github.com/spf13/cobra/blob/master/bash_completions.md
to resolve Issue #299

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,10 +338,22 @@ To enable bash completion, run the following, or put it in `~/.bashrc` or `~/.pr
 . <(eksctl completion bash)
 ```
 
-Or for zsh:
+Or for zsh, run:
 ```
-. <(eksctl completion zsh)
+mkdir -p ~/.zsh/completion/
+eksctl completion zsh > ~/.zsh/completion/_eksctl
 ```
+and put the following in `~/.zshrc`:
+```
+fpath=($fpath ~/.zsh/completion)
+```
+Note if you're not running a distribution like oh-my-zsh you may first have to enable autocompletion:
+```
+autoload -U compinit
+compinit
+```
+
+To make the above persistent, run the first two lines, and put the
 
 
 ## Project Roadmap
--- a/README.md
+++ b/README.md
@@ -331,6 +331,19 @@ eksctl apply --cluster-config advanced-cluster.yaml
 ```
 -->
 
+### Shell Completion
+
+To enable bash completion, run the following, or put it in `~/.bashrc` or `~/.profile`:
+```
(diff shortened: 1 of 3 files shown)


rndstr:
nice, thanks!

would you be able to also add support for zsh? that would be `GenZshCompletion()`. maybe we can have `eksctl completion <bash|zsh>`.

would it make sense to add a hint about this to the README?

ozzieba:
I'm not sure how to put completion in its own package, since that would create a cyclic dependency between `main` and `completion` (since I need `rootCmd`)...

ozzieba:
Hmm, I suppose I could just export a function that takes an arbitrary Command and returns a `completion` command for it

ozzieba:
Not really sure the refactored version is better, let me know what you think...

README.md
```diff
@@ -331,6 +331,19 @@ eksctl apply --cluster-config advanced-cluster.yaml
...
+### Shell Completion
+
+To enable bash completion, run the following, or put it in `~/.bashrc` or `~/.profile`:
+```
+. <(eksctl completion bash)
+```
+
+Or for zsh:
+```
+. <(eksctl completion zsh)
```
rndstr: i'm also fairly new to zsh completion but i just tried it and i don't think that's how it's going to work unfortunately.

My brief research has led me to something like this

```
mkdir ~/.zsh/completion/
eksctl completion zsh > ~/.zsh/completion/_eksctl
```
Then add this to the `fpath` in `~/.zshrc`
```
 fpath=(~/.zsh/completion $fpath)
```

pkg/ctl/completion/completion.go
```diff
@@ -33,10 +33,12 @@ To configure your bash shell to load completions for each session add to your ba
 
 . <(eksctl completion zsh)
 
-To configure your zsh shell to load completions for each session add to your zshrc
+To configure your zsh shell, run:
+
+mkdir -p ~/.zsh/completion/
+eksctl completion zsh > ~/.zsh/completion/_eksctl
+fpath=($fpath ~/.zsh/completion)
```
rndstr: maybe expand on this that it should be put into `~/.zshrc`?

pkg/ctl/completion/completion.go
```diff
@@ -33,10 +33,12 @@ To configure your bash shell to load completions for each session add to your ba
 
 . <(eksctl completion zsh)
```
rndstr: this isn't correct anymore

errordeveloper:
LGTM, thanks a lot!

PR #328, opened 2018-11-30
Group cluster create flags for readability

mumoshu:
Hi! Thanks for maintaining an awesome project :relaxed:

Towards #298, I've tried to group `create cluster` flags as I think appropriate.

I've intentionally left the too-long descriptions as-is. I'll work on it in an another PR if you like.

`eksctl create cluster -h` now looks like:

```
Create a cluster

Usage:
  eksctl create cluster [flags]

Node Flags:
      --asg-access                enable iam policy dependency for cluster-autoscaler
      --full-ecr-access           enable full access to ECR
      --max-pods-per-node int     maximum number of pods per node (set automatically if unspecified)
      --node-ami string           Advanced use cases only. If 'static' is supplied (default) then eksctl will use static AMIs; if 'auto' is supplied then eksctl will automatically set the AMI based on region/instance type; if any other value is supplied it will override the AMI to use for the nodes. Use with extreme care. (default "static")
      --node-ami-family string    Advanced use cases only. If 'AmazonLinux2' is supplied (default), then eksctl will use the offical AWS EKS AMIs (Amazon Linux 2); if 'Ubuntu1804' is supplied, then eksctl will use the offical Canonical EKS AMIs (Ubuntu 18.04). (default "AmazonLinux2")
  -P, --node-private-networking   whether to make initial nodegroup networking private
  -t, --node-type string          node instance type (default "m5.large")
      --node-volume-size int      Node volume size (in GB)
  -N, --nodes int                 total number of nodes (desired capacity of ASG) (default 2)
  -M, --nodes-max int             maximum nodes in ASG (leave unset for a static nodegroup)
  -m, --nodes-min int             minimum nodes in ASG (leave unset for a static nodegroup)
      --ssh-access                control SSH access for nodes
      --ssh-public-key string     SSH public key to use for nodes (import from local path, or use existing EC2 key pair) (default "~/.ssh/id_rsa.pub")
      --storage-class             if true (default) then a default StorageClass of type gp2 provisioned by EBS will be created (default true)

Networking Flags:
      --vpc-cidr ipNet                 global CIDR to use for VPC (default 192.168.0.0/16)
      --vpc-from-kops-cluster string   re-use VPC from a given kops cluster
      --vpc-private-subnets strings    re-use private subnets of an existing VPC
      --vpc-public-subnets strings     re-use public subnets of an existing VPC
      --zones strings                  (auto-select if unspecified)

Stack Flags:
  -r, --region string         AWS region
[...]

PR #330, opened 2018-11-30
feat: CloudFormation Service Role support

### LINKED ISSUE #329: CloudFormation Service Role support
Hey! Thanks for maintaining an awesome project.

I'm raising this issue according to my experience in an another project, and what I believe needed by certain users.

`eksctl` as of today requires an admin-like IAM priv to create a cluster/cfn stack.

This doesn't work in certain organizations with tighter security requirements.

An example would be that there is an IAM role for admins but it has source IP restriction to prevent malicious user w/ stolen AWS creds to do bad things.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "*",
            "Resource": "*",
            "Condition": {
                "IpAddress": {
                    "aws:SourceIp": [
                        "MYIP",
                    ]
                }
            }
        },
```

Add a dedicated flag to tell cfn to use a [service role](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-iam-servicerole.html), either `eksctl create cluster --cloudformation-service-role` or `--cfn-service-role`(which one do you prefer?).

As a sidenote, this is different from #122. #122 is for EKS, while this is for CFN, used by stack creation/update/deletion.

In kube-aws, the exact the same way did work as you can see in https://github.com/kubernetes-incubator/kube-aws/pull/1113/files#diff-6f90e92944e47c5a9899723462d5880bR149.

mumoshu:
**This depends on/includes changes made in #328, to avoid obvious conflicts.**

**Please see only the last commit for effective reviews.**

---

In case your role is too restrictive that `eksctl create cluster` fails due
to cloudformation reporting insufficient permissions, specify a service
role used by CloudFormation to call AWS API while provisioning stacks on
your behalf.

``` eksctl create cluster --cfn-role-arn
arn:aws:iam:YOUR_AWS_ACCOUNT_ID:role/eksctl
```

Also note that eksctl now helps you by printing the guidance like below on
permission errors:

``` 2018-11-30T01:42:20-08:00 [ℹ]  creating cluster stack
"eksctl-ferocious-gopher-1543570938-cluster" 2018-11-30T01:42:21-08:00 [✖]
ensure that the iam role "arn:aws:iam::YOUR_AWS_ACCOUNT_ID:role/eksctl"
exists, it should also have a trust relationship like the below.

{
 "Version": "2012-10-17",
 "Statement": [
   {
     "Sid": "",
     "Effect": "Allow",
     "Principal": {
       "Service": "cloudformation.amazonaws.com"
     },
     "Action": "sts:AssumeRole"
   }
 ]
}
```

Resolves #329

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -70,6 +70,10 @@ func (c *StackCollection) doCreateStackRequest(i *Stack, templateBody []byte, pa
 		input.SetCapabilities(stackCapabilitiesIAM)
 	}
 
+	if cfnRole := c.provider.CloudFormationRoleARN(); cfnRole != "" {
+		input = input.SetRoleARN(cfnRole)
+	}
+
 	for k, v := range parameters {
 		p := &cloudformation.Parameter{
 			ParameterKey:   aws.String(k),
@@ -205,6 +209,10 @@ func (c *StackCollection) DeleteStack(name string) (*Stack, error) {
 				StackName: i.StackId,
 			}
 
+			if cfnRole := c.provider.CloudFormationRoleARN(); cfnRole != "" {
+				input = input.SetRoleARN(cfnRole)
+			}
+
 			if _, err := c.provider.CloudFormation().DeleteStack(input); err != nil {
 				return nil, errors.Wrapf(err, "not able to delete stack %q", name)
 			}
@@ -269,6 +277,10 @@ func (c *StackCollection) doCreateChangeSetRequest(i *Stack, action string, desc
 
 	if withIAM {
 		input.SetCapabilities(stackCapabilitiesIAM)
+	}
+
+	if cfnRole := c.provider.CloudFormationRoleARN(); cfnRole != "" {
+		input.SetRoleARN(cfnRole)
 	}
 
 	for k, v := range parameters {
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -50,6 +50,9 @@ type ProviderServices struct {
 // CloudFormation returns a representation of the CloudFormation API
 func (p ProviderServices) CloudFormation() cloudformationiface.CloudFormationAPI { return p.cfn }
 
+// CloudFormationRoleARN returns, if any,  a service role used by CloudFormation to call AWS API on your behalf
+func (p ProviderServices) CloudFormationRoleARN() string { return p.spec.CloudFormationRoleARN }
+
 // EKS returns a representation of the EKS API
 func (p ProviderServices) EKS() eksiface.EKSAPI { return p.eks }
 
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -45,6 +45,11 @@ func AddRegionFlag(fs *pflag.FlagSet, p *api.ProviderConfig) {
 	fs.StringVarP(&p.Region, "region", "r", "", "AWS region")
 }
 
+// AddCFNRoleARNFlag adds common --cfn-role-arn flag
+func AddCFNRoleARNFlag(fs *pflag.FlagSet, p *api.ProviderConfig) {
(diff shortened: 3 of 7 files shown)


pkg/cfn/manager/api.go
```diff
@@ -54,6 +57,7 @@ func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) *
 	}
 	logger.Debug("tags = %#v", tags)
 	return &StackCollection{
+		cfnSvcRoleArn: provider.CloudFormationRoleARN(),
```
errordeveloper: Seems redundant, as we already have `provider` and can call the `CloudFormationRoleARN` on it to obtain the role ARN.

pkg/cfn/manager/api.go
```diff
@@ -83,13 +91,41 @@ func (c *StackCollection) doCreateStackRequest(i *Stack, templateBody []byte, pa
...
 	if err != nil {
+		logGuidanceOnAssumeRoleFailure(err, input.RoleARN)
 		return errors.Wrapf(err, "creating CloudFormation stack %q", *i.StackName)
 	}
 	logger.Debug("stack = %#v", s)
 	i.StackId = s.StackId
 	return nil
 }
 
+func logGuidanceOnAssumeRoleFailure(err error, roleArn *string) {
```
errordeveloper: This is an advanced use case, I would expect the user to understand very well what they are doing. I don't see much need for this helper, would you mind to omit it for now before we hear from users?

pkg/eks/api.go
```diff
@@ -44,11 +44,16 @@ type ProviderServices struct {
 	eks  eksiface.EKSAPI
 	ec2  ec2iface.EC2API
 	sts  stsiface.STSAPI
+
+	cfnRoleArn string
```
errordeveloper: `s/Arn/ARN/`

pkg/cfn/manager/api.go
```diff
@@ -262,6 +298,7 @@ func (c *StackCollection) doCreateChangeSetRequest(i *Stack, action string, desc
 		StackName:     i.StackName,
 		ChangeSetName: &changeSetName,
 		Description:   &description,
+		RoleARN: aws.String(c.cfnSvcRoleArn),
```
errordeveloper: I'd just use `&c.cfnSvcRoleArn` here, which matches the way it's done above.

errordeveloper:
I've added a few style-related comments, happy to address those myself in this PR, or via a follow-up PR. If you can add functionality for stack updates and deletions, that would be great :)

pkg/cfn/manager/api.go
mumoshu: Agreed! But I ended up removing this line due to the redundancy 0bea25c

mumoshu:
Thanks for reviewing! The support for stack updates was already there, but I fixed a bug in it. Also, I've added the support for deletion.

pkg/cfn/manager/api.go
mumoshu: Completely understood. Removed in 7e45c81

pkg/eks/api.go
mumoshu: Turned out this was also redundant. Removed in 1fcf70d

mumoshu:
This should be good to go now.

errordeveloper:
@mumoshu I am happy to merge this, but will need to rebase and probably squashing the commits would make sense also. Do you mind I just use your branch to do this?

errordeveloper:
It fits quite nicely into a one simple commit: https://github.com/weaveworks/eksctl/commit/1ab84c251853f47b95db0ddf0573098cd214b65f.

mumoshu:
That would be great. Thanks :)

2018年12月19日(水) 4:21 Ilya Dmitrichenko <notifications@github.com>:

Issue #156, opened 2018-08-09
use `KubeleteConfiguration` instead of flags


errordeveloper:
We would benefit from not having to worry about kubelet flags. At least a config file would be helpful, but componentconfig would be even better (need to check status).

errordeveloper:
We switch to kubelet config in #353, but it's still just a static asset, we should generate it and get rid of all the flags.

errordeveloper:
Turns out not all the flags were defined in the config yet, we will migrate what we can.

errordeveloper:
xref https://github.com/kubernetes/kubernetes/issues/61647#issuecomment-474330457

errordeveloper:
This would requite some upstream changes, which are far-fetched at this point.

PR #332, opened 2018-12-03
feat: Nodegroup as a resource

mumoshu:
### Description

This is a successor of #281 but with a bunch of conflicts resolved, successfully rebased onto the #328, and various enhancements to make the scenario works.

The original credit goes to Richard Case <richard.case@outlook.com>. Thanks a lot, Richard!

#### Usage

- [x] Create/delete/scale node groups
- [x] Node labels

##### Creating a cluster along with a nodegroup

`eksctl create cluster` creates the infrastructure, the control-plane, and a nodegroup. The nodegroup is authorized by creating the `aws-auth` configmap containing a `mapRoles` entry for the nodegroup.

##### Adding a node group

`eksctl create nodegroup --cluster CLUSTER_NAME` can be used to add the second and the subsequent nodegroups to the cluster. The `aws-auth` configmap is updated to include the newly added nodegroup.

##### Setting node labels for a nodegroup

Use the `--node-labels KEY_1=VAL_1,KEY_2=VAL_2`, which looks exactly same as the kubelet's flag, to set node labels for the nodegroup being created. That is, `eksctl create nodegroup --cluster CLUSTER_NAME --node-labels foo=bar,bar=baz` results in labeling every node in the nodegroup `foo=bar` and `bar=baz`.

##### Listing all the nodegroups

`eksctl get nodegroup --cluster CLUSTER_NAME` can be used to list all the nodegroups for the cluster.

##### Scaling a nodegroup

`eksctl scale nodegroup --cluster CLUSTER_NAME --nodes DESIREC_COUNT ID` scales the nodegroup, within the cluster CLUSTESR_NAME,  identified by ID, to DESIRED_COUNT.

##### Deleting a nodegroup

`eksctl delete nodegroup --cluster CLUSTER_NAME ID` deletes the specified nodegroup only.

##### Deleting a cluster along with nodegroup(s)

`eksctl delete cluster CLUSTER_NAME` deletes the cluster, along with its nodegroups.

##### `eksctl create nodegroup --cluster CLUSTER_NAME --node-labels foo=bar,bar=baz`

```
$ eksctl create nodegroup --cluster floral-creature-1543934252 --node-labels foo=bar,bar=baz

$ kubectl get no --show-labels
get no --show-labels
NAME                                           STATUS    ROLES     AGE       VERSION   LABELS
*snip*
ip-192-168-59-31.us-west-2.compute.internal    Ready     <none>    1m        v1.10.3   bar=baz,beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=m5.large,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=us-west-2,failure-domain.beta.kubernetes.io/zone=us-west-2b,foo=bar,kubernetes.io/hostname=ip-192-168-59-31.us-west-2.compute.internal
ip-192-168-7-227.us-west-2.compute.internal    Ready     <none>    1m        v1.10.3   bar=baz,beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=m5.large,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=us-west-2,failure-domain.beta.kubernetes.io/zone=us-west-2c,foo=bar,kubernetes.io/hostname=ip-192-168-7-227.us-west-2.compute.internal
```

##### `eksctl get nodegroups --cluster CLUSTER_NAME`

```
eksctl get nodegroup --cluster floral-creature-1543934252
2018-12-05T10:39:07+09:00 [ℹ]  stack eksctl-floral-creature-1543934252-nodegroup-2
2018-12-05T10:39:08+09:00 [ℹ]  stack eksctl-floral-creature-1543934252-nodegroup-0
STACKNAME					SEQ	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
eksctl-floral-creature-1543934252-nodegroup-0	0	2018-12-04T14:47:26Z	2		22			m5.large	ami-0f54a2f7d2e9c88b3
eksctl-floral-creature-1543934252-nodegroup-2	2	2018-12-05T01:29:50Z	2		22			m5.large	ami-0f54a2f7d2e9c88b3
```

##### `eksctl scale nodegroup --cluster CLUSTER_NAME ID`

```
eksctl scale nodegroup --nodes 1 --cluster floral-creature-1543934252 1
2018-12-05T10:35:21+09:00 [ℹ]  scaling nodegroup stack "eksctl-floral-creature-1543934252-nodegroup-1" in cluster eksctl-floral-creature-1543934252-cluster
2018-12-05T10:35:21+09:00 [ℹ]  scaling nodegroup, desired capacity from 2 to 1, min size from 2 to 1
```

##### `eksctl delete cluster CLUSTER_NAME`

- `eksctl delete cluster` deletes all the nodegroups in parallel, and then the control-plane.

```
$ go build ./cmd/eksctl && ./eksctl delete cluster adorable-sheepdog-1543570021 && ./eksctl create cluster
2018-12-04T22:00:07+09:00 [ℹ]  deleting EKS cluster "adorable-sheepdog-1543570021"
2018-12-04T22:00:09+09:00 [ℹ]  will delete stack "eksctl-adorable-sheepdog-1543570021-nodegroup-3"
2018-12-04T22:00:09+09:00 [ℹ]  waiting for stack "eksctl-adorable-sheepdog-1543570021-nodegroup-3" to get deleted
2018-12-04T22:00:09+09:00 [ℹ]  will delete stack "eksctl-adorable-sheepdog-1543570021-nodegroup-1"
2018-12-04T22:00:09+09:00 [ℹ]  waiting for stack "eksctl-adorable-sheepdog-1543570021-nodegroup-1" to get deleted
2018-12-04T22:00:09+09:00 [ℹ]  will delete stack "eksctl-adorable-sheepdog-1543570021-nodegroup-2"
2018-12-04T22:00:09+09:00 [ℹ]  waiting for stack "eksctl-adorable-sheepdog-1543570021-nodegroup-2" to get deleted
```

### Checklist

Signed-off-by: Yusuke Kuoka <ykuoka@gmail.com>


mumoshu:
I've manually verified all the tasks introduced in the `Usage` sections of the PR description works. Any feedback is welcomed!

Also the unit tests are failing terribly. I'll shortly start fixing em.

richardcase:
Great work @mumoshu. i'll take a look.

mumoshu:
FY I've just added `--node-labels k1=v1,k2=v2,...` to `create nodegroup` and `create cluster`, that works as the name says.

mumoshu:
@richardcase Thanks! Looking forward to your comments.

One thing I'm unsure at the moment is, whether we should remove `NodeGroups` from the `ClusterConfig` i.e. actually graduating `NodeGroup` as a top-level resource.

Doing so would prevent us from addressing edge-cases, in the future that eksctl added declarativeness, like `What happens if we create a nodegroup, and then update the cluster along with the nodegroups embedded in the cluster?`

But I'd just wanted someone to confirm before I put more efforts, because it's about changing the API.

mumoshu:
Now idea how those 6 test cases are failing 😢

Raffo:
Hi @mumoshu , this is really great work! I think those feature deserve a bit more of unit tests, don't you think? 
The other failing tests seem now to be related to linting, so easy to get fixed: 
```
pkg/cfn/manager/nodegroup.go:39:1:warning: exported method StackCollection.MakeNodeGroupStackName should have comment or be unexported (golint)
pkg/ctl/get/nodegroup.go:54:21:warning: error strings should not be capitalized or end with punctuation or a newline (golint)
pkg/eks/nodegroup.go:59:1:warning: comment on exported method ClusterProvider.AddNodeGroupToAuthConfigMap should be of the form "AddNodeGroupToAuthConfigMap ..." (golint)
pkg/eks/api/node_labels.go:20:1:warning: exported method NodeLabels.Set should have comment or be unexported (golint)
pkg/eks/api/node_labels.go:38:1:warning: exported method NodeLabels.Type should have comment or be unexported (golint)
```

mumoshu:
@Raffo Thanks for the comment!

The linter has been happy after 7d1439c, and the tests are added in 4da60c8. The documentation is updated in 89fd13f.

I believe this is fully ready to be reviewed.

mumoshu:
@richardcase Nevermind on my [prev concern](https://github.com/weaveworks/eksctl/pull/332#issuecomment-444395728).

I had no problem deferring the decision. eksctl does have two paths to create nodegroups now, one for creating nodegroups along with a cluster, the another is to do it independenty.  But it works without issues in the current use-cases of eksctl.

Also, I can remove the nodegroups nested in the cluster config if necessary, but I'd suggest to do so in an another pull request. This seems already too big to be reviewed :)

README.md
```diff
@@ -162,12 +162,40 @@ To delete a cluster, run:
...
+```
+eksctl get nodegroup --cluster=<cluster name> [<nodegroup SEQ>]
+```
+
+Whereas the nodegroup `SEQ` is an integer value starting from `0` which is used to identify a nodegroup.
+
 ### Scaling nodegroup
 
-The initial nodegroup can be scaled by using the `eksctl scale nodegroup` command. For example, to scale to 5 nodes:
+A nodegroup can be scaled by using the `eksctl scale nodegroug` command:
```
Lazyshot: nitpick on a typo

mumoshu: Thanks! Fixed in the latest commit.

pkg/ctl/create/cluster.go
```diff
@@ -51,38 +48,22 @@ func createClusterCmd() *cobra.Command {
...
-	fs.StringSliceVar(&availabilityZones, "zones", nil, "(auto-select if unspecified)")
-
-	fs.BoolVar(&ng.AllowSSH, "ssh-access", false, "control SSH access for nodes")
-	fs.StringVar(&ng.SSHPublicKeyPath, "ssh-public-key", defaultSSHPublicKey, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
+	addCommonCreateFlags(fs, p, cfg, ng)
 
 	fs.BoolVar(&writeKubeconfig, "write-kubeconfig", true, "toggle writing of kubeconfig")
 	cmdutils.AddCommonFlagsForKubeconfig(fs, &kubeconfigPath, &setContext, &autoKubeconfigPath, exampleClusterName)
 
+	//TODO: is this needed for nodegroup creation????
```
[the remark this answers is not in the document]
mumoshu: I wasn't actually sure, but see the comment at the previous pr https://github.com/weaveworks/eksctl/pull/281#discussion_r232201190

Also, I'm removing this comment as it the `zones` flag for nodegroups just work!

README.md
```diff
@@ -162,12 +162,40 @@ To delete a cluster, run:
...
+```
+eksctl create nodegroup --cluster=<cluser name>
+```
+
+### Listing nodegroups
+
+To list the details about a nodegroup or all of the nodegroups, use:
+
+```
+eksctl get nodegroup --cluster=<cluster name> [<nodegroup SEQ>]
```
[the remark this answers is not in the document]
mumoshu: I actually don't know! Perhaps we'd better ask @richardcase as he seems to have decided to use SEQ.

I'm personally fine with changing it to use auto-generated names instead. https://github.com/weaveworks/eksctl/pull/281#discussion_r238560912

 If I only I can address it in an another PR so that this changes between @richardcase's work and mine become clearer, which may make his review easier.

pkg/cfn/manager/nodegroup.go
```diff
@@ -137,15 +169,71 @@ func (c *StackCollection) ScaleNodeGroup(id int) error {
...
+	for _, stack := range stacks {
+		logger.Info("stack %s\n", *stack.StackName)
+		logger.Debug("stack = %#v", stack)
+
+		template, err := c.GetStackTemplate(*stack.StackName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error getting Cloudformation template for stack %s", *stack.StackName)
+		}
+
+		//TODO: create a map function
```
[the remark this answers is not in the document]
mumoshu: Done in the latest commit. Thanks!

pkg/ctl/create/cluster.go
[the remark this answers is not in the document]
mumoshu: Addressed in the latest commit. The `--nodes-zones` is the flag that works for `create nodegroup`.

README.md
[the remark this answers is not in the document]
Raffo: Let's wait for @richardcase to share his opinion on that then, I'd love to have this from the beginning, but incrementally is also fine.

errordeveloper:
WOW! I'm pretty exciting to this is coming along, I will try to review this tomorrow.

README.md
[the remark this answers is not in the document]
richardcase: Not sure i entirely agree that names are any less error prone. However, it could be easier when identifying specific nodegroups (i.e. a GPU optimized enabled group) but that can be done via labels. 

Perhaps we can have both? We always have a sequence number as that denotes the sequence in which they where created and its easy to to just increment the sequence to make a unique group name. We could then have a name flag that if specified will result in a specific label being attched to the stack that we could then pull out when we do the `get`. WDYT?

pkg/cfn/manager/nodegroup.go
```diff
@@ -137,15 +169,79 @@ func (c *StackCollection) ScaleNodeGroup(id int) error {
...
+			return nil, errors.New("error mapping stack to node gorup summary")
+		}
+
+		summaries = append(summaries, summary)
+	}
+
+	return summaries, nil
+}
+
+func (c *StackCollection) mapStackToNodeGroupSummary(stack *Stack) (*NodeGroupSummary, error) {
```
richardcase: Nice, great to see.

pkg/ctl/create/nodegroup.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if cfg.Metadata.Name == "" {
+		return errors.New("--cluster must be specified. run `eksctl get cluster` to show existing clusters")
+	}
+
+	if ng.SSHPublicKeyPath == "" {
+		return fmt.Errorf("--ssh-public-key must be non-empty string")
+	}
+
+	//TODO:  do we need to do the AZ stuff from create????
```
richardcase: I can't remember why i added that comment.

README.md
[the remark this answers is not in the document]
Raffo: That's pretty much what I meant even though I was super brief 😄 A user provided name is the best we can do and it's what tools like kops are already providing. To give a user perspective, I use this all the time for example to do canary node groups to test new AMIs, would be great to have that!
I'm pretty much agree with what @mumoshu said here :-)

pkg/ctl/create/nodegroup.go
mumoshu: Perhaps you wanted to confirm the new functionally supports `--node-zones` to select which zone(s) to use for the nodegroup?

AFAICS, it just works. `--node-zones` populates `AvailabilityZones` in the nodegroup spec, which is then handled by the cfn builder to either select subnets that corresponds to the zones, or import: https://github.com/weaveworks/eksctl/blob/537f2c413330269cf044ee04f8a1a42058231235/pkg/cfn/builder/nodegroup.go#L141-L162

richardcase: Yes thats right. I have a feeling that @errordeveloper commented back then that for a new nodegroup it would be good to be able to use specific zones (and to overide the default behavior).

README.md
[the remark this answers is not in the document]
mumoshu: Yay 😄

I'm prepared to submit an another PR to change the SEQ number to specific/auto-generated names if we could build a consensus with @richardcase and @errordeveloper.

errordeveloper:
Yes, the nodegroup names sound like a good idea. We might want to have a
tag to determine sequence/generation, but it doesn't need to be numeric,
can be e.g. timestamp.

mumoshu:
@errordeveloper Thanks! Just curious, but what's the use-case of the sequence/generation?

errordeveloper:
I mostly thinking of upgrades, we would want to easily create a new
nodegroup and retire the old one once we are happy with the new one. It
just seems reasonable to have generation tag for that, unless you have an
alternative in mind, I haven't spent too much time thinking about this yet
:)

mumoshu:
@errordeveloper Thanks! So would there be two nodegroups with the same NAME but different GENERATIONs while you're upgrading nodegrooups?

errordeveloper:
Doesn't need to be the same name, but we would want to indicate
relationship between the two.

mumoshu:
@errordeveloper Ah, I think I've figured it out now :)

So, probably you want something that works like a k8s deployment and pods managed by it. As similar to k8s creates a new pod and then removes the old one while updating, we'll want create a new stack and then removes the old one while updating a nodegroup.

It would work by allowing to temporarily have multiple cfn stacks per a nodegroup, whereas the cfn stacks are grouped by the nodegroup NAME, but differentiated by GENERATION. `eksctl get nodegroup` would should show up only the "current" cfn stack at the time.

mumoshu:
The sequence/generation doesn't need to be numeric. Inducing the generation from e.g. a creation timestamp of the underlysing cfn stack would be possible. But we need short generation names that can be embedded into stack names, so that stack names are easy to understand, and they don't collide. I personally find SEQ as a sequence number good fit for embedding.

So, I'd change this PR, adding NAME to nodegroups, while changing scope of each SEQ to the group of stacks associated to the corresponding nodegroup.

`eksctl get nodegroup` would show nodegroups like:

`eksctl-floral-creature-1543934252-nodegroup-fabulous-sheepdog` (`fabulous-sheepdog` is the NAME of the nodegroup)

Whereas underlying stacks would named like:

`eksctl-floral-creature-1543934252-nodegroup-fabulous-sheepdog-0`(`0` is the SEQ=generation)

The trailing `0` is the generation number that will be always zero for now, but incremented when we actually implement nodegroup updates. For example, a new, replacement cfn stack for the nodegroup `fabulous-sheepdog` would get a SEQ of `1`.

@errordeveloper WDYT?

errordeveloper:
> The sequence/generation doesn't need to be numeric. Inducing the generation from e.g. a creation timestamp of the underlysing cfn stack would be possible.

> But we need short generation names that can be embedded into stack names, so that stack names are easy to understand, and they don't collide. I personally find SEQ as a sequence number good fit for embedding.

Now that I think about it, I wonder if we really need it. Maybe we can just use a different name generator?

I believe that if we look at cluster that is called `exciting-badger-1542728046` and has two nodegroups called `ridiculous-mushroom-1545222716` and `extravagant-badger-1545222742`, it's hard to remember which name was for what.

If we were to use something like `ng-cdf8139386` `ng-ba8e9fbd08` for the nodegroups, it'd be much easier distinguish the two from each other and the cluster name. 

> So, I'd change this PR, adding NAME to nodegroups, while changing scope of each SEQ to the group of stacks associated to the corresponding nodegroup.

I think we should consider just removing SEQ for now. In my earlier comment I eluded to generational relationships between nodegroups, but I think that was a premature sort of idea and we should just park it until we get to implementing automated upgrades, we might not need it at all, or otherwise it's easy to add back in.

> `eksctl get nodegroup` would show nodegroups like:
>
> `eksctl-floral-creature-1543934252-nodegroup-fabulous-sheepdog` (`fabulous-sheepdog` is the NAME of the nodegroup)
>
> Whereas underlying stacks would named like:
>
> `eksctl-floral-creature-1543934252-nodegroup-fabulous-sheepdog-0`(`0` is the SEQ=generation)

Sounds good, but let's use separate two columns - `CLUSTER` and `NAME`. And there is no need to expose stack names - those are an implementation details as far as the user is concerned.

So currently the output is this:

```
STACKNAME			SEQ	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
eksctl-cluster-1-nodegroup-0	0	2018-12-19T11:41:35Z	1		1		1			m5.large	ami-0a9006fb385703b54
```

I think it would look better if it was this:

```
CLUSTER		NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
cluster-1	nodegroup-0	2018-12-19T11:41:35Z	1		1		1			m5.large	ami-0a9006fb385703b54
```

Also, there is an inconsistency between column names and flags, we will need to fix that.

> The trailing `0` is the generation number that will be always zero for now, but incremented when we actually implement nodegroup updates. For example, a new, replacement cfn stack for the nodegroup `fabulous-sheepdog` would get a SEQ of `1`.

Let's just forget SEQ for now, when we get to upgrades we will find a solution for this (only if it's indeed needed in practice).

pkg/ctl/create/cluster.go
```diff
@@ -63,24 +57,7 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
...
-
-		fs.StringVar(&ng.AMI, "node-ami", ami.ResolverStatic, "Advanced use cases only. If 'static' is supplied (default) then eksctl will use static AMIs; if 'auto' is supplied then eksctl will automatically set the AMI based on version/region/instance type; if any other value is supplied it will override the AMI to use for the nodes. Use with extreme care.")
-		fs.StringVar(&ng.AMIFamily, "node-ami-family", ami.ImageFamilyAmazonLinux2, "Advanced use cases only. If 'AmazonLinux2' is supplied (default), then eksctl will use the offical AWS EKS AMIs (Amazon Linux 2); if 'Ubuntu1804' is supplied, then eksctl will use the offical Canonical EKS AMIs (Ubuntu 18.04).")
-
-		fs.BoolVar(&ng.AllowSSH, "ssh-access", false, "control SSH access for nodes")
-		fs.StringVar(&ng.SSHPublicKeyPath, "ssh-public-key", defaultSSHPublicKey, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
-
-		fs.BoolVarP(&ng.PrivateNetworking, "node-private-networking", "P", false, "whether to make initial nodegroup networking private")
+		addCommonCreateFlags(fs, p, cfg, ng)
```
errordeveloper: Let's make this `cmdutils.AddCommonFlagsForNodeGroup`

pkg/nodebootstrap/userdata_al2.go
```diff
@@ -56,3 +61,11 @@ func NewUserDataForAmazonLinux2(spec *api.ClusterConfig, nodeGroupID int) (strin
 	logger.Debug("user-data = %s", body)
 	return body, nil
 }
+
+func makeKubeletParamsAmazon(spec *api.ClusterConfig, ng *api.NodeGroup) []string {
+	params := makeKubeletParamsCommon(spec, ng)
+	if len(ng.Labels) > 0 {
+		params = append(params, fmt.Sprintf("KUBELET_EXTRA_ARGS=--node-labels=%s", ng.Labels))
```
errordeveloper: I don't think `KUBELET_EXTRA_ARGS` will not work as we override the whole [kubelet `ExecStart` already](https://github.com/weaveworks/eksctl/blob/master/pkg/nodebootstrap/assets/10-eksclt.al2.conf).
We should just add the `NODE_LABELS` to `kubelet.env`, and add `--node-labels=${NODE_LABELS}` to `10-eksclt.al2.conf`.

By the way, we have switched to using config file on AL2 (#353), but it's not generated yet (#156) and it's not clear if Ubuntu supports this yet or not (#355). Once we have the config file fully generated, we will be able to get rid of the `kubelet.env`.

pkg/nodebootstrap/userdata_ubuntu.go
```diff
@@ -53,3 +54,11 @@ func NewUserDataForUbuntu1804(spec *api.ClusterConfig, nodeGroupID int) (string,
 	logger.Debug("user-data = %s", body)
 	return body, nil
 }
+
+func makeKubeletParamsUbuntu(spec *api.ClusterConfig, ng *api.NodeGroup) []string {
+	params := makeKubeletParamsCommon(spec, ng)
+	if len(ng.Labels) > 0 {
+		params = append(params, fmt.Sprintf("KUBELET_EXTRA_ARGS=node-labels=%s", ng.Labels))
```
errordeveloper: This work differently on Ubuntu at the moment, I can look into it.

pkg/ctl/create/cluster.go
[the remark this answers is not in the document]
errordeveloper: I think better to just call it `--zones`.

errordeveloper:
I've added a few commit to address some of the things I mentioned. We should rebase this on master and probably squash most of the commit. It's hard to carry @richardcase's commit as a separate thing, I'd recommend squashing everything we have so far.

Also, it'd be great to see an up-to-date checklist of what functionality this PR implements. Also, I'd add 'WIP: ' prefix to the PR title for now.

pkg/nodebootstrap/userdata_al2.go
mumoshu: My idea was `KUBELET_EXTRA_ARGS` works because I added it for both AL2 and Ubuntu in this PR, https://github.com/weaveworks/eksctl/pull/332/files/aee3198a4afda26ec4b433ceb3e3955758ac341b#diff-6955196fab7d3ecf801e80fe59c65f30R32 and in https://github.com/weaveworks/eksctl/pull/332/files/aee3198a4afda26ec4b433ceb3e3955758ac341b#diff-3efef590951f408a697d6c592491d9c4R43.

But `NODE_LABELS` would also be fine for clarity.

I agree we should migrate to KubeletConfiguration, and then to the kubelet componentconfig.

mumoshu:
@errordeveloper Thanks for reviewing! I agree that we should squash everything into a single commit.

Happy to add a WIP prefix to the title!

But what would be the goal image?

I have been thinking this commit should be mergeable if no further critical design/implementation issues are found, given I have written tests, docs, and implemented all the sub-commands as proposed in #281.

Given your review, I'd guess our latest TODOs should be as follows. WDYT?

- An automated integration test(is this a requirement in this PR?
- Addition to humans.txt
- Everything in https://github.com/weaveworks/eksctl/pull/332#issuecomment-448581759 (All good points!

Raffo:
I'm really looking forward to see this merged as I will start putting my hands more on EKS now that it is finally available in Frankfurt!

mumoshu:
> If we were to use something like ng-cdf8139386 & ng-ba8e9fbd08 for the nodegroups, it'd be much easier distinguish the two from each other and the cluster name.

> I think it would look better if it was this:
> ```
> CLUSTER		NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED > CAPACITY	INSTANCE TYPE	IMAGE ID
> cluster-1	nodegroup-0	2018-12-19T11:41:35Z	1		1		1			m5.large	ami-0a9006fb385703b54
> ```

My interpretation of these two points is that we want the nodegroup to be automatically named `ng-WHATEVER_ID_LIKE`.

My initial step towards the direction is 8f52fb4.

It doesn't remove SEQ, but rather use it when a nodegroup name is unspecified. That is, the nodegroup is named `ng-<MaxSeq + 1>` when `--name` flag is omitted.

So let's say you already had 8 stacks, `MaxSeq + 1` should be 8, which results in a nodegroup name `ng-8` when the nodegroup `--name` is omitted:

```console
$ eksctl create nodegroup --cluster wonderful-party-1545311857 --ssh-access
[ℹ]  using "ami-063650732b3e8b38c" for nodes
[ℹ]  creating nodegroup stack "eksctl-wonderful-party-1545311857-nodegroup-ng-8"
```

And `--name whatever` uses the specified nodegroup name:

```console
$ eksctl create nodegroup --cluster wonderful-party-1545311857 --ssh-access --name myng1
[ℹ]  using "ami-063650732b3e8b38c" for nodes
[ℹ]  creating nodegroup stack "eksctl-wonderful-party-1545311857-nodegroup-myng1"
```

SEQ is not shown in `get nodegroup` anymore:

```console
$ eksctl get nodegroup --cluster wonderful-party-1545311857
[ℹ]  stack eksctl-wonderful-party-1545311857-nodegroup-myng1
[ℹ]  stack eksctl-wonderful-party-1545311857-nodegroup-ng-8
CLUSTER				NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
wonderful-party-1545311857	myng1		2018-12-20T15:49:26Z	2		2		2			m5.large	ami-063650732b3e8b38c
wonderful-party-1545311857	ng-8		2018-12-20T15:44:30Z	2		2		2			m5.large	ami-063650732b3e8b38c
```

mumoshu:
@errordeveloper Which default nodegroup name would you prefer, `ng-<random string>` or `ng-<max seq + 1>`?

Also, do you prefer the stack name being `eksctl-wonderful-party-1545311857-myng1` rather than `eksctl-wonderful-party-1545311857-nodegroup-myng1`?

errordeveloper:
@mumoshu it'd be great if you could extend `eksctl delete cluster` to delete nodegroups recursively, would you mind to try doing that in this PR?

mumoshu:
> it'd be great if you could extend eksctl delete cluster to delete nodegroups recursively, would you mind to try doing that in this PR?

@errordeveloper Yes. I was considering it to be a must-have feature, and I have just implemented it :)

> I managed to get a panic like this, any ideas?

You seem to (probably intentionally?) miss the first arg = nodegroup name. And the problem was that there had been no validation error emitted when the nodegroup name is missing. That should have been fixed in the last commit.

> we probably want to use group name for deletion, not stack name...

Definitely! I also noticed that, and the last commit should fix it as well.

> First one with the random string. When user provides the name we shouldn't add the ng- prefix.

OK. I'll remove all the logic and data relevant to SEQ and try using random strings for nodegroup names, when not specified.

So the remaining TODOs on my side would be:

- [ ] Remove SEQ, use randomly generated strings instead for the suffixes of the auto-generated nodegroup names
- [ ] Squash and rebase onto the master, and submit a clean PR with the more complete description and checklist.

mumoshu:
Now a randomly generated nodegroup name should look like:

```
eksctl get nodegroup --cluster floral-creature-1545459723
[ℹ]  stack eksctl-floral-creature-1545459723-nodegroup-ng-5b08d3b6
CLUSTER				NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
floral-creature-1545459723	ng-5b08d3b6	2018-12-22T06:37:07Z	2		2      		2			m5.large	ami-094fa4044a2a3cf52
```

mumoshu:
Just submitted #362, which is a rebased version of this PR.

romil-punetha:
Is there a config file that i can use to create nodegroups?

errordeveloper:
xref https://github.com/weaveworks/eksctl/issues/508#issuecomment-478510905

Issue #306, opened 2018-11-06
Tracking Proposal #002: Existing VPC


errordeveloper:
This is an umbrella issue is for tracking progress on custom VPC features as proposed in #303.

--- 

- [x] `--vpc-{private,public}-subnets` (initial baseline feature): #305
- [ ] `--vpc-id=default` (use default VPC and all of its existing subnets)
- [ ] `--vpc-id=default` with `--vpc-{private,public}-subnets` (use default VPC, but given subnet IDs)
- [ ] `--vpc-id="vpc-xxx"` (use given VPC, create dedicated subnets)
- [ ] `--vpc-id="vpc-xxx"` with `--vpc-{private,public}-subnets` (use given VPC ID, and given subnets - an explicit/logical option)
- [ ] `eksctl <create|get|delete> vpc --name=<vpcName>` (manage shared VPCs)
- [ ] `eksctl create cluster --vpc-name=<vpcName>` (create cluster in a shared VPC)

mumoshu:
I love eksctl's ability to use pre-created subnets. Nice!

However, I'd suggest not to do this:

> --vpc-id="vpc-xxx" (use given VPC, create dedicated subnets)

This seems like a shoot-youself-in-the-foot. I've once encountered my kube-aws user who unintentionally created an unmovable, important AWS resource - RDS DB instance - in the subnet kube-aws has created and managed by CloudFormation. They can never destroy the cfn stack...

martina-if:
Most of the things in this issue have been implemented or there are other issues for tracking them so I will close this one.

PR #379, opened 2018-12-30
Use an existing vpc, generate subnets (WIP)

ozzieba:
### Description
Added support for providing --vpc-id without providing subnets, ie the fourth item in #306. Subnets will be auto-generated as if this is a new VPC. To do this, I had to move things around a bit, to cleanly separate management of the subnets and routing from the VPC itself. Also, since only one InternetGateway can be attached to a VPC at once, if one exists for the given VPC it must also be passed to eksctl so it will be used.

Tested with:
```
docker build -t eks --build-arg EKSCTL_BUILD_IMAGE=eksbuild .
docker run --env AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY --env AWS_REGION=$AWS_REGION eks eksctl create cluster -n $NAME --vpc-cidr $VPC_CIDR --vpc-id $VPC_ID --igw-id $IGW_ID 
```
If the approach makes sense, I can polish the various edges, eg get the CIDR from the VPC, add handling for default VPC, add test cases and documentation etc. But first let me know if this looks useful.
### Checklist

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha3/vpc.go
+++ b/pkg/apis/eksctl.io/v1alpha3/vpc.go
@@ -22,6 +22,7 @@ type (
 		// private subnets or any ad-hoc subnets
 		// +optional
 		ExtraCIDRs []*ipnet.IPNet `json:"extraCIDRs,omitempty"`
+		IGW        InternetGateway
 	}
 	// SubnetTopology can be SubnetTopologyPrivate or SubnetTopologyPublic
 	SubnetTopology string
@@ -31,6 +32,10 @@ type (
 		ID string `json:"id,omitempty"`
 		// +optional
 		CIDR *ipnet.IPNet `json:"cidr,omitempty"`
+	}
+	// InternetGateway holds the ID of the Internet Gateway for that VPC
+	InternetGateway struct {
+		ID string
 	}
 )
 
--- a/pkg/cfn/builder/api_test.go
+++ b/pkg/cfn/builder/api_test.go
@@ -277,7 +277,7 @@ var _ = Describe("CloudFormation template builder API", func() {
 			Status: &api.ClusterStatus{
 				Endpoint:                 endpoint,
 				CertificateAuthorityData: caCertData,
-				ARN: arn,
+				ARN:                      arn,
 			},
 			AvailabilityZones: testAZs,
 			VPC:               testVPC(),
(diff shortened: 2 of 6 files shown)


errordeveloper:
@ozzieba thanks a lot for working on this! I'll review this in once it's ready from your point of view. Also @mumoshu might like to look at this also.

In the meantime, just a heads-up - I'm about to land #381 very soon (potentially today), so you will need to rebase. Also, while working on that PR I just opened #383, in case you might want to tackle that in the process, but not to worry.

mumoshu:
One more workaround would be to allow eksctl create create vpc and subnets and relevant aws resources only. So that those resources can be reused by later eksctl calls. This covers more use-cases while avoiding to maintain both eksctl and the reference terraform project.

ozzieba:
@mumoshu Thanks for looking over this!

My basic thinking is that eksctl is meant to allow users to bring up an EKS cluster easily from scratch, without having to worry about details like VPC, subnets, routing, ACLs, etc, unless/until they have a special need for customization.

Unfortunately in my particular case I do need to use an existing VPC, so I can mount an existing EFS volume from a GPU instance (see [here](https://docs.aws.amazon.com/efs/latest/ug/manage-fs-access-vpc-peering.html)). I imagine there are other reasons people may want to reuse VPCs, while having everything else auto-generated from scratch. It seems to me this shouldn't add too much of a maintenance burden, since the interface to existing resources (VPC and IGW) is still fairly small.

Obviously, I could simply create the subnets etc manually, or use Terraform / Cloudformation; but I think explicitly letting eksctl manage them is likely to be less error-prone and cleaner. In particular, I'd expect that having users manage their own Terraform / Cloudformation templates would result in more issues, like users forgetting to add an InternetGateway or correct routing. Indeed, this is a significant reason to use eksctl in the first place, rather than the AWS-provided Cloudformation [scripts](https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html). I also expect this somewhat reduces the likelihood of users adding unrelated resources to subnets that should be reserved for EKS. Likewise, letting eksctl clean up unneeded resources when deleting a cluster is convenient.

To further reduce errors, eksctl could attempt to check for configurations that won't work, eg missing routing, subnets that already contain other resources, etc, and warn the user before creating the cluster. In extreme cases, eksctl could delete the cloudformation stack while [retaining](https://aws.amazon.com/premiumsupport/knowledge-center/delete-cf-stack-retain-resources/) resources that are still needed.

errordeveloper:
@ozzieba have you tried rebasing yet?

ozzieba:
Sorry, it's been a busy week, I hope to get to this in the next few days

errordeveloper:
@ozzieba just let me know if you need help, we can talk on [slack](https://slack.weave.works).

ozzieba:
I've done the rebase, unfortunately I don't have time today to get testing and docs done properly, will try during the week.

errordeveloper:
@ozzieba I'm gonna have to close this, if you want to come back to it - please go ahead and open a new PR! :)

Issue #19, opened 2018-06-04
Cluster API CLI


errordeveloper:
We would like to implement [Cluster API](https://github.com/kubernetes-sigs/cluster-api). One of the modes in which one can use Cluster API would be `eksctl apply -f cluster.yaml`, this would let users create and update clusters using a declarative format of a Cluster API object.
Another mode of using Cluster API would be via a controller running in a management cluster (#20).

errordeveloper:
## What needs to be done here?

- [ ] `eksctl create cluster -f <configFile>` – #381
- [ ] `eksctl create  nodegroup -f <configFile>`
- [ ] `eksctl create -f <configFile>`
- [ ] `eksctl get/delete cluster -f <configFile>`
- [ ] `eksctl get/delete  nodegroup -f <configFile>`
- [ ] `eksctl get/delete -f <configFile>`
- [ ] `eksctl utils upgrade-{cluster,nodegroup} -f <configFile>` (in case we may need it as a temporary solution for upgrades, before we can implement fully idempotent `eksctl apply`)
- [ ] `eksctl apply -f <configFile>`

mumoshu:
@errordeveloper It seems like we have `eksctl upgrade nodegroup` since #1566

michaelbeaumont:
I suspect this issue isn't relevant anymore, there's already the [AWS provider](https://github.com/kubernetes-sigs/cluster-api-provider-aws/). We have an issue for declarative apply https://github.com/weaveworks/eksctl/issues/462

michaelbeaumont:
Closing for now

PR #381, opened 2019-01-02
Create cluster via config file

### LINKED ISSUE #377: automatic label for nodegroups
We currently use `eksctl.cluster.k8s.io/v1alpha1/nodegroup-name` and `eksctl.cluster.k8s.io/v1alpha1/cluster-name` for EC2 tags, and the later certainly needs to get mapped to a label also. It'd be nice to make this more user-friendly.
Perhaps we should call it `eksctl.io/v1alpha2/nodegroup-name` to start with, and see if people would be happy with that or they might want to have shorter alias (like `nodegroup-name`).

errordeveloper:
### Description

This is initial step towards #19. Let's add this in to start with and aim to evolve it as an experimental feature.

This change adds some of the Kubernetes API tooling, and adds JSON tags to `api.ClusterConfig`.
From user's perspective, there is now a `--config-file/-f` flag in `eksctl create cluster`. The format of the object right now is based directly on `api.ClusterConfig`, we may eventually mould it into Cluster API `ProvideSpec`, but for now let's keep it simple.

### Checklist

Closes #377.


errordeveloper:
TODOs:
- [ ] squash
- [ ] re-read the diff
- [ ] complete iniline TODOs

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -122,32 +129,57 @@ type ProviderConfig struct {
...
+	VPC *ClusterVPC `json:"vpc,omitempty"`
 
-	NodeGroups []*NodeGroup
+	// +optional
+	NodeGroups []*NodeGroup `json:"nodeGroups,omitempty"`
 
+	// +optional
+	AvailabilityZones []string `json:"availabilityZones,omitempty"`
+
+	// TODO: refactor and move IAM addons to nodegroup
```
errordeveloper: Will save for another PR.

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -181,31 +215,47 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
...
+	PrivateNetworking bool `json:"privateNetworking"`
+
+	DesiredCapacity int `json:"desiredCapacity"`
+	// +optional
+	MinSize int `json:"minSize,omitempty"`
+	// +optional
+	MaxSize int `json:"maxSize,omitempty"`
+
+	// +optional
+
```
errordeveloper: blank line

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -181,31 +215,47 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
...
+	VolumeSize int `json:"volumeSize"`
+	// +optional
+	MaxPodsPerNode int `json:"maxPodsPerNode,omitempty"`
+
+	Labels NodeLabels
```
errordeveloper: Also optional, missing JSON tag.

pkg/ctl/create/cluster.go
```diff
@@ -91,23 +109,129 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
...
+		if err != nil {
+			return err
+		}
+
+		cfg, ok := obj.(*api.ClusterConfig)
+		if !ok {
+			return fmt.Errorf("decoded object of wrong type")
+		}
+
+		// config
```
errordeveloper: useless comment

errordeveloper:
This PR now has a fix for #385 (734f097).

errordeveloper:
I'm rather happy with this, but would like to re-test `examples/05-two-nodegroups.yaml` once again.

errordeveloper:
Integration tests passed, `./eksctl create cluster -f ./examples/05-two-nodegroups.yaml` passed and nodes have right name labels set correctly.

pkg/eks/api.go
```diff
@@ -195,8 +221,23 @@ func (c *ClusterProvider) EnsureAMI(version string, ng *api.NodeGroup) error {
...
+func (c *ClusterProvider) SetNodeLabels(ng *api.NodeGroup, meta *api.ClusterMeta) error {
+	if ng.Labels == nil {
+		ng.Labels = make(api.NodeLabels)
+	}
+
+	ng.Labels[api.ClusterNameLabel] = meta.Name
+	ng.Labels[api.NodeGroupNameLabel] = ng.Name
+
+	for l := range ng.Labels {
+		if len(strings.Split(l, "/")) > 2 {
```
errordeveloper: We should actually call proper validation function here.

errordeveloper: https://github.com/weaveworks/eksctl/issues/394

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -181,31 +241,48 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
...
+	// +optional
+	InstanceType string `json:"instanceType,omitempty"`
+	// +optional
+	AvailabilityZones []string `json:"availabilityZones,omitempty"`
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
+	// +optional
+	PrivateNetworking bool `json:"privateNetworking"`
+
+	DesiredCapacity int `json:"desiredCapacity"`
```
mumoshu: I believe `DesiredCapacity` should be optional. When an external autoscaler that controls the capacity, like`cluster-autoscaler`, is enabled, you don't want to re-add/delete ec2 instances removed/added by the autoscaler.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-desiredcapacity

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -181,31 +241,48 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
...
+	// +optional
+	MaxPodsPerNode int `json:"maxPodsPerNode,omitempty"`
+
+	// +optional
+	Labels NodeLabels `json:"labels,omitempty"`
+
+	// +optional
+	PolicyARNs []string `json:"policyARNs,omitempty"`
+	// +optional
+	InstanceRoleARN string `json:"instanceRoleARN,omitempty"` // TODO: read-only
```
mumoshu: I wish this field kept writable.

The use-case is that the user pre-creates role used by nodes to provide proper necessary permissions  including iam policy statements and trust relationship, to make kube2iam/kiam/[aws-iam-controller](mikkeloscar/kube-aws-iam-controller) work.

pkg/apis/eksctl.io/v1alpha3/types.go
errordeveloper: At the moment this is entirely open, and we have #385, please suggest there or feel free to open another issue to discuss.

pkg/apis/eksctl.io/v1alpha3/types.go
errordeveloper: Very good point, but isn't `PolicyARNs` actually sufficient for that?

Actually, the TODO needs to be removed. I've added it initially thinking that there was some kind of Kubernetes tag (like `+optional`) that sets a field as read-only, but after digging-in I failed to find such tag. So TODO is to remove the TODO comment :)

mumoshu: Thanks for the response! `PolicyARNs` has potential to partly solve the problem. I've written a more complete description in #398

errordeveloper:
@mumoshu thanks for looking over this PR! I've removed stale TODO note, and made `DesiredCapacity` field `+optional`.

pkg/apis/eksctl.io/v1alpha3/types.go
mumoshu: Opened #395 for this 👍

pkg/apis/eksctl.io/v1alpha3/types.go
```diff
@@ -181,31 +227,49 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
...
+	InstanceType string `json:"instanceType,omitempty"`
+	// +optional
+	AvailabilityZones []string `json:"availabilityZones,omitempty"`
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
+	// +optional
+	PrivateNetworking bool `json:"privateNetworking"`
+
+	// +optional
+	DesiredCapacity int `json:"desiredCapacity"`
```
marccarre: Could the "desired" capacity not be met?
Would it be clearer to have something named like, e.g., the below?
```
Count int `json:"count"`
```

pkg/cfn/builder/outputs.go
```diff
@@ -141,7 +141,7 @@ func (r *resourceSet) GetAllOutputs(stack cfn.Stack, obj interface{}) error {
 			return errors.Wrap(err, "processing stack outputs")
 		}
 	}
-	logger.Debug("obj = %#v", obj)
+	//logger.Debug("obj = %#v", obj)
```
marccarre: Should we just remove these commented-out lines?

pkg/ctl/create/cluster.go
```diff
@@ -92,26 +103,133 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	return cmd
 }
 
-func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, ng *api.NodeGroup, nameArg string) error {
+func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, ng *api.NodeGroup, nameArg string, cmd *cobra.Command) error {
```
marccarre: Nit: this is a very long function. Is there a way to split it in smaller ones? (in order to reduce scope & mental load)

errordeveloper: Yes, indeed! It would have to be re-factored very soon, e.g. to improve testing (see #393).

pkg/apis/eksctl.io/v1alpha3/types.go
errordeveloper: Yes, we had it as `NodeCount` at the top level some time ago, but now it's called `DesiredCapacity` to match the AWS API. However, the flag is still `--nodes`, so there is a little inconsistency there too. Some refactoring is pending here also (#385).

pkg/cfn/builder/outputs.go
errordeveloper: Yes, good idea. I just thought these maybe useful another time, but these are rather trivial, and some pre-date unit tests.

pkg/ctl/create/cluster.go
errordeveloper: I opened https://github.com/weaveworks/eksctl/issues/399.

pkg/apis/eksctl.io/v1alpha3/types.go
marccarre: Fair enough 😅 
Matching the AWS API probably takes precedence over any other argument/preference 👍

Issue #333, opened 2018-12-04
Add `node-volume-type` to create cluster


arbourd:
**Why do you want this feature?**
When using `node-volume-size` flag, eksctl creates a new volume with a default set volume type. In the case of EBS-only machines, like c5xlarge, this defaults to `standard`.

Adding a new flag like `node-volume-type` would allow the user to select what type of EBS volume they would like (standard, gp2, etc)

**What feature/behavior/change do you want?**
- The addition of a new flag `node-volume-type` which can set the type

_Implementation Concerns_
- How does this work with nodes that already have volumes like c5d?
- Does setting a type without a size create that volume or must both flags be present? (Default size is 0)

mumoshu:
@arbourd This sounds like a necessary feature! I work on it after #332.

> Does setting a type without a size create that volume or must both flags be present? (Default size is 0)

I'd be ok with allowing to omit the size, so that it uses a sane default.

errordeveloper:
> so that it uses a sane default

Yes, but that means we need to define a default, and use-cases vary dramatically, so I am honestly not sure what we could possibly pick. I'd rather simply requite both flags. Also current default for `--node-volume-size` is 0, so we would have to introduce another level of defaulting (outside fo Cobra), it would also make flag usage summary more complex to read.

mumoshu:
Makes sense.

Requiring both flags when either is specified sounds good.

mukaibot:
hey @mumoshu I know you said that you were going to work on it but I needed it and it was a small thing in the end.. :)

PR #403, opened 2019-01-08
Allow configuring node volume type

mukaibot:
### Description
Allow specifying the node volume type when configuring a non-default node volume size. Related to #333

### Checklist
- [X] All tests passing (i.e. `make test`)
- [X] Added/modified documentation as required (such as the README)
- [X] Added yourself to the `humans.txt` file

### DIFF
--- a/pkg/ctl/cmdutils/nodegroup.go
+++ b/pkg/ctl/cmdutils/nodegroup.go
@@ -1,6 +1,9 @@
 package cmdutils
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/spf13/pflag"
 
 	"github.com/weaveworks/eksctl/pkg/ami"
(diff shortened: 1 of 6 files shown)


pkg/ctl/cmdutils/nodegroup.go
```diff
@@ -22,6 +22,7 @@ func AddCommonCreateNodeGroupFlags(fs *pflag.FlagSet, p *api.ProviderConfig, cfg
 	fs.IntVarP(&ng.MaxSize, "nodes-max", "M", 0, "maximum nodes in ASG")
 
 	fs.IntVarP(&ng.VolumeSize, "node-volume-size", "", 0, "Node volume size (in GB)")
+	fs.StringVarP(&ng.VolumeType, "node-volume-type", "", "", "Node volume type (one of gp2, io1, st1, sc1 or standard)")
```
errordeveloper: Please add a default value to what it was before, I think it has effectively defaulted to `gp2`.

errordeveloper: In fact, there will need to be more change for default via config file, it's gonna be easier for me to ammend your branch, unless you'd like to do it yourself (happy to explain).

mukaibot: Nah I'm cool for you to amend the branch @errordeveloper ! I'm not really a Go developer! Thank you :)

errordeveloper: I can see that the thread is not too clear about what the actual defaults are supposed to be.

PR #405, opened 2019-01-08
Add exec helper in integration tests

errordeveloper:


integration/creategetdelete_test.go
```diff
@@ -48,6 +48,17 @@ func newKubeTest() (*harness.Test, error) {
...
 }
 
+func eksctl(args ...string) {
+	command := exec.Command(eksctlPath, args...)
+	fmt.Fprintf(GinkgoWriter, "calling %q with %v", eksctlPath, args)
+	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	if err != nil {
+		Fail(fmt.Sprintf("error starting process: %v", err), 1)
+	}
+	session.Wait(scaleTimeout)
```
errordeveloper: It runs, but doesn't work properly and timesout too early.

PR #407, opened 2019-01-08
Move status fields from ClusterConfig to ClusterStatus

### LINKED ISSUE #363: refactor status fields
We have a few status fields, those should go under `status`.

```
status:
  endpoint: "..."
  certificateAuthorityData: "..."
  arn: "..."
```

errordeveloper:
### Description

Fixes #363.

### Checklist


pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go
```diff
@@ -55,10 +55,10 @@ func (in *ClusterConfig) DeepCopyInto(out *ClusterConfig) {
...
 		copy(*out, *in)
 	}
-	if in.CertificateAuthorityData != nil {
-		in, out := &in.CertificateAuthorityData, &out.CertificateAuthorityData
-		*out = make([]byte, len(*in))
-		copy(*out, *in)
+	if in.Status != nil {
+		in, out := &in.Status, &out.Status
+		*out = new(ClusterStatus)
+		(*in).DeepCopyInto(*out)
```
rndstr: isn't this block the same as
```
out.Status = new(ClusterStatus)
in.Status.DeepCopyInto(out.Status)
```

pkg/nodebootstrap/userdata_al2.go
```diff
@@ -15,7 +15,7 @@ func makeAmazonLinux2Config(spec *api.ClusterConfig, ng *api.NodeGroup) (configF
 		return nil, err
 	}
 
-	if spec.CertificateAuthorityData == nil || len(spec.CertificateAuthorityData) == 0 {
+	if spec.Status.CertificateAuthorityData == nil || len(spec.Status.CertificateAuthorityData) == 0 {
```
rndstr: can be just
```
if len(spec.Status.CertificateAuthorityData) == 0 {
```

pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go
```diff
@@ -137,6 +137,27 @@ func (in *ClusterMeta) DeepCopy() *ClusterMeta {
 	return out
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *ClusterStatus) DeepCopyInto(out *ClusterStatus) {
+	*out = *in
+	if in.CertificateAuthorityData != nil {
+		in, out := &in.CertificateAuthorityData, &out.CertificateAuthorityData
```
rndstr: it seems that you wouldn't need to assign the addresses of slices here since you dereference their use below

pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go
errordeveloper: Maybe, but this code is generated by [`deepcopy-gen`](https://github.com/kubernetes/code-generator/tree/master/cmd/deepcopy-gen).

pkg/nodebootstrap/userdata_al2.go
errordeveloper: 👍

pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go
errordeveloper: Also, `deepcopy-gen`, so no need for us to worry.

errordeveloper:
@rndstr thanks for the review!

pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go
rndstr: :+1: I was wondering where it was _autogenerated_ from, maybe put that in the comment if you happen to come across next time

errordeveloper: Generaotor does leave a comment here: https://github.com/weaveworks/eksctl/blob/aca8ec28fa192b2f0653fb1e25ca3a756d8f6ef2/pkg/apis/eksctl.io/v1alpha3/zz_generated.deepcopy.go#L19

But there is a lot of noise, of course... And less obvious from the diff view. Not sure if we need to do anything about this right now, like automating code generation in CI.

PR #412, opened 2019-01-09
feat: associate additional security groups on node creation

### LINKED ISSUE #397: feat: associate additional "glue" security groups on node creation
One of common use-cases I've figured out while maintaining and using kube-aws is allowing the pods and containers on the nodes to access existing AWS resources secured by security groups, including RDS db instances, ELBs, EC2 instances, ENIs, and so on.

A kind of security groups I've calling "glue" works very well here. A `glue` security security is created and the security group associated to existing AWS resources is configured to accept ingress traffic from the glue sg, manually by the user of another tool like terraform/cloudformation, before the nodes are provisioned by a tool like eksctl.

You, as the user, just tells the cluster provisioning tool which glue sg to be associated to the nodes. This way, the cluster provisioning doesn't need to be enhanced to support every protocol and every use-case regarding access control. The user can just creates the glue sgs in their preferred way, the glue sg id is provided, eksctl do the rest.

Implementation-wise, I'd enhance the `NodeGroup` struct https://github.com/weaveworks/eksctl/pull/381/files#diff-5f6afeab74ee405a58ea460f4a061ac7R285 to have the `SecurityGroupIds` or `SecurityGroups` that accepts additional security group ids to be associated to the ec2 instances. A corresponding command-line flag for `create cluster` and `create nodegroup` would also be a good-to-have.

mumoshu:
Adds the `--node-security-groups` flag, along with the corresponding config key `securityGroups`, that accepts ARNs of the additional security groups attached to the nodes.

The additional security groups can be used for allowing extra egress from/ingress to pods running on the nodes.

Resolves #397

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha3/types.go
+++ b/pkg/apis/eksctl.io/v1alpha3/types.go
@@ -252,6 +252,8 @@ type NodeGroup struct {
 	// +optional
 	AvailabilityZones []string `json:"availabilityZones,omitempty"`
 	// +optional
+	SecurityGroups []string `json:"securityGroups,omitempty"`
+	// +optional
 	Tags map[string]string `json:"tags,omitempty"`
 	// +optional
 	PrivateNetworking bool `json:"privateNetworking"`
--- a/pkg/cfn/builder/vpc.go
+++ b/pkg/cfn/builder/vpc.go
@@ -135,6 +135,12 @@ func (n *NodeGroupResourceSet) addResourcesForSecurityGroups() {
 	})
 	n.securityGroups = []*gfn.Value{refSG}
 
+	if len(n.spec.SecurityGroups) > 0 {
+		for _, arn := range n.spec.SecurityGroups {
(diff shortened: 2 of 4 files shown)


mumoshu:
I've created a brand new cluster with eksctl built from the branch running `eksctl create cluster --node-security-groups SG_ID`. All the EC2 instances in the node group has the security group attached, and the stack template includes the specified security group ID in it:

```
{"AssociatePublicIpAddress":true,"IamInstanceProfile":{"Ref":"NodeInstanceProfile"},"ImageId":"ami-094fa4044a2a3cf52","InstanceType":"m5.large","SecurityGroups":[{"Ref":"SG"},"sg-0f4e477e7e496aa26"], ...
```

mumoshu:
Verified this to work by (1) manually creating a new cluster with `--node-security-groups sg-abcd1234` and (2) confirming the specified security group is attached to the EC2 instances in the nodegroup/ASG.

whereisaaron:
I was looking for this option, and it is already done, which is great, but it is not  currently documented in the `README.md`

PR #415, opened 2019-01-10
fix duplicate tags set via --tags

### LINKED ISSUE #413: duplicate tags when set via --tags
I observed this in the output of our integration tests:

```
2019-01-09T15:54:05Z [▶]  CreateStackInput = {
  Capabilities: ["CAPABILITY_IAM"],
  StackName: "eksctl-hilarious-rainbow-1547049243-cluster",
  Tags: [{
      Key: "eksctl.cluster.k8s.io/v1alpha1/cluster-name",
      Value: "hilarious-rainbow-1547049243"
    },{
      Key: "eksctl.cluster.k8s.io/v1alpha1/description",
      Value: "eksctl integration test"
    },{
      Key: "eksctl.cluster.k8s.io/v1alpha1/description",
      Value: "eksctl integration test"
    }],
  TemplateBody: ....
```

mumoshu:
Fixes #413

### DIFF
--- a/pkg/cfn/manager/cluster.go
+++ b/pkg/cfn/manager/cluster.go
@@ -18,7 +18,8 @@ func (c *StackCollection) CreateCluster(errs chan error, _ interface{}) error {
 		return err
 	}
 
-	return c.CreateStack(name, stack, c.spec.Metadata.Tags, nil, errs)
+	// Unlike with `CreateNodeGroup`, all tags are already set for the cluster stack
+	return c.CreateStack(name, stack, nil, nil, errs)
 }
 
 // DeleteCluster deletes the cluster


mumoshu:
Verified this to work as expected, by creating a brand new cluster and looking into the stack tags to contain the tag specified via `--tags`. This should be good to go.

pkg/cfn/manager/api.go
```diff
@@ -92,6 +92,8 @@ func (c *StackCollection) doCreateStackRequest(i *Stack, templateBody []byte, ta
 // any errors will be written to errs channel, when nil is written,
 // assume completion, do not expect more then one error value on the
 // channel, it's closed immediately after it is written to
+//
+// StackCollection doesn't dedup the tags passed as an argument and the StackCollection's shared tags.
```
errordeveloper: I don't think this is really important, let's keep documentation here as is, otherwise there is much more that can be documented about this or any other function.

mumoshu: Understood! Fixed in the latest commit.

pkg/cfn/manager/cluster.go
```diff
@@ -18,7 +18,12 @@ func (c *StackCollection) CreateCluster(errs chan error, _ interface{}) error {
 		return err
 	}
 
-	return c.CreateStack(name, stack, c.spec.Metadata.Tags, nil, errs)
+	// All the tags contained in `c.spec.Metadata.Tags` are already contained in the shared tags.
+	// Providing it to CreateStack results in duplicated stack tags.
+	// See https://github.com/weaveworks/eksctl/issues/413 for more details.
+	tags := map[string]string{}
+
+	return c.CreateStack(name, stack, tags, nil, errs)
```
errordeveloper: I think passing `nil` and simple one line comment will do
```suggestion
    // Unlike with `CreateNodeGroup`, all tags are already set for the cluster stack
	return c.CreateStack(name, stack, nil, nil, errs)
```

mumoshu: Much simpler, and makes sense much more! Thanks for the suggestion.

pkg/cfn/manager/cluster.go
```diff
@@ -18,7 +18,13 @@ func (c *StackCollection) CreateCluster(errs chan error, _ interface{}) error {
 		return err
 	}
 
-	return c.CreateStack(name, stack, c.spec.Metadata.Tags, nil, errs)
+	// All the tags contained in `c.spec.Metadata.Tags` are already contained in the shared tags.
+	// Providing it to CreateStack results in duplicated stack tags.
+	// See https://github.com/weaveworks/eksctl/issues/413 for more details.
+	tags := map[string]string{}
+
```
errordeveloper: Thanks for updating this, but please remove the lines above. For the future, no need to reference issues in comments.

Issue #414, opened 2019-01-09
DNS not working as expected


schottsfired:
**What happened?**
Cluster DNS resolution isn't working on a nodegroup:

I deployed an EKS cluster as follows:
`eksctl create cluster --name foo --tags "key=val" --region us-east-1 --zones us-east-1a,us-east-1b --nodegroup-name foo --node-type m5.large --nodes-min 2 --nodes-max 4 --ssh-access --ssh-public-key=foo --node-ami auto --node-private-networking --node-labels "partition=foo" --asg-access --cfn-role-arn arn:aws:iam::xyz`

Then deployed a nodegroup as follows:
`eksctl create nodegroup --cluster foo --region us-east-1 --name foo-bar --node-type m5.large --nodes 1 --nodes-min 1 --nodes-max 2 --ssh-access --ssh-public-key=foo --node-ami auto --node-private-networking --node-labels "partition=bar" --asg-access --cfn-role-arn arn:aws:iam::xyz`

Then ran `kubectl create -f` with [this yaml](https://gist.github.com/schottsfired/bf5059d1c5647f850839cb23de1ca5ba).

Then ran `kubectl exec -ti busybox -- nslookup kubernetes.default` to test DNS on a `partition: foo` node and the output is OK:
```
Server:    10.100.0.10
Address 1: 10.100.0.10 ip-10-100-0-10.ec2.internal

Name:      kubernetes.default
Address 1: 10.100.0.1 ip-10-100-0-1.ec2.internal
```
But after modifying the pod yaml to run on a `partition: bar` node, the above command fails:
```
Server:    10.100.0.10
Address 1: 10.100.0.10

nslookup: can't resolve 'kubernetes.default'
command terminated with exit code 1
```

**What you expected to happen?**
I expected the pod on the "bar" node to be able to resolve cluster DNS.

**How to reproduce it?**
See above

**Anything else we need to know?**
eksctl 0.1.17 (latest) installed via homebrew on OSX Mojave.

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.17"}
$ uname -a
Darwin dschott-mbp.local 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64
$ kubectl version
[...]

Issue #370, opened 2018-12-22
drain nodegroup on deletion


errordeveloper:
After #369 lands we should add drain capability on deletion, to cater for upgrades nicely (#100, #357, #348). It seems plausible to add `--drain` and keep on by default.

mumoshu:
@errordeveloper `--drain` sounds great if what it does is basically calling `kubectl drain` for each node in the nodegroup being deleted!

Adding a node-drainer addon just makes the flag optional, and may also adds the possibility to handle instance termination external to eksctl, like spot instance termination. Am I correct?

errordeveloper:
Looks like we could try using [`github.com/kubernetes/autoscaler/cluster-autoscaler/utils/drain`](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go) package, or otherwise we can refactor [kubectl code](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain/drain.go) into a library that we can then call.

errordeveloper:
I've started working on https://github.com/kubernetes/kubernetes/pull/72827.

errordeveloper:
To facilitate for upgrades, I will try and get this implemented before I finish [kubernetes/kubernetes#72827](https://github.com/kubernetes/kubernetes/pull/72827).

whereisaaron:
Discussion and builds of automatic node draining for ASG in-scaling and Spot Fleet shutdowns, similar to `kube-aws` https://github.com/awslabs/amazon-eks-ami/issues/66

errordeveloper:
@whereisaaron that seems to be more related to #586, don't you think?

whereisaaron:
Oh yes, didn’t spot it. Moved it over.

PR #422, opened 2019-01-11
Implement nodegroup checks

errordeveloper:
### Description

This is a follow up to #418, which solved the DNS access problem (#414) for new clusters.
This change adds checks so  that we can detect the DNS issue in existing clusters.

### Checklist


errordeveloper:
We should be able to cut a release after this, but I might see if #370 would be easy enough.

PR #424, opened 2019-01-11
Update README.md

prageethw:
### Description
added autoscaling option to the readme.

### Checklist
tested it works fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,17 @@ To delete a nodegroup, run:
 ```
 eksctl delete nodegroup --cluster=<clusterName> --name=<nodegroupName>
 ```
+### Enable Autoscaling
+
+```bash
+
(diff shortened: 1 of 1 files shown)


prageethw:
\ @errordeveloper

README.md
```diff
@@ -208,6 +208,17 @@ To delete a nodegroup, run:
 
 ```
 eksctl delete nodegroup --cluster=<clusterName> --name=<nodegroupName>
+```
+### Enable Autoscaling
+
+```bash
+
+eksctl create cluster --asg-access=true
```
mumoshu: Nit but I think `=true` is redundant here :)

errordeveloper: I'll fix this on master, thanks!

PR #431, opened 2019-01-14
Set tags for cluster autoscaler to discover nodegroups with AGS access

errordeveloper:
### Description

This adds automatic EC2 tags that cluster autoscaler examples mandate.

cc @prageethw @mumoshu 

### Checklist

### DIFF
--- a/pkg/cfn/builder/nodegroup.go
+++ b/pkg/cfn/builder/nodegroup.go
@@ -146,6 +146,32 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
 			},
 		}
 	}
+	tags := []map[string]interface{}{
+		{
+			"Key":               "Name",
+			"Value":             fmt.Sprintf("%s-%s-Node", n.clusterSpec.Metadata.Name, n.nodeGroupName),
+			"PropagateAtLaunch": "true",
+		},
+		{
+			"Key":               "kubernetes.io/cluster/" + n.clusterSpec.Metadata.Name,
+			"Value":             "owned",
+			"PropagateAtLaunch": "true",
+		},
+	}
+	if n.spec.IAM.WithAddonPolicies.AutoScaler {
+		tags = append(tags,
+			map[string]interface{}{
+				"Key":               "k8s.io/cluster-autoscaler/enabled",
+				"Value":             "true",
+				"PropagateAtLaunch": "true",
+			},
+			map[string]interface{}{
+				"Key":               "k8s.io/cluster-autoscaler/" + n.clusterSpec.Metadata.Name,
+				"Value":             "owned",
+				"PropagateAtLaunch": "true",
+			},
+		)
+	}
 	n.newResource("NodeGroup", &awsCloudFormationResource{
 		Type: "AWS::AutoScaling::AutoScalingGroup",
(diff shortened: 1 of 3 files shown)


pkg/cfn/builder/nodegroup.go
```diff
@@ -146,6 +146,32 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
...
+	}
+	if n.spec.IAM.WithAddonPolicies.AutoScaler {
+		tags = append(tags,
+			map[string]interface{}{
+				"Key":               "k8s.io/cluster-autoscaler/enabled",
+				"Value":             "true",
+				"PropagateAtLaunch": "true",
+			},
+			map[string]interface{}{
+				"Key":               "k8s.io/cluster-autoscaler/" + n.clusterSpec.Metadata.Name,
```
mumoshu: You can safely omit this tag in favor of `kubernetes.io/cluster/<CLUSTER NAME>` tag that is already there.

cluster-autoscaler itself isn't actually hard-coded to use the key for identifying the cluster.

In addition to that, the cluster-autoscaler chart is hard-coded to use the `kubernets.io/cluster/CLUSTER_NAME` tag :)

https://github.com/helm/charts/blob/d845bfa6a234817d8a184db1f3860714da13c04f/stable/cluster-autoscaler/templates/deployment.yaml#L49-L52

errordeveloper: Thanks! Sounds like a good idea to match the chart. By the way, there seems to be a new line, have you tested the chart?
```
            {{- if .Values.autoDiscovery.clusterName }}
            - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,kubernetes.io/cluster/
              {{- .Values.autoDiscovery.clusterName }}
            {{- end }}
```

> You can safely omit this tag ...

I'd rather keep both, the chart has them both of them.

errordeveloper:
I've added unit tests to cover these tags.

pkg/cfn/builder/nodegroup.go
mumoshu: Honestly, I have never used the chart!

I only used the almost equivalent CA deployment managed by kube-aws, and leaving the above comment based on my experience as the original author of the autodiscovery feature.

It looks ok to me though as

```
- --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,kubernetes.io/cluster/
              {{- .Values.autoDiscovery.clusterName }}
```

is equivalent to:

```
- --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,kubernetes.io/cluster/{{ .Values.autoDiscovery.clusterName }}
```

due to that the `{{-` symbol that removes any heading whitespaces(Am I missing something?

> I'd rather keep both, the chart has them both of them.

To me, the chart seemed to have `k8s.io/cluster-autoscaler/enabled`, but lack `k8s.io/cluster-autoscaler/<cluster-name>`(if that's what we're talking about

prageethw:
I just tested with official helm chart, can confirm it works as expected. Though i agree k8s.io/cluster-autoscaler/<cluster-name> is redundant by k8s.io/cluster/<cluster-name> in default setting.

Issue #419, opened 2019-01-10
review cross-nodegroup ingress rules


errordeveloper:
At present, we only allow access to majority of ports below 1025 within a single nodegroup. We need to review this, as user may wish to run pods that listen on port 80, for example. We need probably need to open this up, perhaps we can use a shared SG in the cluster stack, or simply allow access on the basis of VPC CIDR (which is what we have for DNS - #418).

errordeveloper:
I think there should be no difference between node-to-node rules within a nodegroup vs accross nodegroups. It's actually a question of what node ports should be open between nodes. At present 

However, I do think we should provide a way for users to implement advanced security by blocking all ports that are not used to provide vital functionality. It's unfortunate that pods cannot be segregated into an SG of their own that would be separate from host network SG.
Also, it should be noted that with an overlay network we would be able to close most of the node ports, and in case of Weave Net, overlay traffic can be actually encrypted.

In any case, the question here should be really about which ports should be kept close between nodes by default. At the moment we open SSH (only when needed), we always open DNS and then only high ports are open between nodegroups and to the outside world.

I believe the case of outside world should remain the same, and we should actually add a mode where none of the ports are open, yet without private resorting to private subnets (because NAT doesn't come for free - #392).

But certainly we should allow use of all ports starting with e.g. starting with 21, so that one can run an FTP server if they must, and we can specify provide a config parameter that lets user specify a range of ports or something like that. But I don't think these ports have to be open outside the VPC CIRD is sufficient. As I said above, I believe we should only open higher ports outside the VPC by default, and we do need a way for users to close those easily.

errordeveloper:
> I think there should be no difference between node-to-node rules within a nodegroup vs accross nodegroups.

I suspect there is a bug at the moment, but I've not check it. I think right now you can run a pod listening on e.g. port 80 and you will get access to it between nodes in one nodegroup, but not between nodegroups.

errordeveloper:
I think we should have a shared SG, as well as per-nodegroup SG. 

By default:
- shared SG would allow inter-node traffic on all ports and all protocols, no CIDR-based rules, it would live inside the cluster stack
- per-nodegroup SGs would be used to only control the following types of traffic:
  - inter-VPC traffic (e.g. for SSH)
  - access to high ports from outside the cluster
  - bindings for control-plane SG
  - opening any ports to outside world

Options for sealing a nodegroup:
- opt-out from shared SG
- custom ingress rules

[...]

PR #438, opened 2019-01-15
Shared nodegroup SG

errordeveloper:
### Description

Initial changes to addresses cross-nodegroup SG configuration (see #419).

We will need to add something like `eksctl utils create-shared-node-security-group` to help users, it's still TBD. This version only adds checks that produce error messages.

There are two main secenarios:

1. old cluster without a shared SG: cannot create new nodegroup

```
 [0] >> ./eksctl create nodegroup --nodes=1 --region=eu-north-1  --cluster=test-shared-nodegroup-check
[ℹ]  using region eu-north-1
[ℹ]  nodegroup "ng-6d61935c" will use "ami-06ee67302ab7cf838" [AmazonLinux2/1.11]
[✖]  cluster compatibility check failed: cluster "test-shared-nodegroup-check" does not not have shared node security group
 [1] >>
```

2. cluster has shared SG, but some nodegroups are still old - not the end of the world issue

```
 [0] >> ./eksctl create nodegroup --nodes=1 --region=eu-north-1  --cluster=test-shared-nodegroup-new  
[ℹ]  using region eu-north-1
[ℹ]  nodegroup "ng-74c42138" will use "ami-06ee67302ab7cf838" [AmazonLinux2/1.11]
[ℹ]  will create a Cloudformation stack for nodegroup ng-74c42138 in cluster test-shared-nodegroup-new
[ℹ]  creating nodegroup stack "eksctl-test-shared-nodegroup-new-nodegroup-ng-74c42138"
[ℹ]  nodegroup "ng-74c42138" has 0 node(s)
[ℹ]  waiting for at least 1 node(s) to become ready in "ng-74c42138"
[ℹ]  nodegroup "ng-74c42138" has 1 node(s)
[ℹ]  node "ip-192-168-62-219.eu-north-1.compute.internal" is ready
[✔]  created nodegroup "ng-74c42138" in cluster "test-shared-nodegroup-new"
[ℹ]  will inspect security group configuration for all nodegroups
[✖]  2 nodegroups (ng-57c4c5a0, ng-dad92af5) don't use shared security group, so cluster networking maybe be broken
[✖]  it's recommended to delete these nodegroup and create new ones instead
[✖]  alternatively, you can patch the configuration and add each of these nodegroups to "sg-0252c495955e6af96"
 [0] >> 
```

### Checklist

### DIFF
--- a/pkg/eks/nodegroup.go
+++ b/pkg/eks/nodegroup.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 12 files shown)


pkg/eks/nodegroup.go
```diff
@@ -128,59 +126,3 @@ func (c *ClusterProvider) WaitForNodes(clientSet *clientset.Clientset, ng *api.N
 
 	return nil
 }
-
-// ValidateConfigForExistingNodeGroups looks at each of the existing nodegroups and
```
errordeveloper: This was introduced as a temporary fix to what seemed like DNS-only issue (#414), clearly a little too involved and I think we have a better answer now.

errordeveloper:
@dlespiau we only have a few examples of the config files, we really need to create proper reference docs, but currently the priority is to make API work properly in different use-cases and add validation etc.

Issue #449, opened 2019-01-18
Add build instructions


errordeveloper:
For a Go developer building the project is not a major question, but we need instructions that would be helpful for someone who's a biginner with Go.

This is what I think those instructions should be:
```
mkdir eksctl
cd eksctl
export GOPATH="$(pwd)"
go get -d github.com/weaveworks/eksctl
cd src/github.com/weaveworks/eksctl
make build
```

Also, we don't have much in terms of docs around testing. Contributor docs would be great! Even if we are to say that one can run `make` to get help, and that there unit tests and integrations tests and point them at the code.

dcherman:
This would definitely be helpful.  Speaking as someone that is literally brand new to go (and also developing on windows at the moment), it took a bit of finagling to be able to build this project.  I ended up resorting to using the Dockerfile in the `build` directory, building that, then mounting my src dir as a docker volume and building in that environment.

errordeveloper:
I believe you can just run 'make eksctl-image' to get dockerized build
going, but there is a chance it might not work exactly as expected on
Windows.

dcherman:
@errordeveloper So I just ran the build again since I couldn't remember all of the steps that I had to take to get things running.  You're correct that we should be able to use that image, but when I build from master (freshly pulled just now),

I think that if you want cross-platform instructions, providing a Dockerfile (as you have) is likely the easiest way since on Windows, you cannot build in `cmd` even with Make installed, you have to use a bash like git bash or something.

```
C:\Users\dherman\Documents\src\eksctl>make build

'CGO_ENABLED' is not recognized as an internal or external command,
operable program or batch file.
make: *** [build] Error 1
```

Building with `make eksctl-image` without modifying anything else yields the following:

https://gist.github.com/dcherman/a9bba39c763a2fa7c686be28300b7a20

I didn't really triage why it was deleting some of the .md files, or why any diff existed in the Gopkg.lock.  Since it should be building in the identical environment that you are (because Docker), I shouldn't be getting any different behavior there.

In order to submit the PR that I did the other day, I actually ended up removing the `git diff` part of the tests in order to get the build passing (didn't commit it of course).

errordeveloper:
On Mon, 21 Jan 2019, 7:38 pm dherman <notifications@github.com wrote:

> I think that if you want cross-platform instructions, providing a
> Dockerfile (as you have) is likely the easiest way since on Windows, you
> cannot build in cmd even with Make installed, you have to use a bash like
[...]

PR #486, opened 2019-01-29
Skip diff tests on Windows

errordeveloper:
### Description

As discussed in #449.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,11 +29,13 @@ COPY . $EKSCTL
 ARG COVERALLS_TOKEN
 ENV COVERALLS_TOKEN $COVERALLS_TOKEN
 
+ARG TEST_TARGET
+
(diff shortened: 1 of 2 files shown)


errordeveloper:
@dcherman did you have a chance to take this for a quick test?

dcherman:
@errordeveloper  I pulled down that branch and gave it a test run this morning.  It definitely gets by the diff tests now, so that is good to go.  It still hangs on the `lint` part of it, but I haven't done a lot of triage on that yet.  One thing that I did note is that unless I'm mistaken, it's linting all of the `vendor` folder which takes a significant amount of time and is also 3rd party code.

errordeveloper:
> ...it's linting all of the vendor folder which takes a significant amount of time and is also 3rd party code.

That's not supposed to happen, see linter config:

https://github.com/weaveworks/eksctl/blob/380d6fc5a8fc3ef8939e9d0a7486044ec2e6599d/.gometalinter.json#L1-L16

I'll merge this, and we can move on with build instructions.

Issue #510, opened 2019-02-06
[0.1.20-rc.3] For 'get nodegroup' the 'CLUSTER' column is blank


whereisaaron:
**What happened?**
The `get nodegroup` command produces an empty `CLUSTER` column.

**What you expected to happen?**
Expected the cluster name specified with the requires `--cluster` parameter to appear in that column

**How to reproduce it?**

```
$ bin/eksctl get nodegroup
[✖]  --cluster must be set
$ eksctl get nodegroup --cluster foo
CLUSTER NODEGROUP       CREATED                 MIN SIZE        MAX SIZE        DESIRED CAPACITY        INSTANCE TYPE   IMAGE ID
        group-1         2019-02-05T23:05:07Z    1               4               2                       m5.large        ami-0c2e8d28b1f854c68
```

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.20-rc.3"}
$ uname -a
Linux 4.4.0-17763-Microsoft #253-Microsoft Mon Dec 31 17:49:00 PST 2018 x86_64 x86_64 x86_64 GNU/Linux
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T07:10:00Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.5-eks-6bad6d", GitCommit:"6bad6d9c768dc0864dab48a11653aa53b5a47043", GitTreeState:"clean", BuildDate:"2018-12-06T23:13:14Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```
Also include your version of `heptio-authenticator-aws` -> N/A

**Logs**

```
2019-02-05T19:09:34-05:00 [▶]  role ARN for the current session is "arn:aws:iam::123456789012:user/bar"
2019-02-05T19:09:34-05:00 [▶]  nodegroups = [{
  Capabilities: ["CAPABILITY_NAMED_IAM"],
  CreationTime: 2019-02-05 23:05:07.696 +0000 UTC,
[...]

PR #513, opened 2019-02-06
make subnet names lowercase

### LINKED ISSUE #459: make subnet names lowercase
Right now we have `vpc.subnets.{Public,Private}`, these should be lowercase. We probably want to move away for using a map and use a struct actually.

errordeveloper:
### Description

Fixes #459.

This changes subnets from `map[SubnetTopology]map[string]Network` to a struct with `Private` and `Public` fields. That is in order to make those lowercase in JSON/YAML, so that it's consistent with the rest of the API.

### Checklist


errordeveloper:
I actually want to make it fail when old capitalised names are used, but I'm not too sure if I must. Perhaps we can try using `gjson.Get(configData, "vpc.subnets.Private")`, but that means we need turn YAML into JSON first... I'm undecided.

jrryjcksn:
Thanks for giving be the background! The more I can learn sooner, the better. :-)

pkg/cfn/manager/nodegroup.go
```diff
@@ -223,7 +223,7 @@ func (c *StackCollection) mapStackToNodeGroupSummary(stack *Stack) (*NodeGroupSu
 		return nil, errors.Wrapf(err, "error getting Cloudformation template for stack %s", *stack.StackName)
 	}
 
-	cluster := getClusterName(stack)
+	cluster := getClusterNameTag(stack)
```
mumoshu: Doesn't this revive #510? 

`stack` here is a nodegroup stack whose name is `<cluster name>-nodegroup-<nodegroup name>`, but then `getClusterNameTag` will end up returning a empty string.

errordeveloper: No, this will get just the tag, and nodegroup stack has cluster name tag.

mumoshu: Seems like I'm not following your idea correctly. Anyway, running eksctl built from your branch(at 2bc7cef2f793025eca9211ffc9b1d0a19f63ff93) produces empty cluster names in `get nodegroup`:

```
$ ./eksctl get nodegroup --cluster mycluser
CLUSTER	NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
	nodegroup1	2019-02-02T13:06:24Z	2		2		0			t3.large	ami-0f0e8066383e7a2cb
```

mumoshu: Probably you wanted a dedicated `getClusterName` func for nodegroups?

errordeveloper:
Ah, so looks like I should make a separate PR to address this issue, not as
trivial as I hoped!

pkg/cfn/manager/nodegroup.go
errordeveloper: Ah, so looks like I should make a separate PR to address this issue, not as trivial as I hoped!

> Probably you wanted a dedicated `getClusterName` func for nodegroups?

Not as such, `getClusterName` was intended to be sufficient for this particular case.

pkg/cfn/manager/cluster.go
```diff
@@ -161,3 +157,14 @@ func getClusterName(s *Stack) string {
 	}
 	return ""
 }
+
+func getClusterNameTag(s *Stack) string {
+	for _, tag := range s.Tags {
+		if *tag.Key == api.ClusterNameTag {
+			if strings.HasSuffix(*s.StackName, "-cluster") {
```
errordeveloper: Should have removed this suffix check, that's all. Will open another PR, so that we keep this PR nicely scoped also.

mumoshu:
Makes sense! Thanks for confirming.
2019年2月7日(木) 18:18 Ilya Dmitrichenko <notifications@github.com>:

errordeveloper:
@mumoshu https://github.com/weaveworks/eksctl/pull/517

errordeveloper:
I don't get why, but somehow this has affected integration tests and I have fix coming up in another PR.

PR #530, opened 2019-02-12
feat: Instance ID Node Label

### LINKED ISSUE #500: feat: Instance ID Node Label

#411 (via #237) allows us to deploy AWS SSM agents as a daemonset, so that we can use Sessions Manager to open terminal sessions into nodes, as an alternative to SSH.

You need to run `start-session` providing the EC2 instance ID of the node:

```console
$ aws ssm start-session --target i-04ffadbaae98a5bd0
```

But there's no straightforward way to obtain the instance of a specific Kubernetes node. One way would be to query it by runnign `aws ec2 describe-instances --filter ...`, but I want something easier.

Let's enhance `eksctl` so that every node has its instance ID in a dedicated node label, which can is shown on `kubectl get node --show-labels`:

```
$ kubectl get node --show-labels
NAME                                                 STATUS    ROLES     AGE       VERSION   LABELS
ip-192-168-119-97.ap-northeast-1.compute.internal    Ready     <none>    16h       v1.11.5   alpha.eksctl.io/cluster-name=mycluster1,alpha.eksctl.io/nodegroup-name=nodegroup1,alpha.eksctl.io/instance-id=i-04ffadbaae98a5bd0,...(snip)
```

Implementation-wise, I'd just add a key-value to `kubelet`'s `--node-labels` flag:

```
--node-labels alpha.eksctl.io/instance-id=$(curl http://169.254.169.254/latest/meta-data/ instance-id)
```

mumoshu:
Resolves #500

Manually verified to work by creating a new cluster and looking into node labels:

```console
$ k get no --show-labels
NAME                                             STATUS    ROLES     AGE       VERSION   LABELS
ip-10-1-11-239.ap-northeast-1.compute.internal   Ready     <none>    7m        v1.11.5   alpha.eksctl.io/cluster-name=staging-sre-1,alpha.eksctl.io/instance-id=i-08553f41b298f910e,alpha.eksctl.io/nodegroup-name=nodegroup1,beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=t3.large,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=ap-northeast-1,failure-domain.beta.kubernetes.io/zone=ap-northeast-1a,kubernetes.io/hostname=ip-10-1-11-239.ap-northeast-1.compute.internal
```

Each node's instance ID is set for the key `alpha.eksctl.io/instance-id`. Because the key is preficed with `alpha`, I'm fine with this being changed in a backward-incompatible way.

This should remain an useful feature until the instance-id-as-node-name feature is implemented in the upstream. See #509 to track the progress towards it.


pkg/nodebootstrap/assets/10-eksclt.al2.conf
```diff
@@ -5,15 +5,15 @@
...
+# Local non-static parameters: NODE_IP, INSTANCE_ID
 EnvironmentFile=/etc/eksctl/kubelet.local.env
 
 ExecStart=
 ExecStart=/usr/bin/kubelet \
   --node-ip=${NODE_IP} \
   --cluster-dns=${CLUSTER_DNS} \
   --max-pods=${MAX_PODS} \
-  --node-labels=${NODE_LABELS} \
+  --node-labels=${NODE_LABELS},alpha.eksctl.io/instance-id=${INSTANCE_ID} \
```
errordeveloper: How does this work when `NODE_LABELS` is empty?

mumoshu: `kubelet` just ignores it.

Also, practically speaking, it never becomes empty because of the cluster-name and the node-group-name labels.

A typical `kubelet.env` that fills the NODE_LABELS env looks like this:

```
NODE_LABELS=alpha.eksctl.io/nodegroup-name=nodegroup1,alpha.eksctl.io/cluster-name=mycluster
```

errordeveloper: You are right, it used to be empty until recently, and now we always set those labels. Thanks!

mumoshu: Thanks for confirming and providing the context!

mumoshu:
I found a regression introduced by the initial version( #531 ) that is now fixed. This should be ready-to-go.

PR #532, opened 2019-02-12
Support updated GPU AMI naming convention

toricls:
### Description

It seems that the new GPU AMIs naming convention has been changed by AWS side. They're named with timestamp value instead of “v{YYYYMMDD}” format.
This PR fixes the problem that the AMI resolver cannot find the new GPU AMIs due to the change above.

### Checklist


errordeveloper:
Thanks a lot, well spotted!

PR #533, opened 2019-02-12
 Fix an issue with referencing of a map in `ctl.ImportSubnet`

### LINKED ISSUE #518: Can't add nodegroup for existing cluster which created with old eksctl
Please [search existing issues (open & closed)](https://github.com/weaveworks/eksctl/issues?utf8=%E2%9C%93&q=is%3Aissue) to see if there is a similar one. If there is, add comments or vote where appropriate.

If you did not find a similar issue or if it did not help then ask for help in our [slack channel](https://weave-community.slack.com/messages/eksctl/)

**What help do you need?**

Hey there.

So I recently updated my local eksctl to latest (0.1.20-rc.3) and tried to add new nodegroup for existing cluster which created with way way older version of eksctl (I'm not sure exactly which version i used), but got this error at first, 

```
cluster compatibility check failed: shared node security group missing, to fix this run 'eksctl utils update-cluster-stack --name=testcluster --region=eu-west-1'
```

and after I ran `eksctl utils update-cluster-stack --name=testcluster --region=eu-west-1` to solve that problem, I got this.

```
2019-02-07T18:48:11+09:00 [✖]  inssuficient number of subnets, at least 2x public and/or 2x private subnets are required
```

I already got 3 subnets for my cluster, but it's still telling me there's no subnets.

I believe this is caused by difference between templates for older version and latest version.

Is there any way to solve this conflict (or bug) without messing around everything? I don't want to create a new cluster and migrate everything just for it.

Thanks a lot.

errordeveloper:
### Description

This fixes #518, which is a bug that got introduced as part of #513.

### Checklist

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha4/vpc.go
+++ b/pkg/apis/eksctl.io/v1alpha4/vpc.go
@@ -92,22 +92,28 @@ func (c *ClusterConfig) PublicSubnetIDs() []string {
 
 // ImportSubnet loads a given subnet into cluster config
 func (c *ClusterConfig) ImportSubnet(topology SubnetTopology, az, subnetID, cidr string) error {
-	subnetCIDR, _ := ipnet.ParseCIDR(cidr)
-
 	if c.VPC.Subnets == nil {
 		c.VPC.Subnets = &ClusterSubnets{}
 	}
 
-	var subnets map[string]Network
 	switch topology {
 	case SubnetTopologyPrivate:
-		subnets = c.VPC.Subnets.Private
+		if c.VPC.Subnets.Private == nil {
+			c.VPC.Subnets.Private = make(map[string]Network)
+		}
+		return doImportSubnet(c.VPC.Subnets.Private, az, subnetID, cidr)
 	case SubnetTopologyPublic:
-		subnets = c.VPC.Subnets.Public
+		if c.VPC.Subnets.Public == nil {
+			c.VPC.Subnets.Public = make(map[string]Network)
+		}
+		return doImportSubnet(c.VPC.Subnets.Public, az, subnetID, cidr)
+	default:
+		return fmt.Errorf("unexpected subnet topology: %s", topology)
 	}
-	if subnets == nil {
-		subnets = make(map[string]Network)
-	}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/apis/eksctl.io/v1alpha4/vpc.go
```diff
@@ -92,22 +92,28 @@ func (c *ClusterConfig) PublicSubnetIDs() []string {
...
 	if c.VPC.Subnets == nil {
 		c.VPC.Subnets = &ClusterSubnets{}
 	}
 
-	var subnets map[string]Network
 	switch topology {
 	case SubnetTopologyPrivate:
-		subnets = c.VPC.Subnets.Private
+		if c.VPC.Subnets.Private == nil {
+			c.VPC.Subnets.Private = make(map[string]Network)
```
mumoshu: Just trying to understand what was going on here.

The cause of the issue seems to be https://github.com/weaveworks/eksctl/pull/513/files#diff-9cf0956b5581521f18ebe4dedc1f1552L90 - we had been mistakenly throwing away the imported `subnets` when `c.VPC.Subnets.Private` or `c.VPC.Subnets.Public`was empty?

errordeveloper: Thanks for asking! I've looked into this some more, and actually have a better idea of what is going on.

I was assuming that maps are always treated as references, which is indeed the case. However, if you assign a nil map to another nil map, nothing will actually happen. Namely, maps are treated as references (well, sort of), but only after they were initialised. Take a look here: https://play.golang.org/p/OTCUWDoH9HK. I didn't realise this until now, or at least I forgot this.

errordeveloper: Actually, it's even simpler then this, any nil reference can be assigned to another one, but nothing gets copied really: https://play.golang.org/p/KFefMZEabWQ.

errordeveloper: So it sort of boils down to `i1 = i2` being the same as `i1 = nil`, so the compiler cannot really tell these apart, hence it has no way of telling me that I'm probably doing something wrong... TIL :)

Issue #462, opened 2019-01-25
eksctl apply: Declarative way to manage clusters


mumoshu:
**TL;DR;**

I want `eksctl apply -f cluster.yaml` for creating a cluster and creating/deleting nodegroups.

Updates to clusters and nodegroups is out of scope of this feature request.

**Why do you want this feature?**

While managing Kubernetes clusters, I believe it is common to just add nodegroup(s) to an existing cluster, to start serving different workloads on different infrastructure(instance type, volume size, host OS, kernel parameters, etc).

Kubernetes, by its nature, allows doing this without recreating the whole cluster. And I believe it is preferable to just add nodegroups rather than recreating the whole cluster every time you need to add nodegroups, because it takes just fewer time and financial cost.

I also prefer managing my cluster configuration as a whole in code, including Kubernetes control-planes provided by EKS and nodegroups provided by eksctl.

eksctl as of today has an API object for clusters that is called `ClusterConfig. It has dedicated field for nodegroup, which can be used for creating zero or more nodegroups along with the control-plane.

```
apiVersion: eksctl.io/v1alpha3
kind: ClusterConfig

metadata:
  name: cluster-1
  region: eu-north-1

nodeGroups:
  - name: ng-1
    instanceType: m5.large
    desiredCapacity: 10
```

However, there's no way to add nodegroups declaratively because:

- There's no way to nodegroups independently and declaratively, e.g. `eksctl create nodegroup --config-file nodegroup.yaml`, atm.
- There's no way to update `nodeGroups`inside a `ClusterConfig` object after the initial creation.

Also note that, I'm unsure if the former actually works.

If something like `eksctl cluster apply -f cluster.yaml` is implemented to allow adding/removing embedded nodegroups, what happens when you run `eksctl create nodegroup -f ng.yaml` and then `eksctl apply cluster -f cluster.yaml`? Should the latter update the nodegroup created by the former command if the nodegroup names were equivalent? Then, which would be said to be the desired state of the nodegroup, `ng.yaml` or a nodegroup embedded in `cluster.yaml`?

**What feature/behavior/change do you want?**

I'd like to add `eksctl apply [-f|--config-file] cluster.yaml`, which supports creating a cluster and creating/deleting nodegroups only. Updates to clusters and nodegroups is out of scope of this feature request.

To me it isn't necessary to support updating every possible field in `ClusterConfig`. At least, initially, only addition/deletion of items under `nodeGroups` would be enough for this specific use-case.

[...]

Issue #538, opened 2019-02-12
Initial nodegroup not showing up in "eksctl get nodegroups"


nesv:
**What happened?**

Our EKS cluster was initially set up using an older version of eksctl (0.1.8, I _think_...). When that cluster was set up, the initial nodegroup was managed by a CloudFormation template called `eksctl-${CLUSTER_NAME}-nodegroup-0`.

After upgrading to eksctl 0.1.20, I started to create new nodegroups to take the place of the initial one, so we could migrate services and eventually decommission the initial nodegroup. However, when I run `eksctl get nodegroups --cluster=${CLUSTER_NAME}`, that initial, unnamed nodegroup does not show up in the list. 

However, I am able to scale the nodegroup by running: ```eksctl scale nodegroup --cluster=${CLUSTER_NAME} --nodes=${N} --name=0```

**What you expected to happen?**

When I run `eksctl get nodegroups --name=${CLUSTER_NAME}`, I hope to see an entry for a nodegroup called `0` (or something to that effect).

**How to reproduce it?**

:man_shrugging: 

**Anything else we need to know?**

- Tested on Linux
- Binary was downloaded from the Releases page
- AWS credentials: default/named profile

**Versions**

```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.20"}
```

```
$ uname -a
Linux xxx 4.20.7-arch1-1-ARCH #1 SMP PREEMPT Wed Feb 6 18:42:40 UTC 2019 x86_64 GNU/Linux
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:08:12Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.5-eks-6bad6d", GitCommit:"6bad6d9c768dc0864dab48a11653aa53b5a47043", GitTreeState:"clean", BuildDate:"2018-12-06T23:13:14Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

**`eksctl utils describe-stacks` output**

```
[ℹ]  stack/eksctl-testing-nodegroup-QA = {
[...]

PR #545, opened 2019-02-14
Design Proposal #003: Extending Use of `--config-file`

errordeveloper:
### Description

This is a design proposal towards #462.

- xref #19

### DIFF
--- a/docs/proposal-003_cluster-and-nodegroups-config.md
+++ b/docs/proposal-003_cluster-and-nodegroups-config.md
@@ -1,22 +1,22 @@
 # Design Proposal #003: ClusterConfig and NodeGroupConfig
 
 > **STATUS**: This proposal is a _working draft_, it will get refined and augment as needed.
-> If any non-trivial changes are need to functionality defined here, in particular the user
+> If any non-trivial changes are needed to functionality defined here, in particular the user
 > experience, those changes should be suggested via a PR to this proposal document.
 > Any other changes to the text of the proposal or technical corrections are also very welcome.
 
 Support for config files had been added in eksctl 0.1.17 as an experimental alpha feature, it has
 become popular amongs users.
 Users currently can specify `ClusterConfig` object in JSON or YAML, which allows them to set more
-different fields then they currently can with CLI flags. They can alos specify several nodegroups,
-and create a cluster in one go. Besides many missing features in `v1alpha4` incornation (where there
+different fields then they currently can with CLI flags. They can also specify several nodegroups,
+and create a cluster in one go. Besides many missing features in `v1alpha4` incarnation (where there
 is little of  management functionality beyond initial cluster creation), one of commonly requested
 features is management of nodegroups through config files. 
 
 This proposal aims to define how existing `ClusterConfig` object can be enhanced to support new type
 of objects that allow users to define and manage nodegroups via a separate step.
 
-Let's conside a simple config:
+Let's consider a simple config:
 
 ```YAML
 # cluster.yaml
@@ -38,7 +38,7 @@ nodeGroups:
     privateNetworking: true
 ```
 
-There are two nodegroups are defined inline.
+There are two nodegroups defined inline.
 
 It should be possible to create this cluster and add more nodegroups based on a separate configs later.
 
@@ -62,14 +62,14 @@ spec:
   privateNetworking: true
 ```
 
-Two way that these config files can be use to create cluster:
+Two ways that these config files can be used to create cluster:
 
 - `eksctl create cluster --config-file=cluster.yaml && eksctl create ng --config-file=extra-nodegroup.yaml`
 - `cat cluster.yaml extra-nodegroup.yaml | eksctl create cluster --config-file -`
 
 ### Questions
 
-- Is it appropriate to also allow referencing nodegroup that is defined externally withing `ClusterConfig`?
+- Is it appropriate to also allow referencing nodegroup that is defined externally within `ClusterConfig`?
 - Where in standalone `NodeGroup` config should lables live? (If we copy the same struct, we will get `spec.labels`,
   while it might make more sense to have them as `metadata.labels`).
 - How will `ClusterConfig` and `NodeGroup` comapare when ispected in a running cluster? Namely, are all nodegroups
@@ -86,7 +86,7 @@ a flag to force creation of node or some of the nodegroups.
 Examples:
 
 ```
-eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*-private' ## only nodegroups matching regex will be created
+eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*-private' ## only nodegroups not matching regex will be created
 eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*' ## no nodegroups will be created all
 eksctl create ng --config-file=cluster.yaml ## all nodegroups will be created
 eksctl create ng --config-file=cluster.yaml --only=ng1-public ## only one nodegroup will be created
--- a/docs/proposal-003_cluster-and-nodegroups-config.md
+++ b/docs/proposal-003_cluster-and-nodegroups-config.md
@@ -75,3 +75,20 @@ Two way that these config files can be use to create cluster:
 - How will `ClusterConfig` and `NodeGroup` comapare when ispected in a running cluster? Namely, are all nodegroups
   supposed to be inline in `ClusterConfig` as well as represented as distinct object? (This is currently hard to
   say,  but in the future it will become importnat that there is coherent representation).
+
+### Alternative Approach
+
+One alternative could be to avoid creating a separate object, and keep objects and relationships simpler.
+
+Given the `cluster.yaml` above, we could allow creating cluster with an empty list of nodegroups, as well as
+a flag to force creation of node or some of the nodegroups.
+
+Examples:
+
+```
+eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*-private' ## only nodegroups matching regex will be created
+eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*' ## no nodegroups will be created all
+eksctl create ng --config-file=cluster.yaml ## all nodegroups will be created
+eksctl create ng --config-file=cluster.yaml --only=ng1-public ## only one nodegroup will be created
+eksctl create ng --config-file=cluster.yaml --only=ng1-public,ng2-private ## both nodegroups will be created
+```
--- a/docs/proposal-003_cluster-and-nodegroups-config.md
+++ b/docs/proposal-003_cluster-and-nodegroups-config.md
@@ -1,0 +1,77 @@
+# Design Proposal #003: ClusterConfig and NodeGroupConfig
+
+> **STATUS**: This proposal is a _working draft_, it will get refined and augment as needed.
+> If any non-trivial changes are need to functionality defined here, in particular the user
+> experience, those changes should be suggested via a PR to this proposal document.
+> Any other changes to the text of the proposal or technical corrections are also very welcome.
+
+Support for config files had been added in eksctl 0.1.17 as an experimental alpha feature, it has
+become popular amongs users.
+Users currently can specify `ClusterConfig` object in JSON or YAML, which allows them to set more
+different fields then they currently can with CLI flags. They can alos specify several nodegroups,
+and create a cluster in one go. Besides many missing features in `v1alpha4` incornation (where there
+is little of  management functionality beyond initial cluster creation), one of commonly requested
+features is management of nodegroups through config files. 
+
+This proposal aims to define how existing `ClusterConfig` object can be enhanced to support new type
+of objects that allow users to define and manage nodegroups via a separate step.
+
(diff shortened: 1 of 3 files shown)


docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+- Is it appropriate to also allow referencing nodegroup that is defined externally withing `ClusterConfig`?
+- Where in standalone `NodeGroup` config should lables live? (If we copy the same struct, we will get `spec.labels`,
+  while it might make more sense to have them as `metadata.labels`).
+- How will `ClusterConfig` and `NodeGroup` comapare when ispected in a running cluster? Namely, are all nodegroups
+  supposed to be inline in `ClusterConfig` as well as represented as distinct object? (This is currently hard to
+  say,  but in the future it will become importnat that there is coherent representation).
+
+### Alternative Approach
+
+One alternative could be to avoid creating a separate object, and keep objects and relationships simpler.
```
errordeveloper: I kind of like it more actually.

errordeveloper: It seems like separating object will be less helpful when it comes to implementing more complex features in the future, and having similar kind of filter flags in 'eksctl apply' maybe of use too. We can always split objects later, bit merging object would be kind of harder (or at least it seems like that now).

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+
+> **STATUS**: This proposal is a _working draft_, it will get refined and augment as needed.
+> If any non-trivial changes are need to functionality defined here, in particular the user
+> experience, those changes should be suggested via a PR to this proposal document.
+> Any other changes to the text of the proposal or technical corrections are also very welcome.
+
+Support for config files had been added in eksctl 0.1.17 as an experimental alpha feature, it has
+become popular amongs users.
+Users currently can specify `ClusterConfig` object in JSON or YAML, which allows them to set more
+different fields then they currently can with CLI flags. They can alos specify several nodegroups,
```
D3nn: NIT: /alos/also/

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+and create a cluster in one go. Besides many missing features in `v1alpha4` incornation (where there
```
D3nn: NIT: /incornation/incarnation/

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+is little of  management functionality beyond initial cluster creation), one of commonly requested
+features is management of nodegroups through config files. 
+
+This proposal aims to define how existing `ClusterConfig` object can be enhanced to support new type
+of objects that allow users to define and manage nodegroups via a separate step.
+
+Let's conside a simple config:
```
D3nn: NIT: /conside/consider/

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+  name: ng3-extra
+
+spec:
+  instanceType: c4.xlarge
+  minSize: 10
+  maxSize: 20
+  privateNetworking: true
+```
+
+Two way that these config files can be use to create cluster:
```
D3nn: NIT:  /way/ways/ and /use/used/

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+
+- `eksctl create cluster --config-file=cluster.yaml && eksctl create ng --config-file=extra-nodegroup.yaml`
+- `cat cluster.yaml extra-nodegroup.yaml | eksctl create cluster --config-file -`
+
+### Questions
+
+- Is it appropriate to also allow referencing nodegroup that is defined externally withing `ClusterConfig`?
```
D3nn: I would like to see that.
Also another NIT: /withing/within/

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+- Where in standalone `NodeGroup` config should lables live? (If we copy the same struct, we will get `spec.labels`,
+  while it might make more sense to have them as `metadata.labels`).
```
D3nn: I could see metadata.labels on NodeGroupConfig and metadata.labelSelectors to _*almost*_ match K8s config, but not sure how much extra work that would add to the controller.

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
+# Design Proposal #003: ClusterConfig and NodeGroupConfig
+
+> **STATUS**: This proposal is a _working draft_, it will get refined and augment as needed.
+> If any non-trivial changes are need to functionality defined here, in particular the user
```
jrryjcksn: NIT: need -> needed

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+  - name: ng1-public
+    instanceType: m5.xlarge
+    desiredCapacity: 4
+  - name: ng2-private
+    instanceType: m5.large
+    desiredCapacity: 10
+    privateNetworking: true
+```
+
+There are two nodegroups are defined inline.
```
jrryjcksn: NIT: 'are defined' -> 'defined'

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+
+One alternative could be to avoid creating a separate object, and keep objects and relationships simpler.
+
+Given the `cluster.yaml` above, we could allow creating cluster with an empty list of nodegroups, as well as
+a flag to force creation of node or some of the nodegroups.
+
+Examples:
+
+```
+eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*-private' ## only nodegroups matching regex will be created
```
jrryjcksn: NIT: matching regex --> not matching regex

jrryjcksn: I wonder if maybe we should leave out the object type when using a config file? It would be more consistent with kubectl.

`eksctl create --config-file=cluster.yaml --exclude='nodegroup:.*-private' --exclude=vpc` or
`eksctl create --config-file=cluster.yaml --exclude='nodegroup:.*-private',vpc`
`eksctl create --config-file=cluster.yaml --include=cluster,nodegroup`

docs/proposal-003_cluster-and-nodegroups-config.md
mumoshu: I prefer this, too. As I've noted in #462, separating object is not a requirement.

docs/proposal-003_cluster-and-nodegroups-config.md
mumoshu: At glance `--exclude-nodegroups='.*-private'` isn't a must-have as we already have `--without-nodegroup`.

Can we start the implementation without `--exclude-nodegroups`?

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,94 @@
...
+eksctl create cluster --config-file=cluster.yaml --exclude-nodegroups='.*' ## no nodegroups will be created all
+eksctl create ng --config-file=cluster.yaml ## all nodegroups will be created
+eksctl create ng --config-file=cluster.yaml --only=ng1-public ## only one nodegroup will be created
```
mumoshu: A nice-to-have feature would be globs in `--only`, like `--only=-*public`, so that we can omit `create cluster --exclude-nodegroups` from the initial implementation without any bad effect to the u/x.

docs/proposal-003_cluster-and-nodegroups-config.md
errordeveloper: Yes, I think it could be a reasonable temporary measure. The `--without-nodegroup` flag is really about  the initial nodegroup and is meant to be incompatible with `--config-file`, but at present is simply ignored when passed along with `--config-file`. We might not actually need this temporary measure, it'd make sense to start with `eksctl create nodegroup` part, and then add the filter flags to `eksctl create cluster` as a final step.

docs/proposal-003_cluster-and-nodegroups-config.md
errordeveloper: Yes, indeed! And it seems like globs would be a better idea over regex. Perhaps we could allow list of globs, so user can specify either literal names explicitly or a chain of globs if they needed to. We've seen people using quite a few nodegroups already (#538).

errordeveloper:
> Would something like `eksctl apply --config-file cluster.yaml` that automatically creates nodegroups missing in the AWS region, and deletes nodegroups missing in the config file be covered in this proposal?

@mumoshu I think `eksctl apply` is technically out of scope at the moment, but it's certainly is a on the horizon. My strategy at the moment is to focus on bitesize functionality, but pave the way for `eksctl apply` to become more tangible. I think we will start discussing `eksctl apply` more after we have upgrades.
Also, the idea of this proposal is to keep it small enough in scope. We somehow we ended-up without a proper `ClusterConfig` proposal to begin with, but maybe we can create one once we are getting closer to a beta version of the API, and use it as a means of driving beta to GA.

docs/proposal-003_cluster-and-nodegroups-config.md
mumoshu: List of globs sounds good!

mumoshu:
I agree with your strategy 👍

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,128 @@
...
+  - name: ng3-dev
+    instanceType: m4.large
+    desiredCapacity: 2
+    privateNetworking: true
+  - name: ng2-test
+    instanceType: m5.large
+    desiredCapacity: 4
+    privateNetworking: true
+```
+`eksctl create nodegroup` will be updated to support glob pattern matching for nodegroup names contained within the cluster config.  Therefore, a user can create these additional nodegroups using `eksctl create nodegroup --config-file=cluster.yaml --only="*dev,*test" --cluster="cluster-5"`.  Eksctl will create two nodegroups in cluster named cluster-5.
```
jrryjcksn: No `--cluster` with `--config-file`

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,128 @@
...
+
+Here are some additional examples:
+
+```
+eksctl create ng --config-file=cluster.yaml --cluster=cluster5 ## all nodegroups will be created
```
jrryjcksn: All of these need to have `--cluster` removed

docs/proposal-003_cluster-and-nodegroups-config.md
```diff
@@ -0,0 +1,128 @@
...
+eksctl create ng --config-file=cluster.yaml --only=ng1-public  --cluster=cluster5 ## only one nodegroup will be created
+eksctl create ng --config-file=cluster.yaml --only=ng1-public,ng2-private  --cluster=cluster5 ## both nodegroups will be created
+```
+
+The `eksctl create cluster` cli will be updated to optionally ignore all nodegroups in the cluster config. https://github.com/weaveworks/eksctl/issues/555
+
+By using a single configuration file, users can keep their cluster definition together and most importantly checked into source control.
```
jrryjcksn: definition -> definitions

docs/proposal-003_cluster-and-nodegroups-config.md
palemtnrider: regex has been replaced with glob patterns.

palemtnrider: Where we landed was to keep the grammar similar to nodegroup creation.  Meaning we would be inclusive instead of exclusive.  i.e. `create config --only "glob pattern(s)"` to create the nodegroups identified by the glob pattern.  _We may want a better flag name_ 

Also, if `--without-nodegroup` is passed along with the config name, the cluster create will ignore all nodegroups defined in the config file.

docs/proposal-003_cluster-and-nodegroups-config.md
palemtnrider: Proposal now reflects globs instead of regex.

errordeveloper:
I think this is good to go, we can always make improvements later. Will try to re-run the job in Circle, maybe just a flake..

errordeveloper:
Let's merge it, and if anyone feels like it needs an update - please open another PR :)

voxxit:
Did anything ever come of the `NodeGroupConfig`?

errordeveloper:
It was divided to keep it to a single file/object for now, as splitting is
easy, but merging back would be harder. Do you have a use-case for it being
defined separately?

PR #550, opened 2019-02-18
feat: Node-local DNS cache support

### LINKED ISSUE #542: Node-local DNS cache support
TL;DR; I'd like to add a new configuration key for nodegroups to specify the value for the kubelet `--cluster-dns` flag.

Currently both internal and external DNS lookups are processed by the CoreDNS service deployed by EKS. This isn't ideal from a reliability perspective.

An internal lookup is for resolving cluster-internal names of Kubernetes services. It may fail in various conditions, like:

- kube-proxy failed on either (1) the node on which your pod sent the DNS query is running or (2) the node on which the CoreDNS pods are running
- kube-dns(CoreDNS) pod failed
- Kubernetes endpoint-controller failed
- Other node-to-node communication failure(EC2 or VPC issue)

An external lookup is for resolving names managed external to the cluster, like RDS DB clusters/instances, SQS endpoints, S3, your own services served via Route 53. It may fail in conditions like:

- All the above conditions
- The Amazon DNS in your VPC failed
- Route 53 failed(Applied in a case that the name is served by it)

The only thing I want to propose for eksctl is - Add an ability to override the value for the `--cluster-dns` flag  passed to the kubelet.

It allows us to deploy a node-local dns cache, that is proposed to resolve this issue. You can find the upstream proposal at https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md. It is implemented in https://github.com/kubernetes/kubernetes/pull/70555.

Seeing how it works, and the implementation, only thing we can't do it outside of eksctl is passing a "magic" IP address that forwards any DNS lookups to the node-local cache, to kubelet's `--cluster-dns` https://github.com/kubernetes/kubernetes/blob/a3877b1776cc55f5a32103d7a072a73e18c3d939/hack/local-up-cluster.sh#L704.

As we're leaning towards exposing new configurations only to config files, I would be glad if it adds a new field named e.g. `clusterDNS`. For example, it will be specified like:

```
nodeGroups:
- name: nodegroup1
  clusterDNS: 169.254.20.10
  # snip
```

mumoshu:
TL;DR; This is the smallest change to allow enabling node-local DNS cache on eksctl-created nodes.

## What

Add a new field named `clusterDNS` that accepts the IP address to the DNS server used for all the internal/external DNS lookups a.k.a the `--cluster-dns` flag of `kubelet`.

```yaml
nodeGroups:
- name: nodegroup1
  clusterDNS: 169.254.20.10
  # snip
```

This, in combination with `k8s-dns-node-cache` deployed as a daemonset on your cluster, allows all the DNS lookups from your pods to firstly routed to the node-local DNS server, which adds more reliability.

## Notes

The configuration key `clusterDNS` is intentionally made per-nodegroup, not per-cluster, so that you can selectively use the node-local DNS. It, in combination with the proper use of node labels/taints, allows you to test the node-local DNS in only a subet of your workload.
It would also be nice to add `clusterDNS` as a cluster-level config key later. But I believe it isn't a must-have in this change.

## Usage

See the [cluster/addons/dns/nodelocaldns](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) in the upstream repository for more details.

A concrete steps to enable node-local DNS would look like the below:

- Decide the which IP addr to be used for binding the node-local DNS. Typically this is `169.254.20.10`
- Add `clusterDNS: 169.254.20.10` to your nodegroup in the cluster config
- Deploy [nodelocaldns.yaml](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml), replacing:
  `__PILLAR__LOCAL__DNS__` with `169.254.169.254`, `__PILLAR__DNS__DOMAIN__` with `cluster.local`, `__PILLAR__DNS__SERVER__` with [`10.100.0.10` or `172.20.0.10`](https://github.com/weaveworks/eksctl/blob/master/pkg/nodebootstrap/userdata.go#L87-L94) according to your VPC CIDR

Resolves #542

### DIFF
--- a/pkg/nodebootstrap/userdata.go
+++ b/pkg/nodebootstrap/userdata.go
@@ -84,7 +84,10 @@ func makeClientConfigData(spec *api.ClusterConfig, ng *api.NodeGroup) ([]byte, e
 	return clientConfigData, nil
 }
 
-func clusterDNS(spec *api.ClusterConfig) string {
+func clusterDNS(spec *api.ClusterConfig, ng *api.NodeGroup) string {
+	if ng.ClusterDNS != "" {
+		return ng.ClusterDNS
+	}
 	// Default service network is 10.100.0.0, but it gets set 172.20.0.0 automatically when pod network
 	// is anywhere within 10.0.0.0/8
 	if spec.VPC.CIDR != nil && spec.VPC.CIDR.IP[0] == 10 {
@@ -106,7 +109,7 @@ func makeKubeletParamsCommon(spec *api.ClusterConfig, ng *api.NodeGroup) []strin
 	// TODO: use componentconfig or kubelet config file – https://github.com/weaveworks/eksctl/issues/156
 	return []string{
 		fmt.Sprintf("MAX_PODS=%d", ng.MaxPodsPerNode),
-		fmt.Sprintf("CLUSTER_DNS=%s", clusterDNS(spec)),
+		fmt.Sprintf("CLUSTER_DNS=%s", clusterDNS(spec, ng)),
 		fmt.Sprintf("NODE_LABELS=%s", strings.Join(labels, ",")),
 	}
 }
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -358,6 +358,9 @@ type NodeGroup struct {
 
 	// +optional
 	OverrideBootstrapCommand *string `json:"overrideBootstrapCommand,omitempty"`
+
+	// +optional
+	ClusterDNS string `json:"clusterDNS,omitempty"`
 }
 
 // ListOptions returns metav1.ListOptions with label selector for the nodegroup
--- a/pkg/cfn/builder/api_test.go
+++ b/pkg/cfn/builder/api_test.go
@@ -835,6 +835,8 @@ var _ = Describe("CloudFormation template builder API", func() {
 
 		cfg.NodeGroups[0].OverrideBootstrapCommand = &overrideBootstrapCommand
 
+		cfg.NodeGroups[0].ClusterDNS = "169.254.20.10"
+
 		rs := NewNodeGroupResourceSet(p, cfg, "eksctl-test-123-cluster", ng)
 		err := rs.AddAllResources()
 		It("should add all resources without errors", func() {
@@ -868,7 +870,7 @@ var _ = Describe("CloudFormation template builder API", func() {
 			Expect(kubeletEnv.Permissions).To(Equal("0644"))
 			Expect(strings.Split(kubeletEnv.Content, "\n")).To(Equal([]string{
 				"MAX_PODS=29",
-				"CLUSTER_DNS=10.100.0.10",
+				"CLUSTER_DNS=169.254.20.10",
 				"NODE_LABELS=os=al2",
 			}))
 
@@ -977,6 +979,8 @@ var _ = Describe("CloudFormation template builder API", func() {
 			"os": "ubuntu",
 		}
 
+		cfg.NodeGroups[0].ClusterDNS = "169.254.20.10"
+
 		cfg.NodeGroups[0].OverrideBootstrapCommand = &overrideBootstrapCommand
 
 		rs := NewNodeGroupResourceSet(p, cfg, "eksctl-test-123-cluster", ng)
@@ -1018,7 +1022,7 @@ var _ = Describe("CloudFormation template builder API", func() {
 			Expect(kubeletEnv.Permissions).To(Equal("0644"))
 			Expect(strings.Split(kubeletEnv.Content, "\n")).To(Equal([]string{
 				"MAX_PODS=29",
-				"CLUSTER_DNS=172.20.0.10",
+				"CLUSTER_DNS=169.254.20.10",
 				"NODE_LABELS=os=ubuntu",
 			}))
 


mumoshu:
For anyone interested, this is how I've verified this to work.

1. Add your nodegroup `clusterDNS: 169.254.20.10`

```yaml
nodeGroups:
  - name: nodegroup1
    clusterDNS: "169.254.20.10"
    instanceType: m4.large
   # and whatever you like
```

2. After bringing up your cluster, `kubectl run` a pod and try resolving any host name from it. I've used an ubuntu container and a `yum update`, but whatever works.

This should fail, as kubelet points cluster DNS to `169.254.20.10` as we configured in the cluster.yaml, but nothing binds it yet:

3. Deploy `node-local-dns`. After the deployment, any DNS lookup against 169.254.20.10 should work, as the node-local-dns binds it!

```
$ curl -L https://raw.githubusercontent.com/kubernetes/kubernetes/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml | sed 's/__PILLAR__DNS__DOMAIN__g/cluster.local/' | sed 's/__PILLAR__LOCAL__DNS__/169.254.20.10/' | sed 's/__PILLAR__DNS__SERVER__/172.20.0.10/' > node-local-dns.yaml

$ k apply -f node-local-dns.yaml
```

4. Repeat the step 2, but expect successful DNS lookup this time:

```
$ kru
If you don't see a command prompt, try pressing enter.
root@xenial-1550475807:/#
root@xenial-1550475807:/# apt-get update -y
Get:1 http://archive.ubuntu.com/ubuntu xenial InRelease [247 kB]
Get:2 http://security.ubuntu.com/ubuntu xenial-security InRelease [109 kB]
Get:3 http://security.ubuntu.com/ubuntu xenial-security/main amd64 Packages [785 kB]
Get:4 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [109 kB]
Get:5 http://archive.ubuntu.com/ubuntu xenial-backports InRelease [107 kB]
Get:6 http://archive.ubuntu.com/ubuntu xenial/main amd64 Packages [1558 kB]
*snip*
```

errordeveloper:
Thanks @mumoshu, also thanks for keeping this small.

pkg/nodebootstrap/userdata.go
```diff
@@ -84,7 +84,10 @@ func makeClientConfigData(spec *api.ClusterConfig, ng *api.NodeGroup) ([]byte, e
...
 }
 
-func clusterDNS(spec *api.ClusterConfig) string {
+func clusterDNS(spec *api.ClusterConfig, ng *api.NodeGroup) string {
+	if ng.ClusterDNS != "" {
+		return ng.ClusterDNS
+	}
 	// Default service network is 10.100.0.0, but it gets set 172.20.0.0 automatically when pod network
 	// is anywhere within 10.0.0.0/8
 	if spec.VPC.CIDR != nil && spec.VPC.CIDR.IP[0] == 10 {
```
errordeveloper: As a future improvement, we should probably move this into struct defaulting code path, and set the field to the default value. That is so the struct fully represents what's going on.

rsyvarth:
Just in case anyone else is implementing DNS caching using instructions here, there is a typo in mumoshu's instructions. Replace `__PILLAR__LOCAL__DNS__` with `169.254.20.10` not `169.254.169.254`

mumoshu:
@rsyvarth Oh! Good catch! Seems like I was too used to write 169.254.169.254.

mumoshu:
For anyone interested in this feature - an alternative way to use the node local DNS would be to specify `dnsPolicy: None` and `dnsConfig` in your pod spec:

For example, this:

```yaml
    dnsConfig:
        nameservers:
        - 169.254.20.10
        searches:
        - default.svc.cluster.local
        - svc.cluster.local
        - cluster.local
        - ap-northeast-1.compute.internal
        - us-west-2.compute.internal
        options:
        - name: attempts
          value: "3"
        - name: timeout
          value: "1"
        - name: rotate
      dnsPolicy: None
```

would result in pods having `/etc/resolv.conf` whose content is:

```
$ cat /etc/resolv.conf
nameserver 169.254.20.10
search default.svc.cluster.local svc.cluster.local cluster.local ap-northeast-1.compute.internal us-west-2.compute.internal
options attempts:3 timeout:1 rotate
```

whereas the default one used when I omit `dnsConfig` in my region(ap-northeast-1) is:

```
$ cat /etc/resolv.conf
nameserver 172.20.0.10
search default.svc.cluster.local svc.cluster.local cluster.local ap-northeast-1.compute.internal us-west-2.compute.internal
options ndots:5
```

The benefit of `dnsConfig` over the usage of kubelet's `--cluster-dns` flag would be that you can gradually migrate to the node-local dns cache.

You can even add the default cluster DNS as the secondary DNS server:

```
    dnsConfig:
        nameservers:
        - 169.254.20.10
        - 172.20.0.10
```

This way, even while the node-local dns is down due to a rolling update or a transient failure you'll be unlikely to notice the down time because your DNS client(from e.g. glibc) is likely to handle the failure with retries and/or parallel query(I believe it depends on the DNS client you rely on).

StevenACoffman:
@mumoshu 

Is it possible to add the default cluster DNS as the secondary DNS server for the node this way:
```
nodeGroups:
  - name: nodegroup1
    clusterDNS: "169.254.20.10,172.20.0.10"
    instanceType: m4.large
   # and whatever you like
```
Since kubelets options are: 

| Kubelet Option  | Description |
| ------------- | ------------- |
| `--cluster-dns stringSlice` | Comma-separated list of DNS server IP address. |

StevenACoffman:
Seems like it works! Hurrah!

StevenACoffman:
The [KEP warns against this for some reason](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md#risks-and-mitigations):
@prameshj From [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md#risks-and-mitigations)
> Populating both the nodelocal cache ip address and kube-dns ip address in resolv.conf is not a reliable option. Depending on underlying implementation, this can result in kube-dns being queried only if cache ip does not respond, or both queried simultaneously.

If musl queries nodelocal cache via tcp and kube-dns via udp in parallel, and one responds quickly and the other times out, why is that a problem?

mumoshu:
@StevenACoffman I guess you'll be interested on this line in the updated KEP https://github.com/kubernetes/enhancements/pull/1005/files#diff-a43ddcc01ee886cc9ca7c60a0900e436R166

The problem:

> so if we use both kube-dns IP as well as the link-local IP used by NodeLocal DNSCache, we could make the DNS query explosion problem worse. More queries means more conntrack entries and more DNATs.

The updated KEP also note:

> This workaround could be viable for client implementations that do round-robin.

But when I experimented I saw on recent distros the resolver worked parallely.

So my best bet that would work TODAY is that you have two nodelocal-dns daemonset, each listens on different virtual IP. All the application pods that wants to leverage the H/A nodelocal-dns-cache, the pod spec must be updated to include a `dnsConfig` section with the two virtual IPs in it.

mumoshu:
I have one more alternative: Just try running two nodelocaldns, both listening on the same virtual IP address and the port. Surveying github issues and the codedns code, it turned out that coredns [sets SO_REUSEPORT whereas available](https://github.com/coredns/coredns/blob/db34c1058937a421cc2769d5308d91ba841329c7/core/dnsserver/listen_go111.go#L18).

So two or more nodelocal dns pods with the same IP:Port should just work. This allows you to use H/A nodelocaldns behind a single IP address for your cluster dns(set via a kubelet option), which is automatically used when `dnsConfig` is inexistent in your pod specs.

StevenACoffman:
Thank you! That last alternative is very interesting.
By the way, In order to successfully start a fresh cluster with `clusterDNS: "169.254.20.10"` initially:
1. `eksctl create cluster --config-file=$EKSCONFIGFILE --without-nodegroup`
2. Apply nodelocaldns and any aws-node fixes using kubectl
3. `eksctl create nodegroup --config-file=$EKSCONFIGFILE`

If the nodegroups are created before nodelocaldns is applied, they will not come up successfully.

PR #552, opened 2019-02-18
Speedup tests

errordeveloper:
### Description

Tweaks to `Dockerfile`, add experimental `skaffold.yaml`.

### Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ build
 *.md
 DCO
 LICENSE*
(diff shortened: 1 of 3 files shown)


errordeveloper:
Still only a minor improvement for CircleCI, but something to begin with. However, with the given Skaffold config I can run `skaffold build --profile=local` and it only takes a few seconds whenever there are no changes, otherwise it builds only what's needed.

Issue #539, opened 2019-02-13
detect old cluster stack early


errordeveloper:
As it was seen in https://github.com/weaveworks/eksctl/issues/518#issuecomment-463122474 and https://github.com/weaveworks/eksctl/issues/538, users don't see an early warning of their clusters being outdated. Essentially it's to do with the fact that many eksctl commands still work for them, so they think they can carry on using all other commands. We need to find a way to detect a shape of an outdated cluster, and warn the user early.

errordeveloper:
I think it would very interesting to explore this from a perspective where actual detection is done by an external service that may apply machine learning instead of meticulously checking CloudFormation stacks for features. Having an external service would also help as it would decouple the features that a given version of eksctl can detect and say e.g. "your cluster is missing a critical feature, you must upgrade to latest version of ekctl and run <cmd>" or "your cluster has a security vulnerability, you must run <cmd>, and ...".

errordeveloper:
If we do decide to provide an external services, we still some general capabilities built in.
First we must identify eksctl releases that had distinct features/bugs - see #598.

PR #571, opened 2019-02-21
make new nodegroups inherit control plane version

errordeveloper:
### Description

- this allow two new modes:
  - `--version=auto` (new default) will inherit version of CP
  - `--version=latest` forces use of latest known version

### Checklist

### DIFF
--- a/pkg/ctl/create/nodegroup.go
+++ b/pkg/ctl/create/nodegroup.go
@@ -28,6 +28,8 @@ func createNodeGroupCmd(g *cmdutils.Grouping) *cobra.Command {
 	cfg := api.NewClusterConfig()
 	ng := cfg.NewNodeGroup()
 
+	cfg.Metadata.Version = "auto"
+
 	cmd := &cobra.Command{
 		Use:     "nodegroup",
 		Short:   "Create a nodegroup",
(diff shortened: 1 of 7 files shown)


errordeveloper:
cc @whereisaaron @christopherhein

errordeveloper:
I was wondering if we should add a check that go over existing nodegroups
and warns user if version drift is detected. Maybe another PR would be
best, but this looks like a good place to discuss it!

christopherhein:
> I was wondering if we should add a check that go over existing nodegroups and warns user if version drift is detected. Maybe another PR would be best, but this looks like a good place to discuss it!

I would say yes, didn't we have another issue that was about this type of check system? Maybe I'm confusing the one with checking if upgrades would be do-able given the state of CFN.

whereisaaron:
A drift warning seems an out of scope bell/whistle for this PR. But a PR for a `eksctl utils cluster-health` that checked versions and maybe things like ASG health and node status would be great.

Would it be possible to display the k8s version as a column in the `eksctl get nodegroup` output? Then you can see at a glance if there are differences across the node groups. You can kinda do that now with the AMI. 

```
$ if [[ $(eksctl get nodegroup --cluster=cave -o json | jq ".[].ImageID" | sort | uniq -u | wc -l) -gt 1 ]]; then echo 'DRIFT!'; else echo 'ALL GOOD!'; fi
ALL GOOD!
```

errordeveloper:
@christopherhein yes, you are thinking of #539.

@whereisaaron you are right, it's probably a good idea to add a separate command that checks state of cluster configuration overall. 

> Would it be possible to display the k8s version as a column in the `eksctl get nodegroup` output?

No, at the moment we don't store version information directly, we'd have ask Kubernetes for it. You can actually already do this yourself using `kubectl`. If it's the AMI that you want to know, I don't recall if that gets tracked by Kubernetes at all, but version of kubelet seems most relevant really.

pkg/ctl/create/nodegroup.go
```diff
@@ -96,6 +98,41 @@ func filterNodeGroups(cfg *api.ClusterConfig) error {
 	return nil
 }
 
+func checkVersion(ctl *eks.ClusterProvider, meta *api.ClusterMeta) error {
```
errordeveloper: Ended up with a simple function here to begin with, we might want to find a better place for it later. I think we might actually promote nodegroup-level version field to reflect the real world, so I'm not too worried about generalising this.

PR #584, opened 2019-02-26
eksctl utils describe-stacks --trail

errordeveloper:
### Description

When different users create clusters in a single AWS account, it's not very easy to tell who owns which cluster, and we shouldn't need to invent our own means, CloudTrail works nicely for this (albeit with 90 day limit). It's not very easy to "just use CloudTrail", as one needs to understand query parameters. Filtering by cluster name is also very untrivial, as one needs CloudFormation stack ID (not just stack name).

This PR makes an initial attempt of helping user with `eksctl utils describe stacks --trail`.

### Checklist

- [ ] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~ – the output of `eksctl utils describe-stacks` is very raw, I'd rather make it better before we promote it to the readme (see #585)


errordeveloper:
@palemtnrider I'll run the integration tests now, but these change don't really touch anything that the test suite currently covers...

errordeveloper:
Integration tests passed.

pkg/cfn/manager/api.go
```diff
@@ -310,6 +311,37 @@ func (c *StackCollection) DescribeStackEvents(i *Stack) ([]*cloudformation.Stack
...
+	input := &cloudtrail.LookupEventsInput{
+		LookupAttributes: []*cloudtrail.LookupAttribute{{
+			AttributeKey:   aws.String(cloudtrail.LookupAttributeKeyResourceName),
+			AttributeValue: id,
+		}},
+	}
+
+	events := []*cloudtrail.Event{}
+
+	pager := func(p *cloudtrail.LookupEventsOutput, last bool) (shouldContinue bool) {
```
errordeveloper: @christopherhein @tiffanyfay could you please help me understand if this is required, of Go SDK does this automatically for me?

errordeveloper: From our slack conversation, it turns out we should be using pagination (I'm still not 100% clear, but @tiffanyfay's reading of the docs is same as mine, so better stay on the safe side). I've opened #590 also, so we shall look into this later on for various other API calls.

PR #596, opened 2019-02-28
Design Proposal #005: Upgrades

errordeveloper:

PR #599, opened 2019-03-01
Add `eksctl update cluster` command

errordeveloper:
### Description

This is first cut, it moves `eksctl utils update-cluster-stack` functionality to `eksctl update cluster`.

### Checklist

### DIFF
--- a/pkg/ctl/update/cluster.go
+++ b/pkg/ctl/update/cluster.go
@@ -48,18 +48,6 @@ func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 }
 
 func doUpdateClusterCmd(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string) error {
-	ctl := eks.New(p, cfg)
-
-	printer := printers.NewJSONPrinter()
-
-	if err := api.Register(); err != nil {
-		return err
-	}
-
-	if err := ctl.CheckAuth(); err != nil {
-		return err
-	}
-
 	if cfg.Metadata.Name != "" && nameArg != "" {
 		return fmt.Errorf("--name=%s and argument %s cannot be used at the same time", cfg.Metadata.Name, nameArg)
 	}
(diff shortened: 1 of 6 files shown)


pkg/ctl/update/cluster.go
```diff
@@ -0,0 +1,99 @@
...
+		fs.BoolVar(&updateClusterDryRun, "dry-run", updateClusterDryRun, "do not apply any change, only show what resources would be added")
+	})
+
+	cmdutils.AddCommonFlagsForAWS(group, p, false)
+
+	group.AddTo(cmd)
+	return cmd
+}
+
+func doUpdateClusterCmd(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string) error {
```
stefanprodan: Can we print the Kubernetes version and ask for confirmation? 

Something like:
```
The gopher cluster will be upgrade to 1.13.3 do you want to proceed? yest/no
```

pkg/ctl/update/cluster.go
```diff
@@ -0,0 +1,99 @@
...
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/eks"
+	"github.com/weaveworks/eksctl/pkg/printers"
+)
+
+var updateClusterDryRun = true
+
+func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
```
stefanprodan: We could have a `--version` optional flag, when not specified it will use the latest stable version of EKS.

pkg/ctl/update/cluster.go
errordeveloper: We have `--dry-run` on by default. This doesn't actually do version updates, that's for another PR.

pkg/ctl/update/cluster.go
errordeveloper: This comment should go to #596 really. I agree we will probably need it, but not right away as right now there is only one version to upgrade to, but this code doesn't handle update yet anyway.

PR #600, opened 2019-03-01
Add functionality to increment control plane version

errordeveloper:
### Description

This carries some refactoring, but crucially it adds a call to `eks.UpdateClusterVersion` from `eksctl update cluster`.

### Checklist

- [ ] ~Added tests that cover your change (if possible)~ - should look testing upgrades separately
- [ ] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~ - will make another PR

### DIFF
--- a/pkg/ctl/update/cluster.go
+++ b/pkg/ctl/update/cluster.go
@@ -15,11 +15,16 @@ import (
 	"github.com/weaveworks/eksctl/pkg/printers"
 )
 
-var updateClusterDryRun = true
+var (
+	updateClusterDryRun = true
+	updateClusterWait   = true
+)
 
 func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	p := &api.ProviderConfig{}
 	cfg := api.NewClusterConfig()
+
+	// cfg.Metadata.Version = "next"
 
 	cmd := &cobra.Command{
 		Use:   "cluster",
@@ -37,8 +42,9 @@ func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	group.InFlagSet("General", func(fs *pflag.FlagSet) {
 		fs.StringVarP(&cfg.Metadata.Name, "name", "n", "", "EKS cluster name (required)")
 		cmdutils.AddRegionFlag(fs, p)
-		cmdutils.AddVersionFlag(fs, cfg.Metadata, "")
+		// cmdutils.AddVersionFlag(fs, cfg.Metadata, `"next" and "latest" can be used to automatically increment version by one, or force latest`)
 		fs.BoolVar(&updateClusterDryRun, "dry-run", updateClusterDryRun, "do not apply any change, only show what resources would be added")
+		cmdutils.AddWaitFlag(&updateClusterWait, fs, "all update operations to complete")
 	})
 
 	cmdutils.AddCommonFlagsForAWS(group, p, false)
@@ -62,6 +68,11 @@ func doUpdateClusterCmd(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg s
 
 	ctl := eks.New(p, cfg)
 
+	if !ctl.IsSupportedRegion() {
+		return cmdutils.ErrUnsupportedRegion(p)
+	}
+	logger.Info("using region %s", cfg.Metadata.Region)
+
 	printer := printers.NewJSONPrinter()
 
 	if err := api.Register(); err != nil {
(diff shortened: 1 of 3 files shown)


errordeveloper:
This works, but hits the timeout. We might need to review the value of the default timeout, for now should probably test `--timeout=30m` or something in that regions (default is 20m).

pkg/ctl/update/cluster.go
```diff
@@ -15,12 +15,17 @@ import (
...
+var (
+	updateClusterDryRun = true
+	updateClusterWait   = true
+)
 
 func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	p := &api.ProviderConfig{}
 	cfg := api.NewClusterConfig()
 
+	// cfg.Metadata.Version = "next"
```
christopherhein: What does `next` represent?

christopherhein: Ah, nevermind I see the other comment below. I'm not super keen `next` that feels a little off. What is your thinking?

pkg/ctl/update/cluster.go
```diff
@@ -37,8 +42,9 @@ func updateClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	group.InFlagSet("General", func(fs *pflag.FlagSet) {
 		fs.StringVarP(&cfg.Metadata.Name, "name", "n", "", "EKS cluster name (required)")
 		cmdutils.AddRegionFlag(fs, p)
-		cmdutils.AddVersionFlag(fs, cfg.Metadata, "")
+		// cmdutils.AddVersionFlag(fs, cfg.Metadata, `"next" and "latest" can be used to automatically increment version by one, or force latest`)
```
tiffanyfay: Why the // ?

pkg/ctl/update/cluster.go
tiffanyfay: I'm not really sure what else would be good for saying next version. I don't think we'd be able to have no version given have a default.

pkg/ctl/update/cluster.go
```diff
@@ -72,6 +83,23 @@ func doUpdateClusterCmd(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg s
...
+	currentVersion := ctl.ControlPlaneVersion()
+	switch currentVersion {
+	case "":
+		return fmt.Errorf("unable to get control plane version")
+	case api.Version1_10:
+		cfg.Metadata.Version = api.Version1_11
+	case api.LatestVersion:
+		cfg.Metadata.Version = api.LatestVersion
+	default:
+		return fmt.Errorf("control plane version version %q is not upgradable by this version of eksctl, try to upgrade eksctl first", currentVersion)
```
tiffanyfay: What about also mentioning that eks might not support the version and to check the docs?

pkg/ctl/update/cluster.go
errordeveloper: I wanted to live it out for a follow-up PR.

pkg/ctl/update/cluster.go
errordeveloper: 👍

pkg/ctl/update/cluster.go
errordeveloper: This is documented in #596. The idea is to have `next` (default), `latest` and _semver_. But I'm not entirely sure I'll manage to implement more then one option in this PR, so I left it commented out as as there isn't much use to a flag that only supports a default value.

errordeveloper:
> It doesn't look like this support the config file since we have `metadata.version` is there anyway we could easily add support for that OOTB?

Yes, I'll make sure to work on that today. I do prefer to leave it for another PR.

errordeveloper:
@christopherhein @tiffanyfay thanks a lot for your input, I'm gonna go ahead and merge this, open a follow-up issue to document what needs doing before we can cut the release.

errordeveloper:
Follow-up: https://github.com/weaveworks/eksctl/issues/608.

pkg/ctl/update/cluster.go
errordeveloper: Actually, this is about version you currently have, but something is not right about this code. I'll have a look at what it improving this before I merge the PR.

errordeveloper: I added comment and tweaked the error message.

pkg/ctl/update/cluster.go
errordeveloper: Another reasons for leaving this until later is that currently there is only one upgrade path to test _1.10 - 1.11_, which implies that `--version=next` as well as  `--version=latest` as well as  `--version=1.11` do the same thing.

errordeveloper:
Integration tests have passed.

Issue #443, opened 2019-01-17
add simple way to replace a nodegroup


errordeveloper:
We can create and delete nodegroups, but one needs to really script their upgrade and it can be tedious. If they are to use e.g. a custom AMI and want to test it first, they probably know how to do tests and will split the phases themeselves.

In many cases users probably want to create a new nodegroup based on an existing one. I think a good way to do this would be with `eksctl create nodegroup --replaces=<oldNodeGroup>`, which would take most configuration from `oldNodeGroup` and use it for the new nodegroup, and make sure to delete that old nodegroup once new nodes joined successfully. We can add a mode where latest AMI from the same family is applied, and a parameter to delayed the deletion by some number of minutes, and perhaps scale the old nodegroup down slowly instead of deleting it immediately (just an idea for those that prefer to play safe).

mumoshu:
Ideally we should finish #396 beforehand and integrate it into this feature. So that the node draining ensures all the old nodes are safely drained (hence pods are migrated to the new nodes).

errordeveloper:
cc @tiffanyfay

pawelprazak:
an interesting strategy would be to do a blue/green with two node pools, esp. with built in self-tests like:
- create new node pool with two nodes, mark as non-schedulable
- deploy some smoke test pods (with node selectors and taint tolerations)
- mark as schedulable
- drain one old node in the old node pool
- continue the rolling update if all new pods are running and ready
- (optionally) pause at 50/50 and 99/1 for a predetermined delay
- delete old node pool

this would require an operator and a lot of effort, but would automate a lot of QA work

errordeveloper:
I think it is a good idea, definitely! We also have though today a few test
workloads for functional verification, essentially something that we could
probably extract from the integration suite with a few tweaks.

errordeveloper:
This would depend on #642.

polanfong:
Any progress on this? I've been looking for the exact same thing. The closest I can find is the `eksctl upgrade nodegroup ...` command but I have a feeling this doesn't result in the same thing. Can anyone comment on the difference(s)?

michaelbeaumont:
Duplicate of #2774

Dhruv-Garg79:
any plans to add this?

PR #610, opened 2019-03-05
Document upgrades

errordeveloper:
### Description

Adds documentation for upgrades.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,8 +249,9 @@ To delete a nodegroup, run:
 ```
 eksctl delete nodegroup --cluster=<clusterName> --name=<nodegroupName>
 ```
+> NOTE: this will drain all pods from that nodegroup before the instances are deleted.
 
-All node are cordoned and all pods are evicted from a nodegroup on deletion,
+All nodes are cordoned and all pods are evicted from a nodegroup on deletion,
 but if you need to drain a nodegroup without deleting it, run:
 ```
 eksctl drain nodegroup --cluster=<clusterName> --name=<nodegroupName>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+
+##### Updating multiple nodegroups
+
+If you have multiple nodegroups, it's your responsibility to track how each one was configured.
+You can do this by using config files, but if you haven't used it already, you will need to inspect
+your cluster to find out how each nodegroup was configured.
+
+In general terms, you are looking to:
+- review nodegroups you have and which ones can be delete or must be replaced with new version
+- note down configuration of each nodegroup, consider using config file to ease upgrades next time
```
rndstr: is it obvious where to find the configuration for someone upgrading or do we need to give a hand here?

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+```
+eksctl create nodegroup --cluster=<clusterName>
+```
+
+Delete old nodegroup:
+```
+eksctl delete nodegroup --cluster=<clusterName> --name=<oldNodeGroupName>
+```
+
+##### Updating multiple nodegroups
```
rndstr: do we want to split this section into `Updating multiple nodegroups with config file` and `Updating multiple nodegroups without config file`?

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+
+Edit config file to add new nodegroups.
+
+If you chose to deleted definition of old nodegoups from the config file, then you can run:
+```
+eksctl create nodegroup --config-file=<path>
+```
+
+Alternatively, you can pass `--only` with a list of the new nodegroup names that you want to
+add (you can use glob expressions, or list specific names - see above for details):
```
rndstr: not sure where `see above` refers to

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+To delete old nodegroup:
+```
+eksctl create nodegroup --cluster=<clusterName>
+```
+
+If you are using config file, you will need to do the following.
+
+Get the list of old nodegroups:
+```
+old_nodegroups="$(eksctl get ng --cluster=<clusterName> --output=json | jq -r '.[].Name')"
```
rndstr: why would i need to keep track of the old nodegroups when i use the config file to delete the old ones?

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+```
+
+Edit config file to add new nodegroups.
+
+If you chose to deleted definition of old nodegoups from the config file, then you can run:
+```
+eksctl create nodegroup --config-file=<path>
```
rndstr: I'm not sure I understand this part, will this command delete nodegroups not in the config file?

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+```
+
+Once you have new nodegroups in place, you can delete old ones:
+```
+for ng in $old_nodegroups ; do eksctl delete nodegroup --cluster=<clusterName> --name=$ng ; done
+```
+
+#### Updating default add-ons
+
+There 3 default add-ons that get included in each EKS cluster, the process for updating each of them is different, hence
```
rndstr:
```suggestion
There are 3 default add-ons that get included in each EKS cluster, the process for updating each of them is different, hence
```

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+eksctl utils update-kube-proxy
+```
+
+
+To update `kube-system:ds/aws-node`, run:
+```
+eksctl utils update-aws-node
+```
+
+If you have upgraded from 1.10 to 1.11, you will need to replace `kube-system:deployment/kube-dns` with `kube-system:deployment/coredns`.
```
rndstr: what if i'm going from 1.11 to 1.12, how do i upgrade `kube-dns`?

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
...
+If you are using config file, you will need to do the following.
+
+Get the list of old nodegroups:
+```
+old_nodegroups="$(eksctl get ng --cluster=<clusterName> --output=json | jq -r '.[].Name')"
+```
+
+Edit config file to add new nodegroups.
+
+If you chose to deleted definition of old nodegoups from the config file, then you can run:
```
rndstr:
```suggestion
If you choose to delete the definition of old nodegroups with the config file, then you can run:
```
(assuming here that nodegroups not found in the config file will be automatically deleted)

README.md
errordeveloper: Yes, we will need to, but cannot at the moment. We should move on to caching config inside the cluster, perhaps in a configmap to start with. This would need to be solved as we get closer to `eksctl apply` (also see #443).

README.md
errordeveloper: Not sure at the moment, can always improve later on. We actually need a proper website for docs somewhat urgently.

README.md
errordeveloper: Well spotted, will try to clarify.

README.md
errordeveloper: I guess it all depends which way you edit the file and how you apply the edits, so you could delete old ones from the file to begin with avoid having to set `--only`. There is definitely much room for improvement of the UX overall, this one of the somewhat tricky things.

README.md
errordeveloper: No, it will only create new ones, and unless you've deleted the definitions of the old ones before hand - it will complain about those. You'd want to set `--only=<listOfNewNGs>` otherwise.

README.md
errordeveloper: That's unknown at present.

README.md
```diff
@@ -261,11 +261,130 @@ To uncordon a nodegroup, run:
 eksctl drain nodegroup --cluster=<clusterName> --name=<nodegroupName> --undo
 ```
 
+### Cluster Upgrades
+
+An _`eksctl`-managed_ cluster can be upgraded in 3 easy steps:
+
+1. update control plane version with `eksctl update cluster`
+2. replace each of the nodegroups by creating a new one and deleting the old one
+3. update default add-ons:
```
errordeveloper: There a few objects there, we should just name the add-ons instead of referring to the workload objects...

README.md
errordeveloper: I think `--only` is documented quite clearly just below this.

README.md
```diff
@@ -261,11 +261,133 @@ To uncordon a nodegroup, run:
...
+If you are using config file, you will need to do the following.
+
+Get the list of old nodegroups:
+```
+old_nodegroups="$(eksctl get ng --cluster=<clusterName> --output=json | jq -r '.[].Name')"
+```
+
+Edit config file to add new nodegroups.
+
+If you choose to delete the definition of old nodegroups from the config file, then you can run:
```
rndstr: i understood now, i think the `delete` made me think of `eksctl delete`, I would reword as such:
```suggestion
If you have removed the definition of the old nodegroups from the config file, then you can run:
```

README.md
```diff
@@ -261,11 +261,133 @@ To uncordon a nodegroup, run:
...
+```
+eksctl create nodegroup --config-file=<path>
+```
+
+Alternatively, you can keep the old definitions and pass `--only` with a list of the new nodegroup
```
rndstr:
```suggestion
Alternatively, you can keep the old definitions in the config-file and pass `--only` with a list of the new nodegroup
```

whereisaaron:
Let me hear it for doc-u-men-tation! Woop! Woop!

PR #611, opened 2019-03-05
Improve printers, use Go 1.12

### LINKED ISSUE #603: Switch to Go 1.12
It would be good to stay up to date. We may wish to try modules later, but it is not a priority.

errordeveloper:
### Description

Turns out Go 1.12 has `strings.ReplaceAll`, which is quite handy. The way `LogObj` worked wasn't particularly great, this allows user to specify full format string (they must include one `%s` for the actual object. We now also ensure that the actual objects in question don't get use as formatting strings (e.g. some Kubernetes field values use `%` character), and we also try to serialise directly whenever API-aware serialiser fails to (e.g. unregistered APIs, missing `apiVersion` etc).

Closes #603.

### Checklist

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -7,6 +7,7 @@
         "misspell"
     ],
     "Exclude": [
+        "\/src\/(crypto|internal|net|os|runtime|syscall)\/",
         "^pkg\/nodebootstrap\/assets.go",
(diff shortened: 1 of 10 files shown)


errordeveloper:
Will run integration tests, just in case :)

PR #623, opened 2019-03-11
Add command `eksctl utils aws-auth`

rndstr:
Adds the command
```
eksctl utils aws-auth --name=<name> [--add-admin-role=<arn>] [--remove-role=<arn>] [--add-account=<number>] [--remove-account=<number>]
```
to allow modifying the `kube-system/aws-auth` object which is a mapping from IAM entities to Kubernetes groups.

PR #632, opened 2019-03-13
Initialize ObjectMeta in authconfigmap

rndstr:
If a ConfigMap doesn't exist we also need to initialize the metadata.

Follow-up to #623 and #629


errordeveloper:
This fixed the integration test! thank you :)

rndstr:
@errordeveloper great, so i guess they are working as intended then :)

Issue #625, opened 2019-03-13
authconfigmap: command to manipulate aws auth configmap


rndstr:
PR #623 attempted to add a command to modify the auth configmap but its location and command hierarchy needs a bit more discussion.

The auth ConfigMap is in `kube-system` and named `aws-auth`. It contains various mappings of IAM/AWS entities to Kubernetes groups.

More information at
- https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html
- https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/README.md#full-configuration-format

## A. command design
The command should provide the following invocations (with description of current implementation of feature)

- display the current config
- add a role by arn, username, and groups
  - no de-duplication with existing mappings (related to #544)
- remove a role by arn
  - remove a single role even if there are duplicates (see #544)
  - fail if role not found
- add an account number
  - de-duplicate
- remove an account number
  - fail if account number not found

Some ideas and the start of the discussion can be found in #623.

It seems like that the following fulfills most of our constraints:
```
eksctl get aws-auth
eksctl update aws-auth add-role <arn> --username <string> --groups <csv>
eksctl update aws-auth remove-role <arn>
eksctl update aws-auth add-account <number>
eksctl update aws-auth remove-account
```
some variations:
- `add => map && remove => unmap`
- `aws-auth => auth`

## B. command behavior

Further discussion can be had about changing the way commands behave:

- remove a role
  - TBD: remove by role arn only? or match username and groups?

errordeveloper:
For the record, we have a712049 that reverted the first implementation, so that can be reverted back.

errordeveloper:
xref #62

rndstr:
Somre more thoughts on naming 

- the flag `--update-auth-config-map` for `eksctl create|delete nodegroup` describes the same thing
- `iam-auth-bindings` has been [brought up](https://github.com/weaveworks/eksctl/pull/623#issuecomment-472229878) to bring it on par with upcoming IAM features

So more options
- `eksctl update iam-auth-bindings`
- `eksctl update iam-auth`
- `eksctl update auth-config-map`

StevenACoffman:
If it is helpful, our use case is that we have four IAM roles: eksctl, read-only, developer, and admin. Our AWS accounts have no access rights except to assume one (or more) of those roles based on group membership.

When we create a new cluster, we have a horrible bash script that inserts into the aws-auth configmap a few lines after the line: `  mapRoles: |`:
[...]

Issue #544, opened 2019-02-14
authconfigmap: shared roles and duplication


errordeveloper:
Currently, it's possible to create an NG that borrows an IAM role from another NG. This effectively results in duplicated entries. #528 doesn't deal with this, but we should devise a method for that soon enough.

errordeveloper:
Our roles are named as `arn:aws:iam::<account>:role/eksctl-<cluster>-nodegroup-<ng>-NodeInstanceRole-<prefix>`, so we can detect it and only cleanup roles that match the nodegroup in question, but this would still remove roles for other nodegroups.

I am not sure what `mapRoles` API currently allows us to store, perhaps there are useful fields that we can add... but actually, we should be able to add another top-level field in the configmap.

E.g. something like this:

```
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-syste
data:
  mapRoles: |
    - username: 'system:node:{{EC2PrivateDNSName}}'
      groups: [ 'system:bootstrappers', 'system:nodes' ]
      rolearn: 'arn:aws:iam::122333:role/eksctl-cluster-5a-nodegroup-ng1-p-NodeInstanceRole-NNH3ISP12CX'
   eksctlInfo: |
      - rolearn: 'arn:aws:iam::122333:role/eksctl-cluster-5a-nodegroup-ng1-p-NodeInstanceRole-NNH3ISP12CX'
        owned: true
        cluster: cluster-5a
        nodegroups: ['ng1-public']
```

christopherhein:
I think #528 did account for this by only deleting one entry then breaking out of the loop, @tiffanyfay isn't that right?

tiffanyfay:
Yes
---
Case 1: Duplicate role in configmap
```
  mapRoles: |
    - username: 'system:node:{{EC2PrivateDNSName}}'
      groups: [ 'system:bootstrappers', 'system:nodes' ]
      rolearn: 'arn:aws:iam::122333:role/eksctl-cluster-5a-nodegroup-ng1-p-NodeInstanceRole-NNH3ISP12CX'
    - username: 'system:node:{{EC2PrivateDNSName}}'
      groups: [ 'system:bootstrappers', 'system:nodes' ]
      rolearn: 'arn:aws:iam::122333:role/eksctl-cluster-5a-nodegroup-ng1-p-NodeInstanceRole-NNH3ISP12CX'
```
`eksctl delete ng --cluster cluster-5a ng1-public` results in:
```
[...]

PR #635, opened 2019-03-14
Allow setting `--update-auth-config-map` on nodegroup creation

### LINKED ISSUE #634: Add --update-auth-config-map to create nodegroup
We have it in `eksctl delete ng`, so it makes sense to add it in `eksctl create ng` also. It would allow user to have a way to opt-out from config map management entirely, perhaps they could use separate commands or do something else.

errordeveloper:
### Description

We have it in `eksctl delete ng`, so it makes sense to add it in `eksctl create ng` also. It would allow user to have a way to opt-out from automatic configmap management entirely, perhaps they could use separate commands (see #625) or do something else.

- close #634 
- xref #544

### Checklist

### DIFF
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -62,6 +62,11 @@ func AddWaitFlag(wait *bool, fs *pflag.FlagSet, description string) {
 	fs.BoolVarP(wait, "wait", "w", *wait, fmt.Sprintf("wait for %s before exiting", description))
 }
 
+// AddUpdateAuthConfigMap adds common --update-auth-configmap flag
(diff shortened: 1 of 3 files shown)


rndstr:
this changes the behavior of `eksctl create nodegroup` to no longer add the instance role ARN to `aws-auth`, something to keep in mind.

errordeveloper:
> this changes the behavior of `eksctl create nodegroup` to no longer add the instance role ARN to `aws-auth`, something to keep in mind.

Actually the default behaviour will be the same.

rndstr:
> Actually the default behaviour will be the same.

My bad, I blindly assumed a boolean flag defaults to `false`.

Issue #396, opened 2019-01-06
feat: add taints to the node on creation


mumoshu:
As @whereisaaron suggested in another issue at https://github.com/weaveworks/eksctl/issues/394#issuecomment-451508249, this is necessary to avoid misplacing pods after the kubelet created the node resource but before the user adds a taint(whose purpose is to avoid the pods being scheduled!)

kubelet provides the `--taints` option to adds taints on node creation, which completely resolves this problem.

I'd suggest adding `--taints` or `--node-taints` to the command-line flags for `create cluster` and `create nodegroup`, and possibly in another PR `taints` or `nodeTaints` in another PR after #381 lands.

aparamon:
It is proposed to provide a more general feature of passing arbitrary [kubelet options](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#options), similar to 
[EKS BootstrapArguments](https://docs.aws.amazon.com/en_us/eks/latest/userguide/getting-started.html#eks-launch-workers).

Specifying taints would indeed be very useful, in order to implement dedicated Load Balancer nodes:
https://github.com/weaveworks/eksctl/issues/440.

errordeveloper:
It'd be good to make use this change as an opportunity to switch over from flags to kubelet config. We already have config, we just need to start generating and move labels and some of the other flags into it.

errordeveloper:
There had been a few requests for this, so I might start working on this for the next release.

errordeveloper:
We should switch to kubelet config files (https://github.com/weaveworks/eksctl/issues/156) before implementing this.

PR #644, opened 2019-03-19
Move all supported flags to `KubeletConfiguration`

errordeveloper:
### Description

Move all supported flags to `KubeletConfiguration` (towards #156).

- remaining flags are not supported yet
- cosmetic improvements to test suite

### Checklist


pkg/nodebootstrap/userdata.go
```diff
@@ -96,20 +98,35 @@ func clusterDNS(spec *api.ClusterConfig, ng *api.NodeGroup) string {
...
 }
 
-func makeKubeletParamsCommon(spec *api.ClusterConfig, ng *api.NodeGroup) []string {
+func makeKubeletConfigYAML(spec *api.ClusterConfig, ng *api.NodeGroup) ([]byte, error) {
+	data, err := Asset("kubelet.yaml")
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: use "k8s.io/kubernetes/pkg/kubelet/apis/kubeletconfig" and runtime.Decode/Encode
```
errordeveloper: Current approach is okay, but probably best to try switching over to using the struct.

errordeveloper:
@rndstr FYI, this is towards #396, however it turned out we still have to use flags to set taints.

PR #653, opened 2019-03-21
ignore existing nodegroups when creating with config file

### LINKED ISSUE #650: ignore existing nodegroups when creating with config file
At the moment `eksctl create ng --config-file=<path>` will trip over any existing nodegroups, we should add a flag that allows user to avoid that without having to pass `--only=<filter>`.

A good name would be `--ignore-existing`, we could make a default also.

errordeveloper:
### Description

Closes #650.

### Checklist

### DIFF
--- a/pkg/cfn/manager/tasks.go
+++ b/pkg/cfn/manager/tasks.go
@@ -67,38 +67,27 @@ func (c *StackCollection) CreateClusterWithNodeGroups() []error {
 		return errs
 	}
 
-	return c.CreateAllNodeGroups(false)
+	return c.CreateAllNodeGroups(nil)
 }
 
 // CreateAllNodeGroups runs all tasks required to create the node groups;
 // any errors will be returned as a slice as soon as one of the tasks
 // or group of tasks is completed
-func (c *StackCollection) CreateAllNodeGroups(ignoreExisting bool) []error {
+func (c *StackCollection) CreateAllNodeGroups(onlySubset sets.String) []error {
 	errs := []error{}
 	appendErr := func(err error) {
 		errs = append(errs, err)
 	}
 
-	existing := sets.NewString()
-
-	if ignoreExisting {
-		existingStacks, err := c.DescribeNodeGroupStacks()
-		if err != nil {
-			return []error{err}
-		}
-		for _, s := range existingStacks {
-			existing.Insert(getNodeGroupName(s))
-		}
-	}
 	createAllNodeGroups := []Task{}
 	for i := range c.spec.NodeGroups {
 		ng := c.spec.NodeGroups[i]
+		if onlySubset != nil && !onlySubset.Has(ng.Name) {
+			continue
+		}
 		t := Task{
 			Call: c.CreateNodeGroup,
 			Data: ng,
-		}
-		if existing.Has(ng.Name) {
-			continue
 		}
 		createAllNodeGroups = append(createAllNodeGroups, t)
 	}
--- a/pkg/cfn/manager/tasks.go
+++ b/pkg/cfn/manager/tasks.go
@@ -3,6 +3,8 @@ package manager
 import (
 	"sync"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"github.com/kris-nova/logger"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
@@ -65,23 +67,38 @@ func (c *StackCollection) CreateClusterWithNodeGroups() []error {
 		return errs
 	}
 
-	return c.CreateAllNodeGroups()
+	return c.CreateAllNodeGroups(false)
 }
 
 // CreateAllNodeGroups runs all tasks required to create the node groups;
 // any errors will be returned as a slice as soon as one of the tasks
 // or group of tasks is completed
-func (c *StackCollection) CreateAllNodeGroups() []error {
+func (c *StackCollection) CreateAllNodeGroups(ignoreExisting bool) []error {
 	errs := []error{}
 	appendErr := func(err error) {
 		errs = append(errs, err)
 	}
 
+	existing := sets.NewString()
+
+	if ignoreExisting {
+		existingStacks, err := c.DescribeNodeGroupStacks()
+		if err != nil {
+			return []error{err}
+		}
+		for _, s := range existingStacks {
+			existing.Insert(getNodeGroupName(s))
+		}
+	}
 	createAllNodeGroups := []Task{}
 	for i := range c.spec.NodeGroups {
+		ng := c.spec.NodeGroups[i]
 		t := Task{
 			Call: c.CreateNodeGroup,
-			Data: c.spec.NodeGroups[i],
+			Data: ng,
+		}
+		if existing.Has(ng.Name) {
+			continue
 		}
 		createAllNodeGroups = append(createAllNodeGroups, t)
 	}
--- a/pkg/cfn/manager/tasks.go
+++ b/pkg/cfn/manager/tasks.go
@@ -62,12 +62,12 @@ func (c *StackCollection) RunSingleTask(t Task) []error {
 // the stacks (a cluster and one or more nodegroups); any errors
 // will be returned as a slice as soon as one of the tasks or group
 // of tasks is completed
-func (c *StackCollection) CreateClusterWithNodeGroups() []error {
+func (c *StackCollection) CreateClusterWithNodeGroups(onlySubset sets.String) []error {
 	if errs := c.RunSingleTask(Task{c.CreateCluster, nil}); len(errs) > 0 {
 		return errs
 	}
 
-	return c.CreateAllNodeGroups(nil)
+	return c.CreateAllNodeGroups(onlySubset)
 }
 
 // CreateAllNodeGroups runs all tasks required to create the node groups;
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -10,7 +10,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	"github.com/weaveworks/eksctl/pkg/ami"
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
 	"github.com/weaveworks/eksctl/pkg/authconfigmap"
 	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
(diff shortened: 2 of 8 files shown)


errordeveloper:
TODOs:

- [ ] perform more manual tests
- [ ] fix output messages about nodegroup count

pkg/ctl/create/cluster.go
```diff
@@ -387,7 +305,7 @@ func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
 				return err
 			}
 
-			if err := checkEachNodeGroup(cfg, canUseForPrivateNodeGroups); err != nil {
+			if err := CheckEachNodeGroup(ngFilter, cfg, canUseForPrivateNodeGroups); err != nil {
```
martina-if: Unrelated: I would move these checks like `canUseForPrivateNodeGroups` to actual "private" functions at the end of these file so this function is shorted and contains the high level stuff. Then details like `canUseForPrivateNodeGroups` can be somewhere else and looked up later if needed. That way the function stays more relevant for the reader and less cluttered.  What do you think about that?

pkg/ctl/create/nodegroup.go
```diff
@@ -52,8 +52,11 @@ func createNodeGroupCmd(g *cmdutils.Grouping) *cobra.Command {
 		cmdutils.AddRegionFlag(fs, p)
 		cmdutils.AddVersionFlag(fs, cfg.Metadata, `for nodegroups "auto" and "latest" can be used to automatically inherit version from the control plane or force latest`)
 		fs.StringVarP(&clusterConfigFile, "config-file", "f", "", "load configuration from a file")
-		fs.StringVarP(&nodeGroupFilter, "only", "", "",
+
+		fs.StringSliceVar(&nodeGroupOnlyFilters, "only", nil,
 			"select a subset of nodegroups via comma-separted list of globs, e.g.: 'ng-*,nodegroup?,N*group'")
+		fs.BoolVar(&ngFilter.IgnoreAllExisting, "ignore-existing", ngFilter.IgnoreAllExisting, "ignore all existing nodegroups")
```
martina-if: I would make this the default behavior, and then print a message saying "skipped creation of existing nodegroup X". And  then possibly add a "strict" flag? Hmmm I would need to think a bit about this,

pkg/ctl/create/utils.go
```diff
@@ -0,0 +1,230 @@
...
+	return &NodeGroupFilter{
+		IgnoreAllExisting: true,
+
+		existing: sets.NewString(),
+	}
+}
+
+// ApplyOnlyFilter parses given globs for exclusive filtering
+func (f *NodeGroupFilter) ApplyOnlyFilter(globs []string, cfg *api.ClusterConfig) error {
+	for _, gs := range globs {
```
martina-if: Can we maybe alias the library so we can actually name one of these `glob` instead of `gs` or `g` to make it less confusing? (Maybe compiledGlob instead of `g`?)
Or name it ngGlob or something like that, perhaps.

pkg/ctl/create/utils.go
```diff
@@ -0,0 +1,230 @@
...
+	// if no globs were given, match everything,
+	// if false - we haven't matched anything so far
+	return len(f.only) == 0
+}
+
+// MatchAll checks all nodegroups against the filter and returns all of
+// matching names as set
+func (f *NodeGroupFilter) MatchAll(cfg *api.ClusterConfig) sets.String {
+	names := sets.NewString()
+	for i := range cfg.NodeGroups {
```
martina-if: nitpick: use `ng` instead of referencing it by the index?

pkg/cfn/manager/tasks.go
```diff
@@ -65,23 +67,27 @@ func (c *StackCollection) CreateClusterWithNodeGroups() []error {
...
 
 	createAllNodeGroups := []Task{}
 	for i := range c.spec.NodeGroups {
+		ng := c.spec.NodeGroups[i]
 		t := Task{
 			Call: c.CreateNodeGroup,
-			Data: c.spec.NodeGroups[i],
+			Data: ng,
+		}
+		if onlySubset != nil && !onlySubset.Has(ng.Name) {
```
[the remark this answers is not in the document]
errordeveloper: Yes, good idea! 👍

pkg/ctl/create/cluster.go
```diff
@@ -46,11 +45,13 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	cfg := api.NewClusterConfig()
 	ng := cfg.NewNodeGroup()
 
+	ngFilter := NewNodeGroupFilter()
```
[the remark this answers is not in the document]
errordeveloper: This was mostly to mirror `createNodeGroupCmd`/`doCreateNodeGroup`, but also eventually we'd like to add `--only` to all of the commands (see [upgrades proposal doc](https://github.com/weaveworks/eksctl/blob/master/docs/proposal-005-upgrades.md#cli-design), for example). We might rename the filter, but it would need to be set based on flags, so it seems to makes sense to keep it here.

pkg/ctl/create/cluster.go
errordeveloper: Indeed, most of it needs refactoring! There is a lot of stuff that could be done in a better way. This PR only moves some stuff over to another file, but it would be great if we could refine this mess into a few packages. A stretch goal would be to turn it all into something that can be consumed as a library more easily, and something we can unit-test too!

pkg/ctl/create/nodegroup.go
errordeveloper: This is actually on by default already (see [`NewNodeGroupFIlter`](https://github.com/weaveworks/eksctl/pull/653/files#diff-ed6d54018faf3cc05074f043771cea47R31)). I'm not sure what would could be a real use-case for the strict behaviour, so I'm tempted to remove the flag.

pkg/ctl/create/utils.go
errordeveloper: Indeed, thanks! 👍

pkg/ctl/create/utils.go
errordeveloper: Yes, we can. I just copied it from `CheckEachNodeGroup` below, where we have to use index due to [issue with callbacks](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables).

pkg/ctl/create/utils.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+	return names
+}
+
+// CheckEachNodeGroup iterates over each nodegroup and calls check function
+// (this is need to avoid common goroutine-for-loop pitfall)
+func CheckEachNodeGroup(f *NodeGroupFilter, cfg *api.ClusterConfig, check func(i int, ng *api.NodeGroup) error) error {
+	for i := range cfg.NodeGroups {
+		if f.Match(cfg.NodeGroups[i]) {
```
errordeveloper: Well, actually, this would look nicer if we do `ng := cfg.NodeGroups[i]` at the beginning, cause we now refer to it twice... I'll change it here, and above.

pkg/ctl/create/utils.go
```diff
@@ -0,0 +1,230 @@
...
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("no nodegroups match filter specification: %q", f.onlySpec)
+}
+
+// GetExisting uses stackManager to list existing nodegroup stacks and configures
+// the filter accordingly
+func (f *NodeGroupFilter) GetExisting(stackManager *manager.StackCollection) error {
```
jrryjcksn: Nit: Any time I see a function or method named "Get<xxx>" it makes me expect it to return some fetched data, rather than perform initialization

pkg/ctl/create/utils.go
```diff
@@ -0,0 +1,230 @@
...
+	for i := range cfg.NodeGroups {
+		if f.Match(cfg.NodeGroups[i]) {
+			names.Insert(cfg.NodeGroups[i].Name)
+		}
+	}
+	return names
+}
+
+// CheckEachNodeGroup iterates over each nodegroup and calls check function
+// (this is need to avoid common goroutine-for-loop pitfall)
```
jrryjcksn: Nit: need -> needed

pkg/cfn/manager/tasks.go
```diff
@@ -65,23 +67,27 @@ func (c *StackCollection) CreateClusterWithNodeGroups() []error {
...
+	return c.CreateAllNodeGroups(nil)
 }
 
 // CreateAllNodeGroups runs all tasks required to create the node groups;
 // any errors will be returned as a slice as soon as one of the tasks
 // or group of tasks is completed
-func (c *StackCollection) CreateAllNodeGroups() []error {
+func (c *StackCollection) CreateAllNodeGroups(onlySubset sets.String) []error {
 	errs := []error{}
 	appendErr := func(err error) {
```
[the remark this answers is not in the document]
errordeveloper: As far as I recall, we do it this way because we want to collect errors from multiple parallel tasks. There may be another way to do it, I've not looked into this in a long time.

pkg/ctl/create/cluster.go
```diff
@@ -103,90 +104,7 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
...
-func checkSubnetsGiven(cfg *api.ClusterConfig) bool {
-	return cfg.VPC.Subnets != nil && len(cfg.VPC.Subnets.Private)+len(cfg.VPC.Subnets.Public) != 0
-}
-
-func checkSubnetsGivenAsFlags() bool {
-	return len(*subnets[api.SubnetTopologyPrivate])+len(*subnets[api.SubnetTopologyPublic]) != 0
-}
-
-func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string, cmd *cobra.Command) error {
+func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string, cmd *cobra.Command, ngFilter *NodeGroupFilter) error {
```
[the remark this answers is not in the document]
errordeveloper: As I mentioned above, we do it here to mirror what `doCreateNodeGroup` does. We cannot add it to `api.ClusterConfig`, but we could consider moving the filter elsewhere... However, it seems to fit well here for now, as it's only applicable to cluster/nodegroup creation, perhaps we can move if/whenever we find other uses?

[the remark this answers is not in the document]
errordeveloper: > That would save us from doing the filtering again and again everywhere.

Very good point indeed! The only issue is that we cannot filter out existing NGs until later on.
We could consider creating a shadow copy of `cfg`, or something else of that sort... but at the end of the day we want a config that represents the whole cluster. Anyways, looks like it's something we can revisit later on :)

pkg/ctl/create/utils.go
errordeveloper: Yeah, I was thinking to rename this... perhaps `ApplyExistingFilter` (so it matches the `ApplyOnlyFilter`)?

pkg/ctl/create/nodegroup.go
martina-if: Yes, I think I would remove the flag as well.

pkg/ctl/create/utils.go
martina-if: Hm, why not just get it from ` range` like `for _, ng := range cfg.NodeGroups {...`

pkg/ctl/create/nodegroup.go
errordeveloper: 👍

pkg/ctl/create/utils.go
errordeveloper: That's to avoid [issue with callbacks](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables).

rade: I cannot see any goroutine in here. What am I missing?

errordeveloper: You are right, there isn't one here. However, that issue applies when you have a closure, and I'm generally very precocious of it as I was bitten by it a number of times. I'll double check the history now, I think there was a closure before this code was moved around.

errordeveloper: I've added unit test and confirm this was indeed a hangover from the time when there used to be a closure.

errordeveloper:
Thanks @jrryjcksn!

PR #657, opened 2019-03-22
Enable allow-ssh flag when a key is specified

martina-if:
### Checklist

### DIFF
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -199,6 +199,10 @@ func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
 				return fmt.Errorf("--ssh-public-key must be non-empty string")
 			}
 
+			if cmd.Flag("ssh-public-key").Changed {
+				ng.AllowSSH = true
+			}
+
 			// generate nodegroup name or use flag
 			ng.Name = utils.NodeGroupName(ng.Name, "")
 
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -59,7 +59,7 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	group := g.New(cmd)
 
 	exampleClusterName := utils.ClusterName("", "")
-	exampleNodeGroupName := utils.NodeGroupName("", "")
+	exampleNodeGroupName := NodeGroupName("", "")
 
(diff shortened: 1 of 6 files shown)


pkg/ctl/create/cluster.go
```diff
@@ -150,10 +150,12 @@ func newNodeGroupChecker(i int, ng *api.NodeGroup) error {
...
 
-	if ng.AllowSSH {
-		if ng.SSHPublicKeyPath == "" {
-			ng.SSHPublicKeyPath = defaultSSHPublicKey
-		}
+	if ng.AllowSSH && ng.SSHPublicKeyPath == "" {
+		ng.SSHPublicKeyPath = defaultSSHPublicKey
+	}
+
+	if ng.SSHPublicKeyPath != defaultSSHPublicKey {
```
errordeveloper: So I'm actually not sure what's supposed to happen here in the current state of things.
We should change the field from `string` to `*string`, so when it wasn't specified it would be `nil`.

errordeveloper: Actually, I think for now you just want to do this:
```suggestion
	if ng.SSHPublicKeyPath != defaultSSHPublicKey && ng.SSHPublicKeyPath != ""  {
```

pkg/ctl/create/nodegroup.go
```diff
@@ -153,22 +153,10 @@ func doCreateNodeGroups(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg s
...
-		err := CheckEachNodeGroup(ngFilter, cfg, func(i int, ng *api.NodeGroup) error {
-			if ng.AllowSSH && ng.SSHPublicKeyPath == "" {
-				return fmt.Errorf("--ssh-public-key must be non-empty string")
-			}
-
-			// generate nodegroup name or use either flag or argument
-			if utils.NodeGroupName(ng.Name, nameArg) == "" {
-				return cmdutils.ErrNameFlagAndArg(ng.Name, nameArg)
-			}
-			ng.Name = utils.NodeGroupName(ng.Name, nameArg)
```
errordeveloper: I missed this... the behaviour here is different from the other version, we will need to reinstate it somehow.

martina-if: Maybe I messed up in the rebase, but initially, I did the renaming of the ng.Name with the nameArg when it was provided. I can fix that

errordeveloper: No to worry, I have a fix for this in #671.

PR #660, opened 2019-03-23
Remove 'required' from the cluster name help message

### LINKED ISSUE #641: name is not required when config is used
We have this right now
```
Usage: eksctl delete cluster [flags]

General flags:
  -n, --name string          EKS cluster name (required)
  -r, --region string        AWS region
  -w, --wait                 wait for deletion of all resources before exiting
  -f, --config-file string   load configuration from a file 
```

We could say something like this
```
  -n, --name string          EKS cluster name (required, unless --config-file/-f is given)
```
But perhaps simpler is better
```
  -n, --name string          EKS cluster name
```
If someone doesn't pass `--name`, they will get an error anyway.

lucioveloso:
### Description

Closes #641.

Before the usage of config file it was required to provide the cluster
name with --name parameter. Now, with config file it can be optional.
Given that, '(required)' is no longer needed.

### Checklist

### DIFF
--- a/pkg/ctl/delete/cluster.go
+++ b/pkg/ctl/delete/cluster.go
@@ -40,7 +40,7 @@ func deleteClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 	group := g.New(cmd)
 
 	group.InFlagSet("General", func(fs *pflag.FlagSet) {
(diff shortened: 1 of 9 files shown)


errordeveloper:
Thanks a lot @lucioveloso! Join us on slack.weave.works if you are looking for things to contribute :)

PR #667, opened 2019-03-26
Use `cmdutils.ConfigFileLoader`

errordeveloper:
### Description

Start using `cmdutils.ConfigFileLoader` where possible.

### Checklist

### DIFF
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -103,69 +103,40 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
 }
 
 func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string, cmd *cobra.Command) error {
-	meta := cfg.Metadata
-
-	printer := printers.NewJSONPrinter()
-
-	if err := api.Register(); err != nil {
-		return err
-	}
 
 	ngFilter := cmdutils.NewNodeGroupFilter()
 
-	if clusterConfigFile != "" {
-		if err := eks.LoadConfigFromFile(clusterConfigFile, cfg); err != nil {
-			return err
-		}
-		meta = cfg.Metadata
+	cfgLoader := cmdutils.NewClusterConfigLoader(p, cfg, clusterConfigFile, cmd)
 
-		incompatibleFlags := []string{
-			"name",
-			"tags",
-			"zones",
-			"version",
-			"region",
-			"nodes",
-			"nodes-min",
-			"nodes-max",
-			"node-type",
-			"node-volume-size",
-			"node-volume-type",
-			"max-pods-per-node",
-			"node-ami",
-			"node-ami-family",
-			"ssh-access",
-			"ssh-public-key",
-			"node-private-networking",
-			"node-security-groups",
-			"node-labels",
-			"node-zones",
-			"asg-access",
-			"external-dns-access",
-			"full-ecr-access",
-			"storage-class",
-			"vpc-private-subnets",
-			"vpc-public-subnets",
-			"vpc-cidr",
-			"vpc-from-kops-cluster",
-		}
+	cfgLoader.FlagsIncompatibleWithConfigFile.Insert(
+		"tags",
+		"zones",
+		"nodes",
+		"nodes-min",
+		"nodes-max",
+		"node-type",
+		"node-volume-size",
+		"node-volume-type",
+		"max-pods-per-node",
+		"node-ami",
+		"node-ami-family",
+		"ssh-access",
+		"ssh-public-key",
+		"node-private-networking",
+		"node-security-groups",
+		"node-labels",
+		"node-zones",
+		"asg-access",
+		"external-dns-access",
+		"full-ecr-access",
+		"storage-class",
+		"vpc-private-subnets",
+		"vpc-public-subnets",
+		"vpc-cidr",
+		"vpc-from-kops-cluster",
+	)
 
-		for _, f := range incompatibleFlags {
-			if cmd.Flag(f).Changed {
-				return fmt.Errorf("cannot use --%s when --config-file/-f is set", f)
-			}
-		}
-
-		if meta.Name == "" {
-			return fmt.Errorf("metadata.name must be set")
-		}
-
-		if meta.Region == "" {
-			return fmt.Errorf("metadata.region must be set")
-		}
-
-		p.Region = meta.Region
-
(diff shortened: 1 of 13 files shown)


pkg/ctl/create/cluster.go
```diff
@@ -103,69 +103,40 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
...
-		if meta.Region == "" {
-			return fmt.Errorf("metadata.region must be set")
-		}
-
-		p.Region = meta.Region
-
+	cfgLoader := cmdutils.NewClusterConfigLoader(p, cfg, clusterConfigFile, cmd)
+
+	cfgLoader.FlagsIncompatibleWithConfigFile.Insert(
+		"tags",
```
errordeveloper: `name`, `version` and `region` are already included (see #665).

pkg/ctl/create/cluster.go
```diff
@@ -176,11 +147,11 @@ func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
 
 		skipNodeGroupsIfRequested(cfg)
 
-		if err := CheckEachNodeGroup(ngFilter, cfg, NewNodeGroupChecker); err != nil {
-			return err
-		}
-	} else {
-		// validation and defaulting specific to when --config-file is unused
+		return CheckEachNodeGroup(ngFilter, cfg, NewNodeGroupChecker)
```
errordeveloper: It's a callback, so we can return this way

pkg/ctl/create/nodegroup.go
```diff
@@ -49,7 +49,7 @@ func createNodeGroupCmd(g *cmdutils.Grouping) *cobra.Command {
 		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "name of the EKS cluster to add the nodegroup to")
 		cmdutils.AddRegionFlag(fs, p)
 		cmdutils.AddVersionFlag(fs, cfg.Metadata, `for nodegroups "auto" and "latest" can be used to automatically inherit version from the control plane or force latest`)
-		fs.StringVarP(&clusterConfigFile, "config-file", "f", "", "load configuration from a file")
+		cmdutils.AddConfigFileFlag(&clusterConfigFile, fs)
```
errordeveloper: A thing we missed earlier...

pkg/ctl/create/cluster.go
```diff
@@ -204,11 +175,10 @@ func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
 
 			return nil
 		})
-		if err != nil {
-			return err
-		}
 	}
 
+	meta := cfg.Metadata
```
errordeveloper: TODO: actually call `cfgLoader.Load()`

pkg/ctl/delete/nodegroup.go
```diff
@@ -56,31 +56,36 @@ func deleteNodeGroupCmd(g *cmdutils.Grouping) *cobra.Command {
...
+		}
 
-	if ng.Name != "" && nameArg != "" {
-		return cmdutils.ErrNameFlagAndArg(ng.Name, nameArg)
+		return nil
 	}
 
-	if nameArg != "" {
-		ng.Name = nameArg
+	if err := cfgLoader.Load(); err != nil {
```
errordeveloper: TODO: I think this one actually needs much more work, will leave it out for another PR

pkg/ctl/create/cluster.go
```diff
@@ -103,69 +103,40 @@ func createClusterCmd(g *cmdutils.Grouping) *cobra.Command {
...
+		"external-dns-access",
+		"full-ecr-access",
+		"storage-class",
+		"vpc-private-subnets",
+		"vpc-public-subnets",
+		"vpc-cidr",
+		"vpc-from-kops-cluster",
+	)
+
+	cfgLoader.ValidateWithConfigFile = func() error {
```
martina-if: I think I would move this validation functionality into the ClusterConfigLoader. It should be the config loader the one who knows how to validate stuff, IMO, not the caller. I should be able to delegate everything related to configuration loader to an "entity" that does it for me. Otherwise it looks like a leaky abstraction to me. What do you think?

errordeveloper: I agree, however currently the validation logic differers slightly in each of the commands, hence it was does as a custom callback. I agree that it's leaky at present.
I've tried to do something like `cfgLoader.ValidateWithConfigFile = cmdutils.ValidateCreateClusterWithConfigFile`, but wasn't sure what exactly was the benefit of that.
I suppose mapping commands to validator implementations inside of `cmdutils.ClusterConfigLoader` maybe an okay thing to do and we could move the lists of flags there also, as the bulk of the actually flags is already defined in `cmdutils` package.

errordeveloper: We could do something like
```
var validatorsWithConfigFile = map[string]func() error{
	"eksctl create cluster": func() error {
		// do stuff
		return nil
	},
}

// NewClusterConfigLoader constructs a standard loader
func NewClusterConfigLoader(provider *api.ProviderConfig, spec *api.ClusterConfig, clusterConfigFile string, cmd *cobra.Command) *ClusterConfigLoader {
	nilValidatorFunc := func() error { return nil }

	logger.Debug("cmd.CommandPath = %s", cmd.CommandPath())
	l := &ClusterConfigLoader{
		provider: provider,
		path:     clusterConfigFile,
		cmd:      cmd,

		Spec: spec,

		ValidateWithConfigFile:             nilValidatorFunc,
		FlagsIncompatibleWithConfigFile:    sets.NewString("name", "region", "version"),
		ValidateWithoutConfigFile:          nilValidatorFunc,
		FlagsIncompatibleWithoutConfigFile: sets.NewString(),
	}

	cmdPath := cmd.CommandPath()

	if validatorFunc, ok := validatorsWithConfigFile[cmdPath]; ok {
		l.ValidateWithConfigFile = validatorFunc
	}

	return l
}
```

martina-if: I think this would be the classic example of one interface with multiple implementations. So we could have one interface ConfigLoader, and then CreateClusterConfigLoader whose validation methods are specific for that. The rest of the common code, `Load()` and `LoadMetadata` would be delegated to the common config loader.

errordeveloper: Yeah, definitely a good idea! Will do that, thanks :)

errordeveloper:
Since we are back to improving the loaded, I'm gonna open another PR

PR #670, opened 2019-03-27
Move nodegroup filters

errordeveloper:
### Description

Move `NodeGroupFilter` so it can be used inside of `cmdutils` package.

### Checklist


pkg/ctl/cmdutils/nodegroup_filter_test.go
```diff
@@ -0,0 +1,126 @@
+package cmdutils_test
```
errordeveloper: As discussed offline, we might want to do some mixed public/private unit tests, but for now public-only tests remain conventional...

pkg/ctl/cmdutils/nodegroup_filter_test.go
```diff
@@ -0,0 +1,126 @@
...
+	. "github.com/onsi/gomega"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
+
+	. "github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+var _ = Describe("nodegroup filter", func() {
+
+	newClusterConfig := func() *api.ClusterConfig {
```
errordeveloper: These lambdas are duplicated, but I don't mind some amount duplication in tests, it's convenient to be able to change these in the same place.

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		if existingCount := f.existing.Len(); existingCount > 0 {
+			reasons = append(reasons, fmt.Sprintf("%d nodegroup(s) (%s) already exist", existingCount, strings.Join(f.existing.List(), ", ")))
+		}
+		logger.Info("%d nodegroup(s) were filtered out: %s", filteredOutCount, strings.Join(reasons, ", "))
+	}
+}
+
+// CheckEachNodeGroup iterates over each nodegroup and calls check function if it matches the filter
+func (f *NodeGroupFilter) CheckEachNodeGroup(nodeGroups []*api.NodeGroup, check func(i int, ng *api.NodeGroup) error) error {
```
errordeveloper: @martina-if I though it's a bit better to make it a method on `NodeGroupFilter` for now, so we don't have to pass `ngFilter` explicitly each time...

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -0,0 +1,134 @@
...
+	f.onlySpec = strings.Join(globExprs, ",")
+	return f.onlyFilterMatchesAnything(cfg)
+}
+
+func (f *NodeGroupFilter) onlyFilterMatchesAnything(cfg *api.ClusterConfig) error {
+	if len(f.only) == 0 {
+		return nil
+	}
+	for i := range cfg.NodeGroups {
+		for _, g := range f.only {
```
martina-if: Can we rename `g` to `nodeGroup` or `ng`?

martina-if: wait, what is `g` here?

martina-if: same for line 85

pkg/ctl/cmdutils/nodegroup_filter_test.go
martina-if: I'm ok with some duplication in tests, it can be better for readability. Do they need to be lambdas though? Can we move them to actual functions to make the tests a bit cleaner?

pkg/ctl/cmdutils/nodegroup_filter_test.go
```diff
@@ -0,0 +1,126 @@
...
+		ng.SecurityGroups.WithLocal = api.NewBoolFalse()
+		ng.Labels = map[string]string{"group": "b", "seq": "1"}
+	}
+
+	Context("CheckEachNodeGroup", func() {
+
+		It("should iterate over unique nodegroups", func() {
+			cfg := newClusterConfig()
+			addGroupA(cfg)
+			f := NewNodeGroupFilter()
```
martina-if: rename `f` to `filter`?

pkg/ctl/cmdutils/nodegroup_filter.go
errordeveloper: `g` for "glob", will rename to something more readable - thanks!

pkg/ctl/cmdutils/nodegroup_filter_test.go
errordeveloper: yeah, happy to move these

pkg/ctl/cmdutils/nodegroup_filter_test.go
errordeveloper: 👍

PR #671, opened 2019-03-27
Improve cluster config loader

errordeveloper:
### Checklist


pkg/ami/api.go
```diff
@@ -4,27 +4,27 @@ import (
...
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
 )
 
 const (
-	// DefaultImageFamily defines the default image family for the worker nodes
-	DefaultImageFamily = ImageFamilyAmazonLinux2
-
 	// ImageFamilyAmazonLinux2 represents Amazon Linux 2 family
-	ImageFamilyAmazonLinux2 = "AmazonLinux2" // Owner 602401143452
+	ImageFamilyAmazonLinux2 = api.NodeImageFamilyAmazonLinux2 // Owner 602401143452
```
errordeveloper: The constants are being relocated to `api` package, so that they can be used there without dragging AWS dependencies from here, as the `api` package is meant to be free of dependencies.

pkg/apis/eksctl.io/v1alpha4/defaults.go
```diff
@@ -0,0 +1,56 @@
+package v1alpha4
+
+// SetNodeGroupDefaults will set defaults for a given nodegroup
```
errordeveloper: This wasn't moved here before due to the concerns regarding `ami` package and its dependencies.

martina-if: Why not keep this in the nodegroup_filter package as private and only expose `ValidateNodeGroupsAndSetDefaults()`? Since this is less abstracted and not really useful by itself, or is it? :/

errordeveloper: So `NodeGroupFilter` is currently designed for CLI use-cases, hence it lives in `cmduitls`.
This is here so that one can construct `ClusterConfig`, and apply defaults and validation in a deamon/operator context or whatever other context. I'm just trying to draw a clear line between CLI and API.

pkg/apis/eksctl.io/v1alpha4/defaults.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if ng.SecurityGroups.WithLocal == nil {
+		ng.SecurityGroups.WithLocal = NewBoolTrue()
+	}
+	if ng.SecurityGroups.WithShared == nil {
+		ng.SecurityGroups.WithShared = NewBoolTrue()
+	}
+
+	// Enable SSH when a key is provided
+	if ng.SSHPublicKeyPath != "" && ng.SSHPublicKeyPath != DefaultNodeSSHPublicKeyPath {
```
errordeveloper: @martina-if this is a temporary fix, I think we should be able to get rid of `!= DefaultNodeSSHPublicKeyPath` when we switch to references.

Issue #386, opened 2019-01-03
refactor ng.AllowSSH and related fields into a separate struct


errordeveloper:
At the moment config file filed for SSH are very confusing.

We have `allowSSH`, `sshPublicKeyPath`, `sshPublicKeyName`, and `SSHPublicKey`.

https://github.com/weaveworks/eksctl/blob/ca30d425158b2106a00892cd084f489e09770158/pkg/apis/eksctl.io/v1alpha4/types.go#L361-L369

However, `sshPublicKeyName` and `SSHPublicKey` are effectively ignored, and the only two way you can make it work are these:
```YAML
nodeGroups:
- name: ng-1
  allowSSH: true # will use '~/.ssh/id_rsa'
- name: ng-2
  allowSSH: true
  sshPublicKeyPath: <localPathOrNameOfExistingKeyInEC2>
```

We should turn this to something like:
```YAML
- name: ng-1
  ssh:
    allow: true  # will use '~/.ssh/id_rsa'
- name: ng-2
  ssh:
    sshPublicKeyPath: <localPath>
- name: ng-3
  ssh:
    sshPublicKeyName: <NameOfExistingKeyInEC2>
- name: ng-4
  ssh:
    sshPublicKey: <PublicKeyBody>
```

PR #673, opened 2019-03-28
Move SSH configuration into its own structure

martina-if:
Move SSH configuration into it's own struct in preparation for  https://github.com/weaveworks/eksctl/issues/386

- [x] Manually tested both the config file and the cli parameters


pkg/cfn/builder/api_test.go
```diff
@@ -608,10 +608,10 @@ var _ = Describe("CloudFormation template builder API", func() {
 		})
 	})
 
-	Context("NodeGroup{PrivateNetworking=true AllowSSH=true}", func() {
+	Context("NodeGroup{PrivateNetworking=true Allow=true}", func() {
```
errordeveloper: nit: use `SSH.Allow`

errordeveloper:
> closes #386

There quite a few things in that issue, do you want to make it all happen in one PR? Up to you.

pkg/eks/auth.go
```diff
@@ -77,29 +78,30 @@ func (c *ClusterProvider) importSSHPublicKeyIfNeeded(clusterName string, ng *api
...
-	logger.Debug("SSH public key %s already exists", ng.SSHPublicKeyName)
+	logger.Debug("SSH public key %s already exists", *ng.SSH.PublicKeyName)
 	return nil
 }
 
 // LoadSSHPublicKey loads the given SSH public key
 func (c *ClusterProvider) LoadSSHPublicKey(clusterName string, ng *api.NodeGroup) error {
-	if !ng.AllowSSH {
+	if !ng.SSH.Allow {
 		// TODO: https://github.com/weaveworks/eksctl/issues/144
```
errordeveloper: Please remove the todo, the issue has been closed for a while now :)

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -445,4 +437,16 @@ type (
 		// +optional
 		ALBIngress *bool `json:"albIngress"`
 	}
+
+	// SSHConfig holds all the ssh access configuration to a NodeGroup
+	SSHConfig struct {
+		// +optional
+		Allow bool `json:"allow"`
```
errordeveloper: I think we should make this one `*bool` also, so that we can do meaningful defaulting for it :)

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -445,4 +437,16 @@ type (
 		// +optional
 		ALBIngress *bool `json:"albIngress"`
 	}
+
+	// SSHConfig holds all the ssh access configuration to a NodeGroup
+	SSHConfig struct {
```
errordeveloper: Perhaps we could call it `NodeGroupSSH`, so it is consistent with `NodeGroupIAM` and `NodeGroupSGs`...?

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -378,15 +378,7 @@ type NodeGroup struct {
...
-	// TODO move to separate struct
-	// +optional
-	AllowSSH bool `json:"allowSSH"`
-	// +optional
-	SSHPublicKeyPath string `json:"sshPublicKeyPath,omitempty"`
-	// +optional
-	SSHPublicKey []byte `json:"SSHPublicKey,omitempty"` // TODO: right now it's kind of read-only, but one may wish to use key body in a config file so we will need recognise that
-	// +optional
-	SSHPublicKeyName string `json:"sshPublicKeyName,omitempty"`
+	SSH SSHConfig `json:"ssh"`
```
errordeveloper: Seems plausible to make this one a reference also, at least for consistency. Some of the fields above aren't references, but arguably that's just because we didn't get to refactor those yet.

pkg/ctl/cmdutils/configfile.go
```diff
@@ -180,12 +180,12 @@ func NewCreateClusterLoader(provider *api.ProviderConfig, spec *api.ClusterConfi
 		}
 
 		return ngFilter.CheckEachNodeGroup(l.spec.NodeGroups, func(i int, ng *api.NodeGroup) error {
-			if ng.AllowSSH && ng.SSHPublicKeyPath == "" {
+			if ng.SSH.Allow && ng.SSH.PublicKeyPath == nil {
```
errordeveloper: In the case of flags it this should remain as an actual empty string check, actually we might move it down below...

pkg/ctl/cmdutils/configfile.go
```diff
@@ -180,12 +180,12 @@ func NewCreateClusterLoader(provider *api.ProviderConfig, spec *api.ClusterConfi
...
 				return fmt.Errorf("--ssh-public-key must be non-empty string")
 			}
 
 			if cmd.Flag("ssh-public-key").Changed {
-				ng.AllowSSH = true
+				ng.SSH.Allow = true
```
errordeveloper: This should be sufficient:
```Go
				if rng.SSH.PublicKeyPath == "" {
								return fmt.Errorf("--ssh-public-key must be non-empty string")
				}
```

pkg/ctl/cmdutils/configfile.go
errordeveloper: What I think we should do is this:
- make `cfg.NewNodeGroup()` should construct `&NodeGroupSSH{Allow: NewBoolFlase(), PublicKeyPath: DefaultSSHPublicKeyPath}`
- `ng.SSH.Allow` gets passed as default value of `--ssh-access`
- `ng.SSH.PublicKeyPath` gets passed as default value of `--ssh-public-key`

The idea is that all of defaulting logic is defined in the API package, albeit it there are `NewClusterConfig` and `NewNodeGroup` for flags and `SetNodeGroupDefaults` for config files, those are still in the same package.

Right now we have this:
https://github.com/weaveworks/eksctl/blob/653cd2a174120ddb38f9025f928543fbe04b3f75/pkg/ctl/cmdutils/nodegroup_flags.go#L39-L40

And based on the above it would become:
```Go
fs.BoolVar(&ng.SSH.Allow, "ssh-access", ng.SSH.Allow, "control SSH access for nodes")
fs.StringVar(&ng.SSH.PublicKeyPath, "ssh-public-key", ng.SSH.PublicKeyPath, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
```

The `nodegroup_flags.go` doesn't implement this pattern everywhere yet, but that's what I've been trying to do for every flag that pairs with a config file filed.

pkg/ctl/cmdutils/nodegroup_flags.go
```diff
@@ -36,8 +36,8 @@ func AddCommonCreateNodeGroupFlags(cmd *cobra.Command, fs *pflag.FlagSet, p *api
 
 	fs.IntVar(&ng.MaxPodsPerNode, "max-pods-per-node", 0, "maximum number of pods per node (set automatically if unspecified)")
 
-	fs.BoolVar(&ng.AllowSSH, "ssh-access", false, "control SSH access for nodes")
-	fs.StringVar(&ng.SSHPublicKeyPath, "ssh-public-key", api.DefaultNodeSSHPublicKeyPath, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
+	fs.BoolVar(&ng.SSH.Allow, "ssh-access", false, "control SSH access for nodes")
```
errordeveloper: See my comment above...

pkg/ctl/cmdutils/configfile.go
martina-if: Ah, I thought that if one enables allow-ssh then the keypath would be ~/.ssh/id_rsa.pub by default. Should we then always force the user to define which key?

errordeveloper: > I thought that if one enables allow-ssh then the keypath would be ~/.ssh/id_rsa.pub by default.

Yes, indeed. What I'm saying is that we should move that empty string check here.

pkg/apis/eksctl.io/v1alpha4/types.go
martina-if: I thought about it, then I thought It could represent any SSH key as well. What do you think? I don't have a strong opinion here.

errordeveloper: I think whatever SSH key it is, it still belongs to a nodegroup :)

phlukman:
hi @martina-if 

I'm running version 0.1.29 and I'm configuring the yaml file as follows, according to the examples:

 nodeGroups:
  - name: security-cluster-ng1-public
    instanceType: t2.2xlarge
    desiredCapacity: 1
    ssh:
      publicKeyName: security-es-cluster-key
  - name: security-cluster-ng2-private
    instanceType: t2.2xlarge
    desiredCapacity: 2
    privateNetworking: true
    ssh:
      publicKeyName: security-es-cluster-key


and I'm getting error:
[✖]  loading config file "secure-cluster-config.yaml": error unmarshaling JSON: while decoding JSON: json: unknown field "ssh"

What am I missing here?

martina-if:
Hi @phlukman,

The new ssh schema hasn't been released yet. We will make a release with those changes tomorrow at the latest. In the meantime try using it with the old schema, like this:

```
nodeGroups:
  - name: security-cluster-ng1-public
    instanceType: t2.2xlarge
    desiredCapacity: 1
    sshPublicKeyPath: security-es-cluster-key

  - name: security-cluster-ng2-private
    instanceType: t2.2xlarge
    desiredCapacity: 2
    privateNetworking: true
    sshPublicKeyPath: security-es-cluster-key
```

phlukman:
thanks for your quick reply @martina-if !

PR #680, opened 2019-03-28
Adding Support for 1.12

### LINKED ISSUE #679: Support for Amazon EKS 1.12 Clusters
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

Everyone should be able to use the latest EKS version. :)

https://github.com/aws/containers-roadmap/issues/24

Support for `create`'ing and `update`'ing existing clusters with Kubernetes 1.12.

Do not hesitate, when appropriate, to share the exact commands or API you would like, and/or to share a diagram (e.g.: asciiflow.com): "a picture is worth a thousand words".

christopherhein:
### Description

closes #679

This adds support for deploying Amazon EKS with Kubernetes 1.12. Default is still set to `1.11` if you'd like to use this use:

```yaml
apiVersion: eksctl.io/v1alpha4
kind: ClusterConfig
metadata:
  name: <name>
  region: us-west-2
  version: "1.12"
```

or:

```bash
eksctl create cluster --name <name> --version 1.12
```

### Checklist


whereisaaron:
It looks like it is always the case we need to update `eksctl` before we can upgrade to a new version. Is there sufficient AW API to enable discovering the available control plane version and matching AMI versions? Or it inevitably a manual task?

errordeveloper:
> discovering the available control plane version

I'm not aware of one, I'd be very keen to see it!

> matching AMI versions?

We have `make generate-ami`, so that part is automated and we can improve that a bit more also.

whereisaaron:
Thanks @errordeveloper. Sounds like that is a 'no'! There is an 'auto' option for AMIs maybe we could speculatively at '1.13' and '1.14 'to `eksctl` right now, and just report the error until it starts working 😄

Issue #682, opened 2019-03-29
improve handling of deletions, add dry-run


errordeveloper:
We currently have it done in a rather complicated way, we tend to try deleting things and then retry. We should instead collect a list of things we want to delete first and then try deleting those things. This will make the logic cleaner, but also help us to implement dry-run capability.

errordeveloper:
It's important to add dry-run as we are going to allow deleting multiple NGs with a config file. Currently CLI works okay for single NG, as user types it all explicitly.

mreferre:
I am not sure if dry-run will cut it but I believe it would be useful to have some sort of delete/termination protection mechanism similar to what we have in Cloudformation (or actually leverage the CF mechanism?). Deleting a cluster today is way to easy and we should consider make it less error prone. Dry run may help but I am not sure it will be enough.

errordeveloper:
Yes, I think we should have deleting protection mode and have it enabled by
default. We will ultimately need to use CloudFormation for that.

errordeveloper:
Closed via #695.

Issue #646, opened 2019-03-19
improve output messages in 'eksctl update cluster'


errordeveloper:
Right now we print this
```
[ℹ]  using region us-east-2
[ℹ]  re-building cluster stack "eksctl-sample-cluster"
[ℹ]  (dry-run) updating stack to add new resources [ClusterSharedNodeSecurityGroup IngressInterNodeGroupSG] and ouputs [ServiceRoleARN SharedNodeSecurityGroup]
[ℹ]  checking security group configuration for all nodegroups
[✖]  found 1 nodegroup(s) (0) without shared security group, cluster networking maybe be broken
[✖]  it's recommended to create new nodegroups, then delete old ones
```
1. there is typo - `ouputs`
2. it's not clear that `--dry-run=false` has to be used

And with `--dry-run=false` we print this
```
[ℹ]  using region us-east-2
[ℹ]  re-building cluster stack "eksctl-sample-cluster"
[ℹ]  updating stack to add new resources [ClusterSharedNodeSecurityGroup IngressInterNodeGroupSG] and ouputs [ServiceRoleARN SharedNodeSecurityGroup]
[ℹ]  checking security group configuration for all nodegroups
[✖]  found 1 nodegroup(s) (0) without shared security group, cluster networking maybe be broken
[✖]  it's recommended to create new nodegroups, then delete old ones
```
Which also doesn't make it clear if all is well at the end and if shared SG gets added or what happens really.

venkyvb:
@errordeveloper Is this issue still open? If yes, I would like to work on this.

errordeveloper:
Hi! Yes, please do feel free to start on this, and open a WIP PR, so you
can get feedback early enough :)

venkyvb:
Re 1) This can easily be corrected
Re 2) From the code it seems like --dry-run is now deprecated and --approve is to be used. 

Also when I analyze the usage of the --dry-run seems like this is just used in the output and it doesn't seem to control the StackResource creation itself. Proposal would be to remove this flag in the output itself.

michaelbeaumont:
Closing, quite old, see also #476

Issue #476, opened 2019-01-28
Add dry-run to all commands


errordeveloper:
We should add `--dry-run`, which would be especially helpful to config file users, but it'd be good to make it generally useful.

For me personally, it seems that showing defaulted config would be quite useful, but I am pretty certain it won't be very friendly to most other users.

Perhaps we could do something where we show a summary of stack features, it'd be good to make it general enough, so that we could use it for `eksctl get cluster|nodegroup`, or `eksctl describe cluster|nodegroup`.

Key items to show will be:

- use of dedicated VPC vs existing one
- use of new security groups vs existing ones
- use of dedicated IAM roles vs existing ones

We should still show defaulted config, perhaps at the top of the output. It would be good to respect the format that user has used. Current debug logs use JSON as it's easy, but if user doesn't know the internals, they may get confused by seeing JSON instead of YAML.

pawelprazak:
another interesting use case is an export functionality, where user can get all templates/manifests and version them in git and apply them manually if they want

sbernheim:
@gemagomez - Shutterstock asked for a dryrun feature as well, and I was surprised to hear that it's not already implemented.  @errordeveloper confirmed that it is not implemented for createCluster and most other actions.

This feature is generally useful for automated dev/unit/functional testing.  It's a fairly standard feature for this sort of tool.

mrPsycho:
it will help to configure policies like:
* PolicyNLB
* PolicyCloudWatchMetrics 
and some additional, which sometimes are needed to be more strict, than allowing to all resources.

alimeerutech:
+1

mattvonrocketstein:
> another interesting use case is an export functionality, where user can get all templates/manifests and version them in git and apply them manually if they want

I'm also interested in this functionality, and the related issue https://github.com/weaveworks/eksctl/issues/1888 was closed so I guess this is the best place to discuss.

It's possible to just grab these templates back from cloudformation after eksctl creates them, but for node groups the template has a huge block of b64 embedded.  So I checked in the code base for a obvious templates/user-data.sh file, but only found a collection of related code in pkg/nodebootstrap/userdata_al2.go  (no template).

Thus it seems without a dry-run, even if we accept the "pull cloudformation templates from AWS" solution, the only simple option for showing the user-data template is to actually apply changes and pull the template from the debugger.

Callisto13:
Dry run now available, see docs: https://eksctl.io/usage/dry-run/

[...]

PR #688, opened 2019-04-01
[WIP] More outputs support, a small step towards real dry-run/export feature

pawelprazak:
### Description

Added `--output` support to `create`, `update`, `delete`, a small step towards real dry-run/export feature.

As mentioned in #476:
> We should still show defaulted config, perhaps at the top of the output. It would be good to respect the format that user has used. Current debug logs use JSON as it's easy, but if user doesn't know the internals, they may get confused by seeing JSON instead of YAML.

The end goal would be to allow for versioning of all templates/manifests to examine all changes made by `eksctl` and version a cluster state in git, not just the config file.

Might be also useful for #646, #682

### Checklist


pawelprazak:
I wanted to make minimal amount of changes to achieve the goal of respecting `--output`.

This can lead to some ugly/weird/redundant code in places where the existing code does not accept changes very well.

Plead advise on the amount of refactoring expected in this PR and the overall direction.

pkg/cfn/builder/api.go
```diff
@@ -30,6 +30,7 @@ type ResourceSet interface {
 	WithNamedIAM() bool
 	RenderJSON() ([]byte, error)
 	GetAllOutputs(cfn.Stack) error
+	Template() gfn.Template
```
errordeveloper: Why not use `RenderJSON()`? I suppose the format is not too pretty, but we could add a parameter that make it prettier, like `RenderJSON(pretty bool)`...

pkg/cfn/manager/api.go
```diff
@@ -110,13 +111,17 @@ func (c *StackCollection) DoCreateStackRequest(i *Stack, templateBody []byte, ta
...
 // channel, it's closed immediately after it is written to
-func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, tags, parameters map[string]string, errs chan error) error {
+func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, tags, parameters map[string]string, printer printers.OutputPrinter, errs chan error) error {
 	i := &Stack{StackName: &name}
 	templateBody, err := stack.RenderJSON()
 	if err != nil {
 		return errors.Wrapf(err, "rendering template for %q stack", *i.StackName)
 	}
 
+	if err := printer.LogObj(logger.Debug, "stack = \\\n%s\n", stack.Template()); err != nil {
```
errordeveloper: This logger is actually for Kubernetes-style objects, albeit it might work for other things too.

pkg/cfn/manager/cluster.go
```diff
@@ -98,6 +102,10 @@ func (c *StackCollection) AppendNewClusterStackResource(dryRun bool) (bool, erro
 		return false, err
 	}
 
+	if err := printer.LogObj(logger.Debug, "newStack = \\\n%s\n", newStack.Template()); err != nil {
```
errordeveloper: There is `logger.Debug("newTemplate = %s", newTemplate)` down below, it does the same thing, doesn't it?

pkg/cfn/manager/tasks.go
```diff
@@ -87,7 +100,10 @@ func (c *StackCollection) CreateAllNodeGroups(onlySubset sets.String) []error {
 		}
 		t := Task{
 			Call: c.CreateNodeGroup,
-			Data: ng,
+			Data: DataWithPrinter{
```
errordeveloper: I'm actually reworking how this is done, we are moving away from `interface{}` and adding a few other things, I'm due about to open a PR sometime today :)

errordeveloper:
hey @pawelprazak! thanks for opening a PR and showing the direction you are thinking of going, there are a few things I'll need to explain about this, and as I mention, I'm making `Task` abstraction more useful for dry-run, can we connect on Slack?

pkg/cfn/manager/cluster.go
pawelprazak: yes, I was not sure what to do with it, the debug log shows JSON only, I can remove it, no problem.

pkg/cfn/manager/tasks.go
pawelprazak: nice!

pkg/cfn/builder/api.go
pawelprazak: because gfn.Template implements JSON and YAML marshaling

pkg/cfn/manager/api.go
pawelprazak: thank you, didn't know, so what function should I use?

pawelprazak:
so as we've discussed, I'll start working on a separate `export` PR

do you want me to finish this output changes?

pkg/cfn/manager/cluster.go
pawelprazak: if I understand correctly the code, `PrintObj` will just marshall and indent any object

pkg/cfn/manager/api.go
pawelprazak: if I understand correctly the code, `PrintObj` will just marshall and indent any object

errordeveloper:
@pawelprazak it's probably best to focus on #718. I'll close this.

Issue #536, opened 2019-02-12
clean nodegroup deletion


errordeveloper:
As mentioned earlier in  https://github.com/weaveworks/eksctl/pull/523#issuecomment-461623452, we need to devise a better strategy for nodegroup deletion, which should also help us with #103 and #274 (as well as other related issues).

So we should try implementing the following strategy:

* for all nodegroups
  * taint the node and drain pods
* delete all daemonsets
* for all nodegroups
  * scale nodes to zero
  * delete dangling ENIs
  * look for ELBs, ALBs and EBS volumes and delete those
    * also ALBs created through `Ingress`
* delete each nodegroup stack
* delete cluster stack

We may do something different for deleting individual nodegroups, e.g. only scale drain, taint and down, but we do need to implement this first, as deletion is an important aspect of user experience.

related: #2278, #2172

errordeveloper:
Maybe related to https://github.com/aws/amazon-vpc-cni-k8s/issues/69.

errordeveloper:
Fixing this would pave the road towards fixing other issue with deletions.

gemagomez:
@2opremio is this finished?

2opremio:
Nope, I fixed #103 in a different way and haven't tackled #274 yet for which we may follow a different approach.

programmer04:
The deletion of entries in DNS (automatically with cluster) created by [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) (deployed in a specific cluster) would be useful too.

Probably `eksctl` would be able to clean only entries created in [Route 53](https://aws.amazon.com/route53/). Every entry in [Route 53](https://aws.amazon.com/route53/) created automatically by Kubernetes with [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) has some metadata in `TXT` record which gives the possibility of tracking e.g.
```
"heritage=external-dns,external-dns/owner=test-cluster,external-dns/resource=ingress/default/app-ingress"
```

Probably the simplest solution is to remove with the cluster all entries from [Route 53](https://aws.amazon.com/route53/) which match below condition
```
heritage == external-dns && external-dns/owner == NAME_OF_DELETED_CLUSTER
```


Behaviour for `eksctl 0.23.0` with [ExternalDNS](https://github.com/kubernetes-sigs/external-dns):
[...]

PR #695, opened 2019-04-02
rework stack management

errordeveloper:
### Description

- cater for dry-run
- cleaner main functions for create and delete commands
- more general abstraction for tasks
- expose all common operations through functions
- remove special purpose functions

### Checklist

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -33,7 +33,7 @@ type StackInfo struct {
 	Template  *string
 }
 
-// ChangeSet represents a Cloudformation changeSet
+// ChangeSet represents a CloudFormation ChangeSet
 type ChangeSet = cloudformation.DescribeChangeSetOutput
 
 // StackCollection stores the CloudFormation stack information
@@ -62,7 +62,7 @@ func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) *
 	}
 }
 
-// DoCreateStackRequest requests the creation of a cloudformation stack.
+// DoCreateStackRequest requests the creation of a CloudFormation stack
 func (c *StackCollection) DoCreateStackRequest(i *Stack, templateBody []byte, tags, parameters map[string]string, withIAM bool, withNamedIAM bool) error {
 	input := &cloudformation.CreateStackInput{
 		StackName: i.StackName,
@@ -121,12 +121,14 @@ func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, ta
 		return err
 	}
 
+	logger.Info("deploying stack %q", name)
+
 	go c.waitUntilStackIsCreated(i, stack, errs)
 
 	return nil
 }
 
-// UpdateStack will update a cloudformation stack by creating and executing a ChangeSet.
+// UpdateStack will update a CloudFormation stack by creating and executing a ChangeSet
 func (c *StackCollection) UpdateStack(stackName string, changeSetName string, description string, template []byte, parameters map[string]string) error {
 	logger.Info(description)
 	i := &Stack{StackName: &stackName}
@@ -226,8 +228,8 @@ func defaultStackStatusFilter() []*string {
 	)
 }
 
-// DeleteStack kills a stack by name without waiting for DELETED status
-func (c *StackCollection) DeleteStack(name string, force bool) (*Stack, error) {
+// DeleteStackByName sends a request to delete the stack
+func (c *StackCollection) DeleteStackByName(name string) (*Stack, error) {
 	i := &Stack{StackName: &name}
 	s, err := c.DescribeStack(i)
 	if err != nil {
@@ -243,38 +245,15 @@ func (c *StackCollection) DeleteStack(name string, force bool) (*Stack, error) {
 		}
 		return nil, err
 	}
-	if *s.StackStatus == cloudformation.StackStatusDeleteFailed && !force {
-		return nil, fmt.Errorf("stack %q previously couldn't be deleted", name)
-	}
 	i.StackId = s.StackId
-	for _, tag := range s.Tags {
-		if *tag.Key == api.ClusterNameTag && *tag.Value == c.spec.Metadata.Name {
-			input := &cloudformation.DeleteStackInput{
-				StackName: i.StackId,
-			}
-
-			if cfnRole := c.provider.CloudFormationRoleARN(); cfnRole != "" {
-				input = input.SetRoleARN(cfnRole)
-			}
-
-			if _, err := c.provider.CloudFormation().DeleteStack(input); err != nil {
-				return nil, errors.Wrapf(err, "not able to delete stack %q", name)
-			}
-			logger.Info("will delete stack %q", name)
-			return i, nil
-		}
-	}
-
-	return nil, fmt.Errorf("cannot delete stack %q as it doesn't bare our %q tag", *s.StackName,
-		fmt.Sprintf("%s:%s", api.ClusterNameTag, c.spec.Metadata.Name))
+	return c.DeleteStackBySpec(i)
 }
 
-// WaitDeleteStack kills a stack by name and waits for DELETED status;
-// any errors will be written to errs channel, when nil is written,
-// assume completion, do not expect more then one error value on the
-// channel, it's closed immediately after it is written to
-func (c *StackCollection) WaitDeleteStack(name string, force bool, errs chan error) error {
-	i, err := c.DeleteStack(name, force)
+// WaitDeleteStackByName sends a request to delete the stack, and waits until status is DELETE_COMPLETE;
+// any errors will be written to errs channel, assume completion when nil is written, do not expect
+// more then one error value on the channel, it's closed immediately after it is written to
+func (c *StackCollection) WaitDeleteStackByName(name string, errs chan error) error {
+	i, err := c.DeleteStackByName(name)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


errordeveloper:
@rndstr would you mind to take a look at this please?

errordeveloper:
Still need to rewrite the unit tests.

errordeveloper:
I think there is a bug, like a possible race condition... I'll need to investigate, perhaps unit tests will show.

pkg/ctl/delete/cluster.go
```diff
@@ -76,75 +100,46 @@ func doDeleteCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
...
-					logger.Critical("%s\n", err.Error())
-				}
-				return fmt.Errorf("failed to delete nodegroup(s)")
-			}
-			return nil
-		}
-		if err := tryDeleteAllNodeGroups(false); err != nil {
-			logger.Info("will retry deleting nodegroup")
+		tasks, err := stackManager.DeleteTasksForClusterWithNodeGroups(wait, func(_ chan error, _ string) error {
 			logger.Info("trying to cleanup dangling network interfaces")
```
rndstr: this cleanup function is being called once for each nodegroup now. it looks like it was intended to be just run once previously?

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,208 @@
...
+	stack *Stack
+	call  func(*Stack, chan error) error
+}
+
+func (t *taskWithStackSpec) Describe() string { return t.info }
+func (t *taskWithStackSpec) Do(errs chan error) error {
+	return t.call(t.stack, errs)
+}
+
+type asyncTaskWithStackSpec struct {
```
rndstr: can you elaborate what the purpose for this struct is and what is intended to happen when the error channel is closed in `Do()`?

pkg/cfn/manager/deprecated.go
```diff
@@ -1,57 +1,98 @@
...
-	return err
+	cpStackFound := false
+	for _, s := range stacks {
+		if strings.HasSuffix(*s.StackName, "-ControlPlane") {
+			cpStackFound = true
+		}
+	}
+	tasks := &TaskSet{}
+
+	for _, suffix := range deprecastedStackSuffices() {
```
martina-if: Typo: ~deprecastedStackSuffices~ -> deprecatedStackSuffices

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,208 @@
...
-		if onlySubset != nil && !onlySubset.Has(ng.Name) {
-			continue
-		}
-		t := Task{
-			Call: c.CreateNodeGroup,
-			Data: ng,
-		}
-		createAllNodeGroups = append(createAllNodeGroups, t)
+	if t.Parallel {
+		doParallelTasks(appendErr, t.tasks)
```
martina-if: I think there is still a race condition here: there will be several goroutines in parallel appending to the `errs` slice, which I think is not thread safe. I think `doParallelTasks()` should collect the errors through channels, then combine them and return them. (Or use a mutex, which I think may look uglier)

errordeveloper: I think you are right, I'm seeing misbehaviour in some of my tests. I was hoping unit tests will bring it to justice! :)
It sounds like a channel here would help, but let me just write up some unit tests first.

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,208 @@
...
+	Do(chan error) error
+	Describe() string
+}
 
-// Task has a function with an opaque payload
-type Task struct {
-	Call taskFunc
-	Data interface{}
+// TaskSet wraps a set of tasks
+type TaskSet struct {
```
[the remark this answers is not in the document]
errordeveloper: Yeah, good point, thanks! I just called it somethings, cause `Tasks` would be confusing, given we have `Task` already.

pkg/ctl/delete/cluster.go
errordeveloper: Yeah, that's true, but I think it'd be okay for now, we need to do more here anyway (see #536). It shouldn't hurt to call this one multiple times.

pkg/cfn/manager/tasks.go
errordeveloper: So the error channel is closed here right after the handler returns, to indicate the task has completed. Maybe we could give this one a better name... It's a fire-and-forget type of task.

pkg/cfn/manager/deprecated.go
errordeveloper: 🤣

pkg/cfn/manager/delete_tasks.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/cloudformation"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// DeleteTasksForClusterWithNodeGroups defines tasks required to delete all the nodegroup
+// stacks and the cluster
+func (c *StackCollection) DeleteTasksForClusterWithNodeGroups(wait bool, cleanup func(chan error, string) error) (*TaskSet, error) {
```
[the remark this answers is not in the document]
errordeveloper: How about `NewTasksToDeleteNodeGroups`, `NewTasksToDeleteCluster` etc?

errordeveloper:
@martina-if thanks for spotting the race condition! As I mentioned, I was able to reproduce it with unit tests. I've fixed it a we have quite a few tests now. This PR is in a pretty good shape to merge next week :)

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,233 @@
...
-	for _, s := range nodeGroupStacks {
-		if err := c.DeleteNodeGroup(getNodeGroupName(s)); err != nil {
-			errs = append(errs, err)
-		}
+	allErrs := []error{}
+	for err := range errs {
+		allErrs = append(allErrs, err)
 	}
+	return allErrs
+	// for {
```
errordeveloper: @rndstr would you happen to know if `for x := range <channel>` is exactly the same as above? It seems like it is, but I'm not 100% sure.

pkg/cfn/manager/tasks_test.go
```diff
@@ -1,104 +1,328 @@
...
 
-					return nil
+					Expect(tasks.Describe()).To(Equal("2 sequential tasks: { 2 sequential sub-tasks: { t1.1, 2 parallel sub-tasks: { t3.1, t3.2 } }, t2.1 }"))
+
+					errs := tasks.DoAllSync()
+					Expect(errs).To(HaveLen(2))
+					Expect(errs[0].Error()).To(Equal("t3.2 always fails"))
+					Expect(errs[1].Error()).To(Equal("t2.1 always fails"))
+
+					Expect(called11).To(BeTrue())
```
errordeveloper: This checks if tasks have been called, but it'd be nice to test the effective order of completion... I was wondering would be a good way to test that? Perhaps a channel or thread-safe slice with a mutex?

pkg/cfn/manager/tasks_test.go
```diff
@@ -1,104 +1,328 @@
...
+
+					tasks.DryRun = true
+
+					Expect(tasks.DoAllSync()).To(HaveLen(0))
+
+					tasks.DryRun = false
+					errs := tasks.DoAllSync()
+					Expect(errs).To(HaveLen(2))
+					Expect(errs[0].Error()).To(Equal("t1.1 always fails"))
+					Expect(errs[1].Error()).To(Equal("t1.3 always fails"))
```
errordeveloper: Should add another case where a task return an error right way, which should also come on the channel...

pkg/cfn/manager/tasks.go
rndstr: yep looks the same to me.
(with the caveat that if you were to have code after that block you'd need a `break <label>` to get out of the for _and_ select  I think)

errordeveloper: Yeah... I think select just let's you handle multiple channels nicely, it makes sense to have `for/range` shorthand syntax for a single channel.

errordeveloper:
Looks like we cannot use `go test -race` in alpine container due to CGO/musl:
```
# runtime/race
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: race_linux_amd64.syso: in function `__sanitizer::GetArgv()':
gotsan.cc:(.text+0x4183): undefined reference to `__libc_stack_end'
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: race_linux_amd64.syso: in function `__sanitizer::ReExec()':
gotsan.cc:(.text+0x9797): undefined reference to `__libc_stack_end'
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: race_linux_amd64.syso: in function `__sanitizer::InternalAlloc(unsigned long, __sanitizer::SizeClassAllocatorLocalCache<__sanitizer::SizeClassAllocator32<__sanitizer::AP32> >*, unsigned long)':
gotsan.cc:(.text+0xaac1): undefined reference to `__libc_malloc'
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: race_linux_amd64.syso: in function `__sanitizer::InternalRealloc(void*, unsigned long, __sanitizer::SizeClassAllocatorLocalCache<__sanitizer::SizeClassAllocator32<__sanitizer::AP32> >*)':
gotsan.cc:(.text+0xca20): undefined reference to `__libc_realloc'
/usr/lib/gcc/x86_64-alpine-linux-musl/8.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: race_linux_amd64.syso: in function `__sanitizer::InternalFree(void*, __sanitizer::SizeClassAllocatorLocalCache<__sanitizer::SizeClassAllocator32<__sanitizer::AP32> >*)':
gotsan.cc:(.text+0x66e8): undefined reference to `__libc_free'
collect2: error: ld returned 1 exit status
```

pkg/cfn/manager/tasks_test.go
errordeveloper: Tried a channel, and it didn't work too well. Switched to a slice with a mutex, seems to work fine.

pkg/cfn/manager/delete_tasks.go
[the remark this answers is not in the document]
errordeveloper: I've renamed these to `NewTasksToDeleteX` and `NewTasksToCreateX`, let me know how this looks!

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,215 @@
...
-// Task has a function with an opaque payload
-type Task struct {
-	Call taskFunc
-	Data interface{}
+// TaskTree wraps a set of tasks
+type TaskTree struct {
+	tasks    []Task
+	Parallel bool
+	DryRun   bool
+	Sub      bool
```
martina-if: What is Sub?

pkg/cfn/manager/tasks.go
```diff
@@ -1,158 +1,215 @@
...
+func (t *asyncTaskWithStackSpec) Do(errs chan error) error {
+	_, err := t.call(t.stack)
+	close(errs)
+	return err
+}
+
+func doSingleTask(allErrs chan error, task Task) {
+	desc := task.Describe()
+	logger.Debug("started task: %s", desc)
+	errs := make(chan error)
```
martina-if: Why not share the same allErrs channel?

martina-if: defer errs.close()?

pkg/cfn/manager/tasks.go
errordeveloper: A sub-task, i.e. child not a parent. It's just so we can print user-friendly messages. Would `Child` be better? Besides, please do review the message formatting, it should evident from the tests and the readme changes.

pkg/cfn/manager/tasks.go
errordeveloper: > Why not share the same allErrs channel?

The task itself is expected to close this channel, and what we see on this channel gets copied into shared `allErrs` channel. I don't think we should change that now.

pkg/cfn/manager/api.go
```diff
@@ -243,38 +245,55 @@ func (c *StackCollection) DeleteStack(name string, force bool) (*Stack, error) {
...
-		return nil, fmt.Errorf("stack %q previously couldn't be deleted", name)
-	}
 	i.StackId = s.StackId
+	return c.DeleteStackBySpec(i)
+}
+
+// WaitDeleteStackByName sends a request to delete the stack, and waits until status is DELETE_COMPLETE;
+// any errors will be written to errs channel, assume completion when nil is written, do not expect
+// more then one error value on the channel, it's closed immediately after it is written to
+func (c *StackCollection) WaitDeleteStackByName(name string, errs chan error) error {
```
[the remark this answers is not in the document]
errordeveloper: 👍

pkg/cfn/manager/tasks.go
martina-if: maybe something like `isChild`?

martina-if: or isSubtask?

errordeveloper: Let's go with `isSubtask`, cause that doesn't imply changing words that are used in output.

Issue #649, opened 2019-03-20
Support Private/Public Endpoints


christopherhein:
##  ❗️ Currently this is blocked on CloudFormation supporting the Params - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-cluster.html

**Announcement** - https://aws.amazon.com/about-aws/whats-new/2019/03/amazon-eks-introduces-kubernetes-api-server-endpoint-access-cont/

**Why do you want this feature?**
This would add support for enabling Private and/or Public API Server endpoints allowing you to gate access to your clusters via the VPC. This allows you to isolate the API server limiting exposure.

**What feature/behavior/change do you want?**
Add support for `endpointPrivateAccess` and `endpointPublicAccess` on the ResourcesVpcConfig`.

https://docs.aws.amazon.com/eks/latest/APIReference/API_VpcConfigRequest.html#AmazonEKS-Type-VpcConfigRequest-endpointPrivateAccess

https://docs.aws.amazon.com/eks/latest/APIReference/API_VpcConfigRequest.html#AmazonEKS-Type-VpcConfigRequest-endpointPublicAccess

errordeveloper:
Thanks @christopherhein! Do you think we should consider enabling private access by default? Seems like we could actually do it quite safely.

mindfulmonk:
I don't see any cons to enabling both by default.

It is worth tracking https://github.com/aws/containers-roadmap/issues/22 for the private only situation.

christopherhein:
@mindfulmonk @errordeveloper I agree with this for the default to having both turned on, with flags/config to turn off.

whereisaaron:
_Not_ enabling the private access seems odd to me, why would you want worker nodes to have to loop through the Internet to get to their own control plane? So when they added this feature, my first thought was 'Wha? They weren't doing this already?' 😄

whereisaaron:
Although CF does not have these setting yet, you can change the settings to [enable private access on existing clusters from the AWS API/cli](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html)

`aws eks --region region update-cluster-config --name dev --resources-vpc-config endpointPublicAccess=true,endpointPrivateAccess=true`

So we could apply the setting after CF creation completes. Does that mess up CF? Given it is unaware of the settings would it keep reverting the setting on updates, or just not touch it?

I would wait, but sometimes the AWS CF team takes actual years to add the simplist things, or just never does.

[...]

PR #701, opened 2019-04-04
 Enable access from control plane to all node/pod ports

### LINKED ISSUE #699: Proxy Request URLs don't work
**What happened?**
`curl` requests through the port created with `kubectl proxy` work fine for service description endpoints but do not work for `proxy` endpoints. I am referring specifically to "apiserver proxy URLs" described here: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls 

I expect apiserver proxy URLs to work in my EKS cluster.

**How to reproduce it?**
1. Create a new EKS cluster using: `eksctl create cluster -n proxy-test -N 1 --ssh-access --ssh-public-key my-aws-public-key`
2. When finished, create a service and deployment that listens on port 80. Here is the yaml i used with `kubectl apply -f hello-world.yaml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-world
spec:
  replicas: 1
  selector:
    matchLabels:
      app: hello-world
  template:
    metadata:
      labels:
        app: hello-world
    spec:
      containers:
      - name: hw-container
        image: nginx
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: hello-world
spec:
  ports:
  - name: www
    protocol: TCP
    port: 80
    targetPort: 80
  selector:
    app: hello-world
  type: ClusterIP
```
3. You can verify that this service is up and serving traffic by getting the clusterIP from `kubectl get svc`, and then ssh-ing to the one node by doing `kubectl get nodes -owide` and then `ssh ec2-user@<ip-from-prev-cmd>`, and then while on the node doing `curl <cluster-ip-from-kubectl-get-svc>`. When you do this, i see a nice `<title>Welcome to nginx!</title>` html response. 
4. Exit the `ssh` session, and run `kubectl proxy`. This should proxy `localhost:8001` to the cluster api.
5. run `curl "localhost:8001/api/v1/namespaces/default/services/hello-world/"` You should see a nice output of information about the hello world service. `kubectl proxy` is working great!
6. run `curl "localhost:8001/api/v1/namespaces/default/services/hello-world:www/proxy/"` THIS REQUEST HANGS AND THEN FAILS with:
```
Error: 'dial tcp 192.168.11.3:80: connect: connection timed out'
Trying to reach: 'http://192.168.11.3:80/'
```
This is unexpected, i would expect this curl to work and to print out the same `<title>Welcome to nginx!</title>` html response from step 3 above. I am able to use `kops` to create a cluster where this DOES work, so it seems like an `eksctl` and/or `eks` issue.

Ultimately, my goal is to interact internal services within my cluster programmatically from golang using `ProxyGet` https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/service_expansion.go#L30 but I believe that these steps to recreate reveal the issue. If i can get these steps to work, it will likely fix the `ProxyGet` issues i'm having as well.

Should be using vanilla everything, i am on OSX 10.14.3, i used `brew` to install `eksctl`.

**Versions**
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.23"}
$ uname -a
Darwin benno-mbp 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.4", GitCommit:"c27b913fddd1a6c480c229191a087698aa92f0b1", GitTreeState:"clean", BuildDate:"2019-03-01T23:34:27Z", GoVersion:"go1.12", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.8-eks-7c34c0", GitCommit:"7c34c0d2f2d0f11f397d55a46945193a0e22d8f3", GitTreeState:"clean", BuildDate:"2019-03-01T22:49:39Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
```
$ eksctl create cluster -n proxy-test -N 1 --ssh-access --ssh-public-key spell-shared
[ℹ]  using region us-west-2
[ℹ]  setting availability zones to [us-west-2a us-west-2b us-west-2d]
[ℹ]  subnets for us-west-2a - public:192.168.0.0/19 private:192.168.96.0/19
[ℹ]  subnets for us-west-2b - public:192.168.32.0/19 private:192.168.128.0/19
[ℹ]  subnets for us-west-2d - public:192.168.64.0/19 private:192.168.160.0/19
[ℹ]  nodegroup "ng-a0af582d" will use "ami-0c28139856aaf9c3b" [AmazonLinux2/1.11]
[ℹ]  SSH public key file "spell-shared" does not exist; will assume existing EC2 key pair
[ℹ]  found EC2 key pair "spell-shared"
[ℹ]  creating EKS cluster "proxy-test" in "us-west-2" region
[ℹ]  will create 2 separate CloudFormation stacks for cluster itself and the initial nodegroup
[ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=us-west-2 --name=proxy-test'
[ℹ]  creating cluster stack "eksctl-proxy-test-cluster"
[ℹ]  creating nodegroup stack "eksctl-proxy-test-nodegroup-ng-a0af582d"
[ℹ]  --nodes-min=1 was set automatically for nodegroup ng-a0af582d
[ℹ]  --nodes-max=1 was set automatically for nodegroup ng-a0af582d
[✔]  all EKS cluster resource for "proxy-test" had been created
[✔]  saved kubeconfig as "/Users/ben/.kube/config"
[ℹ]  nodegroup "ng-a0af582d" has 0 node(s)
[ℹ]  waiting for at least 1 node(s) to become ready in "ng-a0af582d"
[ℹ]  nodegroup "ng-a0af582d" has 1 node(s)
[ℹ]  node "ip-192-168-3-221.us-west-2.compute.internal" is ready
[ℹ]  kubectl command should work with "/Users/ben/.kube/config", try 'kubectl get nodes'
[✔]  EKS cluster "proxy-test" in "us-west-2" region is ready
```

errordeveloper:
### Description

Fixes #699.

This had been copied from the original nodegroup template, arguably it breaks Kubernetes proxy feature.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,8 +456,7 @@ By default, `eksctl create cluster` will build a dedicated VPC, in order to avoi
 variety of reasons, including security, but also because it's challenging to detect all the settings in an existing VPC.
 Default VPC CIDR used by `eksctl` is `192.168.0.0/16`, it is divided into 8 (`/19`) subnets (3 private, 3 public & 2 reserved).
 Initial nodegroup is create in public subnets, with SSH access disabled unless `--allow-ssh` is specified. However, this implies
-that each of the EC2 instances in the initial nodegroup gets a public IP and can be accessed on ports 1025 - 65535, which is
-not insecure in principle, but some compromised workload could risk an access violation.
+that each of the EC2 instances in the initial nodegroup gets a public IP.
 
 If that functionality doesn't suit you, the following options are currently available.
 
--- a/pkg/cfn/builder/vpc.go
+++ b/pkg/cfn/builder/vpc.go
@@ -126,6 +126,7 @@ func (c *ClusterResourceSet) addOutputsForVPC() {
 
 var (
 	sgProtoTCP           = gfn.NewString("tcp")
+	sgProtoAny           = gfn.NewString("-1")
 	sgSourceAnywhereIPv4 = gfn.NewString("0.0.0.0/0")
 	sgSourceAnywhereIPv6 = gfn.NewString("::/0")
 
@@ -159,7 +160,7 @@ func (c *ClusterResourceSet) addResourcesForSecurityGroups() {
 			GroupId:               refClusterSharedNodeSG,
 			SourceSecurityGroupId: refClusterSharedNodeSG,
 			Description:           gfn.NewString("Allow nodes to communicate with each other (all ports)"),
-			IpProtocol:            gfn.NewString("-1"),
+			IpProtocol:            sgProtoAny,
 			FromPort:              sgPortZero,
 			ToPort:                sgMaxNodePort,
 		})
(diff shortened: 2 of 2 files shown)


errordeveloper:
cc @mhausenblas

christopherhein:
Hmm, should this be a default? I might make this a flag for the VPC config.

mhausenblas:
I agree with @christopherhein, this should be on an opt-in basis.

errordeveloper:
I agree, we could make it optional, but the tricky part is how is the user supposed to discover that they need to enable such an option? From #699 it is very clear that UX is very very confusing. Also, the special case of port 443 is kind of odd too, if you think about it.

Does private EKS endpoint (#649) change anything here? Seems like it may, but actually EKS->node comms are already private, as EKS is linked to the same set of subnets as nodes.

errordeveloper:
Arguably, anything that is being done here should be reinforced by some kind of admission policy, if you cannot do much with a pod that exposes port 80, you shouldn't be able to create such a pod in the first place. One of the challenges is that a lot of examples out there tell you to use nginx, so you will have to use custom config or remap to another port.

bensussman:
If we are considering making it user customizable, I would like to understand the security difference: what situation (presumably security) is helped by closing those lower port ranges?

README.md
```diff
@@ -456,8 +456,7 @@ By default, `eksctl create cluster` will build a dedicated VPC, in order to avoi
 variety of reasons, including security, but also because it's challenging to detect all the settings in an existing VPC.
 Default VPC CIDR used by `eksctl` is `192.168.0.0/16`, it is divided into 8 (`/19`) subnets (3 private, 3 public & 2 reserved).
 Initial nodegroup is create in public subnets, with SSH access disabled unless `--allow-ssh` is specified. However, this implies
```
martina-if: typo: is ~create~ -> is created

errordeveloper:
So [the official EKS nodegroup temple](https://github.com/awslabs/amazon-eks-ami/blob/2caf69b451b9f18126253f70e742075a3b1aa818/amazon-eks-nodegroup.yaml#L220-L262) only allowed ports 443 and 1025-65535 to be open nodegroup-to-nodegroup and nodegroup-to-CP. Please do note this excludes UDP.

We have discovered the hard way that this mean pods in different nodegroup weren't getting fair level of access to (most notably) DNS, but more generally a pod on any node should be able to communicate with a pod on any other node. A cluster with restricted comms between nodegroups is a plausible option, but users going down that route should be aware of what they are doing and thereby we don't place such restrictions by default.

With respect to node-to-CP comms, I'm very fuzzy on whether there is any real risk, user already trusts EKS control plane to do many things. Restricting comms between nodes and the control plane seems to only break the port forwarding feature in a non-obvious way, unless I'm missing something?

martina-if:
Closing until this comes back into our planning

PR #702, opened 2019-04-04
Enable strict parsing

errordeveloper:
### Description

This allows us to error when config has unknown fields.

### Checklist

### DIFF
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -17,6 +17,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/yaml"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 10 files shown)


pkg/eks/api.go
```diff
@@ -158,6 +159,10 @@ func LoadConfigFromFile(configFile string, cfg *api.ClusterConfig) error {
 		return errors.Wrapf(err, "reading config file %q", configFile)
 	}
 
+	if err := yaml.UnmarshalStrict(data, &api.ClusterConfig{}); err != nil {
```
martina-if: Where is the old unmarshal call? was it the `runtime.Decode()` below? should we keep both?

errordeveloper: Yes, and we still want to use that one. The output of this one is not stored. A bit of a hack, but couldn't find a way to make `runtime.Decode()` do the same, but it does other things that are useful for us.

martina-if: Oh, but then this will be confusing, can we add a comment that the actual thing is used is `obj` from the `Decode()`?

errordeveloper: Yes, good idea, I added a comment with more notes.

PR #703, opened 2019-04-05
Add taints to nodegroup config

### LINKED ISSUE #396: feat: add taints to the node on creation

rndstr:
### Description

This adds a section `taints` similar to `labels` to nodegroups so they
can be registered during creation.

### Checklist
- [ ] confirm an empty `--register-with-taints=` does not break

Closes #396


pkg/nodebootstrap/assets/10-eksclt.al2.conf
```diff
@@ -12,6 +12,7 @@ ExecStart=
 ExecStart=/usr/bin/kubelet \
   --node-ip=${NODE_IP} \
   --node-labels=${NODE_LABELS},alpha.eksctl.io/instance-id=${INSTANCE_ID} \
+  --register-node=true --register-with-taints=${NODE_TAINTS} \
```
rndstr: `--register-node=true` is required for `--register-with-taint` to take any effect.

`--register-node=true` was removed in #644 from this file (but hasn't been from https://github.com/weaveworks/eksctl/blob/cf75cb1288a9ac0d42e2850f301aafa8999bb139/pkg/nodebootstrap/assets/bootstrap.ubuntu.sh btw)

@errordeveloper will re-introducing this here break things?

errordeveloper: No, it was removed as it was assumed to be of no effect, thanks for find this out!

errordeveloper:
> confirm an empty `--register-with-taints=` does not break

Integration tests should confirm it. I'll try to run it here, maybe we can merge this along with another PR and cut a release today.

PR #713, opened 2019-04-08
Support other ssh options

martina-if:
closes https://github.com/weaveworks/eksctl/issues/386
Support different ways to use SSH keys: by local file, by the key content and through the name of a key already imported in EC2

### Checklist

- [x] rebase
- [x] Manually tested


pkg/apis/eksctl.io/v1alpha4/validation.go
```diff
@@ -132,5 +133,33 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
 		return err
 	}
 
+	if err := validateNGSSHFlags(ng.SSH); err != nil {
+		return errors.New("only one ssh public key can be specified per node-group")
+	}
+	return nil
+}
+
+func validateNGSSHFlags(SSH *NodeGroupSSH) error {
```
martina-if: this name looks a bit weird, any ideas for a better name?

StevenACoffman: Suggestion 1: Omit `Flags` because it is validating the NodeGroupSSH (right now only flags, sure).
Suggestion 2: Either spell out NodeGroup like `validateNodeGroupIAM` or all the private functions of ValidateNodeGroup could omit the implied 'NodeGroup', and just be `validateSSH`?
NOTE: Feel free to ignore both.

martina-if: Thanks!

pkg/apis/eksctl.io/v1alpha4/validation.go
```diff
@@ -135,5 +136,33 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
...
+		SSH.PublicKey,
+		SSH.PublicKeyName)
+
+	if numSSHFlagsEnabled > 1 {
+		return errors.New("only one ssh public key can be specified per node-group")
+	}
 	return nil
 }
+
+func countEnabledFlags(flags ...*string) int {
```
errordeveloper: Let's try to use word flags when it's about the CLI flags, you can use 'fields' or 'params', whichever you like :)

pkg/apis/eksctl.io/v1alpha4/zz_generated.deepcopy.go
```diff
@@ -299,6 +299,13 @@ func (in *NodeGroup) DeepCopyInto(out *NodeGroup) {
 			(*out)[key] = val
 		}
 	}
+	if in.Taints != nil {
```
errordeveloper: I wonder why is this one getting added... would you mind trying to delete and regenerate it? Seems like it could be a rebase artefact.

README.md
```diff
@@ -628,6 +628,8 @@ To delete this cluster, run:
 eksctl delete cluster -f cluster.yaml
 ```
 
+
```
errordeveloper: Did you want to add something here?

examples/07-ssh-keys.yaml
```diff
@@ -0,0 +1,34 @@
+# A simple example of ClusterConfig object:
+---
+apiVersion: eksctl.io/v1alpha4
+kind: ClusterConfig
+
+metadata:
+  name: test-cluster
```
errordeveloper: By current convention this would be `cluster-7` :)

examples/07-ssh-keys.yaml
```diff
@@ -0,0 +1,34 @@
+# A simple example of ClusterConfig object:
```
errordeveloper: How about this:
```suggestion
# An example of ClusterConfig showing different ways of setting nodegroup SSH keys:
```

pkg/apis/eksctl.io/v1alpha4/validation.go
```diff
@@ -1,6 +1,7 @@
 package v1alpha4
 
 import (
+	"errors"
```
errordeveloper: Would you mind to switch from `errors.New` to `fmt.Errorf`, so it's consistent with other packages? `errors` is usually the `github.com/pkg/errors` in all other places...

pkg/ctl/create/utils.go
```diff
@@ -50,3 +51,44 @@ func checkVersion(ctl *eks.ClusterProvider, meta *api.ClusterMeta) error {
 
 	return nil
 }
+
+// loadSSHKey loads the ssh public key specified in the NodeGroup. The key should be specified
+// in only one way: by name (for a key existing in EC2), by path (for a key in a local file)
+// or by its contents (in the config-file). It also assumes that if ssh is enabled (SSH.Allow
+// == true) then one key was specified
+func loadSSHKey(ng *api.NodeGroup, clusterName string, provider api.ClusterProvider) error {
```
errordeveloper: Why not make it part of `pkg/ssh`?

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		logger.Debug("deleting key %q", *matching[i])
+		if _, err := provider.EC2().DeleteKeyPair(input); err != nil {
+			logger.Debug("key pair couldn't be deleted: %v", err)
+		}
+	}
+}
+
+// CheckKeyExistsInEc2 returns whether a public ssh key already exists in EC2 or error if it couldn't be checked
+func CheckKeyExistsInEc2(sshPublicKeyName string, provider api.ClusterProvider) error {
```
errordeveloper:
```suggestion
func CheckKeyExistsInEC2(sshPublicKeyName string, provider api.ClusterProvider) error {
```

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	"k8s.io/kops/pkg/pki"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha4"
+	"github.com/weaveworks/eksctl/pkg/utils"
+)
+
+// LoadSSHKeyFromFile loads and imports a public SSH key from a file provided a path to that file
+func LoadSSHKeyFromFile(filePath, clusterName, ngName string, provider api.ClusterProvider) error {
```
errordeveloper: We can actually strip `SSH` from function names, as it's the `ssh` package :)

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if _, err := provider.EC2().ImportKeyPair(input); err != nil {
+		return errors.Wrap(err, "importing SSH public key")
+	}
+	return nil
+}
+
+// getSSHKeyName generates the name of an SSH key based on the cluster name, nodegroup name and fingerprint
+// in the form "eksctl-<clusterName>-nodegroup-<nodeGroupName>-<fingerprint>"
+func getSSHKeyName(clusterName, nodeGroupName, fingerprint string) string {
```
errordeveloper: Let's use either "Key" or "PublicKey" consistently 👍

README.md
martina-if: Nope, seems like a mistake, thanks

examples/07-ssh-keys.yaml
martina-if: right!

pkg/apis/eksctl.io/v1alpha4/validation.go
martina-if: Done. I also use `errors.New()` in tests. I'll change those as well, or those can be from the `errors` package?

pkg/ctl/create/utils.go
martina-if: I put it here because I consider this to be on a different layer. This functionality does logic related to how the CLI and the config file should work (closer to the UI), while the other package should focus, IMO, on how to do SSH related things, without caring about the fine details of the user interface.

pkg/apis/eksctl.io/v1alpha4/validation.go
errordeveloper: It's not a big deal, let's keep it that way.

pkg/ctl/create/utils.go
errordeveloper: 👍

pkg/apis/eksctl.io/v1alpha4/defaults.go
```diff
@@ -30,17 +30,19 @@ func SetNodeGroupDefaults(_ int, ng *NodeGroup) error {
...
-	// Enable SSH when a key is provided
-	if ng.SSH.PublicKeyPath != nil {
+	if numSSHFlagsEnabled > 0 {
 		ng.SSH.Allow = Enabled()
-	}
-
-	if *ng.SSH.Allow && ng.SSH.PublicKeyPath == nil {
-		ng.SSH.PublicKeyPath = &DefaultNodeSSHPublicKeyPath
+	} else {
+		if ng.SSH.Allow != nil && *ng.SSH.Allow {
```
errordeveloper:
```suggestion
		if IsEnabled(ng.SSH.Allow) {
```

pkg/ctl/cmdutils/nodegroup_flags.go
```diff
@@ -37,7 +37,7 @@ func AddCommonCreateNodeGroupFlags(cmd *cobra.Command, fs *pflag.FlagSet, p *api
 	fs.IntVar(&ng.MaxPodsPerNode, "max-pods-per-node", 0, "maximum number of pods per node (set automatically if unspecified)")
 
 	ng.SSH.Allow = fs.Bool("ssh-access", *ng.SSH.Allow, "control SSH access for nodes")
-	ng.SSH.PublicKeyPath = fs.String("ssh-public-key", *ng.SSH.PublicKeyPath, "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
+	ng.SSH.PublicKeyPath = fs.String("ssh-public-key", "", "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
```
errordeveloper: The thing with previous approach that user sees this:
```
      --ssh-public-key string          SSH public key to use for nodes (import from local path, or use existing EC2 key pair) (default "~/.ssh/id_rsa.pub")
```
Now they won't see what's the default, while there is a default.

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if existing == nil {
+		return fmt.Errorf("cannot find EC2 key pair %q", sshKeyName)
+	}
+
+	return nil
+}
+
+// FileExists returns true if there is a local file with that path
+func FileExists(filePath string) bool {
```
errordeveloper: I just remove function with the same name that was defined in `utils` (7203b1f1392668e462ef293f28a3203140772579) 🙄🤣...

errordeveloper: Perhaps we can create a new package in `pkg/utils/file` and move `ExpandPath` and this one there? We could also make sure that other instances of `if _, err := os.Stat(path); os.IsNotExist(err) {`  use the function, and it seems like  only `pkg/utils/kubeconfig` needs it basically.

pkg/utils/file.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+
+	return false, err
+}
+
+// CheckFileExists returns an error if the file doesn't exist
```
errordeveloper: Would returning a boolean be insufficient?

pkg/utils/file.go
```diff
@@ -0,0 +1,32 @@
+package utils
+
+import (
+	kopsutils "k8s.io/kops/upup/pkg/fi/utils"
+	"os"
+)
+
+// FileExists checks to see if a file exists.
```
errordeveloper: Looks like this one isn't actually used anywhere...

pkg/ctl/cmdutils/nodegroup_flags.go
martina-if: Changed in the line above, what do you think?

pkg/utils/file.go
```diff
@@ -0,0 +1,32 @@
+package utils
```
errordeveloper: I was think it should go into its own package, as that would avoid cyrcular dependency between `pkg/utils` and `pkg/utils/kubeconfig`, as `pkg/utils`  already depends on `pkg/utils/kubeconfig`...

pkg/utils/file.go
martina-if: Yeah, I just returned an error because that's usually the convention with the "checkSomething()" methods but the boolean is nicer to use anyway. Will change

PR #718, opened 2019-04-09
[WIP] Feature utils export

pawelprazak:
### Description

As discussed previously, this PR provides support for advanced use cases:
- versioning cloud formation templates in git for transparency and auditability (should help users with compliance)
- manual deployment of cloud formation templates
- help with supporting custom/derivative distributions/scripts
- help with porting changes to other projects

### Checklist

### DIFF
--- a/pkg/ami/static_resolver.go
+++ b/pkg/ami/static_resolver.go
@@ -16,7 +16,11 @@ type StaticDefaultResolver struct {
 func (r *StaticDefaultResolver) Resolve(region, version, instanceType, imageFamily string) (string, error) {
 	logger.Debug("resolving AMI using StaticDefaultResolver for region %s, version %s, instanceType %s and imageFamily %s", region, version, instanceType, imageFamily)
 
-	regionalAMIs := StaticImages[version][imageFamily][ImageClassGeneral]
+	regionalAMIs, ok := StaticImages[version][imageFamily][ImageClassGeneral]
+	if !ok {
+		logger.Critical("image family %s doesn't have default image class", imageFamily)
+		return "", NewErrFailedResolution(region, version, instanceType, imageFamily)
+	}
 	return regionalAMIs[region], nil
 }
 
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -366,6 +366,7 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
 		},
 		SSH: &NodeGroupSSH{
 			Allow:         Disabled(),
+			PublicKeyName: new(string),
 			PublicKeyPath: &DefaultNodeSSHPublicKeyPath,
 		},
 	}
@@ -424,6 +425,8 @@ type NodeGroup struct {
 	PreBootstrapCommands []string `json:"preBootstrapCommand,omitempty"`
 	// +optional
 	OverrideBootstrapCommand *string `json:"overrideBootstrapCommand,omitempty"`
+	// +optional
+	OverrideUserData *string `json:"overrideUserData,omitempty"`
 
 	// +optional
 	ClusterDNS string `json:"clusterDNS,omitempty"`
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -155,6 +155,12 @@ func IsDisabled(v *bool) bool { return v != nil && !*v }
 // IsSetAndNonEmptyString will only return true if s is not nil and not empty
 func IsSetAndNonEmptyString(s *string) bool { return s != nil && *s != "" }
 
+// NewInt return pointer to an int value
+// for use in defaulters of *int fields
+func NewInt(i int) *int {
+	return &i
+}
+
 // SupportedRegions are the regions where EKS is available
 func SupportedRegions() []string {
 	return []string{
@@ -338,8 +344,12 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
 			WithLocal:  Enabled(),
 			WithShared: Enabled(),
 		},
-		DesiredCapacity: nil,
+		DesiredCapacity: NewInt(DefaultNodeCount),
+		MinSize:         NewInt(DefaultNodeCount),
+		MaxSize:         NewInt(DefaultNodeCount),
 		InstanceType:    DefaultNodeType,
+		AMI:             DefaultResolver,
+		AMIFamily:       DefaultNodeImageFamily,
 		VolumeSize:      0,
 		VolumeType:      DefaultNodeVolumeType,
 		IAM: &NodeGroupIAM{
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -57,7 +57,7 @@ const (
 	// RegionAPSouth1 represents the Asia-Pacific South Region Mumbai
 	RegionAPSouth1 = "ap-south-1"
 
-	// DefaultRegion defines the default region, where to deploy the EKS cluster
+	// DefaultRegion defines the default region, where to deploy the EKS cluster, see also DefaultAvailabilityZones
 	DefaultRegion = RegionUSWest2
 
 	// Version1_10 represents Kubernetes version 1.10.x
@@ -130,6 +130,13 @@ var (
 
 	// DefaultNodeSSHPublicKeyPath is the default path to SSH public key
 	DefaultNodeSSHPublicKeyPath = "~/.ssh/id_rsa.pub"
+
+	// DefaultAvailabilityZones for the DefaultRegion
+	DefaultAvailabilityZones = []string{
+		DefaultRegion + "a",
+		DefaultRegion + "b",
+		DefaultRegion + "c",
+	}
 )
 
 // Enabled return pointer to true value
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -89,6 +89,8 @@ const (
 	// NodeVolumeTypeST1 is Cold HDD
 	NodeVolumeTypeST1 = "st1"
 
+	// DefaultResolver defines the default AMI resolver to use
+	DefaultResolver = NodeImageResolverStatic
 	// DefaultNodeImageFamily defines the default image family for the worker nodes
 	DefaultNodeImageFamily = NodeImageFamilyAmazonLinux2
 	// NodeImageFamilyAmazonLinux2 represents Amazon Linux 2 family
--- a/pkg/ctl/utils/export.go
+++ b/pkg/ctl/utils/export.go
@@ -87,6 +87,8 @@ func doExport(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string, cmd
 		if err := ctl.CheckAuth(); err != nil {
 			return err
 		}
+	} else {
+		logger.Info("offline mode: skipping STS access check")
 	}
 
 	if !ctl.IsSupportedRegion() {
@@ -130,7 +132,6 @@ func doExport(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg string, cmd
 		}
 
 		if !cfg.HasAnySubnets() {
-			// default: create dedicated VPC
 			if len(cfg.AvailabilityZones) == 0 {
 				cfg.AvailabilityZones = api.DefaultAvailabilityZones
 			}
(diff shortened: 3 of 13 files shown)


pawelprazak:
FYI no support for Ubuntu AMIs at the moment.

pkg/ctl/utils/utils.go
```diff
@@ -7,7 +7,8 @@ import (
 )
 
 var (
-	clusterConfigFile = ""
+	clusterConfigFile string
+	output            string
```
martina-if: Please move the `output` variable to `export.go` where it is used.

martina-if:
Hi @pawelprazak , thanks for the PR! I added some feedback.

pkg/cfn/manager/tasks.go
```diff
@@ -156,3 +157,42 @@ func (c *StackCollection) WaitDeleteAllNodeGroups(force bool) []error {
...
+
+	templates := map[string]gfn.Template{}
+	templates[name] = stack.Template()
+
+	errs := []error{}
+	if err := stack.AddAllResources(); err != nil {
+		errs = append(errs, err)
+	}
+
+	for i := range c.spec.NodeGroups {
```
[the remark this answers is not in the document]
pawelprazak: sure, but that's actually copy-pasted for here ;)
https://github.com/weaveworks/eksctl/blob/ca9f7332a759a328e44d64a7d47f6dfddb65f17a/pkg/cfn/manager/create_tasks.go#L36

pkg/ctl/utils/export.go
```diff
@@ -0,0 +1,273 @@
...
+		if !validVersion {
+			return fmt.Errorf("invalid version, supported values: %s", strings.Join(api.SupportedVersions(), ", "))
+		}
+	}
+	if cfg.Metadata.Region == "" {
+		cfg.Metadata.Region = api.DefaultRegion
+	}
+
+	// VPC and subnets
+	createOrImportVPC := func() error {
```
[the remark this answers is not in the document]
pawelprazak: again, that's copy pasted from create code, don't like it either but didn't wanted to refactor too much in one go, original code:
https://github.com/weaveworks/eksctl/blob/master/pkg/ctl/create/cluster.go#L162

pkg/ctl/utils/utils.go
pawelprazak: again, that follows existing code for consistency:
https://github.com/weaveworks/eksctl/blob/master/pkg/ctl/get/get.go#L11

pkg/ctl/utils/export.go
```diff
@@ -0,0 +1,273 @@
...
+				return fmt.Errorf("none or too few private subnets to use with --node-private-networking")
+			}
+			return nil
+		}
+
+		if !cfg.HasAnySubnets() {
+			// default: create dedicated VPC
+			if len(cfg.AvailabilityZones) == 0 {
+				cfg.AvailabilityZones = []string{
+					cfg.Metadata.Region + "a",
```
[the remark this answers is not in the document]
pawelprazak: yes, I've threw out a lot of complex code (not sure what kops is doing here anyway) because existing code was unreadable and replaced it with simple default to test it works for simple use case

please advise on the implementation that is acceptable and in scope for this PR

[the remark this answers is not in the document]
pawelprazak: for now I will make it static, just like region

pawelprazak:
thank you @martina-if I've addressed the comments and pushed a new commit, please have a look :)

pkg/cfn/manager/tasks.go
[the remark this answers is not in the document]
pawelprazak: fixed in new commit

pawelprazak:
added rebase on master

pawelprazak:
rebased, again

pkg/ctl/utils/export.go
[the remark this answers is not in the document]
martina-if: I agree, I also find it tricky to find the defaults for different parameters. We should take a look at that. If you have any ideas please let us know :)

Regarding the regions, let's see what @errordeveloper says.

martina-if:
LGTM, but @errordeveloper should also take a look

pkg/ctl/utils/export.go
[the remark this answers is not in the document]
voxxit: Why not use the AWS API to grab AZs for a region? https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-availability-zones.html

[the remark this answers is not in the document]
errordeveloper: Please use `pkg/az` for picking AZs, and we have `pkg/vpc` for picking subnets as well.

errordeveloper:
Still didn't look at everything, but planning on doing it soon (hopefully this week).

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -336,8 +351,12 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
 			WithLocal:  Enabled(),
 			WithShared: Enabled(),
 		},
-		DesiredCapacity: nil,
+		DesiredCapacity: NewInt(DefaultNodeCount),
+		MinSize:         NewInt(DefaultNodeCount),
+		MaxSize:         NewInt(DefaultNodeCount),
 		InstanceType:    DefaultNodeType,
+		AMI:             DefaultResolver,
+		AMIFamily:       DefaultNodeImageFamily,
```
errordeveloper: 👍

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -336,8 +351,12 @@ func (c *ClusterConfig) NewNodeGroup() *NodeGroup {
 			WithLocal:  Enabled(),
 			WithShared: Enabled(),
 		},
-		DesiredCapacity: nil,
+		DesiredCapacity: NewInt(DefaultNodeCount),
```
errordeveloper: So actually having no default size is a feature, which is to do with cluster autoscaler. I believe the idea is that user should be able to set just one of the 3 parameters.

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -412,6 +432,8 @@ type NodeGroup struct {
 	PreBootstrapCommands []string `json:"preBootstrapCommand,omitempty"`
 	// +optional
 	OverrideBootstrapCommand *string `json:"overrideBootstrapCommand,omitempty"`
+	// +optional
+	OverrideUserData *string `json:"overrideUserData,omitempty"`
```
errordeveloper: I think this would be a separate feature to discuss. So far, we have stayed away from exposing this...

pkg/nodebootstrap/userdata.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"strings"
 
+	gfn "github.com/awslabs/goformation/cloudformation"
```
errordeveloper: I see what you wanted to do here, but actually it's better to avoid introducing package dependencies like this.

pkg/nodebootstrap/assets/simple.al2.userdata
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -o xtrace
+/etc/eks/bootstrap.sh ${ClusterName} ${BootstrapArguments}
```
errordeveloper: I this implies that you will have to bypass gzip+base64 encoding of user data... isn't it?

pkg/ami/static_resolver.go
```diff
@@ -16,7 +16,11 @@ type StaticDefaultResolver struct {
 func (r *StaticDefaultResolver) Resolve(region, version, instanceType, imageFamily string) (string, error) {
 	logger.Debug("resolving AMI using StaticDefaultResolver for region %s, version %s, instanceType %s and imageFamily %s", region, version, instanceType, imageFamily)
 
-	regionalAMIs := StaticImages[version][imageFamily][ImageClassGeneral]
+	regionalAMIs, ok := StaticImages[version][imageFamily][ImageClassGeneral]
```
errordeveloper: Do you want make a PR to just add this few lines?

pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -128,6 +130,13 @@ var (
 
 	// DefaultNodeSSHPublicKeyPath is the default path to SSH public key
 	DefaultNodeSSHPublicKeyPath = "~/.ssh/id_rsa.pub"
+
+	// DefaultAvailabilityZones for the DefaultRegion
+	DefaultAvailabilityZones = []string{
```
errordeveloper: Please consider using `pkg/az`, and do feel free to suggest any changes to that. E.g., you might want to have deterministic behaviour, as it currently pick AZs randomly. But let's not add another AZ selection mechanism outside of that package.

errordeveloper:
@pawelprazak are you intending to come back to this soon? I'm keen to get it merged in! :)

pkg/apis/eksctl.io/v1alpha4/types.go
pawelprazak: I require this, because the current content of user data is over-complicated and not usable when exported

pkg/nodebootstrap/assets/simple.al2.userdata
pawelprazak: `gzip+base64` is not useful when exported anyway

pkg/ami/static_resolver.go
pawelprazak: I can do it the question is how much it will postpone an already long-running PR

pkg/apis/eksctl.io/v1alpha4/types.go
pawelprazak: this one works, and is actually missing, because you have a `DefaultRegion`, `pkg/az` is dufferent story

pkg/apis/eksctl.io/v1alpha4/types.go
pawelprazak: defaults are the only sane way to write this kind of things, otherwise you'll get tons of subtle bugs every time you want to change something (as I've experienced while working on this PR)

pawelprazak:
it staled for so long I kind of run out of steam ;)

pkg/nodebootstrap/userdata.go
pawelprazak: it is dependant on `gfn` just not explicitly, returning string was not helpful, so I've changed it to explicit type instead of string-wrapped one ;)

pkg/apis/eksctl.io/v1alpha4/types.go
pawelprazak: also I need this code have offline capability if requested (for UX/DX reasons)

pkg/apis/eksctl.io/v1alpha4/types.go
pawelprazak: this is the minimal output you need:
```
      UserData:
        Fn::Base64:
          !Sub |
            #!/bin/bash
            set -o xtrace
            /etc/eks/bootstrap.sh ${ClusterName}
            /opt/aws/bin/cfn-signal --exit-code $? \
                     --stack  ${AWS::StackName} \
                     --resource NodeGroup  \
                     --region ${AWS::Region}
```

pawelprazak:
all of the problems with this PR are from pre-existing technical debt, a major refactoring is required to actually fix those problems

pawelprazak:
and I would make things a little bit more functional
and a lot less mutable state :wink:
basically at the high lvl all wee need i series of transformations from one type to another
this would help with transition to operator later on
basically as a first step (with no dependency on connectivity nor aws credentials)
we should be able to transform ClusterConfig into sets of gfn.Template and k8s manifest (edited) 
as a second step we'd take those stacks and manifests and apply them with aws and k8s credentials
and get the cluster status as a result
the place with the '+' sign is where every command would use different operations to achieve the last step
I know it's a very simplified view
but I'm confident this refactoring would improve things a lot
as a side quest, there is also an opportunity to make it even easier to reuse `eksctl` as a library in other projects (edited) 
(oh, I forgot about step zero, transforming/merging command line args with the config file into a cluster config and provider config)

What is needed:
- clear responsibility separation
- elimination of shared mutable state
- transformation of types (instead of procedural instructions)
```
        /--> ClusterConfig -> 
cli -> |
        \--> ProviderConfig -> 
```

errordeveloper:
> all of the problems with this PR are from pre-existing technical debt, a major refactoring is required to actually fix those problems

That's understandable, I am sorry about that. Let me see where we can pick up. However, it might have to wait until after KubeCon.

pawelprazak:
I'm also at KubeCon so feel free to track me down in Barcelona ;)

errordeveloper:
I'm having a go at rebasing this, so far so good. Wondering if I should push to this branch or open another PR...

errordeveloper:
I've opened #1039.

PR #719, opened 2019-04-09
Add reference value helpers for *bool and *string

errordeveloper:
### Description

This PR adds helper for checking if a `*string` value was set and isn't an empty string, as well as helpers for checking when a `*bool` value was set and is either `true` or `false`.

This is a follow-up to #673.

### Checklist

- [ ] ~All integration tests passing (i.e. `make integration-test`)~ - blocked by #720

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha4/types.go
+++ b/pkg/apis/eksctl.io/v1alpha4/types.go
@@ -123,19 +123,28 @@ var (
 	DefaultNodeSSHPublicKeyPath = "~/.ssh/id_rsa.pub"
 )
 
-// NewBoolTrue return pointer to true value
+// Enabled return pointer to true value
 // for use in defaulters of *bool fields
-func NewBoolTrue() *bool {
+func Enabled() *bool {
 	v := true
 	return &v
 }
 
-// NewBoolFalse return pointer to false value
+// Disabled return pointer to false value
 // for use in defaulters of *bool fields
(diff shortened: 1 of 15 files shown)


pkg/apis/eksctl.io/v1alpha4/types.go
```diff
@@ -137,6 +137,15 @@ func NewBoolFalse() *bool {
 	return &v
 }
 
+// BoolIsSetToTrue will only return true if v is not nil and true
+func BoolIsSetToTrue(v *bool) bool { return v != nil && *v }
```
martina-if: Maybe it could be named `FlagIsEnabled(...)`?

errordeveloper: Yeah, maybe `IsEnabled` would be a nice and short. Word 'flag' belongs to parts where it's about CLI flags (like `pkg/ctl/cmduits`), so let's not mix up words here it's the API package :)

errordeveloper:
@martina-if could you please take another look?

I also added another commit that fixes a minor case that we've missed in #673. I don't know if you might have fixed it in your other PR, I'm happy to drop the commit from here if you feel that we should. From the perspective of this PR that commit fixes integration tests.

errordeveloper:
> From the perspective of this PR that commit fixes integration tests.

Actually, I said that too early! It's not true, the commit doesn't fix it. Let's work on that separately.

PR #714, opened 2019-04-08
delete nodegroups with config file

errordeveloper:
### Description

Enhance `eksctl delete nodegroup`:

- add functionality to delete multiple nodegroups with `--config-file`
- and `--include`/`--exclude` filters
- add `--only-missing` (inverted fitering)
- add `--approve` (only with `--config-file`)

Fixes #664.

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] Added tests that cover your change (if possible)
- [x] All unit tests passing (i.e. `make test`)
- [x] All integration tests passing (i.e. `make integration-test`)
- [x] Added/modified documentation as required (such as the `README.md`, and `examples` directory)

PR #723, opened 2019-04-11
Replace `--only` with `--include`, add `--exclude`

errordeveloper:
### Description

- generalise `NodeGroupFilter`
- rename `CheckEachNodeGroup` to `ForEach`
- add tests

### Checklist

### DIFF
--- a/pkg/ctl/cmdutils/nodegroup_filter.go
+++ b/pkg/ctl/cmdutils/nodegroup_filter.go
@@ -16,54 +16,84 @@ import (
 
 // NodeGroupFilter holds filter configuration
 type NodeGroupFilter struct {
-	SkipAll           bool
-	IgnoreAllExisting bool
+	ExcludeAll bool // highest priority
 
-	existing sets.String
-	only     []glob.Glob
-	onlySpec string
+	// include filters take precedence
+	includeNames    sets.String
+	includeGlobs    []glob.Glob
+	rawIncludeGlobs []string
+
+	excludeNames    sets.String
+	excludeGlobs    []glob.Glob
+	rawExcludeGlobs []string
 }
 
 // NewNodeGroupFilter create new NodeGroupFilter instance
 func NewNodeGroupFilter() *NodeGroupFilter {
 	return &NodeGroupFilter{
-		IgnoreAllExisting: true,
-		SkipAll:           false,
-		existing:          sets.NewString(),
+		ExcludeAll:   false,
+		includeNames: sets.NewString(),
+		excludeNames: sets.NewString(),
 	}
 }
 
-// ApplyOnlyFilter parses given globs for exclusive filtering
-func (f *NodeGroupFilter) ApplyOnlyFilter(globExprs []string, cfg *api.ClusterConfig) error {
+// AppendGlobs appends globs for inclusion and exclusion rules
+func (f *NodeGroupFilter) AppendGlobs(includeGlobExprs, excludeGlobExprs []string, nodeGroups []*api.NodeGroup) error {
+	if err := f.AppendIncludeGlobs(nodeGroups, includeGlobExprs...); err != nil {
+		return err
+	}
+	return f.AppendExcludeGlobs(excludeGlobExprs...)
+}
+
+// AppendIncludeGlobs sets globs for inclusion rules
+func (f *NodeGroupFilter) AppendIncludeGlobs(nodeGroups []*api.NodeGroup, globExprs ...string) error {
 	for _, expr := range globExprs {
 		compiledExpr, err := glob.Compile(expr)
 		if err != nil {
 			return errors.Wrapf(err, "parsing glob filter %q", expr)
 		}
-		f.only = append(f.only, compiledExpr)
+		f.includeGlobs = append(f.includeGlobs, compiledExpr)
+		f.rawIncludeGlobs = append(f.rawIncludeGlobs, expr)
 	}
-	f.onlySpec = strings.Join(globExprs, ",")
-	return f.onlyFilterMatchesAnything(cfg)
+	return f.includeGlobsMatchAnything(nodeGroups)
 }
 
-func (f *NodeGroupFilter) onlyFilterMatchesAnything(cfg *api.ClusterConfig) error {
-	if len(f.only) == 0 {
+func (f *NodeGroupFilter) includeGlobsMatchAnything(nodeGroups []*api.NodeGroup) error {
+	if len(f.includeGlobs) == 0 {
 		return nil
 	}
-	for _, ng := range cfg.NodeGroups {
-		for _, compiledExpr := range f.only {
-			if compiledExpr.Match(ng.Name) {
-				return nil
-			}
+	for _, ng := range nodeGroups {
+		if f.matchGlobs(ng.Name, f.includeGlobs) {
+			return nil
 		}
 	}
-	return fmt.Errorf("no nodegroups match filter specification: %q", f.onlySpec)
+	return fmt.Errorf("no nodegroups match include glob filter specification: %q", strings.Join(f.rawIncludeGlobs, ","))
 }
 
-// ApplyExistingFilter uses stackManager to list existing nodegroup stacks and configures
+// AppendIncludeNames appends explicit names to the include filter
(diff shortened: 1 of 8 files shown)


errordeveloper:
I've picked this out from #714.

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -16,54 +16,79 @@ import (
...
 // NodeGroupFilter holds filter configuration
 type NodeGroupFilter struct {
-	SkipAll           bool
-	IgnoreAllExisting bool
+	ExcludeAll bool // highest priority
 
-	existing sets.String
-	only     []glob.Glob
-	onlySpec string
+	// include filters take presidence
```
martina-if: filter for the presidence \( ﾟヮﾟ)/ (or precedence)

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -72,70 +97,109 @@ func (f *NodeGroupFilter) ApplyExistingFilter(stackManager *manager.StackCollect
...
+		}
+	}
 	return nil
 }
 
-// Match checks given nodegroup against the filter
-func (f *NodeGroupFilter) Match(ng *api.NodeGroup) bool {
-	if f.SkipAll {
-		return false
+func (*NodeGroupFilter) matchGlobs(name string, exprs []glob.Glob, result *bool) {
```
martina-if: Any particular reason to have the result as an argument instead of return value? I think it would be nicer if this method returned a bool result instead of having to pass an output argument. So it could be used like:


```if shouldInclude := f.matchGlobs(...); shouldInclude {...}```

File: pkg/ctl/cmdutils/nodegroup_filter.go [hunk 3]
```diff
@@ -72,70 +97,109 @@ func (f *NodeGroupFilter) ApplyExistingFilter(stackManager *manager.StackCollect
...
 	}
 
-	f.existing.Insert(existing...)
+	f.ExcludeNames.Insert(existing...)
 
+	for _, name := range existing {
+		isAlsoIncluded := false
+		f.matchGlobs(name, f.includeGlobs, &isAlsoIncluded)
+		if isAlsoIncluded {
+			return fmt.Errorf("existing nodegroup %q should be excluded, but matches include fileter: %q", name, f.includeGlobsSpec)
```
martina-if: typo: fileter -> filter

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -72,70 +97,109 @@ func (f *NodeGroupFilter) ApplyExistingFilter(stackManager *manager.StackCollect
...
 	}
 
-	for _, compiledExpr := range f.only {
-		if compiledExpr.Match(ng.Name) {
-			return true // return first match
+	if hasExcludeRules {
+		exclude := f.ExcludeNames.Has(name)
+		f.matchGlobs(name, f.excludeGlobs, &exclude)
+		if exclude && !mustInclude {
+			// no inclusion rules to override
```
martina-if: If the exclude is supposed to override the include, then shouldn't the `if` be like `if exclude { return false }`?

martina-if: Maybe the comment is outdated?

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -72,70 +97,109 @@ func (f *NodeGroupFilter) ApplyExistingFilter(stackManager *manager.StackCollect
...
 }
 
-// MatchAll checks all nodegroups against the filter and returns all of
-// matching names as set
-func (f *NodeGroupFilter) MatchAll(cfg *api.ClusterConfig) sets.String {
-	names := sets.NewString()
-	if f.SkipAll {
-		return names
+// MatchAll nodegroups against the filter and return two sets of names
+func (f *NodeGroupFilter) MatchAll(nodeGroups []*api.NodeGroup) (sets.String, sets.String) {
```
martina-if: We can make this signature a bit more explicit by naming the return values like this:
```func ...MatchAll(...) (included sets.String, excluded sets.String) {```
Also then we can just do `return` instead of `return included, excluded`

Also adding to the doc that the first set is the included and the second the excluded ones.

File: pkg/ctl/cmdutils/nodegroup_filter_test.go [hunk 6]
```diff
@@ -14,9 +14,122 @@ import (
...
+		It("should match exclude filter with ExcludeAll", func() {
+			filter.ExcludeAll = true
+			included, excluded := filter.MatchAll(cfg.NodeGroups)
+			Expect(included).To(HaveLen(0))
+			Expect(excluded).To(HaveLen(6))
+			Expect(excluded.HasAll("test-ng1a", "test-ng2a", "test-ng3a", "test-ng1b", "test-ng2b", "test-ng3b")).To(BeTrue())
+		})
+
+		It("should match exclude filter with names and globs", func() {
+			filter.ExcludeNames.Insert("test-ng3b")
```
martina-if: now that I see how this is used, maybe this insert should be exposed as a method too like `filter.AddExcludedName(name)`. That way we don't leak an impl detail (the set, which later could be an array or it could be made thread safe if needed by adding a mutex, etc)

pkg/ctl/cmdutils/nodegroup_filter.go
```diff
@@ -16,54 +16,79 @@ import (
...
-	IgnoreAllExisting bool
+	ExcludeAll bool // highest priority
 
-	existing sets.String
-	only     []glob.Glob
-	onlySpec string
+	// include filters take presidence
+	IncludeNames     sets.String
+	includeGlobs     []glob.Glob
+	includeGlobsSpec string
```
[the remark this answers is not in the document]
errordeveloper: We need to store raw glob expression to print error message and log info about how the filter got applied...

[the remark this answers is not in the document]
errordeveloper: I've rename it to `rawIncludeGlobs` and `rawExcludeGlobs`, and extended the info logger to make better use of this.

pkg/ctl/cmdutils/nodegroup_filter.go
errordeveloper: It just saves a few extra lines and a variable, e.g.:
```Go
mustInclude = f.includeNames.Has(name)
f.matchGlobs(name, f.includeGlobs, &mustInclude)
```

pkg/ctl/cmdutils/nodegroup_filter.go
errordeveloper: Include is supposed to override, via `mustInclude`... comment is somewhat correct, but isn't very clearly phrased.

pkg/ctl/cmdutils/nodegroup_filter.go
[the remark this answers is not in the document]
martina-if: ah, so this is exactly what the user provided?

pkg/ctl/cmdutils/nodegroup_filter.go
martina-if: Hm, I think this is a confusing API. Also is easier to read code when there's less state changes (so variables, when possible are assigned as fewer times as possible, typically just once).

errordeveloper: I agree, but it's just sooo many ifs... Will change.

pkg/ctl/cmdutils/nodegroup_filter.go
errordeveloper: Named returns are evil... I'll update the comment.

pkg/ctl/cmdutils/nodegroup_filter_test.go
errordeveloper: Yeah, I was just lazy, thanks.

pkg/ctl/cmdutils/nodegroup_filter.go
martina-if: Ahaaa, I had a suspicion that they could be dangerous but because I saw them in effective go I thought they were recommended. Never mind then

martina-if:
These --include and --exclude look a lot nicer :+1:

PR #727, opened 2019-04-12
Replace `--dry-run=false` with `--approve`

errordeveloper:
### Description

Having to set `--dry-run=false` is a little unintuitive.
This change deprecates `--dry-run` flag from `eksctl update cluster`
command with new `--approve` flag. The new flag is also added in the
commands that already had most of the plumbing in place to support it.

This also fixes an issue with `cmd.PreRun`.

### Checklist

### DIFF
--- a/pkg/addons/default/aws_node.go
+++ b/pkg/addons/default/aws_node.go
@@ -23,33 +23,33 @@ const (
 )
 
 // UpdateAWSNode will update the `aws-node` add-on
-func UpdateAWSNode(rawClient kubernetes.RawClientInterface, region string) error {
+func UpdateAWSNode(rawClient kubernetes.RawClientInterface, region string, plan bool) (bool, error) {
 	_, err := rawClient.ClientSet().AppsV1().DaemonSets(metav1.NamespaceSystem).Get(AWSNode, metav1.GetOptions{})
 	if err != nil {
 		if apierrs.IsNotFound(err) {
 			logger.Warning("%q was not found", AWSNode)
-			return nil
+			return false, nil
 		}
-		return errors.Wrapf(err, "getting %q", AWSNode)
+		return false, errors.Wrapf(err, "getting %q", AWSNode)
 	}
 
 	// if DaemonSets is present, go through our list of assets
 	list, err := LoadAsset(AWSNode, "yaml")
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	for _, rawObj := range list.Items {
 		resource, err := rawClient.NewRawResource(rawObj)
 		if err != nil {
-			return err
+			return false, err
 		}
 		if resource.GVK.Kind == "DaemonSet" {
 			image := &resource.Info.Object.(*appsv1.DaemonSet).Spec.Template.Spec.Containers[0].Image
 			imageParts := strings.Split(*image, ":")
 
 			if len(imageParts) != 2 {
-				return fmt.Errorf("unexpected image format %q for %q", *image, KubeProxy)
+				return false, fmt.Errorf("unexpected image format %q for %q", *image, KubeProxy)
 			}
 
 			if strings.HasPrefix(imageParts[0], awsNodeImagePrefix) &&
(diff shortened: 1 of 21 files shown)


pkg/addons/default/aws_node.go
```diff
@@ -58,13 +58,18 @@ func UpdateAWSNode(rawClient kubernetes.RawClientInterface, region string) error
...
 		if err != nil {
-			return err
+			return false, err
 		}
 		logger.Info(status)
 	}
 
+	if plan {
+		logger.Critical("(plan) %q is not up-to-date", AWSNode)
+		return true, nil
```
martina-if: What is this boolean that the function returns?

pkg/addons/default/coredns.go
```diff
@@ -145,68 +147,73 @@ func InstallCoreDNS(rawClient kubernetes.RawClientInterface, region string, wait
 		}
 		watcher.Stop()
 		if timeout {
-			return fmt.Errorf("timed out (after %s) waiting for %q pods to become ready", waitTimeout, CoreDNS)
+			return false, fmt.Errorf("timed out (after %s) waiting for %q pods to become ready", waitTimeout, CoreDNS)
 		}
 	}
 
+	if plan {
```
martina-if: I would move this to line 108 so it's out of the way as soon as possible

pkg/addons/default/kube_proxy.go
```diff
@@ -20,55 +20,55 @@ const (
...
+		return false, fmt.Errorf("unexpected image format %q for %q", *image, KubeProxy)
 	}
 
 	desiredTag := "v" + controlPlaneVersion
 
 	if imageParts[1] == desiredTag {
 		logger.Debug("imageParts = %v, desiredTag = %s", imageParts, desiredTag)
 		logger.Info("%q is already up-to-date", KubeProxy)
-		return nil
+		return false, nil
```
martina-if: Seems like the return value is completely dependent of the input argument `plan`. I would remove the return boolean. It's a bit confusing.

pkg/ctl/utils/update_kube_proxy.go
```diff
@@ -75,5 +75,12 @@ func doUpdateKubeProxy(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg st
 		return err
 	}
 
-	return defaultaddons.UpdateKubeProxyImageTag(clientSet, cprv, false)
+	updateRequired, err := defaultaddons.UpdateKubeProxyImageTag(clientSet, cprv, plan)
+	if err != nil {
+		return err
+	}
+
+	cmdutils.LogPlanModeWarning(plan && updateRequired)
```
martina-if: I think we would get cleaner APIs if we remove the return boolean from `UpdateKubeProxyImageTag()`, `UpdateCoreDNSImageTag()`, etc. and instead, just move the call to `cmdutils.LogPlanModeWarning()` inside those functions instead

martina-if:
:+1: I think this will look nice

pkg/addons/default/aws_node.go
errordeveloper: it's to indicate pending changes, so that on the outside we can tell the user whether they want to run with `--approve` or it'd make no difference.

pkg/addons/default/coredns.go
errordeveloper: The idea is that we actually run up to the point where we have done everything, except for updating, so that we can see if all read calls will succeed...

pkg/addons/default/kube_proxy.go
errordeveloper: Yes, almost.

The idea is to print this
```
foo already up-to-date
```

vs this:
```
foo needs an update
re-run with '--approve'
```
A simple solution would be to print that line right away here, but that will break API/CLI boundary. So the idea is that the logic that decides whether to print `re-run with '--approve'` or not must be outside of this function. returning a boolean value may not be ideal, but I didn't want to over-engineer this part.
There is a boolean `plan` argument, and a boolean return value on whether approved plan would make any changes or not, that seems plausible to me. If error value matching was more flexible, perhaps we could have used the error value somehow, but Go is unhelpful with that.

pkg/ctl/utils/update_kube_proxy.go
errordeveloper: Yes, but see my comment above... And more specifically - we should avoid calling any `cmdutils` function from a library function. That package is for CLI contexts only.

Issue #535, opened 2019-02-12
get rid of eksctl.cluster.k8s.io/v1alpha1/ tags


errordeveloper:
We should use new tags, that would be consistent with labels, i.e. use `alpha.eksctl.io/`. We do need to keep it backwards compatible, of course.

PR #728, opened 2019-04-12
Get rid of eksctl.cluster.k8s.io/v1alpha1/ tags

martina-if:
closes https://github.com/weaveworks/eksctl/issues/535
Rename tags from eksctl.cluster.k8s.io/v1alpha1/ to alpha.eksctl.io/ keeping backwards compatibility

- [x] test manually


errordeveloper:
This is good to go! I've opened #731 to discuss whether tags should be upgrade during cluster upgrades, but that is indeed a separate topic!

PR #733, opened 2019-04-15
adding EKS Account ID for security measures when using  for AMIs

### LINKED ISSUE #732: Contain AMI Search Based on EKS Account ID for AL2 images
This could lead to EKS clusters being created with non AWS owned AMIs.

Current search could include AMIs not AWS owned but made public using the same tags with a more recent created at, the request is to add an additional check for Image searching using `auto` and the `make generate-ami` list to only be AWS owned. *Ubuntu images can be from any account for now*

christopherhein:
Signed-off-by: Christopher Hein <me@chrishein.com>

### Description
closes #732 

This checks the right Account ID for AL2 AMIs so that they always come from the EKS team.

### Checklist

### DIFF
--- a/pkg/ami/api.go
+++ b/pkg/ami/api.go
@@ -39,6 +39,12 @@ var ImageClasses = []string{
 	"ImageClassGPU",
 }
 
+// ImageFamilyToAccountID is a map of image families to account Ids
+var ImageFamilyToAccountID = map[string]string {
+	ImageFamilyAmazonLinux2: "602401143452",
+	ImageFamilyUbuntu1804: "099720109477",
+}
+
 // IsAvailable checks if a given AMI ID is available in AWS EC2
 func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 	input := &ec2.DescribeImagesInput{
@@ -60,8 +66,9 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 // FindImage will get the AMI to use for the EKS nodes by querying AWS EC2 API.
 // It will only look for images with a status of available and it will pick the
 // image with the newest creation date.
-func FindImage(api ec2iface.EC2API, namePattern string) (string, error) {
+func FindImage(api ec2iface.EC2API, namePattern string, imageFamily string) (string, error) {
 	input := &ec2.DescribeImagesInput{
+		Owners: []*string{aws.String(ImageFamilyToAccountID[imageFamily])},
 		Filters: []*ec2.Filter{
 			{
 				Name:   aws.String("name"),
--- a/pkg/ami/api.go
+++ b/pkg/ami/api.go
@@ -39,16 +39,10 @@ var ImageClasses = []string{
 	"ImageClassGPU",
 }
 
-// ImageFamilyToAccountID is a map of image families to account Ids
-var ImageFamilyToAccountID = map[string]string {
-	ImageFamilyAmazonLinux2: "602401143452",
-	ImageFamilyUbuntu1804: "099720109477",
-}
-
 // IsAvailable checks if a given AMI ID is available in AWS EC2
 func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 	input := &ec2.DescribeImagesInput{
-		ImageIds: []*string{aws.String(id)},
+		ImageIds: []*string{&id},
 	}
 
 	output, err := api.DescribeImages(input)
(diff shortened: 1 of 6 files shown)


pkg/ami/api.go
```diff
@@ -86,6 +86,10 @@ func FindImage(api ec2iface.EC2API, namePattern string) (string, error) {
 		},
 	}
 
+	if imageFamily == "ImageFamilyAmazonLinux2" {
```
errordeveloper: I think we should match family to owner elsewhere, perhaps we can look it up from another map. But from the standpoint if this function, it's unaware of families of what now, so we just want to pass the owner to it.

errordeveloper:
It would be also great to not spoil the coverage, this package has pretty good coverage and we should keep it that way. So please take a look if you can add a meaningful test for this!

pkg/ami/api.go
errordeveloper: Both owner IDs are documented here already:

https://github.com/weaveworks/eksctl/blob/93cc5eb09b06a7b60ae6cba123fafd82c001be71/pkg/ami/api.go#L17-L21

Please use the Ubuntu one also, and after you have a lookup map, remove those comments.

christopherhein:
Once tests pass let me know what you think @errordeveloper

christopherhein:
@errordeveloper everything looks good now on this if you get a second to merge.

martina-if:
LGTM

nrdlngr:
Any update on this?

pkg/ami/api.go
```diff
@@ -60,8 +66,9 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 // FindImage will get the AMI to use for the EKS nodes by querying AWS EC2 API.
 // It will only look for images with a status of available and it will pick the
 // image with the newest creation date.
-func FindImage(api ec2iface.EC2API, namePattern string) (string, error) {
+func FindImage(api ec2iface.EC2API, namePattern string, imageFamily string) (string, error) {
 	input := &ec2.DescribeImagesInput{
+		Owners: []*string{aws.String(ImageFamilyToAccountID[imageFamily])},
```
errordeveloper: This a little too optimistic, we should check if the key exists, so we don't panic if it doesn't.

pkg/ami/api.go
```diff
@@ -60,8 +66,9 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 // FindImage will get the AMI to use for the EKS nodes by querying AWS EC2 API.
 // It will only look for images with a status of available and it will pick the
 // image with the newest creation date.
-func FindImage(api ec2iface.EC2API, namePattern string) (string, error) {
+func FindImage(api ec2iface.EC2API, namePattern string, imageFamily string) (string, error) {
```
errordeveloper: I think map lookup should be all in the same place, let's just make it like this here:
```suggestion
func FindImage(api ec2iface.EC2API, namePattern, accountID string) (string, error) {
```

pkg/ami/api.go
errordeveloper:
```suggestion
		Owners: []*string{&accountID},
```

errordeveloper: Please move the look-up  into the caller, and the map declaration should live in the same file also:

https://github.com/weaveworks/eksctl/blob/3d20b573fcc5b495ba6729ba4affdc25a285639a/pkg/ami/auto_resolver.go#L50

errordeveloper:
We should be able to include this in the next release, but due to undergoing API changes, we cannot cut it before all of [milestone issues](https://github.com/weaveworks/eksctl/milestone/11) are closed, as we are half-way-through.

errordeveloper:
@christopherhein I'll have to finish this up myself, as we are looking to cut 0.1.30 today.

pkg/ami/auto_resolver.go
```diff
@@ -50,17 +56,23 @@ type AutoResolver struct {
...
 		if !ok {
 			logger.Critical("image family %s doesn't support GPU image class", imageFamily)
 			return "", NewErrFailedResolution(region, version, instanceType, imageFamily)
 		}
 	}
 
-	id, err := FindImage(r.api, p, imageFamily)
+	ownerAccount, knownOwner := ImageFamilyToAccountID[imageFamily]
+	if !knownOwner {
+		logger.Critical("unable to determin account owner for image family %s", imageFamily)
```
martina-if: typo: ~determin~ determine

martina-if: Also: "determine the account owner..." might read better

errordeveloper: 👍

Issue #731, opened 2019-04-15
Should we upgrade tags on cluster upgrade?


errordeveloper:
We have introduce new tags with #728, but after upgrading a cluster old tags remain. This is somewhat deliberate as a general feature, yet we might want to upgrade tag schema as well.

cdenneen:
@errordeveloper @martina-if Not sure if this is best place for it or new issue but we need ability to update cluster/nodegroup/fargateprofile/etc for changes.
Now while replacing NG and Fargate profiles is recommended for those it shouldn't be required for updates to things like Tags or Labels
Same with the cluster resources (stacks, ControlPlane, SG, etc) if you update the tags for all these resources let's say you create all of these with something like:

```
tags:
  Support Team: teamemail@domain.com
  Service: Testing
  CostCenter: 1234567
```

But then you need to change email, costcenter, etc... I shouldn't have to create a new cluster in order to update the tags.

How can this be accomplished? Maybe some sort of eksctl util function?

martina-if:
Hi @cdenneen , this seems like a good use case. Can you open a new issue for this?

PR #748, opened 2019-04-23
Update cluster name tag on cluster upgrade

martina-if:
### Description

closes https://github.com/weaveworks/eksctl/issues/731#event-2276053753

When a cluster is upgraded the tag schema should also be upgraded.

I'm not sure this is the right way to do it or where to call this from exactly. 

### Checklist


errordeveloper:
I'm not sure if this is important at the moment, let's revisit it later on.

Issue #756, opened 2019-04-27
panic on deletion


errordeveloper:
```
 [0] >> eksctl delete cluster scrumptious-rainbow-1556285024   
[ℹ]  using region us-west-2
[ℹ]  deleting EKS cluster "scrumptious-rainbow-1556285024"
[✔]  kubeconfig has been updated
[ℹ]  1 task: { delete cluster control plane "scrumptious-rainbow-1556285024" [async] }
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x1ffe2ca]

goroutine 30 [running]:
github.com/weaveworks/eksctl/pkg/cfn/manager.(*StackCollection).DeleteStackBySpec(0xc000b551a0, 0x0, 0x0, 0x45, 0xc0000fd710)
	/go/src/github.com/weaveworks/eksctl/pkg/cfn/manager/api.go:270 +0x3a
github.com/weaveworks/eksctl/pkg/cfn/manager.(*asyncTaskWithStackSpec).Do(0xc000950220, 0xc000098540, 0xc000098540, 0x1)
	/go/src/github.com/weaveworks/eksctl/pkg/cfn/manager/tasks.go:176 +0x33
github.com/weaveworks/eksctl/pkg/cfn/manager.doSingleTask(0xc0000984e0, 0x3b8a020, 0xc000950220)
	/go/src/github.com/weaveworks/eksctl/pkg/cfn/manager/tasks.go:185 +0xec
github.com/weaveworks/eksctl/pkg/cfn/manager.doSequentialTasks(0xc0000984e0, 0xc00029a0a0, 0x1, 0x1)
	/go/src/github.com/weaveworks/eksctl/pkg/cfn/manager/tasks.go:212 +0x50
created by github.com/weaveworks/eksctl/pkg/cfn/manager.(*TaskTree).DoAllSync
	/go/src/github.com/weaveworks/eksctl/pkg/cfn/manager/tasks.go:115 +0x246
 [2] >> eksctl version                                       
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.29"}
 [0] >>
```
I think this could be to do with some of the nodegroups being an weird state, but nevertheless it needs looking into. Looks like it's fixed in master, but I cannot think of any recent change in this area.

errordeveloper:
We do need to handle nil properly here, but also there is another issue.

[...]

PR #758, opened 2019-04-29
Add old cluster and ng name tags back for forward comp.

martina-if:
related to https://github.com/weaveworks/eksctl/issues/756

In https://github.com/weaveworks/eksctl/pull/728 the tags used for the cluster
and nodegroup names where renamed.  This PR makes sure both new and old tags
are added to new clusters. We will keep this behavior for a while until we
deprecate the old tags.

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -51,6 +51,7 @@ func newTag(key, value string) *cloudformation.Tag {
 func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) *StackCollection {
 	tags := []*cloudformation.Tag{
 		newTag(api.ClusterNameTag, spec.Metadata.Name),
+		newTag(api.OldClusterNameTag, spec.Metadata.Name),
 	}
 	for key, value := range spec.Metadata.Tags {
 		tags = append(tags, newTag(key, value))
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -371,6 +371,10 @@ func fmtStacksRegexForCluster(name string) string {
 	return fmt.Sprintf(ourStackRegexFmt, name)
 }
 
+func (c *StackCollection) errStackNotFound() error {
+	return fmt.Errorf("no eksctl-managed CloudFormation stacks found for %q", c.spec.Metadata.Name)
+}
+
 // DescribeStacks describes the existing stacks
 func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
 	stacks, err := c.ListStacks(fmtStacksRegexForCluster(c.spec.Metadata.Name))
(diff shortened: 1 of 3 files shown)


errordeveloper:
Do you want to improve nil handling as part of this also?

I was thinking to add this:
```patch
diff --git a/pkg/cfn/manager/api.go b/pkg/cfn/manager/api.go
index a50b541e..782034b0 100644
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -315,6 +315,9 @@ func (c *StackCollection) DeleteStackByNameSync(name string, errs chan error) er
 
 // DeleteStackBySpec sends a request to delete the stack
 func (c *StackCollection) DeleteStackBySpec(s *Stack) (*Stack, error) {
+       if s == nil {
+               return nil, fmt.Errorf("cannot delete nil stack")
+       }
        for _, tag := range s.Tags {
                if matchesClusterName(*tag.Key, *tag.Value, c.spec.Metadata.Name) {
                        input := &cloudformation.DeleteStackInput{
```

But perhaps it'd be better to do this:
```patch
diff --git a/pkg/cfn/manager/api.go b/pkg/cfn/manager/api.go
index a50b541e..e31e137f 100644
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -370,6 +370,10 @@ func fmtStacksRegexForCluster(name string) string {
        return fmt.Sprintf(ourStackRegexFmt, name)
 }
 
+func (c *StackCollection) errStackNotFound() error {
+       return fmt.Errorf("no eksctl-managed CloudFormation stacks found for %q", c.spec.Metadata.Name)
+}
+
 // DescribeStacks describes the existing stacks
 func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
        stacks, err := c.ListStacks(fmtStacksRegexForCluster(c.spec.Metadata.Name))
@@ -377,7 +381,7 @@ func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
                return nil, errors.Wrapf(err, "describing CloudFormation stacks for %q", c.spec.Metadata.Name)
        }
        if len(stacks) == 0 {
-               return nil, fmt.Errorf("no eksctl-managed CloudFormation stacks found for %q", c.spec.Metadata.Name)
+               return nil, c.errStackNotFound()
        }
        return stacks, nil
 }
diff --git a/pkg/cfn/manager/cluster.go b/pkg/cfn/manager/cluster.go
index 04d5e438..2e2c8d3c 100644
--- a/pkg/cfn/manager/cluster.go
+++ b/pkg/cfn/manager/cluster.go
@@ -53,7 +53,7 @@ func (c *StackCollection) DescribeClusterStack() (*Stack, error) {
                        return s, nil
                }
        }
-       return nil, nil
+       return nil, c.errStackNotFound()
 }
 
 // AppendNewClusterStackResource will update cluster
```

What do you think?

Issue #774, opened 2019-05-02
Dynamically lookup DeviceName for AMIs


adamjohnson01:
**Why do you want this feature?**
We are using a custom ami which uses /dev/sda1 for the root device, currently there is no way for us extend the root volume of the ami as the launch configuration will add another disk using /dev/xvda 

**What feature/behavior/change do you want?**
We would like the ability to override the DeviceName value using a value in the config file

errordeveloper:
Would you mind sharing more info about the kind AMI that depends on this? Is it an in-house thing for you, or one from the marketplace or elsewhere?

adamjohnson01:
No problem, the official Redhat and Centos AMIs use /dev/sda1 for the root device. We build our images on top of those.

errordeveloper:
That sounds relatively mainstream to me, perhaps we can add image families
for those and set device name based on the family, wdyt?

adamjohnson01:
Just to clarify, when I say official Redhat/Centos I am referring to the base images, there are no eks images available, we build and maintain our own private eks images.

I think rather than hard coding values it may be a better idea to a populate the value dynamically, the information can be retrieved by describing the ami that is going to be used, this would make it a lot more flexible and remove the need to override anything.

errordeveloper:
That is a very good point indeed. Let's focus on retrieving it, rather then
hardcoding.
We may still want to add more parameters later, but I think the issue at
hand is that there is a hardcoded value now that needs to be dynamically
resolved based on AMI used.

Beside that, I think we should open another issue to discuss RH image
family, even if there is no public AMI, I think we should discuss whether a
family should be defined instead of falling back on built-in AL2 family
definition.

adamjohnson01:
Cool, I will close my other pull request and work on getting the value dynamically. 

I think that having a RH family makes sense, happy to discuss it in another issue.

Issue #717, opened 2019-04-09
timing of nodegroup delete/scale-down


errordeveloper:
We often see errors like this in the integration tests:
```
• Failure [75.887 seconds]
(Integration) Create, Get, Scale & Delete
/go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:27
  when creating a cluster with 1 node
  /go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:53
    and add the second nodegroup
    /go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:135
      and delete the second nodegroup
      /go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:223
        should make it 4 nodes total [It]
        /go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:233

        Expected
            <int>: 7
        to equal
            <int>: 4

        /go/src/github.com/weaveworks/eksctl/integration/creategetdelete_test.go:245
```
The nodes eventually go away.

We could probably fix this by waiting longer, but perhaps we could make nodes go away quicker also, e.g. right now we don't delete node  object and let expire, we probably should delete it right after CloudFomration stack gets deleted. See also #536.

errordeveloper:
I've increased the timeout in the test from 5m to 10m, but seems that it still occurs.

PR #775, opened 2019-05-02
Override the launchTemplateData DeviceName for custom AMIs

adamjohnson01:
### Description
Added a config option to allow overriding of the launchTemplateData DeviceName. This will allow custom amis that do not have a root device of /dev/xvda to have the root volume extended.

Issue #774

### Checklist

### DIFF
--- a/examples/08-custom-ami.yaml
+++ b/examples/08-custom-ami.yaml
@@ -1,0 +1,17 @@
+# An example of ClusterConfig showing how to use a custom ami that is not Amazon Linux or Ubuntu that has had its volume size increased.
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: cluster-8
+  region: eu-west-1
+
+nodeGroups:
+  - name: ng-1
+    instanceType: m5.large
+    desiredCapacity: 1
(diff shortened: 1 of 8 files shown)


errordeveloper:
Hi @adamjohnson01! The code looks good to me, but I do have a few concerns about this going forward... Perhaps we should expand the support for EBS volumes a little more, and move it to a struct or an array, what are your thoughts?

adamjohnson01:
@errordeveloper I agree, creating a struct to represent a disk and moving volumetype, volume size and the new device name would be great. This would then allow passing a device array in to config and having multiple volumes. However I'd say it's better to add a separate issue for that to the backlog as implementing it wouldn't be backward compatible. This allows support for non-standard device names without breaking existing usage patterns and is a great stop-gap.

errordeveloper:
We could try and keep `volumeSize` and `volumeType` for basic uses-case, and add another field that maps more closely to what CloudFormation offers. It's certainly a good idea to move discussion to another issue, for sure!

adamjohnson01:
@errordeveloper, I am hitting issue #717 on on of the integration tests, all of the others pass. Can I mark this as All integration tests passing?

adamjohnson01:
Will work on getting the values dynamically as discussed in issue #774

PR #773, opened 2019-05-02
Fixes #772: Adds extraCIDRs to VPC

tiffanyfay:
### Description
Fixes #772: uses extraCIDRs in config file and adds the CIDR blocks to the VPC in CloudFormation.

@errordeveloper should I add it to flags too?

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [ ] Added tests that cover your change (if possible) @errordeveloper where would it go?
- [x] All unit tests passing (i.e. `make test`)
- [x] All integration tests passing (i.e. `make integration-test`)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory) @errordeveloper any particular doc this should go in? Would be cool if we had a blank template of all options that are editable
- [x] Added yourself to the `humans.txt` file

PR #776, opened 2019-05-02
Allocate IPv6 blocks for all subnets

### LINKED ISSUE #566: Create ipv6 subnets for created VPC
To be able to create dualstack load balancers in the created VPC (Our api needs to be accessible via ipv6). I am currently adding the IPv6 CIDR's to the subnets manually but would like this process automated with eksctl.

Options for adding IPv6 CIDR's to vpc/subnets and route table setup for ipv6 to the internet gateway.

errordeveloper:
### Description

Fixes #566.

### Checklist

### DIFF
--- a/examples/02-custom-vpc-cidr-no-nodes.yaml
+++ b/examples/02-custom-vpc-cidr-no-nodes.yaml
@@ -1,5 +1,6 @@
-# An example of ClusterConfig object with custom VPC CIDR,
-# and without any nodegroups:
+# An example of ClusterConfig object with custom VPC IPv4 CIDR,
+# and auto-allocated IPv6 CDIRs for all subnets; also without
+# any nodegroups:
 --- 
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
@@ -8,6 +9,8 @@ metadata:
   name: cluster-2
   region: eu-north-1
 
-vpc: {cidr: 10.10.0.0/16}
+vpc:
+  cidr: 10.10.0.0/16
+  autoAllocateIPv6: true
 
 nodeGroups: []
--- a/pkg/cfn/builder/vpc.go
+++ b/pkg/cfn/builder/vpc.go
@@ -11,6 +11,17 @@ import (
 )
 
 func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTopology, subnets map[string]api.Network) {
+	var subnetIndexForIPv6 int
+	if api.IsEnabled(c.spec.VPC.AutoAllocateIPv6) {
+		// this is same kind of indexing we have in vpc.SetSubnets
+		switch topology {
+		case api.SubnetTopologyPrivate:
+			subnetIndexForIPv6 = len(c.spec.AvailabilityZones)
+		case api.SubnetTopologyPublic:
+			subnetIndexForIPv6 = 0
+		}
+	}
+
 	for az, subnet := range subnets {
 		alias := string(topology) + strings.ToUpper(strings.Join(strings.Split(az, "-"), ""))
 		subnet := &gfn.AWSEC2Subnet{
@@ -18,6 +29,7 @@ func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTop
 			CidrBlock:        gfn.NewString(subnet.CIDR.String()),
 			VpcId:            c.vpc,
 		}
+
 		switch topology {
 		case api.SubnetTopologyPrivate:
 			subnet.Tags = []gfn.Tag{{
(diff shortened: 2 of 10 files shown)


VincentCouzij:
Feature is working as expected during my testing. The ipv6 CIDR are added to the subnets and the route table also includes the ipv6 outgoing rule.

pkg/cfn/builder/vpc.go
```diff
@@ -35,20 +42,52 @@ func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTop
...
+)
+
+func getCIDRv6(n int) *gfn.Value {
+	return gfn.MakeFnSelect(n, refCIDRv6)
+}
+
 //nolint:interfacer
 func (c *ClusterResourceSet) addResourcesForVPC() {
 	internetCIDR := gfn.NewString("0.0.0.0/0")
+	//internetCIDRv6 := gfn.NewString("::/0")
```
martina-if: remember to remove the comment before merging (and the others about DestinationIpv6CidrBlock)

pkg/vpc/vpc.go
```diff
@@ -31,10 +31,16 @@ func SetSubnets(spec *api.ClusterConfig) error {
...
 	}
-	prefix, _ := spec.VPC.CIDR.Mask.Size()
-	if (prefix < 16) || (prefix > 24) {
-		return fmt.Errorf("VPC CIDR prefix must be betwee /16 and /24")
+
+	cidrs := append(spec.VPC.ExtraCIDRs, vpc.CIDR)
+	for _, cidr := range cidrs {
+		prefix, _ := cidr.Mask.Size()
+		if (prefix < 16) || (prefix > 24) {
+			return fmt.Errorf("VPC CIDR (%v) prefix must be between /16 and /24", cidr)
```
martina-if: Should this happen during the validation instead?

errordeveloper: This slipped in from #773.

examples/02-custom-vpc-cidr-no-nodes.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
   name: cluster-2
   region: eu-north-1
 
-vpc: {cidr: 10.10.0.0/16}
+vpc:
+  cidr: 10.10.0.0/16
+  autoAllocatedIPv6: true
```
errordeveloper: Not sure about the name yet, maybe just `autoIPv6`, but that reads as if it'd enable full IPv6 thing...

pkg/cfn/builder/vpc.go
```diff
@@ -35,20 +42,52 @@ func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTop
...
+			SubnetId:      refSubnet,
+			Ipv6CidrBlock: getCIDRv6(counter),
+		})
+
 		c.subnets[topology] = append(c.subnets[topology], refSubnet)
+		counter++
 	}
 }
 
+var refCIDRv6 = gfn.MakeFnCIDR(
```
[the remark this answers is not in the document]
errordeveloper: it's just for readability... I'll move it inside the function.

examples/02-custom-vpc-cidr-no-nodes.yaml
errordeveloper: Maybe be `autoCIDRv6`?

martina-if: I think `autoAllocatedIPv6` is clear and not too long

martina-if: or `autoAllocateIPv6` ?

pkg/cfn/builder/vpc.go
```diff
@@ -35,10 +41,29 @@ func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTop
...
+				Ipv6CidrBlock: getCIDRv6(counter),
+			})
+		}
+
 		c.subnets[topology] = append(c.subnets[topology], refSubnet)
+		counter++
 	}
 }
 
+func getCIDRv6(n int) *gfn.Value {
```
martina-if: What is `n` here? Can we rename it to something more explicit?

martina-if: This is a bit obscure if you don't know what those `MakeFn...()` do. Can you add a comment of what it returns? :/

errordeveloper: It slices the AWS-provided IPv6 CIDR and returns Nth element... comment would be good, for sure.

martina-if: How about `selectIPv6Element(index int)`? Or something like that?

errordeveloper: Let me try a different thing, I'm not sure how useful this function is actually, I might just inline this in the other function.

errordeveloper:
I added a whole bunch of new unit tests for the cluster stack, it's not complete yet, but at least it now goes as far as testing key aspects of the cluster stack builder.

https://github.com/weaveworks/eksctl/pull/776/files#diff-d5202f6261e88644b8b1715122f47c4c

pkg/cfn/builder/vpc.go
```diff
@@ -35,6 +47,26 @@ func (c *ClusterResourceSet) addSubnets(refRT *gfn.Value, topology api.SubnetTop
...
 			RouteTableId: refRT,
 		})
+
+		if api.IsEnabled(c.spec.VPC.AutoAllocateIPv6) {
+			// get 8 of /64 subnets from the auto-allocated IPv6 block,
+			// and pick one block based on subnetIndexForIPv6 counter;
+			// NOTE: this is done inside of CloudFormation using Fn::Cidr,
+			// we don't slice it here, just construct the JSON expression
+			// that does slicing at runtime.
+			refAutoAllocateCIDRv6 := gfn.MakeFnSelect(
```
martina-if: Since this is inside a big loop and these are quite low level details, can we encapsulate lines 57-66 into a function like `addCIDR6Selector(subnetIndex)`

pkg/cfn/builder/vpc.go
martina-if: Oh, I just saw it. I think it was better here, with the same comment as the one you added inlined. Mainly, because if it's an implementation detail of CF it's nicer to abstract it a little and give it a name.

pkg/cfn/builder/vpc.go
errordeveloper: Well, the puzzling thing is that it's all a bit tied into how the outer function works, don't you think? I'd much rather take all of the conditional IPv6 bits out, but that asks for more refactoring...

errordeveloper: I've tried, but it turns more complicated for what its worth... I'd rather live it for later.

PR #779, opened 2019-05-03
Dynamically lookup the RootDeviceName of the AMI being used

### LINKED ISSUE #774: Dynamically lookup DeviceName for AMIs

adamjohnson01:
### Description

The cfn/builder/nodegroup.go was using a hardcoded value for DeviceName. This meant it was not possible to manage the size of the root volume for AMIs that did not use /dev/xvda as the RootDeviceName. I have added a lookup function that will retrieve the value from the AMI being used and populate the value in the launchTemplateData.BlockDeviceMappings. 

Closes #774

### Checklist


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -395,6 +395,8 @@ type NodeGroup struct {
 	// +optional
 	VolumeType string `json:"volumeType"`
 	// +optional
+	RootDevice string `json:"rootDevice,omitempty"`
```
errordeveloper: I think we should probably consider something like `rootVolume.{name,type,size}` going forward, but for the time being, let's keep it consistent with existing fields, so perhaps `volumeName` would be most suitable.

pkg/ami/api.go
```diff
@@ -57,6 +57,20 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
...
+	input := &ec2.DescribeImagesInput{
+		ImageIds: []*string{&id},
+	}
+
+	output, err := api.DescribeImages(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to determine root device name for %q", id)
+	}
+
+	return *output.Images[0].RootDeviceName, nil
```
errordeveloper: We should check for lenght of the list, and return an error if it's not strictly one item. Also,  isn't it entirely possible for `RootDeviceName` to be unset, e.g. when the AMI doesn't actually have any particular requirements about the root volume?

pkg/ctl/create/cluster.go
```diff
@@ -253,6 +253,12 @@ func doCreateCluster(p *api.ProviderConfig, cfg *api.ClusterConfig, nameArg stri
 		}
 		logger.Info("nodegroup %q will use %q [%s/%s]", ng.Name, ng.AMI, ng.AMIFamily, cfg.Metadata.Version)
 
+		// Lookup AMI root device name
+		if err := ctl.GetRootDevice(ng); err != nil {
```
errordeveloper: Let's just roll it up into `ctl.EnsureAMI`.

pkg/apis/eksctl.io/v1alpha5/types.go
adamjohnson01: Cool, sounds good to me.

pkg/ami/api.go
adamjohnson01: I agree that we should also check the length of the list. 

With regards to the RootDeviceName , EBS AMIs will always have RootDeviceName set but  instance-store AMIs will not. I will need to test but I think that the current launch config template will not work with instance-store AMIs if the VolumeSize is set in the config. I think that perhaps we need to check that the AMI is an EBS AMI(the vast majority of AMIs are EBS), what do you think?

pkg/ctl/create/cluster.go
adamjohnson01: Will do!

pkg/ami/api.go
adamjohnson01: I have added the length check

pkg/ami/api.go
```diff
@@ -57,6 +57,24 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
...
+		ImageIds: []*string{&id},
+	}
+
+	output, err := api.DescribeImages(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to determine root device name for %q", id)
+	}
+
+	if len(output.Images) < 1 {
+		return "", errors.Wrapf(err, "unable to find %q", id)
```
martina-if: Perhaps make the log message more explicit: "unable to find root device %q"

pkg/eks/api.go
```diff
@@ -257,6 +257,14 @@ func (c *ClusterProvider) EnsureAMI(version string, ng *api.NodeGroup) error {
 		return ami.NewErrNotFound(ng.AMI)
 	}
 
+	// Get Root Device Name
+	rootDevice, err := ami.LookupRootDeviceName(c.Provider.EC2(), ng.AMI)
+	if err != nil {
+		return errors.Wrapf(err, "%s failed root device name lookup", ng.AMI)
```
martina-if: I think that since `ami.LookupRootDeviceName()` returns a specific error message here we could just return that error without an extra message. 

@errordeveloper what do you think?

pkg/ami/api.go
adamjohnson01: We have not even checked the root device yet, this error will happen if no AMI data is returned at all. If I change it to be root device specific would it not be a bit misleading?

martina-if: Ah, right, so this is the id of the image? then how about `unable to find image %q`?

pkg/ami/api.go
```diff
@@ -57,6 +57,24 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
 	return *output.Images[0].State == "available", nil
 }
 
+// LookupRootDeviceName gets the root block device mapping
+func LookupRootDeviceName(api ec2iface.EC2API, id string) (string, error) {
```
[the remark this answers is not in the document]
errordeveloper: Yeah, we do use `id` in other functions, but here the function name is about device name, so it's not clear if `id` refers to a  device ID of some sort or something else...

pkg/eks/api.go
errordeveloper: I'd say `unable to get root device name for image %q`.

pkg/ami/api.go
```diff
@@ -57,6 +57,24 @@ func IsAvailable(api ec2iface.EC2API, id string) (bool, error) {
...
+	input := &ec2.DescribeImagesInput{
+		ImageIds: []*string{&id},
+	}
+
+	output, err := api.DescribeImages(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to determine root device name for %q", id)
+	}
+
+	if len(output.Images) < 1 {
```
errordeveloper:
```suggestion
	if len(output.Images) != 1 {
```

errordeveloper: We'd need a check for when it's 0 as well, or use `!= 1` (but will need to adjust the error message too).

adamjohnson01:
@errordeveloper as the ami.IsAvailable function can also provide the info that I am looking up in my function, how about merging them, something like this?

`func IsAvailable(api ec2iface.EC2API, id string) (bool, string, string, error) {
	input := &ec2.DescribeImagesInput{
		ImageIds: []*string{&id},
	}

	output, err := api.DescribeImages(input)
	if err != nil {
		return false, "", "", errors.Wrapf(err, "unable to find %q", id)
	}

	if len(output.Images) < 1 {
		return false, "", "", nil
	}

	return *output.Images[0].State == "available", *output.Images[0].RootDeviceName, *output.Images[0].RootDeviceType, nil
}`

We could then check if the type is instance-store and error if VolumeSize has also been set. 
Something like this

`	// Check the AMI is available
	available, devName, devType, err := ami.IsAvailable(c.Provider.EC2(), ng.AMI)
	if err != nil {
		return errors.Wrapf(err, "%s is not available", ng.AMI)
	}

	if !available {
		return ami.NewErrNotFound(ng.AMI)
	}

	if devType == "instance-store" && ng.VolumeSize > 0 {
		return errors.Wrap(err, "Cannot set Volume Size for instance-store AMIs")
	}

	if devName == "ebs" {
		ng.VolumeName = devName
	}`

pkg/ami/api.go
errordeveloper: I suppose that with just the single image ID search parameter, we will only ever get one or none.

errordeveloper:
Thanks, Adam! This is looking good.

pkg/ami/api.go
errordeveloper: >  I think that perhaps we need to check that the AMI is an EBS AMI(the vast majority of AMIs are EBS), what do you think?

Yes, good idea!

pkg/eks/api.go
```diff
@@ -257,6 +257,14 @@ func (c *ClusterProvider) EnsureAMI(version string, ng *api.NodeGroup) error {
...
+	}
+
+	ng.VolumeName = rootDevice
```
errordeveloper: Actually, one should be able to override this by setting `volumeName` (if they must). So please make this lookup optional. You'd want to use `*string` for the field type instead of `string`, and use `if api.IsSetAndNonEmptyString(ng.VolumeName)` here.

errordeveloper: Also, we will need validation of the combination of all 3 fields being set in a sane way, take a look at `pkg/apis/eksctl.io/v1alpha5/validation.go`, e.g. `volumeName` or `volumeType` cannot be set without `volumeSize`, but `volumeSize` alone is okay. The preference is to use pointers for all fields, so do feel free to convert `volumeSize` to `*int` and `volumeType` to `*string` for consistency.

adamjohnson01: > Actually, one should be able to override this by setting `volumeName` (if they must). So please make this lookup optional. You'd want to use `*string` for the field type instead of `string`, and use `if api.IsSetAndNonEmptyString(ng.VolumeName)` here.

I am not sure why you would want to override something that is being looked up. Is that so you can add an extra disk to the node instead of increasing the size of the root volume?

pkg/ami/api.go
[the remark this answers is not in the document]
adamjohnson01: I have now merged this function into the IsAvailable function.

pkg/ami/api.go
adamjohnson01: Would you like me to add the comment to the IsAvailable function as they are now merged.

errordeveloper: Yeah, please do!

pkg/eks/api.go
errordeveloper: > I am not sure why you would want to override something that is being looked up.

So that you can avoid the look-up. Some users prefer to avoid extra APIs calls due to access restrictions. I don't remember, do we make the `DescribeImages` call unconditionally when we call `IsAvailable`, or there is a way to avoid that right now? If we call it unconditionally now, then don't worry too much.

pkg/ami/api.go
adamjohnson01: I have added the comment

pkg/ami/api.go
adamjohnson01: This has been merged into the IsAvailable function.

pkg/eks/api.go
adamjohnson01: This has been merged into IsAvailable.

adamjohnson01:
Please let me know if there is anything else I need to do.

errordeveloper:
@adamjohnson01 do you want to make sure all of your commits show up with your profile image? This usually comes down to email address being use in the commit.

adamjohnson01:
@errordeveloper, I have added the email in my commits to my profile now. Thanks!

pkg/eks/api.go
```diff
@@ -248,7 +248,7 @@ func (c *ClusterProvider) EnsureAMI(version string, ng *api.NodeGroup) error {
 	}
 
 	// Check the AMI is available
-	available, err := ami.IsAvailable(c.Provider.EC2(), ng.AMI)
+	available, devName, devType, err := ami.IsAvailable(c.Provider.EC2(), ng.AMI)
```
errordeveloper: Since this is the only place `ami.IsAvailable` is called, how about renaming it and moving everything there?
So here we will have `return ami.Use(c.Provider.EC2(), ng)`, and let that function populate all the fields in `ng` or produce whatever errors it may?

adamjohnson01: I have now done this.

adamjohnson01:
> Just one last suggestion! :)

No problem, that is done now.

Issue #198, opened 2018-09-10
Provide an option to use spot instances


yhvh:
**Why do you want this feature?**
We are using kubernetes for data science. Since this load is non-critical we would like to use spot instances for worker nodes.

**What feature/behavior/change do you want?**
An option to deploy spot instances instead of on-demand instances

errordeveloper:
I think this should part of node group management epic, along the side with #100. I would rather avoid creating an ad-hoc flag(s) in `eksctl create cluster` before we have `eksctl create nodegroup`.

danielchalef:
I'll second this request: We're using k8 with [polyaxon](polyhttps://github.com/polyaxon/polyaxontores) for ML model experimentation, development, and management. We typically launch a large number of nodes for short periods of time (e.g. for hyperparameter tuning jobs, large training jobs). These nodes are often relatively expensive (high-mem, high-core count, and/or GPUs) but not mission-critical. These workloads are ideal for spot instance usage.

alando46:
+1 EC2spot support would be amazing

whereisaaron:
@errordeveloper can we leverage Launch Templates to provide options like spot instances? That could cover a lot of other requests I see in issues. E.g. I would like to be able to create node pools of nodes that have extra volumes attached. I could provide `eksctl` with a Launch Template.

Anything specified in the `eksctl` config or command line options would override the template, but otherwise everything from extra tags, extra volumes, tenancy, spot instance choices, T2/T3-unlimited option, placement groups etc. could come from the Launch Template, rather than duplicating all these options within `eksctl`?

voxxit:
What is the status on this? We are currently having to resort to _manually_ updating our node groups to support spot fleets using the following tutorial, and it is less than optimal to say the least...

https://eksworkshop.com/spotworkers/workers/

errordeveloper:
Josh, what would be simplest thing that would help to begin with?

voxxit:
@errordeveloper my team and I found that it would be simplest to add the `SpotPrice` option to the `AWS::EC2::LaunchConfiguration` that is generated for node groups. I think a flag on node groups for the maximum spot price would suffice (in fact, I think that is what kops does?)

Longer term: it would be best to use a launch template which supports spot fleets so we can get around service limits/capacity issues for certain instance types, etc.

romil-punetha:
@voxxit Just adding the spot price to the launch configuration doesn't work for me. Are there any other changes to be made in the launch configuration? 
[...]

PR #781, opened 2019-05-07
WIP Add support for spot instances

martina-if:
This is the first step of #198

It adds support in the API to specify a maximum price and uses that to create a nodegroup of spot instances.

These nodegroups will only have spot instances, the next step would be to support 
ngs with both on demand and spot instances.

- [ ] Added PR to update documentation as required


martina-if:
Closed in favor of https://github.com/weaveworks/eksctl/pull/794

Issue #148, opened 2018-08-07
eksctl utils ssh


errordeveloper:
It'd be handy to provide an SSH helper, as right now it's not very trial to jump into a node, as you have to remember kubectl command, or look in EC2 console (or use some other command). Most of the time I just want to SSH to any of the nodes.

errordeveloper:
@karinnainiguez did you intend to unassign this?

errordeveloper:
The intention would be to provide debugging utility to those who are using a custom AMI, or debugging eksctl. To start with, we probably want to SSH into the host, but in the future we may consider SSH into a privileged toolbox container, but there is no particular use for that at the moment.

As far as debugging use-case is concerned, we should be looking at EC2 API for node IPs, because one should be able to debug a node that hasn't joined the cluster. I'd guess default mode should be to SSH into any node, and specific node name should be an option also, but not required.

karinnainiguez:
@errordeveloper thanks for checking! It was intentional, my coworker @kschumy is researching this one. 
:)

errordeveloper:
@gotjosh is willing to take this one!

thapakazi:
Anyone working on this ??

errordeveloper:
@thapakazi not as far as I'm aware, please feel free to have a go! I'm happy to chat on Slack, if you need help finding your way around the codebase.

thapakazi:
Great, I would love to do it :heart_eyes: 
I will ping you on slack as needby. Thanks  :)

thapakazi:
I have started working on this. I am thinking of:
1. generating ssh config based on the tags cluster nodes has
2. use the config from step1 && wrap around *ssh* binary to enter into nodes for 
`eksctl utils ssh node...`
<Details>
<summary>Example WIP</summary>

```
(master)⚡ % go run ./cmd/eksctl/*.go utils list-nodes                                             ~/go/src/github.com/thapakazi/eksctl
[ℹ]  listing the worker nodes...


Host myeks-ng-96d7c0d7-Node
        hostname 192.168.103.219
        User ec2-user
        Port 22

Host myeks-ng-96d7c0d7-Node
        hostname 192.168.167.172
        User ec2-user
        Port 22
```
</Details>

<Details>
<summary>Sample Example HELP</summary>

```
$ go run ./cmd/eksctl/*.go utils list-nodes --help
List the nodes in the cluster in ssh config format

Usage: eksctl utils list-nodes [flags]

Common flags:
[...]

PR #787, opened 2019-05-09
WIP: easy ssh with eksctl utils issue#148

thapakazi:
_**This PR is work in progress.. I will update it accordingly.**_ 

### Description
This PR is to support easy ssh access to the nodes via eksctl utils.  

Initial thread: https://github.com/weaveworks/eksctl/issues/148
### Design: :speaking_head: : 
I would like to break this into:
  1. generating ssh config into a file `~/.eksctl/cluster/ssh.config` based on the tag used in cluster nodes 
```bash
eksctl utils ssh list nodes                # list the nodes
eksctl utils ssh list nodes --update  # or -U similar to update the ssh inventory
```
Sample:<Details>
<summary>Example WIP</summary>

```
(master)⚡ % go run ./cmd/eksctl/*.go utils list nodes                                             ~/go/src/github.com/thapakazi/eksctl
[ℹ]  listing the worker nodes...


Host myeks-ng-96d7c0d7-Node
        hostname 192.168.103.219
        User ec2-user
        Port 22

Host myeks-ng-96d7c0d7-Node
        hostname 192.168.167.172
        User ec2-user
        Port 22
```
</Details>

  2. use the config from step1 && wrap around ssh binary to enter into nodes for
```bash
eksctl utils ssh node_from_list_above...
```
I have built a [library](https://github.com/thapakazi/easyssh-go/blob/master/library/tags_2_ssh_config.go) (a wrapper around aws, nothing fancy) to fetch instance info based on custom tags. I am thinking of re-using it here. Easy Alternatives...?

Suggestion Please...

### Checklist
- [ ] yet to build... idea validation in progress

### Stakes
@errordeveloper

### DIFF
--- a/pkg/ctl/utils/ssh.go
+++ b/pkg/ctl/utils/ssh.go
@@ -1,0 +1,44 @@
+package utils
+
+import (
+	"os"
+
+	"github.com/kris-nova/logger"
+	"github.com/spf13/cobra"
+	"github.com/thapakazi/easyssh-go/library"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+func listNodesCmd(g *cmdutils.Grouping) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list-nodes",
+		Short: "List the nodes in the cluster in ssh config format",
+		Run: func(_ *cobra.Command, args []string) {
+			logger.Info("listing the worker nodes...")
+			generateSSHConfig()
+		},
+	}
+	return cmd
+}
+
+// generateSSHConfig populates the ssh config based on cluster tags
+func generateSSHConfig() {
+
+	// TODO: discover these base on cluster nodes
+	username := "ec2-user"
+	port := "22"
+	var tags = make(map[string][]string)
+	tags["tag:eksctl.cluster.k8s.io/v1alpha1/cluster-name"] = []string{"myeks"}
+	tags["instance-state-name"] = []string{"running", "pending"}
+
+	response, err := library.FetchIps(tags)
+	if err != nil {
+		logger.Critical("Error fetching ips: %q", err.Error())
+		os.Exit(-1)
+	}
+	err = library.GenerateConfig(username, port, response, os.Stdout)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


martina-if:
Hi @thapakazi , thanks a lot for the PR! I'm not sure what @errordeveloper thinks but I am thinking that if we plan to add a command like `eksctl utils ssh...` it might be worth not changing the user's ssh config? I think eksctl is used often for experiments and then we could end up leaving that file full of useless configuration over time. 

What do you think about adding `eksctl utils ssh ...` without touching the .ssh/config file (calling ssh using normal flags), and perhaps add another option like `eksctl utils ssh-write-config`?

PaulMaddox:
It might be nicer to use SSM Session Manager, rather than wrapping the SSH binary. You could then initiate a console session via either the AWS Management Console, or via the awscli (which eksctl could wrap). 

https://cloudonaut.io/goodbye-ssh-use-aws-session-manager-instead/

https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-sessions-start.html#sessions-start-cli

This has some nice benefits:
- Easy support for node groups that sit in private subnets, rather than working with bastion hosts etc
- Better auditing/permission model  
- No need to modify ssh_config, or depend on the user having an ssh binary present

thapakazi:
Hmmm, getting way into ec2 instances via SSM sounds dope. 
Also, my worries about jumping to private instances are no more.

Thanks for sharing the links. I see my friend using this solution in the past and worry less of ssh inventory. I will also implement this util with the SSM and see which fit best to this problem.

Thanks :bowing_man:

errordeveloper:
> Easy support for node groups that sit in private subnets, rather than working with bastion hosts etc

I didn't realise SSM supports this, that's quite nice!

errordeveloper:
I'm going to close this, as there is consensus that SSM is better way to go long term.

whereisaaron:
SSM just added ssh/scp tunneling, so even fewer reasons not to use SSM.

https://aws.amazon.com/about-aws/whats-new/2019/07/session-manager-launches-tunneling-support-for-ssh-and-scp/

PR #794, opened 2019-05-10
Add support for mixed ASG with spot instances

### LINKED ISSUE #198: Provide an option to use spot instances

### LINKED ISSUE #320: Add support for ASG Multi Instance Types & Purchase Options
AWS recently released EC2 Auto Scaling Groups with multiple instance types & Purchase options:
https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

It would be great if these new options could be configured on the worker node autoscaling group through the eksctl tool. 

This would make it easy to get started with spot instance on the EKS cluster.

martina-if:
closes #198
closes #320

It adds support in the API to specify a maximum price and uses that to create a nodegroup of spot instances.

Nodegroups can now have different kinds of instances: on-demand or spot, and different instance types.

- [x] Merge https://github.com/weaveworks/eksctl/pull/802 first
- [x]  Has been manually tested


pkg/cfn/builder/nodegroup.go
```diff
@@ -196,41 +184,134 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
...
+
+	return nil
+}
+
+// GetAllOutputs collects all outputs of the node group
+func (n *NodeGroupResourceSet) GetAllOutputs(stack cfn.Stack) error {
+	return n.rs.GetAllOutputs(stack)
+}
+
+func launchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_LaunchTemplateData {
```
errordeveloper: let's call it `newLaunchTemplateData`, as it constructs the data and returns a reference

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -489,4 +491,21 @@ type (
 		// +optional
 		PublicKeyName *string `json:"publicKeyName,omitempty"`
 	}
+
+	// MixedInstances holds the configuration for spot instances
+	// TODO: validation of base capcity : make required
+	// TODO allocation strategy?
+
+	MixedInstances struct {
```
errordeveloper:
```suggestion
	NodeGroupInstancesDistribution struct {
```
So it's in line with e.g. `NodeGroupSSH`.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -378,6 +378,8 @@ type NodeGroup struct {
 	AMIFamily string `json:"amiFamily,omitempty"`
 	// +optional
 	InstanceType string `json:"instanceType,omitempty"`
+	//+optional
+	SpotInstances *MixedInstances `json:"spotInstances,omitempty"`
```
errordeveloper:
```suggestion
	InstancesDistribution *NodeGroupInstancesDistribution `json:"instancesDistribution,omitempty"`
```

docs/08-spot-instances.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+    name: martina-test-spotinst-mixed-8
+    region: eu-central-1
+
+nodeGroups:
+    - name: ng-8
+      instanceType: t2.small
```
errordeveloper: I think when `instancesDistribution` is set, `instanceType` should be unset, we could also allow a special value (`mixed`) that can be used for explicitness.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -489,4 +491,21 @@ type (
 		// +optional
 		PublicKeyName *string `json:"publicKeyName,omitempty"`
 	}
+
+	// MixedInstances holds the configuration for spot instances
+	// TODO: validation of base capcity : make required
```
errordeveloper: would be good to validate ranges of each of the numeric values 👍

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -489,4 +491,21 @@ type (
...
+	// TODO allocation strategy?
```
errordeveloper: there are a few fields that only have one value, we don't need to include those right now

martina-if: :+1: I didn't include those

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -489,4 +491,18 @@ type (
...
+	// NodeGroupInstancesDistribution holds the configuration for spot instances
+	NodeGroupInstancesDistribution struct {
+		//+required
+		InstanceTypes []string `json:"instanceTypes,omitEmpty"`
+		// +optional
+		MaxPrice *float64 `json:"maxPrice,omitempty"`
+		//+optional
+		OnDemandBaseCapacity *int `json:"onDemandBaseCapacity,omitEmpty"`
+		//+optional
+		PercentageAboveBase *int `json:"percentageAboveBase,omitEmpty"`
```
Jeffwan: Why not use original term in ASG. like `OnDemandPercentageAboveBaseCapacity `

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -146,6 +146,50 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
...
+
+	if distribution.OnDemandBaseCapacity != nil && *distribution.OnDemandBaseCapacity < 0 {
+		return fmt.Errorf("onDemandBaseCapacity should be 0 or more")
+	}
+
+	if distribution.PercentageAboveBase != nil && (*distribution.PercentageAboveBase < 0 || *distribution.PercentageAboveBase > 100) {
+		return fmt.Errorf("percentageAboveBase should be between 0 and 100")
+	}
+
+	if distribution.SpotInstancePools != nil && (*distribution.SpotInstancePools < 1 || *distribution.SpotInstancePools > 20) {
```
Jeffwan: API level, SpotInstancePools allows a value greater than `len(allInstanceTypes)`. I am not sure if it makes sense to scope this value here

pkg/cfn/builder/nodegroup.go
```diff
@@ -196,41 +184,115 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
...
 	}
-	n.newResource("NodeGroup", &awsCloudFormationResource{
+
+	return &awsCloudFormationResource{
 		Type:       "AWS::AutoScaling::AutoScalingGroup",
 		Properties: ngProps,
 		UpdatePolicy: map[string]map[string]string{
 			"AutoScalingRollingUpdate": {
-				"MinInstancesInService": "1",
+				"MinInstancesInService": "0",
```
Jeffwan: Any reason 0 is used here?

docs/08-spot-instances.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+nodeGroups:
+    - name: ng-6
+      ssh:
+          publicKeyPath: ~/.ssh/id_rsa_tests.pub
+      minSize: 2
+      maxSize: 5
+      mixedInstances:
+        maxPrice: 0.017
+        instanceTypes: ["t3.small", "t3.medium", "t3.small"] # Required because we need more instance types
```
Jeffwan: two t3.small?

Jeffwan:
This feature would be really helpful! Thanks!

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: I think we try to keep the names short in the API. In this case, I'm torn because it could be mistaken for the spot percentage. Let's see what @errordeveloper thinks.

pkg/apis/eksctl.io/v1alpha5/validation.go
martina-if: I added this check because I saw that in the documentation it says that the range is 1-20
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-instancesdistribution.html#cfn-autoscaling-autoscalinggroup-instancesdistribution-spotinstancepools
Should it be different?

pkg/cfn/builder/nodegroup.go
martina-if: This is what I've seen in some examples when people were using pure spot instance groups. I'm not sure what's best here. What do you think?

docs/08-spot-instances.yaml
martina-if: Thanks! This is from the last test I made. Will fix.

martina-if:
@Jeffwan thanks for the review!

pkg/apis/eksctl.io/v1alpha5/validation.go
Jeffwan: I think this could be an issue on AWS sdk side, if eksctl want to check following case. 
If I use two different instances, I can still have SpotInstancePools > 2 configured here. Do you think this is an issue?

pkg/cfn/builder/nodegroup.go
Jeffwan: 0-downtime should be the goal. Seems you are right, 0 should be used because it's impossible for AWS to guarantee a minimum number of instances to be kept in service during the rolling update.

voxxit: I think I was one of the people who suggested this. Yes, this is *only* to be used for 100% spot instance node groups using launch _*configurations*_ (not templates) because CloudFormation does not support anything other than this config when updating LCs.

pkg/apis/eksctl.io/v1alpha5/validation.go
martina-if: I'm not sure I completely understand you, but I think it should be ok because a pool consists of a combination of an instance type and an AZ, so there can be more pools than instance types. 
Here I am only checking that the number is between 1-20 regardless of the number of instance types. Is that wrong?

Jeffwan: No. It's correct, API doesn't check it neither.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -152,6 +152,50 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
...
 	}
 
+	if err := validateInstancesDistribution(ng); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TODO What happens with the other instance type. Should we use it as well, ignore it, or fail when both are specified?
```
errordeveloper: I think "fail when both are specified" is the way to go.

pkg/apis/eksctl.io/v1alpha5/types.go
errordeveloper: As far as I recall, we borrowed the short name from kops, but it would be a good idea to match the ASG API as this is 1-1 mapping.

pkg/apis/eksctl.io/v1alpha5/validation.go
errordeveloper: Ah, I can see this was addressed. Let's remove the TODOs.

examples/08-spot-instances.yaml
```diff
@@ -0,0 +1,22 @@
+# A simple example of ClusterConfig object:
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+    name: martina-test-spotinst-mixed-9
```
errordeveloper: this should be `cluster-8` to match other examples...

examples/08-spot-instances.yaml
```diff
@@ -0,0 +1,22 @@
+# A simple example of ClusterConfig object:
```
errordeveloper: Please update this comment to say what this is...

examples/08-spot-instances.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+nodeGroups:
+    - name: ng-6
+      ssh:
+          publicKeyPath: ~/.ssh/id_rsa_tests.pub
+      minSize: 2
+      maxSize: 5
+      instancesDistribution:
+        maxPrice: 0.017
+        instanceTypes: ["t3.small", "t3.medium"] # Required because we need more instance types
```
errordeveloper: Should we say something more specific, e.g. "at least to instance types are required"?

examples/08-spot-instances.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+    name: martina-test-spotinst-mixed-9
+    region: eu-central-1
+
+nodeGroups:
+    - name: ng-6
+      ssh:
```
errordeveloper: I'd say it's better to keep the example focused on `instancesDistribution`

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -46,9 +46,98 @@ var _ = Describe("ConfigFile ssh flags validation", func() {
...
+			Expect(err).To(HaveOccurred())
+
+			ng.InstanceType = "mixed"
+			ng.InstancesDistribution.InstanceTypes = []string{"t3.medium", "t3.small"}
+
+			err = validateInstancesDistribution(ng)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("It fails when the onDemandBaseCapacity is not above 0", func() {
```
ruecarlo: Really cosmetic one. Test description is a bit misleading. Code checks actually for <0. 0 is a supported value for onDemandBaseCapacity.

martina-if: Thanks, I'll fix it.

Issue #791, opened 2019-05-10
Allow the use of existing IAM profile and enable cluster autoscaling for nodegroup


adamjohnson01:
**Why do you want this feature?**
We would like to use existing IAM roles for our nodegroups as well as enabling autoscaling on the nodegroups.

**What feature/behavior/change do you want?**
Currently the only way to enable a nodegroup to be managed by the autoscaler is to set

```
iam:
  withAddonPolicies:
    autoScaler: true
```

This gives the nodegroup IAM role unnecessary permissions. The permissions set are only required if the nodegroup is also running the cluster autoscaler pod and setting this along with iam.instanceProfileARN currently causes an error.

I would like to add and additional config option for autoscaling which would allow autoscaling of the nodegroup without also giving it the IAM permissions required to perform the autoscaling. Something like this.

```
autoScalerEnabled: true
```
I also think that it may be better to use this option to manage the autoscaling and remove the functionality from the the code that runs when iam.withAddonPolicies.autoScaler: true as there could be situations where you do not want the cluster autoscaler to manage the nodegroup where it is running. 

I am happy to work on this.

errordeveloper:
Yes, `iam.withAddonPolicies.autoScaler` is indeed for cluster autoscaler. Could you please clarify what would you be looking to use instead? Is it an external autoscaler of some sort?

adamjohnson01:
> Yes, `iam.withAddonPolicies.autoScaler` is indeed for cluster autoscaler. Could you please clarify what would you be looking to use instead? Is it an external autoscaler of some sort?

I would still be using the cluster autoscaler, I have opened a pull request to show you what I mean. Apologies if my explanation was not very good.

adamjohnson01:
This can be resolved by setting tags explicitly like this

```
nodeGroups:
- name: ng-1
  tags:
    k8s.io/cluster-autoscaler/enabled: "true"
    k8s.io/cluster-autoscaler/<clusterName>: "owned"
```

PR #796, opened 2019-05-10
WIP: Separate autoscaler IAM permissions and autoscaling group autoscaler tagging

adamjohnson01:
### Description

I have added a new config option called autoScalerEnabled which will be responsible for enabling the nodegroup autoscaling group for cluster autoscaling. This was previously managed by iam.withAddonPolicies.autoScaler but it meant that unnecessary permissions were being given to nodegroups and it made it impossible to use existing IAM roles and enable autoscaling at the same time.

Issue #791 

### Checklist

### DIFF
--- a/examples/05-advanced-nodegroups.yaml
+++ b/examples/05-advanced-nodegroups.yaml
@@ -12,13 +12,14 @@ nodeGroups:
     instanceType: m5.xlarge
     minSize: 2
     maxSize: 8
+    autoScalerEnabled: true # This configures the nodegroup so that it can be managed by the cluster autoscaler
     volumeSize: 100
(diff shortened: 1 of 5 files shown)


errordeveloper:
> it made it impossible to use existing IAM roles and enable autoscaling at the same time

Did you try setting tags explicitly?
```YAML
nodeGroups:
- name: ng-1
   tags:
     k8s.io/cluster-autoscaler/enabled: "true"
     k8s.io/cluster-autoscaler/<clusterName>: "owned"
```

adamjohnson01:
I didn't not think of trying that. That should work. Thanks!

PR #802, opened 2019-05-16
Runtime discovery of max pods for an instance type

martina-if:
With the support of mixed instance type ASGs we will need a runtime way to set
the maximum number of pods when lauching the kubelet. This PR achieves this by:

- Packaging and sending the eni_max_pods.txt file as user data. This is so we
 can control the formatting and so that we make sure that the node has the latest 
 version. Otherwise, if the node is based on an old AMI it may not have the newer
 instance types
- Changed the bootrap scripts to determine at runtime the max. number of pods based
 on its instance type using the file described above
- Still make it possible to override the max pods with the parameter --max-pods-per-node

- [x] Manually tested both AL and ubuntu instances


pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,31 @@
-#!/bin/sh -eu
+#!/bin/bash -e
```
mumoshu: 👍 

I agree we should be explicit about what we're actually using.

pkg/nodebootstrap/userdata_ubuntu.go
```diff
@@ -20,22 +20,23 @@ func makeUbuntu1804Config(spec *api.ClusterConfig, ng *api.NodeGroup) (configFil
...
 	files := configFiles{
 		configDir: {
 			"metadata.env": {content: strings.Join(makeMetadata(spec), "\n")},
 			"kubelet.env":  {content: strings.Join(kubeletEnvParams, "\n")},
 			// TODO: https://github.com/weaveworks/eksctl/issues/161
-			"ca.crt":          {content: string(spec.Status.CertificateAuthorityData)},
-			"kubeconfig.yaml": {content: string(clientConfigData)},
+			"ca.crt":           {content: string(spec.Status.CertificateAuthorityData)},
+			"kubeconfig.yaml":  {content: string(clientConfigData)},
+			"max_pods_map.txt": {content: string(maxPodsData)},
```
mumoshu: Note: At first I thought this could potentially be `max_pods.csv`(and making each line separated by `,` instead of a whitetspace) so that we don't need to infer the format of the file.

But that's YAGNI :)

pkg/nodebootstrap/userdata_al2.go
```diff
@@ -24,6 +24,11 @@ func makeAmazonLinux2Config(spec *api.ClusterConfig, ng *api.NodeGroup) (configF
 		return nil, err
 	}
 
+	maxPodsData, err := max_pods_mapTxtBytes()
```
mumoshu: Can we make the `max_pods_map.txt` on provisioning time from the former `pkg/nodebootstrap/maxpods.go`, rather than go-generating it as a txt file?

The rationale is that I have a potential(very sorry but not raised yet) feature that would allow us to tune the "ratio" of max pods per instance type, e.g. `1/2 of default MAX_PODS`. Setting the uniform max pods via configs won't work for mixed instance types.

Leaving `pkg/nodebootstrap/maxpods.go` as-is allows both making this PR smaller, while the code required to implement the potential feature minimum.

WDYT?

martina-if: Done :) The only issue is that the file we are generating does not have a deterministic order. Maybe it would be good to sort the entries but not sure if it's important. @errordeveloper what do you think?

martina-if:
Thanks for the review @mumoshu !

pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,31 @@
...
 
-echo "NODE_IP=$(curl --silent http://169.254.169.254/latest/meta-data/local-ipv4)" > /etc/eksctl/kubelet.local.env
-echo "INSTANCE_ID=$(curl --silent http://169.254.169.254/latest/meta-data/instance-id)" >> /etc/eksctl/kubelet.local.env
+function get_max_pods() {
+  MAX_PODS_FILE="/etc/eksctl/max_pods_map.txt"
+  grep "${INSTANCE_TYPE}" "${MAX_PODS_FILE}" | while read instance_type pods; do
```
errordeveloper:
```suggestion
  "${MAX_PODS_FILE}" > while read instance_type pods; do
```

pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,31 @@
...
+
+    if  [[ "${pods}" =~ ^[0-9]+$ ]] ; then
```
errordeveloper: Let's just add `"${instance_type}" == "${1}"` here...

pkg/nodebootstrap/assets/bootstrap.al2.sh
errordeveloper: Once we switched to bash here, let's get rid of `-e` and use more readable syntax:
```
set -o errexit
set -o pipefail
set -o nounset
```

pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,31 @@
...
+function get_max_pods() {
+  MAX_PODS_FILE="/etc/eksctl/max_pods_map.txt"
+  while read instance_type pods; do
+
+    if  [[ "${instance_type}" == "${1}" ]] && [[ "${pods}" =~ ^[0-9]+$ ]] ; then
+      echo ${pods};
+      return
+    fi ;
+
+    done < "${MAX_PODS_FILE}"
```
errordeveloper: Should we drop indentation here... Also, not quite sure about having `<` at the end here, rather then at the top.

pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,31 @@
...
+
+NODE_IP=$(curl --silent http://169.254.169.254/latest/meta-data/local-ipv4)
+INSTANCE_ID=$(curl --silent http://169.254.169.254/latest/meta-data/instance-id)
+INSTANCE_TYPE=$(curl --silent http://169.254.169.254/latest/meta-data/instance-type)
+
+source /etc/eksctl/kubelet.env
+
+MAX_PODS=${MAX_PODS:-$(get_max_pods ${INSTANCE_TYPE})}
+
+set -o nounset
```
errordeveloper: I see, so this needs to be here because of `MAX_PODS` above... let's find a different way to handle this.

errordeveloper: e.g. if `MAX_PODS` is set - user it, call the function otherwise. The problem with `set -o` in the middle of a script it is that it's too easy to forget about it, so it's best to just do it at the top.

pkg/nodebootstrap/assets/bootstrap.al2.sh
martina-if: Will fix the indentation. the `<` has to go there AFAIK. It doesn't work in the top.

pkg/nodebootstrap/assets/bootstrap.al2.sh
martina-if: I don't know how to check if the variable is set without getting an error for `unbound variable` unless I move it there. Maybe there's a way I don't know about?

errordeveloper: There is a way to do it with a [`+` expansion](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02), e.g.:
```
$ env MAX_PODS=456 bash -eu -c 'if [ -z "${MAX_PODS+x}" ] ; then export MAX_PODS=123; fi ; echo $MAX_PODS'   
456
$ bash -eu -c 'if [ -z "${MAX_PODS+x}" ] ; then export MAX_PODS=123; fi ; echo $MAX_PODS'             
123
```

pkg/nodebootstrap/assets/bootstrap.al2.sh
errordeveloper: Yes, looks like this is the canonical way to do it :)

pkg/nodebootstrap/assets/bootstrap.al2.sh
martina-if: I think this looks worse and more error prone than having the set where it is now. I'd rather keep it how it is or perhaps remove the option of passing the MAX_PODS argument since we don't use it or need it anyway

errordeveloper: I don't mind if it is ugly... It's bash, I find that reliable scripts do get ugly.

We do need to pass it in, as if one wants to use e.g. Weave Net, they need bypass this logic and force a fixed value.

errordeveloper: I will take a look at this in a bit, I think there is better option...

pkg/nodebootstrap/assets/bootstrap.al2.sh
```diff
@@ -1,7 +1,35 @@
...
+source /etc/eksctl/kubelet.env
+
+if [[ -z "${MAX_PODS+x}" ]];
+  then export MAX_PODS=$(get_max_pods ${INSTANCE_TYPE});
+fi
+
+echo "NODE_IP=${NODE_IP}" > /etc/eksctl/kubelet.local.env
+echo "INSTANCE_ID=${INSTANCE_ID}" >> /etc/eksctl/kubelet.local.env
+echo "INSTANCE_TYPE=${INSTANCE_TYPE}" >> /etc/eksctl/kubelet.local.env
+echo "MAX_PODS=${MAX_PODS}" >> /etc/eksctl/kubelet.local.env
```
errordeveloper: Actually, looking at this once again, I found that `:-` handles unset variable case just fine.

So this should be sufficient:
```suggestion
echo "MAX_PODS=${MAX_PODS:-$(get_max_pods ${INSTANCE_TYPE})}" >> /etc/eksctl/kubelet.local.env
```

Also, at this stage, heredoc syntax would be better:
```
cat > /etc/eksctl/kubelet.local.env <<EOF
NODE_IP=${NODE_IP}
INSTANCE_ID=${INSTANCE_ID}
INSTANCE_TYPE=${INSTANCE_TYPE}
MAX_PODS=${MAX_PODS:-$(get_max_pods ${INSTANCE_TYPE})}
EOF
```

pkg/nodebootstrap/assets/bootstrap.ubuntu.sh
```diff
@@ -18,6 +35,10 @@ systemctl reset-failed
   source /etc/eksctl/kubelet.env
   source /etc/eksctl/metadata.env
 
+  if [[ -z "${MAX_PODS+x}" ]];
```
errordeveloper: Let's move it below as `"max-pods=${MAX_PODS:-$(get_max_pods ${INSTANCE_TYPE})}"`

PR #840, opened 2019-06-05
Add various build improvements

2opremio:
* Install `dep` as part of `make install-build-deps`.
* Honor `$GOBIN` variable.
* Stop assuming `$GOPATH` is defined (since it won't be mandatory with go modules).
* Various minor simplifications and refactorings.

### DIFF
--- a/build/install.sh
+++ b/build/install.sh
@@ -1,6 +1,12 @@
 #!/bin/sh -eu
 
-BIN="$1"
+if [ -z "${GOBIN+x}" ]; then
+ GOBIN="$(go env GOPATH)/bin"
+fi
+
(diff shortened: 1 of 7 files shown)


2opremio:
I need to think about how to reconcile this with `build/Dockerfile`

build/install.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -eu
 
+BIN="$1"
```
martina-if: `$1` ? Or `${GOPATH}/bin/`?

2opremio: I am passing it as a parameter from the makefile, but I need to think about how to make it work from the Dockerfile as well. I asked for a review too early, sorry.

2opremio:
@martina-if PTAL

martina-if:
LGTM!

2opremio:
@errordeveloper want to take a look?

PR #847, opened 2019-06-06
Manage CloudWatch logging configuration

### LINKED ISSUE #704: Add config file options to enable EKS control plane logging to CloudWatch
We want to see the EKS control plane logs for monitoring, troubleshooting, and security audit.

The ability to enable/disable in the config file, each of the [five control plane log types available in EKS](https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html):
- `api`
- `audit`
- `authenticator`
- `controllerManager`
- `scheduler`

Documentation on EKS control plane logging API and CLI commands:
https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html

`aws eks --region us-west-2 update-cluster-config --name prod \
--logging '{"clusterLogging":[{"types":["api","audit","authenticator","controllerManager","scheduler"],"enabled":true}]}'`

Ref: https://github.com/aws/containers-roadmap/issues/26

errordeveloper:
### Description

Closes #704

### Checklist


errordeveloper:
TODOs:
- [ ] make `eks.UpdateClusterConfig` call one of the tasks
- [ ] decide on `eksctl update cluster` behaviour (probably leave for another PR)
- [ ] define behaviour without config file (probably enable all by default, optional flag to disable)

errordeveloper:
The overall shape of this is in place now, I just need to write some tests and tidy up a few things.

pkg/eks/eks.go
```diff
@@ -228,18 +230,110 @@ func (c *ClusterProvider) WaitForControlPlane(id *api.ClusterMeta, clientSet *ku
...
 		Name:    &cfg.Metadata.Name,
 		Version: &cfg.Metadata.Version,
 	}
 	output, err := c.Provider.EKS().UpdateClusterVersion(input)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
-	return *output.Update.Id, nil
+	return output.Update, nil
```
martina-if: It's better to keep our own API's free of third party libraries for testability and maintainability. Can we return the id instead? AFAICT that's the only thing we need

pkg/ctl/utils/enable_logging.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/eks"
+	"github.com/weaveworks/eksctl/pkg/printers"
+)
+
+func enableLoggingCmd(rc *cmdutils.ResourceCmd) {
```
martina-if: Can we add a unit test for this since it has some logic?

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+				Update: &awseks.Update{
+					Id:     aws.String("u123"),
+					Type:   aws.String(awseks.UpdateTypeLoggingUpdate),
+					Status: aws.String(awseks.UpdateStatusSuccessful),
+				},
+			}
+
+			p.MockEKS().On("DescribeUpdateRequest", mock.MatchedBy(func(input *awseks.DescribeUpdateInput) bool {
+				*describeUpdateInput = *input
+				return input.UpdateId != nil && *input.UpdateId == *describeUpdateOutput.Update.Id
```
martina-if: What is this for `*input.UpdateId == *describeUpdateOutput.Update.Id`?

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cfg.EnableLogging).To(BeEmpty())
+		})
+
+
+		It("should handle unknown facilities", func() {
+			cfg.EnableLogging = []string{"anything", "anyOtherThing"}
+
+			err = api.SetClusterConfigDefaults(cfg)
+			Expect(err).To(HaveOccurred())
```
martina-if: Also check that they are kept in `cfg.EnableLogging`

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+		})
+
+		It("should have no logging by default", func() {
+			err = api.SetClusterConfigDefaults(cfg)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cfg.EnableLogging).To(BeEmpty())
+		})
+
+
+		It("should handle unknown facilities", func() {
```
martina-if: This seems like a unit test for the api package, doesn't it?

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+var _ = Describe("EKS API wrapper", func() {
+	Describe("can update cluster configuration for logging", func() {
+		var (
+			ctl *ClusterProvider
+
+			cfg *api.ClusterConfig
+			err error
+
+			clusterLogging []*awseks.LogSetup
```
martina-if: Rename to `sentClusterLogging` or something that indicates that this is the one sent to the AWS API

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+			err = ctl.UpdateClusterConfigForLogging(cfg)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(clusterLogging[0].Types).To(BeEmpty())
+
+			Expect(clusterLogging[1].Types).ToNot(BeEmpty())
+			Expect(clusterLogging[1].Types).To(Equal(aws.StringSlice(api.SupportedLoggingFacilities())))
+		})
+
+		It("should expand '*' to _all_", func() {
```
martina-if: Since this is testing logic in the `api` module I would move it there. Those unit tests will probably be shorter and easier to write.

pkg/eks/update_config_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(clusterLogging[0].Types).ToNot(BeEmpty())
+			Expect(clusterLogging[0].Types).To(Equal(aws.StringSlice(cfg.EnableLogging)))
+
+			Expect(clusterLogging[1].Types).ToNot(BeEmpty())
+			Expect(clusterLogging[1].Types).To(Equal(aws.StringSlice([]string{"api", "audit", "scheduler"})))
+		})
+
+		It("should enable some logging facilities and disable others", func() {
```
martina-if: I'm not sure what value is added by this test.

martina-if:
The word facilities for me means buildings or "easyness". I don't see it fitting well in this context. Should it be logging sources or logging types instead?

D3nn:
I think he means it in this context:

an option or service that gives the opportunity to do or benefit from
something.
"the program includes a help facility and interactive windows"

errordeveloper:
Based on the converstion with @kalbir, I've change the config field to `cloudWatch.clusterLogging.enableTypes`, which will allow us to accommodate for other fields (e.g. #1017).
I've also change `facility` to `logType` now. Please have another look!

pkg/eks/eks.go
```diff
@@ -228,18 +230,110 @@ func (c *ClusterProvider) WaitForControlPlane(id *api.ClusterMeta, clientSet *ku
...
+		}
+	}
+	return enabled, disabled, nil
+}
+
+// UpdateClusterConfigForLogging calls UpdateClusterConfig to enable logging
+func (c *ClusterProvider) UpdateClusterConfigForLogging(cfg *api.ClusterConfig) error {
+	enabled := sets.NewString(cfg.EnableLogging...)
+	disabled := sets.NewString()
+	for _, facility := range api.SupportedLoggingFacilities() {
```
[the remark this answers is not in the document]
errordeveloper: Yes, I just realised this one in another place... will simplify!

pkg/eks/update_config_test.go
errordeveloper: true... also, it's not actually producing that error any more, so I must have remove validation in the process of refactoring.

errordeveloper:
@martina-if @D3nn actually I was thinking of https://en.wikipedia.org/wiki/Syslog#Facility... 😄

errordeveloper:
@martina-if I still have a few TODOs and tests to fix, but could you please take another look? It changed significantly since last time you reviewed it.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -1,5 +1,21 @@
 package v1alpha5
 
+// SupportedLoggingFacilities retuls all supported logging facilities
+func SupportedLoggingFacilities() []string {
```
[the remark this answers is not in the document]
errordeveloper: I renamed this in line with the new config filed names.

pkg/eks/update_config_test.go
errordeveloper: Good point, thanks!

pkg/eks/update_config_test.go
errordeveloper: Will do.

pkg/eks/update_config_test.go
errordeveloper: The function matcher needs to return true or false, I guess it would be easier to just return true here as there is only one call to match.

pkg/eks/eks.go
```diff
@@ -228,18 +230,124 @@ func (c *ClusterProvider) WaitForControlPlane(id *api.ClusterMeta, clientSet *ku
...
+
+// UpdateClusterConfigForLogging calls UpdateClusterConfig to enable logging
+func (c *ClusterProvider) UpdateClusterConfigForLogging(cfg *api.ClusterConfig) error {
+	enabled := sets.NewString()
+	disabled := sets.NewString()
+	if cfg.CloudWatch != nil && cfg.CloudWatch.ClusterLogging != nil {
+		enabled.Insert(cfg.CloudWatch.ClusterLogging.EnableTypes...)
+	}
+
+	for _, logType := range api.SupportedCloudWatchClusterLoggingTypes() {
```
martina-if: This could also be all.Difference(enabled)

pkg/eks/eks.go
```diff
@@ -228,18 +230,124 @@ func (c *ClusterProvider) WaitForControlPlane(id *api.ClusterMeta, clientSet *ku
...
+				},
+			},
+		},
+	}
+
+	output, err := c.Provider.EKS().UpdateClusterConfig(input)
+	if err != nil {
+		return err
+	}
+	if err := c.waitForUpdateToSucceed(cfg.Metadata.Name, output.Update); err != nil {
```
martina-if: This is a bit odd: the updateClusterConfig operation is part of `c.Provider.EKS()` but the functionality that waits for it to complete is in the `c`. I think they should be both part of the same API (either EKS() or the `c`), otherwise it seems that there is not a clear responsibility boundary.

pkg/eks/eks.go
errordeveloper: Thanks, will update.

pkg/eks/eks.go
errordeveloper: `c.Provider.EKS()` is just a copy of the EKS client which `c` wraps. There are various places we need consume some of the clients (e.g. `./pkg/vpc`, `./pkg/iam`), that's why clients are exported and not completely hidden by the wrapper. If you have an idea of how we can clean this up, let's open another issue and discuss it there :)

errordeveloper:
I've added this to the main integration test suite.

pkg/eks/eks.go
```diff
@@ -228,18 +230,110 @@ func (c *ClusterProvider) WaitForControlPlane(id *api.ClusterMeta, clientSet *ku
...
+func (c *ClusterProvider) GetClusterConfigForLogging(cl *api.ClusterMeta) (sets.String, sets.String, error) {
+	enabled := sets.NewString()
+	disabled := sets.NewString()
+
+	cluster, err := c.DescribeControlPlane(cl)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, i := range cluster.Logging.ClusterLogging {
```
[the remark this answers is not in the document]
errordeveloper: 👍

pkg/eks/update_config_test.go
errordeveloper: Yeah, I'll remove it. It was a lazy fuzz test. We should do proper fuzz tests instead.

errordeveloper:
I also added docs, will address the rest of review comments now.

pkg/ctl/utils/enable_logging.go
errordeveloper: We have it covered by integration suite now. Do you insist on unit test as well?

pkg/ctl/utils/enable_logging.go
```diff
@@ -0,0 +1,110 @@
...
+			disableFacilities := sets.NewString()
+			for _, facility := range allSupportedLoggingFacilities {
+				facilityFlag := cmd.Flag(facility)
+
+				shouldEnable := facilityFlag.Value.String() == "true"
+				shouldDisable := facilityFlag.Changed && !shouldEnable
+
+				if shouldEnable {
+					cfg.EnableLogging = append(cfg.EnableLogging, facility)
+					if !cmd.Flag("all").Changed {
```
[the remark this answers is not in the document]
errordeveloper: No, this is to force disabling it when `--all` wasn't given, as it's enabled by default. I'll add a comment.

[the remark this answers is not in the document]
errordeveloper: I've simplified the logic.

pkg/eks/update_config_test.go
errordeveloper: Currently we don't discard bad values, we just return an error and the program will terminate.

errordeveloper:
@martina-if @D3nn I'm happy with the state of this PR now, please have another look.

errordeveloper:
@rndstr if you have a moment, please take a look also.

pkg/eks/update.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+// UpdateClusterConfigTasks returns all tasks for updating cluster configuration
+func (c *ClusterProvider) UpdateClusterConfigTasks(cfg *api.ClusterConfig) *manager.TaskTree {
+	tasks := &manager.TaskTree{Parallel: false}
+
+	tasks.Append(&updateClusterConfigTask{
+		info: "update CloudWatch logging configurtaion",
+		skip: !cfg.HasClusterCloudWatchLogging(),
```
martina-if: I would really like to avoid the added complexity of the skip option to an already complex system of concurrent tasks. I think this is probably simpler to implement as an `if cfg.HasclustercloudWatchLogging()` inside the actual task.

Adding more complexity to the management of concurrent tasks makes it even harder to reason about the code and hard to debug when there are issues.

site/content/usage/12-cloudwatch-cluster-logging.md
```diff
@@ -0,0 +1,98 @@
...
+
+[CloudWatch logging][eksdocs] for EKS control plane is not enabled by default due to data
+ingestion and storage costs.
+
+To enable control plane logging when cluster is created, you will need to define **`cloudWatch.clusterLogging.enableTypes`** setting in your `ClusterConfig` (see below for examples).
+
+So if you have a config file with correct **`cloudWatch.clusterLogging.enableTypes`**
+setting, you can create a cluster with `eksctl create cluster --config-file=<path>`.
+
+If you have create a cluster already, you can use `eksctl utils enable-logging`.
```
martina-if: "have created"

site/content/usage/12-cloudwatch-cluster-logging.md
```diff
@@ -0,0 +1,98 @@
...
+
+```YAML
+cloudWatch:
+  clusterLogging:
+    enableTypes: ["*"]
+```
+
+To disable all types, use `[]` or remove `cloudWatch` section completely.
+
+You can enable subset of types by listing the types you want to enable:
```
martina-if: "enable a subset"

errordeveloper:
> The only thing that I'm quite concerned about is the skip flag in the tasks.

I understand you concern. I couldn't see a better solution that would still provide the information to the user. The idea was that the users will get a chance to find out that logs hadn't been enable by default and will be looking at the docs to find out how to enable them.

pkg/eks/update.go
errordeveloper: Maybe we could just rewriting this functions.

Perhaps it could be something like:

```Go
func (c *ClusterProvider) AppendUpdateClusterConfigTasks(cfg *api.ClusterConfig, tasks *manager.TaskTree) {
   newTasks := &manager.TaskTree{Parallel: false}
   if cfg.HasClusterCloudWatchLogging() {
     tasks.Append(&updateClusterConfigTask{
       info: "update CloudWatch logging configurtaion",
     }
   }  else {
     logger.Info("CloudWatch logging will not be enabled for cluster %q in %q", cfg.Metadata.Name, cfg.Metadata.Region)
   if newTasks.Len() == 0 {
     return
   }
   tasks.Append(newTasks)
}
```

errordeveloper: Keep in mind that we are going to add `UpdateClusterConfigForVPC` soon (for #649), that's why I'm adding new `TaskTree`.

pkg/ctl/utils/enable_logging.go
errordeveloper: I absolutely agree, but I won't have the time this week.

pkg/ctl/utils/enable_logging.go
```diff
@@ -0,0 +1,148 @@
...
+
+	rc.FlagSetGroup.InFlagSet("Enable/disable log types", func(fs *pflag.FlagSet) {
+		enableExplicitly := sets.NewString()
+		allSupportedTypes := api.SupportedCloudWatchClusterLogTypes()
+		inherit := sets.NewString()
+
+		enableAll := fs.Bool("all", true, fmt.Sprintf("Enable all supported log types (%s)", strings.Join(allSupportedTypes, ", ")))
+
+		for _, logType := range allSupportedTypes {
+			_ = fs.Bool(logType, false, fmt.Sprintf("Enable %q log type", logType))
```
cPu1: I find the camel case naming for the `--controllerManager` flag inconsistent with other flags which are hyphen-separated, but I can see it saves us some code.

For consistency, WDYT about handling this differently at the expense of repeating the supported logging options here?

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -1,5 +1,18 @@
 package v1alpha5
 
+// SetClusterConfigDefaults will set defaults for a given cluster
+func SetClusterConfigDefaults(cfg *ClusterConfig) error {
+	if cfg.HasClusterCloudWatchLogging() {
+		for _, logType := range cfg.CloudWatch.ClusterLogging.EnableTypes {
+			switch logType {
+			case "all", "*":
```
cPu1: Since we don't validate other log types here, this makes it possible to pass invalid log types and bypass the validation in `ValidateClusterConfig` as the value is reset to `SupportedCloudWatchClusterLogTypes()` . 

So this would pass validation:
```yaml
enableTypes: ["all", "asdf", "asdfasdf"]
```

I think we should only support "all" or "*" when `enableTypes` is a single-element array.


```golang
func SetClusterConfigDefaults(cfg *ClusterConfig) {
	if cfg.HasClusterCloudWatchLogging() && len(cfg.CloudWatch.ClusterLogging.EnableTypes) == 1 {
		switch cfg.CloudWatch.ClusterLogging.EnableTypes[0] {
		case "all", "*":
			cfg.CloudWatch.ClusterLogging.EnableTypes = SupportedCloudWatchClusterLogTypes()
		}
	}
}
```

pkg/eks/update.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return t.info
+}
+func (t *updateClusterConfigTask) Do(errs chan error) error {
+	err := t.call(t.spec)
+	close(errs)
+	return err
+}
+
+// GetCurrentClusterConfigForLogging fetches current cluster logging configuration as two sets - dsiable and enanaled types
```
cPu1: Minor typo: "... as two sets - disabled and enabled"

pkg/ctl/cmdutils/configfile.go
```diff
@@ -315,3 +317,16 @@ func NewDeleteNodeGroupLoader(rc *ResourceCmd, ng *api.NodeGroup, ngFilter *Node
 
 	return l
 }
+
+// NewUtilsEnableLoggingLoader will laod config or use flags for 'eksctl utils enable-logging'
```
cPu1: Minor typo: laod -> load

pkg/ctl/utils/enable_logging.go
errordeveloper: I did think about this, but didn't want to make it complicated... feel free to implement if you have ideas.

In away I'm also struggling with the fact that it's `controllerManager` and not `ControllerManager`, as we are trying to make string constants adhere to Kubernetes-style (#1065). So far it's seems like our constants will be Kubernetes-style, with an exception for the ones defined by AWS.

pkg/apis/eksctl.io/v1alpha5/defaults.go
errordeveloper: Yes, but `enableTypes: ["all", "api", "audit"]` should be valid. Perhaps validation should run before defaulting?

errordeveloper: I think we could just open an issue to address this later.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -1,5 +1,15 @@
 package v1alpha5
 
+// SetClusterConfigDefaults will set defaults for a given cluster
+func SetClusterConfigDefaults(cfg *ClusterConfig) {
+	if cfg.HasClusterCloudWatchLogging() && len(cfg.CloudWatch.ClusterLogging.EnableTypes) == 1 {
+		switch cfg.CloudWatch.ClusterLogging.EnableTypes[0] {
+		case "all", "*":
+			cfg.CloudWatch.ClusterLogging.EnableTypes = SupportedCloudWatchClusterLogTypes()
```
errordeveloper: I guess removing and combining would be a little more correct?

cPu1: Sorry, I didn't get it. Could you please elaborate?

pkg/apis/eksctl.io/v1alpha5/defaults.go
cPu1: >Yes, but enableTypes: ["all", "api", "audit"] should be valid

Not sure if we should allow adding other types when `all` or `*` is present.

pkg/ctl/cmdutils/configfile.go
```diff
@@ -315,3 +317,16 @@ func NewDeleteNodeGroupLoader(rc *ResourceCmd, ng *api.NodeGroup, ngFilter *Node
 
 	return l
 }
+
+// NewUtilsEnableLoggingLoader will load config or use flags for 'eksctl utils enable-logging'
+func NewUtilsEnableLoggingLoader(rc *ResourceCmd) ClusterConfigLoader {
+	l := newCommonClusterConfigLoader(rc)
+
+	l.flagsIncompatibleWithConfigFile.Insert(
+		append(api.SupportedCloudWatchClusterLogTypes(), "all")...,
```
errordeveloper: Please be sure to update this with new flags. The point of this is to prevent user from using `--config-file` along with flags that map to config fields.

pkg/ctl/utils/enable_logging.go
```diff
@@ -103,29 +78,32 @@ func doEnableLogging(rc *cmdutils.ResourceCmd) error {
...
 	}
 
-	shouldEnable := sets.NewString()
-
+	willBeEnabled := sets.NewString()
 	if cfg.HasClusterCloudWatchLogging() {
-		shouldEnable.Insert(cfg.CloudWatch.ClusterLogging.EnableTypes...)
+		willBeEnabled.Insert(cfg.CloudWatch.ClusterLogging.EnableTypes...)
+	} else {
+		baselineEnabled := sets.NewString(currentlyEnabled.List()...).Union(willBeEnabled)
```
cPu1: Minor nit: the `Union` is a no-op since `willBeEnabled` will always be an empty set here.

cPu1:
I think we should rename the `enable_logging.go` file to `update_logging.go` or `update_cluster_logging.go` to reflect the renamed command.

pkg/ctl/utils/enable_logging.go
martina-if: yeah, I think i was tweaking this several times and now it didn't make sense anymore :)

pkg/apis/eksctl.io/v1alpha5/cloudwatch.go
```diff
@@ -0,0 +1,28 @@
...
+// ClusterCloudWatch contains config parameters related to CloudWatch
+type ClusterCloudWatch struct {
+	//+optional
+	ClusterLogging *ClusterCloudWatchLogging `json:"clusterLogging,omitempty"`
+}
+
+// ClusterCloudWatchLogging container config parameters related to cluster logging
+type ClusterCloudWatchLogging struct {
+	//+optional
+	EnableTypes []string `json:"enableTypes,omitempty"`
```
errordeveloper: Once we have `--enable-types` and `--disable-types`, we should add `DisableTypes` here.

site/content/usage/12-cloudwatch-cluster-logging.md
```diff
@@ -0,0 +1,98 @@
...
+Alternatively, you can use CLI flags.
+
+To enable all types of logs, run:
+```
+eksctl utils update-cluster-logging
+```
+
+To enable `audit` logs, run:
+```
+eksctl utils update-cluster-logging --audit
```
errordeveloper: This didn't get updated, I've opened https://github.com/weaveworks/eksctl/issues/1090.

pkg/apis/eksctl.io/v1alpha5/cloudwatch.go
errordeveloper: I've opened https://github.com/weaveworks/eksctl/issues/1091.

PR #855, opened 2019-06-07
Log info for gpu instances in instances distribution

martina-if:

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -9,3 +9,16 @@ import (
 func IsGPUInstanceType(instanceType string) bool {
 	return strings.HasPrefix(instanceType, "p2") || strings.HasPrefix(instanceType, "p3")
 }
+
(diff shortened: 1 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -9,3 +9,15 @@ import (
 func IsGPUInstanceType(instanceType string) bool {
 	return strings.HasPrefix(instanceType, "p2") || strings.HasPrefix(instanceType, "p3")
 }
+
+func HasGPUInstanceType(instanceTypes []string) bool {
```
errordeveloper: Is there another place where we can use this function? Also, seems like it could actually take `ng *api.NodeGroup` instead, but we don't have to do that in this PR.

martina-if: Yeah, I thought about that. but didn't want to drag the dependency. I would put it in the api package, with the hasMixedInstances. WDYT?

PR #865, opened 2019-06-11
Make troubleshooter output less verbose

errordeveloper:
### Description

Small step towards #491.

/cc @mhausenblas 

Before:
```
[ℹ]  2 sequential tasks: { create cluster control plane "beautiful-hideout-1560245357", create nodegroup "ng-398ca642" }
[ℹ]  building cluster stack "eksctl-beautiful-hideout-1560245357-cluster"
[ℹ]  deploying stack "eksctl-beautiful-hideout-1560245357-cluster"
[✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-beautiful-hideout-1560245357-cluster"
[ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
[ℹ]  AWS::CloudFormation::Stack/eksctl-beautiful-hideout-1560245357-cluster: ROLLBACK_IN_PROGRESS – "The following resource(s) failed to create: [RouteTableAssociationPublicUSWEST2B, RouteTableAssociationPublicUSWEST2D, RouteTableAssociationPrivateUSWEST2C, RouteTableAssociationPublicUSWEST2C, RouteTableAssociationPrivateUSWEST2D, NATGateway, RouteTableAssociationPrivateUSWEST2B]. . Rollback requested by user."
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPublicUSWEST2C: CREATE_FAILED – "Resource creation cancelled"
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPrivateUSWEST2C: CREATE_FAILED – "Resource creation cancelled"
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPublicUSWEST2D: CREATE_FAILED – "Resource creation cancelled"
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPrivateUSWEST2D: CREATE_FAILED – "Resource creation cancelled"
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPrivateUSWEST2B: CREATE_FAILED – "Resource creation cancelled"
[✖]  AWS::EC2::SubnetRouteTableAssociation/RouteTableAssociationPublicUSWEST2B: CREATE_FAILED – "Resource creation cancelled"
[...]

PR #866, opened 2019-06-11
Ensure sequential task errors result in termination of all tasks

errordeveloper:
### Description

This is a follow-up from #865.

This is problematic as it results in confusing output message:
```
[✖]  waiting for CloudFormation stack "eksctl-beautiful-sheepdog-1560245255-cluster" to reach "CREATE_COMPLETE" status: ResourceNotReady: failed waiting for successful resource state
[✖]  invalid cluster config: missing CertificateAuthorityData
[✖]  failed to create cluster "beautiful-sheepdog-1560245255"
```
The `missing CertificateAuthorityData` is very misleading.

### Checklist

### DIFF
--- a/pkg/cfn/manager/tasks.go
+++ b/pkg/cfn/manager/tasks.go
@@ -178,19 +178,20 @@ func (t *asyncTaskWithStackSpec) Do(errs chan error) error {
 	return err
 }
 
-func doSingleTask(allErrs chan error, task Task) {
+func doSingleTask(allErrs chan error, task Task) bool {
 	desc := task.Describe()
 	logger.Debug("started task: %s", desc)
 	errs := make(chan error)
 	if err := task.Do(errs); err != nil {
 		allErrs <- err
-		return
+		return false
 	}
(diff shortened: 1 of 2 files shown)


pkg/cfn/manager/tasks.go
```diff
@@ -178,19 +178,20 @@ func (t *asyncTaskWithStackSpec) Do(errs chan error) error {
...
 
-func doSingleTask(allErrs chan error, task Task) {
+func doSingleTask(allErrs chan error, task Task) bool {
 	desc := task.Describe()
 	logger.Debug("started task: %s", desc)
 	errs := make(chan error)
 	if err := task.Do(errs); err != nil {
 		allErrs <- err
-		return
+		return false
```
martina-if: Why not return an error instead?

errordeveloper: Because errors are handled via the channel already and we don't need to worry about the content as it's already take care of by the other mechanism, so this is just a hint, and boolean is easier to check.

PR #878, opened 2019-06-12
Update CoreDNS manifests

### LINKED ISSUE #692: source latest coredns add-on manifest
We currently have this:

https://github.com/weaveworks/eksctl/blob/ed9e7de2ea301fc8e68937bd4e5cd235d59c605b/pkg/addons/default/generate.go#L4

But it's not the latest version, version update is implemented by patching image tag:

https://github.com/weaveworks/eksctl/blob/ed9e7de2ea301fc8e68937bd4e5cd235d59c605b/pkg/addons/default/coredns.go#L163-L212

We should update the URL in `pkg/addons/default/generate.go` and re-generate bindata assets, then replace `UpdateCoreDNSImageTag` with a simpler version that applies the manifest directly.

errordeveloper:
### Description

Close #692

### Checklist


pkg/addons/default/assets/coredns-1.11.json
```diff
@@ -0,0 +1,311 @@
...
+          "eks.amazonaws.com/component": "kube-dns",
+          "k8s-app": "kube-dns",
+          "kubernetes.io/cluster-service": "true",
+          "kubernetes.io/name": "CoreDNS"
+        },
+        "name": "kube-dns",
+        "namespace": "kube-system"
+      },
+      "spec": {
+        "clusterIP": "10.100.0.10",
```
martina-if: is this always the same?

pkg/addons/default/assets/coredns-1.11.json
```diff
@@ -0,0 +1,311 @@
...
+          "k8s-app": "kube-dns"
+        },
+        "name": "coredns",
+        "namespace": "kube-system"
+      }
+    },
+    {
+      "apiVersion": "v1",
+      "data": {
+        "Corefile": ".:53 {\n    errors\n    health\n    kubernetes cluster.local in-addr.arpa ip6.arpa {\n      pods insecure\n      upstream\n      fallthrough in-addr.arpa ip6.arpa\n    }\n    prometheus :9153\n    proxy . /etc/resolv.conf\n    cache 30\n}\n"
```
martina-if: this looks a bit weird, is it normal?

martina-if:
LGTM, 2 minor comments. Merge when you need to

pkg/addons/default/assets/coredns-1.11.json
errordeveloper: No, and we handle here: https://github.com/weaveworks/eksctl/pull/878/files#diff-9db8b5b22bd493bd608250a68252142fR211

pkg/addons/default/assets/coredns-1.11.json
errordeveloper: It's [not JSON](https://coredns.io/2017/07/23/corefile-explained/).

errordeveloper: So this one look like this:
```
.:53 {
    errors
    health
    kubernetes cluster.local in-addr.arpa ip6.arpa {
      pods insecure
      upstream
      fallthrough in-addr.arpa ip6.arpa
    }
    prometheus :9153
    proxy . /etc/resolv.conf
    cache 30
}
```

martina-if:
Would it be better to remove the field so if we ever forget to replace it
it will fail fast?

errordeveloper:
@martina-if I've removed the field now. I kept the logic that sets `metadata.resourceVersion` and `spec.clusterIP`, as otherwise I'm seeing this error:
```
[✖]  Service "kube-dns" is invalid: [metadata.resourceVersion: Invalid value: "": must be specified for an update, spec.clusterIP: Invalid value: "": field is immutable]
```

errordeveloper:
I'd like to do some more manual tests on this tomorrow, as I didn't get a chance to try it in 1.11 –> 1.12 scenario.

errordeveloper:
This is good to go!

pkg/addons/default/coredns.go
```diff
@@ -167,54 +162,81 @@ func InstallCoreDNS(rawClient kubernetes.RawClientInterface, region string, wait
...
-	image := &d.Spec.Template.Spec.Containers[0].Image
-	imageParts := strings.Split(*image, ":")
+	for _, rawObj := range list.Items {
+		resource, err := rawClient.NewRawResource(rawObj)
+		if err != nil {
+			return false, err
+		}
+		switch resource.GVK.Kind {
+		case "Deployment":
+			image := &resource.Info.Object.(*appsv1.Deployment).Spec.Template.Spec.Containers[0].Image
```
martina-if: Is this ok `...Containers[0]`? Do we know it will always have one item?

martina-if: Is this ok `...Containers[0]`? Do we know it will always have one item?

pkg/addons/default/coredns.go
```diff
@@ -167,54 +162,81 @@ func InstallCoreDNS(rawClient kubernetes.RawClientInterface, region string, wait
 	return false, nil
 }
 
-// UpdateCoreDNSImageTag updates image tag for kube-system:deployment/coredns based to match the latest release
-func UpdateCoreDNSImageTag(clientSet k8s.Interface, plan bool) (bool, error) {
-	printer := printers.NewJSONPrinter()
+// UpdateCoreDNS will update the `coredns` add-on
+func UpdateCoreDNS(rawClient kubernetes.RawClientInterface, region, controlPlaneVersion string, plan bool) (bool, error) {
```
martina-if: There is a bunch of logic here now, it would be good to test that it is doing the right thing, not only not throwing an error.

pkg/addons/default/coredns.go
errordeveloper: Yes, because it's defined by the JSON manifest that we control.

pkg/addons/default/coredns_test.go
```diff
@@ -106,41 +109,62 @@ var _ = Describe("default addons - coredns", func() {
...
+		loadSample("1.11", 10)
 
 		It("can load 1.11 sample", func() {
-			check("v1.1.3")
+			checkCoreDNSImage(rawClient, "eu-west-1", "v1.1.3")
 		})
 
 		It("can update based to latest version", func() {
-			_, err := UpdateCoreDNSImageTag(clientSet, false)
+			_, err := UpdateCoreDNS(rawClient, "eu-west-2", "1.12.x", false)
```
errordeveloper: @martina-if what do you make of this test? It checks as much as currently possible with this fake client of ours...

errordeveloper: I suppose we could add tests with other kind of pre-conditions that would result in errors, if you think that's important?

martina-if: The problem I see is that if I break the logic in that function this test won't caught it because it only checks that there is no error returned. I prefer tests that tell you that you actually broke some piece of logic. Maybe it's hard to test with how it looks now. Perhaps it needs to be split in smaller pieces that can be tested. Or test that we call the client with the right information?

errordeveloper: See below, the test looks at what requests were made - does that seem sufficient?

martina-if: ok, I don't see a better way right now to test this :+1:

Issue #881, opened 2019-06-13
Improvements to website


bia:
Building on #857,
from discussion with @martina-if @gemagomez @errordeveloper,

Here are next steps and changes to make in hierarchical order :  

- [x] Add one-click-copying for all code blocks
- [x] Consolidate content between index and introduction pages (this includes merging some info from 'Introduction' to 'Installation'): 
<img width="2881" alt="Screenshot 2019-06-13 at 15 16 53" src="https://user-images.githubusercontent.com/92439/59437764-766d7e00-8df2-11e9-9f73-fd75a25c6f30.png">

- ~Change right side navigation (link) colors~ This was too heavy and distracting :
![Screenshot 2019-06-20 at 12 35 33](https://user-images.githubusercontent.com/92439/59844774-f5ffcd80-935b-11e9-9ba1-d6f9bf256a4a.png)

- [x] Add a border and syntax highlighting to YAMLs like in jk : 
<img width="706" alt="Screenshot 2019-06-12 at 19 32 22" src="https://user-images.githubusercontent.com/92439/59438180-4b375e80-8df3-11e9-9949-bbd85b90f74d.png">

- [x] Add this back to the top of the index page
<img width="491" alt="Screenshot 2019-06-13 at 16 10 5 2" src="https://user-images.githubusercontent.com/92439/59439642-ed584600-8df5-11e9-95d9-635a6b9cc7a3.png">

- [x] Add link for 'Need help? Join [Weave Community Slack][slackjoin]'

- [x] Maintain ~Proxima Nova~ Montserrat (Proxima Nova is not a free font) and Roboto Mono fonts the same between this project, Flux, and Weave Cloud

- [x] Bring back github buttons (though I took some away because it made the page too busy)
<img width="1398" alt="Screenshot 2019-06-17 at 17 52 06" src="https://user-images.githubusercontent.com/92439/59690853-64208500-91e2-11e9-8f8d-e8f9c0a4e308.png">

- [x] Check and complete Features list if needed

- [x] Look into improving syntax highlighting

whereisaaron:
This website sounds cool and useful. But where is it? Only in git or is it hosted somewhere?

martina-if:
Hi @whereisaaron , we haven't published the new site yet because it's still work in progress. If you want to preview it ping me on Slack and I'll send you the link :)

errordeveloper:
I think one thing that definetly catches me is that the title and heading auto-upcase `Eksctl`, can we disable it?

[...]

PR #884, opened 2019-06-14
Improvements to website

bia:
Work towards #881 so that we can formally publish the hugo website.

cc @martina-if


errordeveloper:
I've merged it since it wanted a rebase, which wasn't strictly required.

Issue #745, opened 2019-04-19
Support for cert-manager


Yannig:
**Why do you want this feature?**

When adding cert-manager with a cluster created by eksctl, i'm not able to create a certificate.

This issue will be followed by a PR.

**What feature/behavior/change do you want?**

The problem came from the default policy used with --external-dns-access:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "route53:ListHostedZones",
                "route53:ListResourceRecordSets"
            ],
            "Resource": "*",
            "Effect": "Allow"
        }
    ]
}
```

In this case, the cert-manager pod send error of this kind:

```
I0419 12:14:12.777229       1 controller.go:205] challenges controller: syncing item 'kube-system/bla-bla-961647875-0'
I0419 12:14:12.777482       1 dns.go:89] Presenting DNS01 challenge for domain "bla.bla"
E0419 12:14:13.157374       1 controller.go:207] challenges controller: Re-queuing item "kube-system/bla-bla-961647875-0" due to error processing: Failed to determine Route 53 hosted zone ID: AccessDenied: User: arn:aws:sts::xxx:assumed-role/eksctl-prod-a-nodegroup-ng-baxxxx-NodeInstanceRole-XXX/i-xxx is not authorized to perform: route53:ListHostedZonesByName
        status code: 403, request id: xx-xx-xx-xx-xx
```

If I add "route53:ListHostedZonesByName", the error became:

```
I0419 12:15:33.157661       1 controller.go:205] challenges controller: syncing item 'kube-system/bla-bla-961647875-0'
I0419 12:15:33.157908       1 dns.go:89] Presenting DNS01 challenge for domain "bla.bla"
E0419 12:17:36.674535       1 controller.go:207] challenges controller: Re-queuing item "kube-system/bla-bla-961647875-0" due to error processing: Time limit exceeded. Last error: Failed to query Route 53 change status: AccessDenied: User: arn:aws:sts::121748747150:assumed-role/eksctl-prod-a-nodegroup-ng-ba8014-NodeInstanceRole-XXX/i-xxx is not authorized to perform: route53:GetChange on resource: arn:aws:route53:::change/xxx
        status code: 403, request id: xx-xx-xx-xx-xx
```

To get it working, I need the following policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "route53:ListHostedZones",
                "route53:ListResourceRecordSets",
[...]

PR #885, opened 2019-06-14
Enable policies for cert-manager.

Yannig:
Fix for https://github.com/weaveworks/eksctl/issues/745

### Checklist

- [X] All unit tests passing (i.e. `make test`)
- [X] Added yourself to the `humans.txt` file


errordeveloper:
@Yannig thanks very much! This looks good overall, but I wonder if we should make this controlled by a dedicated config field instead of hanging it on `externalDNS` – what do you think?

Yannig:
@errordeveloper no problem. I'm doing it asap.

Yannig:
@errordeveloper what do you think about it?

errordeveloper:
@Yannig thank you, this looks good! Ideally, we would make this a config-only option, if you are okay with that? We have a strong preference to avoid adding new flags, as there are already too many.

Yannig:
I understand. I'm doing the modification.

Yannig:
@errordeveloper I think it's fine now. What do you think about?

Yannig:
Done

PR #898, opened 2019-06-18
Generate schema spec into the site documentation

martina-if:
- [x] Manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,10 @@ generate-kubernetes-types-test: generate-kubernetes-types ## Test if generated K
 generate-ami: ## Generate the list of AMIs for use with static resolver. Queries AWS.
 	go generate ./pkg/ami
 
+.PHONY: generate-schema
+generate-schema: ## Generate the schema file in the documentation site.
+	@go run ./cmd/schema/generate.go
+
 .PHONY: ami-check
 ami-check: generate-ami ## Check whether the AMIs have been updated and fail if they have. Designed for a automated test
 	git diff --exit-code pkg/ami/static_resolver_ami.go > /dev/null || (git --no-pager diff; exit 1)
--- a/cmd/schema/generate.go
+++ b/cmd/schema/generate.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"github.com/alecthomas/jsonschema"
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"io/ioutil"
+	"sigs.k8s.io/yaml"
+	"strings"
+)
+
+const outputFile = "site/content/usage/20-schema.md"
+
+func main() {
+
+	var document strings.Builder
+	document.WriteString(`---
+title: Config file schema
+weight: 200
+---
+
+`)
+	document.WriteString("```yaml\n")
+
+	schema := jsonschema.Reflect(&api.ClusterConfig{})
+	yamlSchema, err := yaml.Marshal(schema.Definitions)
+	if err != nil {
+		panic(err)
+	}
+	document.Write(yamlSchema)
+	document.WriteString("```\n")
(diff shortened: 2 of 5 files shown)


pkg/schema/generate.go
```diff
@@ -0,0 +1,33 @@
...
+
+const outpufFile = "site/content/usage/20-schema.md"
+
+func main() {
+
+	var document strings.Builder
+	document.WriteString("---\n")
+	document.WriteString("title: Config file schema\n")
+	document.WriteString("weight: 200\n")
+	document.WriteString("---\n\n")
```
cPu1: Minor nit: we could also use a template string and a single `WriteString` call for the content till here.

```golang
	b.WriteString(`---
title: Config file schema
weight: 200
---

`)
```

pkg/schema/generate.go
```diff
@@ -0,0 +1,33 @@
...
+	document.WriteString("```yaml\n")
+
+	schema := jsonschema.Reflect(&api.Config{})
+	yamlSchema, _ := yaml.Marshal(schema.Definitions)
+	document.WriteString(string(yamlSchema))
```
cPu1: Minor nit: the conversion to string can be avoided by using `document.Write`.

pkg/schema/generate.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/alecthomas/jsonschema"
+	"io/ioutil"
+	"k8s.io/client-go/tools/clientcmd/api"
+	"sigs.k8s.io/yaml"
+	"strings"
+)
+
+const outpufFile = "site/content/usage/20-schema.md"
```
cPu1: Minor typo: `outputFile`

pkg/schema/generate.go
```diff
@@ -0,0 +1,33 @@
...
+
+	var document strings.Builder
+	document.WriteString("---\n")
+	document.WriteString("title: Config file schema\n")
+	document.WriteString("weight: 200\n")
+	document.WriteString("---\n\n")
+	document.WriteString("```yaml\n")
+
+	schema := jsonschema.Reflect(&api.Config{})
+	yamlSchema, _ := yaml.Marshal(schema.Definitions)
```
cPu1: Should we ignore the error? Looks like `errcheck` warns if the error isn't checked.

pkg/schema/generate.go
martina-if: Good idea, I forget about these, thanks

errordeveloper:
A couple of errors in CI:
```
pkg/schema/generate.go:4:2:error: cannot find package "github.com/alecthomas/jsonschema" in any of: (vet)
pkg/schema/generate.go:4:2:warning: error return value not checked (could not import github.com/alecthomas/jsonschema (cannot find package "github.com/alecthomas/jsonschema" in any of:) (errcheck)
```

pkg/schema/generate.go
```diff
@@ -0,0 +1,38 @@
...
+	var document strings.Builder
+	document.WriteString(`---
+title: Config file schema
+weight: 200
+---
+
+`)
+	document.WriteString("```yaml\n")
+
+	schema := jsonschema.Reflect(&api.Config{})
```
errordeveloper: Did you try it with our struct?

pkg/schema/generate.go
```diff
@@ -0,0 +1,38 @@
+package main
```
errordeveloper: I'd say we should move it under `cmd/`, as we don't have `main` package in `pkg/` (aside from things that get have build tags and only used with `go generate`, IIRC).

pkg/schema/generate.go
martina-if: yeah, I used the k8s one by mistake. Thanks!

PR #674, opened 2019-03-28
[WIP] Adding go modules

christopherhein:
Signed-off-by: Christopher Hein <me@chrishein.com>

### Description
Adding go module support

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] Added tests that cover your change (if possible)
- [x] All unit tests passing (i.e. `make test`)
- [x] All integration tests passing (i.e. `make integration-test`)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [x] Added yourself to the `humans.txt` file

PR #914, opened 2019-06-20
Start using GitHub Actions to run builds

errordeveloper:
### Description

Start using GitHub Action to see if it's better/faster than CircleCI. This is a minimal change, we may wish to do more, e.g. publish the build image as a versioned action.

We might want to use GitHub Action for contributor workflow automation as well. Let's see how well it works :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 built_at := $(shell date +%s)
 git_commit := $(shell git describe --dirty --always)
 
-version_pkg := github.com/weaveworks/eksctl/pkg/version
+import_prefix := github.com/weaveworks/eksctl
+version_pkg := $(import_prefix)/pkg/version
 
 EKSCTL_BUILD_IMAGE ?= weaveworks/eksctl-build:latest
 EKSCTL_IMAGE ?= weaveworks/eksctl:latest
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -45,6 +46,11 @@ INTEGRATION_TEST_ARGS += -eksctl.version=$(INTEGRATION_TEST_VERSION)
 $(info will launch integration tests for Kubernetes version $(INTEGRATION_TEST_VERSION))
 endif
 
+.PHONY: github-build
+github-build:
+	cp -a . $(GOPATH)/src/$(import_prefix)
```
errordeveloper: @2opremio I believe we won't have to do this, if we switch to modules...

2opremio: Exactly, we won't :)

Makefile
```diff
@@ -45,6 +46,11 @@ INTEGRATION_TEST_ARGS += -eksctl.version=$(INTEGRATION_TEST_VERSION)
 $(info will launch integration tests for Kubernetes version $(INTEGRATION_TEST_VERSION))
 endif
 
+.PHONY: github-build
+github-build:
```
2opremio: Can you move this to the workflow file?

2opremio:
What's the plan/criteria to transition to the GitHub Actions? (i.e. do we want to run both in parallel for a while? what would make us switch?).

In terms of speed, GitHub runs 40% slower (it took ~10.5 minutes vs CircleCI's ~7.45 minutes , which may be an outlier from a single run, but still),  which isn't great. The feedback loop should be fast in PRs. In fact, CircleCI is already very slow (which I plan to address after #674 by persisting the Go build cache between runs), particularly considering we want to include a smoke test.

Any idea of why GitHub is slower and whether it's easy to improve? Does it support caching? (asking since I have never used it myself).

stefanprodan:
@2opremio there is no caching support in GH actions. You can share files between actions in the same workflow but not between workflow runs.

Makefile
errordeveloper: Yeah, with modules it should be just one line, so we won't need to wrap anything.

errordeveloper:
I just think it's a nice feature of GitHub to try, and we should be able to automate more workflows etc. They also have a Docker image registry coming out soon, in theory it could all be very integrated nicely without a 2rd party CI (less clicking, if nothing else).

errordeveloper:
Closing for now

PR #917, opened 2019-06-21
Migrate dependency management to go modules

2opremio:
### Description

As a followup of #603 , this PR is aimed at migrating dependency management from [`dep`](https://github.com/golang/dep) to Go modules. Fixes #917 

It is being quite an endeavor, namely due to the disruption caused in code-generation and tools in general (most of the tools have required an update to cope either with go modules or the removal of `vendor` directories).

Worth mentioning:
* It's not required to define `GOPATH` anymore to build or develop on `eksctl`. This means you don't need to place `eksctl` in your `GOPATH` and in fact, after this change you should move it out of your `go env GOPATH` (otherwise `go` will refuse to find the dependencies, this is intentional on Go's side).
* `vendor` directories are no more (caching of vendored code will happen through the build image)
* I flattened the dependencies into a single module (before we tracked dependencies with `dep` both in the root directory and `build`).
* I moved the remaining content of the `build` directory to the project's root (it was very scarce after the refactoring).

### Checklist
- [x] All unit tests passing (i.e. `make test`). ~I am facing a problem with the generated code.~
- [x] Manually tested
- [-] Push `weaveworks/eksctl-build:latest` to speedup the build (now it's really slow since all the modules are downloaded all the time). To be done (automatically) in #920
- [-] Consider persisting the Go build cache (i.e. `go env GOCACHE`) to speed the builds up even further. To be done in #920


2opremio:
_−2,151,712_ lines of code. That feels good, it may be worth re-writing the Git history to get rid of all the vendored code once we merge.

2opremio:
This is ready to go, performance-aside. The CircleCI build time has increased considerably, due to the lack of persistent dependencies (due to the removal of `vendor` directories).

This PR has already grown out of hand, so I am considering to merge and address performance in another PR.

.gitignore
```diff
@@ -55,7 +24,10 @@ site/resources/_gen/
 coverage.out
 
 # Test results
-test-results/
+test-results
 
 integration/kubeconfig-*
 
+# Generated header file
```
errordeveloper: Does this appear in top-level dir?

CONTRIBUTING.md
```diff
@@ -36,49 +36,41 @@ the latest information.
...
-Select a folder path for your go projects, for example `~/eksctl`
-```bash
-mkdir ~/eksctl
-cd ~/eksctl
-export GOPATH="$(pwd)"
-export GOBIN="${GOPATH}/bin"
-```
-
-> NOTE: Windows users should install Docker for Windows and run `make eksctl-image` to build their code.
+2. Make sure that `$(go env GOPATH)/bin` is in your shell's `PATH`. You can do so by
```
errordeveloper: Why?

2opremio: I recall this was needed during the build. Let me double-check.

CONTRIBUTING.md
```diff
@@ -36,49 +36,41 @@ the latest information.
...
-Clone the repo using go:
+Make a fork of this repository and clone it by running:
 
 ```bash
-go get -d github.com/weaveworks/eksctl/...
+git clone git@github.com:<yourusername>/eksctl.git
 ```
 
-Make a fork of this repository and add it as a remote:
+It is discouraged to clone under `go env GOPATH`. Otherwise you will need to set
```
errordeveloper: Sounds a bit weird and unclear... how about this?
```suggestion
It is not recommended to clone under `GOPATH`, otherwise you will need to set
```

Dockerfile
```diff
@@ -17,7 +17,7 @@ ENV KUBECTL_VERSION v1.11.5
 RUN curl --silent --location "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" --output /out/usr/local/bin/kubectl \
     && chmod +x /out/usr/local/bin/kubectl
 
-ENV EKSCTL $GOPATH/src/github.com/weaveworks/eksctl
+ENV EKSCTL /eksctl
```
errordeveloper: Let's call it `/src`.

.gitignore
2opremio: Yes

Dockerfile.build
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.12-alpine3.9
```
errordeveloper: Let's keep `build/` as is please. It's mean for performance of `docker build`. It prevents having to pass copy the whole tree, only what is in `build/` is copied to build context, and also ensure cache works is used as long as nothing in `build/` changed.

Makefile
```diff
@@ -130,14 +133,18 @@ ami-check: generate-ami ## Check whether the AMIs have been updated and fail if
 
 .PHONY: generate-kubernetes-types
 generate-kubernetes-types: ## Generate Kubernetes API helpers
-	@build/vendor/k8s.io/code-generator/generate-groups.sh deepcopy,defaulter _ github.com/weaveworks/eksctl/pkg/apis eksctl.io:v1alpha5
+	@go mod download k8s.io/code-generator # make sure the code-generator is present
+	@echo "/*\n$$(cat LICENSE)*/\n" > codegenheader.txt
```
errordeveloper: What is this for?

CONTRIBUTING.md
2opremio: It's intentional, with `1.12` you don't need do `GOPATH` env variable anymore

Makefile
```diff
@@ -130,14 +133,18 @@ ami-check: generate-ami ## Check whether the AMIs have been updated and fail if
...
+	@env GOPATH="$$(go env GOPATH)" bash "$$(go env GOPATH)/pkg/mod/k8s.io/code-generator@v0.0.0-20190612205613-18da4a14b22b/generate-groups.sh" \
```
errordeveloper: How is `go env GOPATH` different from simple variable lookup?

CONTRIBUTING.md
2opremio: `mockery` wasn't found otherwise, but I think I can fix it so that it looks for mockery in the right place.

errordeveloper:
> `vendor` directories are no more (caching of vendored code will happen through the build image)

I am not very comfortable about this. I don't actually use build image locally, I use `make build`. I would like to keep `vendor`.

Dockerfile.build
2opremio: The bottleneck is not passing the whole tree, that takes seconds. Let's have a call about this.

Makefile
2opremio: `go env GOPATH` always provides a value, even if you don't declare `GOPATH` (which you don't need to with go >= 1.12)

Makefile
2opremio: It's the header of the generated files, without providing an explicit header (`--go-header-file` ) `generate-groups.sh` doesn't work with go modules (`k8s.io/code-generator` doesn't fully accept go modules).

2opremio:
> I am not very comfortable about this. I don't actually use build image locally, I use `make build`. I would like to keep `vendor`.

What makes you uncomfortable? It works equally well with `make build` locally, `go` will just cache the dependencies under `$(go env GOPATH)/pkg/mod` just like the `vendor` directory, but out of your tree.

CONTRIBUTING.md
2opremio: Fixed

Dockerfile
2opremio: Done

CONTRIBUTING.md
2opremio: Although, we are losing clarify and friendliness for precision. I will change it.

2opremio:
I had a discussion with @errordeveloper offline about this PR. The conclusions are:
1. We agreed it's OK to remove the `vendor` directories and the `build` directory, due to how `go mod` caching works
2. We will improve performance in a later PR and merge now
3. To avoid issues, I am to make sure that we are not upgrading the `k8s.io` dependencies inadvertently.

| dependency | before (with `dep`) | now (with `go mod`) | notes |
| ------ | ---- | ---- | ----|
|  `k8s.io/api`|  `145d52631d00cbfe68490d19ae4f0f501fd31a95` | `145d52631d00` |
| `k8s.io/apiextensions-apiserver` | `bd0469a053ff88529a61145790499fe78a09a49d` | `bd0469a053ff`
| `k8s.io/apimachinery` | `01f179d85dbce0f2e0e4351a92394b38694b7cae` |   `01f179d85dbc`
| `k8s.io/apiserver` | `cf2f1d68202d872417a0a0af0589f40e1bb04458` |`cf2f1d68202d` | indirect
| `k8s.io/cli-runtime` | `082c0831af2b8f6a5ecf4d3a5a666f317c5e1fb2` | `082c0831af2b` 
| `k8s.io/client-go` |  `78295b709ec6fa5be12e35892477a326dea2b5d3` | `78295b709ec6` 
| `k8s.io/code-generator` | `3a2206dd6a78497deceb3ae058417fdeb2036c7e` | `18da4a14b22b` | Intentionally upgraded so that it can cope with go modules
| `k8s.io/csi-api`| `a37926bd22154f4a8e21fda60bec9f3973c29fec` |`a37926bd2215` |  indirect
| `k8s.io/kops` | `278e6606534eef97e9086563f030f6de25fb36ff` | `278e6606534e`
| `k8s.io/kube-openapi` | `e494cc58111187acad93e64529228a2fc0153e39` | `e494cc581111` | indirect
| `k8s.io/kubelet` | `3556bcde967069b8b8bcd57be2cddcae0d8dc072` | `3556bcde9670`
| `k8s.io/kubernetes` | `v1.12.6` | `v1.12.6` | indirect
| `k8s.io/utils` | `a82851343d668ce8cdd50da9f8d565164fad6951` |`a82851343d66` | indirect
| `sigs.k8s.io/yaml` | `fd68e9863619f6ec2fdd8625fe1f02e7c877e480` (`v1.1.0`) | `v1.1.0`

2opremio:
@errordeveloper PTAL

errordeveloper:
Thanks @2opremio, the dependencies look good to me.

Makefile
errordeveloper: Ah, I see. Would you mind to add a comment to clarify this?

2opremio: Done

errordeveloper: Also, would you mind calling it `.kubernetes-code-generator-headers`? So it not too visible.

Dockerfile.build
errordeveloper: To clarify for anyone reading this:
- copying context tool long time because of `vendor/`
- modules provide dependency caching, so no need to use `vendor/`

errordeveloper:
I've restarted the build in Circle, let's see if it's a flake...

Makefile
2opremio: I think it's better for it to be explicit

2opremio:
> I've restarted the build in Circle, let's see if it's a flake...

There was a problem fetching `go.hein.dev/go-version`, hopefully it will be gone now (and won't have a problem anymore once we start caching dependencies again in the build container)

2opremio:
Actually, Circle is failing because of https://github.com/christopherhein/go-version/commit/812a3b5e32acb9effed36765ee30f96295003288 probably due to a transitive dependency

2opremio:
The problem happens when building github.com/kubernetes-sigs/aws-iam-authenticator , whose dependencies I hadn't added to the go modules

tools.go
```diff
@@ -5,9 +5,17 @@ package eksctl
 // Mock imports to enforce their installation by `go mod`.
 // See https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
 import (
+	// This go-version dependency only included to enforce aws-iam-authenticator to use a specific version of go-version.
+	// aws-iam-authenticator v0.4.0 uses `dep`, causing `go mod` to ignore its dependency files and just to pull
+	// github.com/christopherhein/go-version:master. This causes the build to break because go-version:master has recently
+	// switched to module name go.hein.dev/go-version, causing a module name mismatch.
+	// TODO(fons): get rid of this dependency once we bump the version of aws-iam-authenticator
+	//             which now uses go modules
```
2opremio: @errordeveloper FYI, the fact that this is possible makes it obvious that you cannot depend on multiple versions of a library (enforcing a version of a transitive dependency also changes the version of such dependency)

errordeveloper: 👍

errordeveloper:
```
panic: codecgen version mismatch: current: 8, need 10. Re-generate file: /go/pkg/mod/github.com/coreos/etcd@v3.3.10+incompatible/client/keys.generated.go
```

I think this basically means etcd client version has changed...

Issue #874, opened 2019-06-12
Add iam identityMappings to config file


errordeveloper:
We have commands for managing IAM identity mapping now, and config file can be used as input, but we don't yet support definitions of the identity mappings in the config file itself.

I thing this would be the most suitable way to expose it:
```
iam:
  identityMapping: ...
```

At the I'm not sure about how the array/map itself should look like.

We would probably want to keen nodegroups out of the picture for now, i.e. what we do automatically at the moment, can remain invisible at the level of config file. Users who want to manage node IAM roles along with `--update-auth-configmap=false`, should be able to do it and use config file to set authenticate their nodegroup roles.

martina-if:
Should it be a map? or perhaps an array of mappings?

errordeveloper:
@martina-if yes, good point - I've generalised the snippet above.

andreamaruccia:
how is this related to #841 ?

neelakansha85:
+1  to this feature. 

This will make it a lot easier instead of having separate `aws-auth.yaml` config file and applying it to the cluster after cluster creation to having all config in declarative form.

jvosantos:
upvote upvote upvote

pierresteiner:
Upvote as well!

martina-if:
see also #967

muram:
+1, any update on when this feature might be considered for implementation?

aclevername:
@muram this is currently being worked on, feedback/thoughts are welcome 😄  https://github.com/weaveworks/eksctl/pull/3097

aclevername:
Closing. We are expecting an new [API from AWS to handle cluster permissions](https://github.com/aws/containers-roadmap/issues/185#issuecomment-706636321), so we could implement this feature but it would likely be replaced very soon by this new API. I think it makes sense to halt any new development on `iamidentitymappings` functionality

adammw:
the iamidentitymappings is the eksctl feature exposing the identity mapping used by `aws-iam-authenticator`, no? therefore the feature to add it to the config file should remain even if the underlying implementation would need to change...

I also note that `aws-iam-authenticator` now supports configuration via a `IAMIdentityMappings` CRD which should free up some of the use cases suggested..

PR #932, opened 2019-06-24
Feature/iam users

patstrom:
### Description

Added support for modifying the `mapUsers` key in the `aws-auth` configmap.

Opted for changing the `MapRole` type into a `MapIdentity` type with custom (un)marshal functions to handle the different keys (`rolearn`/`userarn`). Since the only time you actually need to know what kind of ARN you are working with is when choosing between `mapRoles` and `mapUsers` (which is centralised to `setIdentities`) as well as when marshaling for `-o (yaml|json)` this seemed like a good choice. Or at least a better choice than juggling two different types for each kind.

I also took the liberty to store the ARN in the `aws-sdk-go/arn.ARN` type which I created an alias for in order to implement the `pflag.Value` interface.

This is a breaking change since I renamed the `--role` flag to `--arn`.

The idea is the implement some tests after initial feedback.

### Checklist

- [ ] Manually tested


patstrom:
I've added both unit tests and some integration tests to also manipulate user mappings!

patstrom:
[Here](https://gist.github.com/patstrom/222e2676fbc30d01aec2285a482d78c4) is the output from the integration test. I ran into the issue with orphaned network interface but everything else passed.

errordeveloper:
Hi @patstrom! I must apologise that your PR has gone unnoticed for so long, we've been very busy getting things ready for 0.2.0 release. We now have this in 0.3.0 milestone, which we are expecting in a couple of weeks from now.

rndstr:
@patstrom this looks really good. It might take me a couple more days to give it a thorough review.

errordeveloper:
I'd would be great if we can get this into 0.3.0!

patstrom:
What is the time-frame for 0.3?

kalbir:
@patstrom We're aiming for 5th August for 0.3.0

errordeveloper:
@kalbir do you think we can get this into 0.4.0?

kalbir:
0.3.0 and 0.4.0 have been a bit busy so after discussion with @gemagomez we are making a concerted effort to get this into 0.5.0!

pkg/authconfigmap/arn.go
```diff
@@ -0,0 +1,35 @@
+package authconfigmap
+
+import "github.com/aws/aws-sdk-go/aws/arn"
```
errordeveloper: I didn't know this package exists, we should adopt it in all places where ARNs are used.

pkg/authconfigmap/mapidentity.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil
+}
+
+// MarshalJSON for MapIdentity marshals the ARN field into either "rolearn" or "userarn"
+// depending on what is appropriate and returns and error if it cannot determine that.
+// All other fields are marshaled with the default marshaler
+func (m *MapIdentity) MarshalJSON() ([]byte, error) {
+	// Marshal everything as one would normally do
+	type alias MapIdentity
+	a := (*alias)(m)
```
errordeveloper: I'd just make a copy without creating a type alias, try: `json.Marshal(MapIdentity(*m))`.

errordeveloper: Ignore that, it doesn't solve the problem. I need to think of this a little more.

pkg/authconfigmap/arn.go
patstrom: So you want to remove this type from the PR and solve it later on? Or keep it in here now as-is and centralize it for the entire project in #1218?

pkg/authconfigmap/mapidentity.go
patstrom: What is the problem? It doesn't create a copy of `MapIdentity`, just `*MapIdentity`. Or at least that is the idea.

errordeveloper: Well, what you are trying to overcome is recursive call to `MarshalJSON`, right? I am not sure if what I suggested with `json.Marshal(MapIdentity(*m))` actually overcomes that or not, I'll need to test a few things and understand why this whole function is needed. I don't yet understand, but I'm keen to hear an explanation, if you have a time :)

pkg/authconfigmap/arn.go
errordeveloper: No, keep it here, refactor later 👍

pkg/authconfigmap/mapidentity.go
patstrom: Yeah the point of the alias is to not have a recursive `MarshalJSON`. The current code works as expected, no?

So the problem is that `MapIdentity` either has the structure

```
{
   "rolearn": "..."
}
```

or

```
{
   "userarn": "..."
}
```

and in order to handle that situation in go you could have a `MapRole` and a `MapUser` struct and write custom unmarshaling logic that determines which type it should be. Since this is super annoying and the only time they need to be distinguished is when (un)marshaling I opted to create one type with custom functions instead. 

Since only the `ARN` field and the `rolearn` and `userarn` keys are relevant we can handle everything else with the default implementation of (un)marshal. In order  to do that we need to create a type alias since otherwise it will get stuck in infinite recursion.

errordeveloper: I think exposing two separate fields would simplify things. We will need to review the design of the struct as part of #874 anyway.
```Go
type Identity struct {
        RoleARN *ARN `json:"rolearn,omitempty"`
        UserARN *ARN `json:"userarn,omitempty"`
	Username *string   `json:"username,omitempty"`
	Groups   []string `json:"groups,omitempty"`
}
```

You could easily check whether `RoleARN` or `UserARN` is `nil` (having both set should probably result in an error).

It may make sense to add a higher-level struct, but let's not worry about it for now.

All in all, please do feel free to make changes to `pkg/iam/` as you see fit, there's no need to work around the API it currently exposes!

errordeveloper: Also, if you don't have the time - just let us know and we can take care of wrapping it up :)

patstrom: I'll take a look at it next week!

patstrom: I have now refactored to use such a struct.

Unfortunately there are new (and better I suppose) linters so I haven't been able to test it. I don't really have time to do the necessary rebase for this, sorry.

Apart from some question marks the functionality shouldn't really have changed though. The biggest question mark is in `pkg/ctl/create/iamidentitymapping.go:49` where an error should be handled. Perhaps it should be moved into the `doCreateIAMIdentityMapping` function?

martina-if:
Hi @patstrom , we finally got some time to look into this, sorry for the delay. And thank you for the work you put into it, we appreciate it.

I thought about the generalization of the Role and User into one struct and I think the code would be simpler and easier to maintain if we created a new struct MapUser similar to MapRole:

```
type MapUser struct {
	iam.Identity `json:",inline"`
	RoleARN      string `json:"userarn"`
}
```

Then I would add a flag `--user` just like the `--role` one. This wold mean that the code is slightly duplicated, but I think the trade off is worth it because:

- We don't break the API
- We only have 2 types: roles and users
- When we actually process the entity we need to distinguish if t's a user or a role, making us have to do switch(){}`...
- We need to maintain the custom marhalling/unmarshalling of the struct.

What do you think? How is your availability to finish the PR? :)

patstrom:
I thought about two different types as well but it just feels unnecessary  to duplicate all code at no benefit to anyone, really.

The ARN itself contains enough information to decide if it should go into the `rolearn` or `userarn` key, and the only time it really matters is when actually marshaling the authconfigmap for storage in the cluster. That's why I first opted for custom (un)marshal functions. Sure, there should be sanity checks to make sure it's either a user of a role when it's entered by the user in the first place but it doesn't *really* matter until marshaling.

Having two different types means that the *entire* (internal) API has to be duplicated just to make sure that an arn of resource `user` is marshaled as `userarn` and an arn of resource `role` is marshaled as `rolearn`. I don't really see how maintaining all that code is better than maintaining custom (un)marshal functions.

As for breaking the API (which I take to mean the command line) you can just alias both `--role` and `--user` to `--arn`.

cPu1:
Thanks for the PR @patstrom. The PR looks good, I just have a few comments on the implementation.

When an `iamidentitymapping` command is processed, the user-supplied ARN is validated to ensure it's either a user ARN or a role ARN, but further down (all the way down to `setIdentities`), the same checks are performed when adding or removing identities. Ideally, the ARN should be validated only once for the user-supplied value. 

The identity manipulation methods treat both types homogeneously, and this ends up having the `Valid()` and user vs role checks. If we have separate methods for manipulating the two different types of identities, we wouldn't need this.

rndstr:
@patstrom i rebased and am currently fixing the tests, would you be able to [allow changes in your fork](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

PR #935, opened 2019-06-25
Fix broken kubernetes types #918

### LINKED ISSUE #918: Kubernetes-generated code is broken
**What happened?**

As a result of #886 , the kubernetes code-generator yields incorrect code.

The following command sequence shouldn't cause any errors:

```
make generate-kubernetes-types 
make build
```

**How to reproduce it?**

```
make generate-kubernetes-types 
make build
```

which results in:

```
pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go:374:23: val.DeepCopyinterface undefined (type interface {} is interface with no methods)
pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go:535:22: val.DeepCopyinterface undefined (type interface {} is interface with no methods)
```

* Similar issue: https://github.com/kubernetes/code-generator/issues/50
* ~This is blocking #917~ (Not anymore, I just reverted the generated code there) . My (poor) attempt at fixing the problem there: https://github.com/weaveworks/eksctl/pull/917/commits/ecca6d5e1c001591c96ddec99044d48be698b671 (which results in unit-test failures).
* To prevent this from happening again, I would strongly suggest adding CI check to make sure the generated code is always up to date,  see https://github.com/weaveworks/flux/blob/1.13.0/Makefile#L130-L132

martina-if:
fixes #918

- [x] Manually tested
- [ ] Double check that it works at runtime


2opremio:
Can you add a CI check so that broken types don't go unnoticed again?

For inspiration:  https://github.com/weaveworks/flux/blob/1.13.0/Makefile#L130-L132

errordeveloper:
> Can you add a CI check so that broken types don't go unnoticed again?

Yes, maybe we can just make it part of `make generate` (which already has a few `git diff` tests as part of it), I don't recall of a reason why we ended-up with a separate make target...

martina-if:
Yes, I was thinking if `generate` should be a precondition of `build` as well so we make sure we build with the right version of everything. What do you think?

errordeveloper:
> Yes, I was thinking if `generate` should be a precondition of `build` as well so we make sure we build with the right version of everything. What do you think?

Ah, that's a bit different from what I was suggesting...

Right now we run `generate` before `test`, so I thought we could just merge `generate-kubernetes-types` and `generate`, so it will get triggered when you run `make test`.

However, some of things in `generate` are time-consuming, specifically the mock generator (which kind rarely changes anything, only when AWS SDK changes). I would rather not add `generate` to `build`, maybe we can separate `make generate` and `make generate-mocks` (and make that dependency of `make test`). How does that sound?

martina-if:
Take a look as it is now. I just added `generate-kubernetes-types` to `build` and left `generate` only called by `test`.

martina-if:
@errordeveloper I renamed `generate` to `generate-mocks` and added the generation of k8s types to 
 `build`

errordeveloper:
Maybe we should just break them all down, I'll make a PR against your branch...

errordeveloper:
Generating these types seems to be pretty slow, maybe we shouldn't include them in `make build`...

errordeveloper:
https://github.com/martina-if/eksctl/pull/1

martina-if:
@errordeveloper how can I test that I didn't break anything by changing the k8s types?

errordeveloper:
@martina-if what do you mean exactly?

Also, looks like I messed up a little will push another commit now.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -547,6 +548,16 @@ type (
 // NodeGroupKubeletConfig contains extra config parameters for the kubelet.yaml
 type NodeGroupKubeletConfig map[string]interface{}
 
+// DeepCopy is needed to generate kubernetes types for NodeGroupKubeletConfig
+func (in *NodeGroupKubeletConfig) DeepCopy() *NodeGroupKubeletConfig {
```
martina-if: @errordeveloper how do I make sure that by introducing this I didn't break anything? What is the eksctl command to run to exercise this `DeepCopy()` functions?

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
```diff
@@ -1,20 +1,19 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright The Kubernetes Authors.
+Copyright 2018 Weaveworks. All rights reserved.
```
martina-if: @errordeveloper @2opremio Is this change ok? seems to be done automatically now.

errordeveloper: Yes, that's fine.

pkg/apis/eksctl.io/v1alpha5/types.go
errordeveloper: I am not so sure to be honest, I'm surprised this didn't break anything, maybe we need to look into it, but let's  open an issue and review that separately. I'd rather get this PR over the line as it stands first.

errordeveloper: Maybe it's possible generator behaviour just changed a little. Maybe the runtime used to deep-copy certain things more auto-magically and became more explicit? Fons bumped the version in his PR earlier.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -547,6 +548,16 @@ type (
...
+	if in == nil {
+		return nil
+	}
+	out := new(NodeGroupKubeletConfig)
+	*out = NodeGroupKubeletConfig(runtime.DeepCopyJSON(*in))
```
errordeveloper: Ah, yes - looks like it just handled `map[string]interface{}` before, and now it wants to use `runtime.DeepCopyJSON` for it.

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
```diff
@@ -382,6 +386,10 @@ func (in *NodeGroup) DeepCopyInto(out *NodeGroup) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.KubeletExtraConfig != nil {
+		in, out := &in.KubeletExtraConfig, &out.KubeletExtraConfig
```
errordeveloper: This is related to the above.

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
```diff
@@ -157,6 +156,11 @@ func (in *ClusterMeta) DeepCopy() *ClusterMeta {
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *ClusterNAT) DeepCopyInto(out *ClusterNAT) {
 	*out = *in
+	if in.Gateway != nil {
+		in, out := &in.Gateway, &out.Gateway
+		*out = new(string)
```
errordeveloper: Also, looks like it got more explicit about strings...

errordeveloper:
I think it should work now! 🤞:)

errordeveloper:
```
Generating deepcopy funcs
F0626 09:22:05.840410     686 main.go:82] Error: Failed executing generator: some packages had errors:
errors in package "./pkg/apis/eksctl.io/v1alpha5":
unable to format file "/src/pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go" (15:33: illegal character U+005C '\').
```
Such an exciting error! I'll continue working on this 🤣

martina-if:
I added the `panic()` but I don't see any panics when running the unit tests...

martina-if:
Also, this doesn't happen on my machine `unable to format file "/src/pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go" (15:33: illegal character U+005C '\').` And I don't see that character in the file. Could it be that the os separator in circleci has changed from `/` to ` \`?

errordeveloper:
The license header file was different in docker, as  `echo` in busybox doesn't expand `\n`, so I changed it for `printf`. It should pass now!

Let's see how else we can trace the callsights of the generated code. I think it would be best to do it separately from this PR - do you agree?

martina-if:
> Let's see how else we can trace the callsights of the generated code. I think it would be best to do it separately from this PR - do you agree?

Sure

Makefile
```diff
@@ -50,11 +50,9 @@ lint: ## Run linter over the codebase
 	"$(GOBIN)/gometalinter" ./pkg/... ./cmd/... ./integration/...
 
 .PHONY: test
-test: generate ## Run unit test (and re-generate code under test)
+test: ## Run unit test (and re-generate code under test)
 	$(MAKE) lint
-	git diff --exit-code pkg/nodebootstrap/assets.go > /dev/null || (git --no-pager diff pkg/nodebootstrap/assets.go; exit 1)
-	git diff --exit-code ./pkg/eks/mocks > /dev/null || (git --no-pager diff ./pkg/eks/mocks; exit 1)
-	git diff --exit-code ./pkg/addons/default > /dev/null || (git --no-pager diff ./pkg/addons/default; exit 1)
+	$(MAKE) generate-aws-mocks-test generate-bindata-assets-test generate-kubernetes-types-test
```
2opremio: @martina-if  @errordeveloper Making `generate-kubernetes-types-test` here causes the code generation (`generate-kubernetes-types`) to happens twice (both here and in `build`).

I think it can be done so that it's only invoked only once, either by removing the duplicated dependency or by depending on the generated type,

Also I think it's bad practice to invoke make from make, that's what dependencies are for.

errordeveloper: Ok, let's fix it. The thing is that dependencies get invoked in parallel if all are phony, so this is is just a lazy option... Maybe we can do something without declaring all the files as dependencies?

2opremio: We don't need to declare all the files, just the `zz_deepcopy.go` one I think.

Can we also use explicit dependencies instead of invoking make. Pretty pretty please? :)

errordeveloper: Sure! But what I was trying to say is that when you have phony targets listed as dependencies make in parallel mode will call them all at once and may not get it right, so that was reason I called make directly... It is not a big deal, there is little use of parallel mode for us right now, go build does most of the work.

2opremio: @errordeveloper where did you resolve this?

errordeveloper: It wasn't fix, I just clicked 'resolve', as in - let's go ahead and fix it per proposal.

PR #941, opened 2019-06-25
Use pagination for DescribeNetworkInterfaces

### LINKED ISSUE #590: review use of pagination in AWS API requests
We should make use of pagination in all requests where we lookup multiple items, and SDK supports it.

For example, we currently use `ec2.DescribeNetworkInterfaces` here:
https://github.com/weaveworks/eksctl/blob/510fb269e62fbd68734df8ab655335ab014aa032/pkg/vpc/cleanup.go#L21-L36

This can easily return too many items, we should use `ec2.DescribeNetworkInterfacesPages` instead. I believe we need to update SDK for that, current version we use doesn't have it.

Also here we use `ec2.DescribeKeyPairs` in attempt of listing all key pairs, but there is no `ec2.DescribeKeyPairsPages`.

https://github.com/weaveworks/eksctl/blob/ee7e72febc80ff960e37648dd32c058f574567d1/pkg/eks/auth.go#L114-L131

cPu1:
### Description

Closes #590 

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/vpc/cleanup.go
+++ b/pkg/vpc/cleanup.go
@@ -9,8 +9,6 @@ import (
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 
-	"context"
-
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 )
@@ -20,7 +18,7 @@ func fmtSecurityGroupNameRegexForCluster(name string) string {
 	return fmt.Sprintf(ourSecurityGroupNameRegexFmt, name)
 }
 
-func findAvailableNetworkInterfaces(ctx context.Context, ec2API ec2iface.EC2API, spec *api.ClusterConfig, fn func(eniID string) error) error {
+func forEachDanglingENI(ec2API ec2iface.EC2API, spec *api.ClusterConfig, fn func(eniID string) error) error {
 	input := &ec2.DescribeNetworkInterfacesInput{
 		Filters: []*ec2.Filter{
 			{
@@ -39,21 +37,23 @@ func findAvailableNetworkInterfaces(ctx context.Context, ec2API ec2iface.EC2API,
 		return errors.Wrap(err, "failed to create security group regex")
 	}
 
-	var lastErr error
+	var fnErr error
 
-	err = ec2API.DescribeNetworkInterfacesPagesWithContext(ctx, input, func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) bool {
+	err = ec2API.DescribeNetworkInterfacesPages(input, func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) bool {
 		for _, eni := range output.NetworkInterfaces {
 			id := *eni.NetworkInterfaceId
 			for _, sg := range eni.Groups {
-				if securityGroupRE.MatchString(*sg.GroupName) {
-					logger.Debug("found %q, which belongs to our security group %q (%s)", id, *sg.GroupName, *sg.GroupId)
-					if err := fn(id); err != nil {
-						lastErr = err
-						return false
-					}
-					break
+				if !securityGroupRE.MatchString(*sg.GroupName) {
+					logger.Debug("found %q, but it belongs to security group %q (%s), which does not appear to be ours", id, *sg.GroupName, *sg.GroupId)
+					continue
 				}
-				logger.Debug("found %q, but it belongs to security group %q (%s), which does not appear to be ours", id, *sg.GroupName, *sg.GroupId)
+
+				logger.Debug("found %q, which belongs to our security group %q (%s)", id, *sg.GroupName, *sg.GroupId)
+				if err := fn(id); err != nil {
+					fnErr = err
+					return false
+				}
+				break
 			}
 		}
 		return !lastPage
@@ -63,13 +63,12 @@ func findAvailableNetworkInterfaces(ctx context.Context, ec2API ec2iface.EC2API,
 		return errors.Wrapf(err, "unable to list dangling network interfaces in %q", spec.VPC.ID)
 	}
 
-	return lastErr
+	return fnErr
 }
 
 // CleanupNetworkInterfaces finds and deletes any dangling ENIs
 func CleanupNetworkInterfaces(ec2API ec2iface.EC2API, spec *api.ClusterConfig) error {
-	ctx := context.TODO()
-	return findAvailableNetworkInterfaces(ctx, ec2API, spec, func(eniID string) error {
+	return forEachDanglingENI(ec2API, spec, func(eniID string) error {
 		input := &ec2.DeleteNetworkInterfaceInput{
 			NetworkInterfaceId: &eniID,
 		}
--- a/pkg/vpc/cleanup.go
+++ b/pkg/vpc/cleanup.go
@@ -9,6 +9,8 @@ import (
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 
+	"context"
+
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 )
 
@@ -17,7 +19,7 @@ func fmtSecurityGroupNameRegexForCluster(name string) string {
 	return fmt.Sprintf(ourSecurityGroupNameRegexFmt, name)
 }
 
-func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.ClusterConfig) ([]string, error) {
+func findAvailableNetworkInterfaces(ctx context.Context, provider api.ClusterProvider, spec *api.ClusterConfig, fn func(eniID string) error) error {
 	input := &ec2.DescribeNetworkInterfacesInput{
 		Filters: []*ec2.Filter{
 			{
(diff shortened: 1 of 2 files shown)


pkg/vpc/cleanup.go
```diff
@@ -30,49 +33,50 @@ func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.Clus
...
-		for _, sg := range eni.Groups {
-			if securityGroupRE.MatchString(*sg.GroupName) {
-				logger.Debug("found %q, which belongs to our security group %q (%s)", id, *sg.GroupName, *sg.GroupId)
-				networkInterfaces = append(networkInterfaces, id)
-				break
-			} else {
+
+	var lastErr error
+
+	err = ec2API.DescribeNetworkInterfacesPagesWithContext(ctx, input, func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) bool {
```
errordeveloper: Do you think we need to pass context? Take a look at these:
```
pkg/cfn/manager/api.go:	if err := c.provider.CloudFormation().ListStacksPages(input, pager); err != nil {
pkg/cfn/manager/api.go:	if err := c.provider.CloudFormation().DescribeStackEventsPages(input, pager); err != nil {
pkg/cfn/manager/api.go:	if err := c.provider.CloudTrail().LookupEventsPages(input, pager); err != nil {
```

Whichever way we decide, let's make sure we use the same pattern in all instances of paged requests :)

pkg/vpc/cleanup.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 
+	"context"
```
2opremio: I don't know if this was intentionally but the customary way to order imports is:

```
stdlibimports
localmoduleimports
external imports
```

@errordeveloper is this supposed to be different for `eksctl`

(BTW, this is configurable in Goland, in case that's the editor you are using)

pkg/vpc/cleanup.go
```diff
@@ -17,7 +20,7 @@ func fmtSecurityGroupNameRegexForCluster(name string) string {
 	return fmt.Sprintf(ourSecurityGroupNameRegexFmt, name)
 }
 
-func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.ClusterConfig) ([]string, error) {
+func findAvailableNetworkInterfaces(ctx context.Context, ec2API ec2iface.EC2API, spec *api.ClusterConfig, fn func(eniID string) error) error {
```
2opremio: This goes through all the network interfaces, right? Is so, I would rename things like this:

```suggestion
func visitAvailableNetworkInterfaces(ctx context.Context, ec2API ec2iface.EC2API, spec *api.ClusterConfig, visitor func(eniID string) error) error {
```

pkg/vpc/cleanup.go
```diff
@@ -30,49 +33,50 @@ func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.Clus
...
-	return networkInterfaces, nil
+
+	return lastErr
 }
 
-func deleteNetworkInterfaces(provider api.ClusterProvider, networkInterfaces []string) error {
-	for _, eni := range networkInterfaces {
+// CleanupNetworkInterfaces finds and deletes any dangling ENIs
+func CleanupNetworkInterfaces(ec2API ec2iface.EC2API, spec *api.ClusterConfig) error {
+	ctx := context.TODO()
```
2opremio: I would pass a context to `CleanupNetworkInterfaces` instead, or, if you are certain we can wait indefinitely (which I doubt) use `context.Background()` instead of `context.TODO()`

pkg/vpc/cleanup.go
```diff
@@ -30,49 +33,50 @@ func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.Clus
...
 		}
+		return !lastPage
+	})
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to list dangling network interfaces in %q", spec.VPC.ID)
 	}
-	return networkInterfaces, nil
+
+	return lastErr
```
2opremio: Why are we interested in the last error? Shouldn't we simply exit early on the first error?

pkg/vpc/cleanup.go
errordeveloper: Yes, I prefer to have it in that sort of order. I do tend to break up external into k8s.io vs anything else. We might have some incosistency, and ordering of the groups should be made more consistent, but we do have `stdlibimports` at the top all the time.

I think VSCode respect this in some way, but I didn't know it can configurable, all IDEs probably use `goimports` (or what was it called?).

pkg/vpc/cleanup.go
cPu1: We do exit on error from the callback function `fn` https://github.com/weaveworks/eksctl/blob/f539f63495f13d76f70d5bc9aa2852986605b9ff/pkg/vpc/cleanup.go#L50 but that error needs to be propagated to the call to `DescribeNetworkInterfacesPagesWithContext` because the signature of its callback function (`func (output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) bool`) does not support an error return value.

pkg/vpc/cleanup.go
cPu1: Good point. This is from `goimports` but apparently it did not group this with the standard library imports because there was a trailing line before the package.

pkg/vpc/cleanup.go
cPu1: It goes through available network interfaces belonging to the cluster's VPC. `visit*` to me sounds more like something used to traverse a tree or graph, and `visitor` (non-verb) as a function name sounds like an interface that'd have a `Visit` method. By convention, most libraries, including the standard library, use just `fn` to denote a callback function.
`forEach*`, IMO, may be more appropriate here.

Also, @errordeveloper what do you think about using *DanglingNetworkInterfaces instead of *AvailableNetworkInterfaces?

pkg/vpc/cleanup.go
cPu1: The idea was to have `CleanupNetworkInterfaces` eventually take a context when it's available in the calling function. Since a context isn't available in the calling function, not sure it'd make sense to move this there.
`context.TODO()` is used here to be caught by a linter (if a rule is configured) and as a reminder to supply a meaningful context when it's available.

2opremio: I would propose to fix it in the calling function now. If it is too much work, please create an issue and fix it soon after this PR.

cPu1: If by fixing it, you mean making the `context.TODO()` or `context.Background()` call from the calling function and passing it, then it's a minor change. If we actually want a useful context to be passed in from the top-level caller and supply that to other function calls that support it (which is what I meant from my comment), that'd require more changes.

I can do the former in this PR but I don't understand the reasoning behind it.

2opremio: The reasoning behind it is not to make the situation worse than it was. I am happy to drop it if the overall context issue is fixed right after this PR.

pkg/vpc/cleanup.go
2opremio: Ah, I see, sorry I missed that. Would you mind renaming the variable so that it is less misleading? (it's not really the last error)

pkg/vpc/cleanup.go
cPu1: I think we should address all context issues in a separate PR as it's more involved and revert the changes here to use the non-context function calls. I'll create a separate issue for reviewing usage of context. WDYT?

pkg/vpc/cleanup.go
cPu1: Yeah, I agree that it's not that clear. What do you suggest renaming it to? I've seen `fnErr` used in some cases.

martina-if:
Do we have tests for this? because it would be nice to have those and see that this change didn't break the behavior of the cleanup function (or the find* function).

pkg/vpc/cleanup.go
```diff
@@ -30,49 +33,50 @@ func findAvailableNetworkInterfaces(provider api.ClusterProvider, spec *api.Clus
...
+		for _, eni := range output.NetworkInterfaces {
+			id := *eni.NetworkInterfaceId
+			for _, sg := range eni.Groups {
+				if securityGroupRE.MatchString(*sg.GroupName) {
+					logger.Debug("found %q, which belongs to our security group %q (%s)", id, *sg.GroupName, *sg.GroupId)
+					if err := fn(id); err != nil {
+						lastErr = err
+						return false
+					}
+					break
```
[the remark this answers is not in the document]
errordeveloper: Yes, we need to find all of them!

pkg/vpc/cleanup.go
errordeveloper: `fnErr` would be fine, I think :)

pkg/vpc/cleanup.go
errordeveloper: Yes, please just use whatever the default context SDK uses, like we have in other instances:
```
pkg/cfn/manager/api.go:	if err := c.provider.CloudFormation().ListStacksPages(input, pager); err != nil {
pkg/cfn/manager/api.go:	if err := c.provider.CloudFormation().DescribeStackEventsPages(input, pager); err != nil {
pkg/cfn/manager/api.go:	if err := c.provider.CloudTrail().LookupEventsPages(input, pager); err != nil {
```

pkg/vpc/cleanup.go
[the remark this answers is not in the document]
errordeveloper: Yes, attempting deletions on the spot has proven to be messy in general. We might want to go all the way and create a list of deletion tasks like we do for most other things.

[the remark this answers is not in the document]
martina-if: So then, having this return an array can be a good API for now and for later if we decide to do that.

pkg/vpc/cleanup.go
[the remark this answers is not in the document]
cPu1: Good question. We stop because we only need to check if the network interface has at least one security group matching the regex. After invoking the callback function, the ENI would no longer exist and if another security group for the same ENI matches the regex, the callback function fail to delete the already-deleted ENI.

pkg/vpc/cleanup.go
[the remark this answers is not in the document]
cPu1: The default page size I believe is around 1000. The use of pagination methods here is to handle the edge case where more than 1000 items exist. If we construct an array of matching items from all pages, all of them will be held in memory (granted, after filtering, they'd be a lot less).
The motivation is to avoid having to create a slice of matching ENIs and do the deletion in a single pass through the loop, instead of running another loop for deleting the items.

So for the normal case where less than 1000 items exist, this code is somewhat identical to the original code with the difference that it deletes ENIs as it finds them rather than creating a slice of matching items and running another loop.

>It also makes error handling and testing easier.

I agree that the code for testing this would be a bit different than the other test cases.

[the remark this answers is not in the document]
martina-if: I see your point that this is more efficient. I typically prefer readability and simplicity over performance, unless this is a very slow operation. Also, we don't have a garbage collection problem so I don't think creating new objects/arrays should be an issue.

I really value code maintenance, specially since we have more people working on this project and external contributors, so simplicity means less bugs, and faster ramp up for new joiners.

I won't block this as it is but I would ask you to consider this option for the reasons I mentioned :)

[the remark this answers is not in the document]
2opremio: I'm with Martina ok this one. Readability and maintenance over performance unless there is empirical evidence of performance being an actual problem.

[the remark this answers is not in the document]
cPu1: @martina-if I agree with your points. I'll make the proposed changes.

martina-if:
LGTM! :tada:

PR #944, opened 2019-06-26
Speed CI and local builds up

### LINKED ISSUE #920: Speed CI up
CircleCI is taking ~8 minutes to validate PRs. We should improve this:

1. The change->validate loop should ideally be much faster (particularly for code reviews).
2. In the future, we plan to run integration tests (or a simplified version of them such a smoke test) in PRs. The build times should have room for this.

2opremio:
This change speeds CI builds up by using a docker image (`weaveworks/eksctl-build:deps-VERSION`) as dependency-cache and CircleCI's caching system for caching go modules (`$GOPATH/pkg/mod`) and the go build artifacts (`go env GOCACHE`). Fixes #920

Dependencies don't change all that often, so caching them should save us a lot of time.

The change also speeds local builds and tests up since:
1. We no longer need to generate code if it's not needed (there are proper file dependencies in place now)
2. I disabled test coverage (it wasn't used) allowing go to cache test results (see 
https://github.com/golang/go/issues/23565 )

Finally, the change makes building the eksctl-image faster since the local go caches are mounted when building.

CI build times are:
* [With this change](https://circleci.com/gh/weaveworks/eksctl/3407): **~4.5 minutes**
* [Master before this change](https://circleci.com/gh/weaveworks/eksctl/3324) (with go modules, after removing the `vendor` directories): **~12 minutes**
* [Master before go modules were introduced](https://circleci.com/gh/weaveworks/eksctl/3251) (with `vendor` directories and dependency caching): **~7 minutes**

So, this change brings a ~66%  speed improvement compared to master and a ~42% improvement compared to before go modules were introduced. Things will get even faster once https://github.com/weaveworks/eksctl/pull/935#discussion_r297793336 is addressed.

~This change follows @errordeveloper's vision of keeping caching independent of the CI-system. It is inspired by [this blog post](https://andrewlock.net/caching-docker-layers-on-serverless-build-hosts-with-multi-stage-builds---target,-and---cache-from/) and this [Stackoverflow question](https://stackoverflow.com/questions/55384975/build-go-dependencies-in-separate-docker-layer).~

For now the dependency-cache image needs to be manually pushed (which I did), just like it was done before with `weaveworks/eksctl-build:latest`, which isn't ideal doesn't worsen the current situation. I will automate it once https://github.com/weaveworks/corp/issues/350 is resolved.

~PS: @errordeveloper FYI, using Docker images as a caching layer is a neat idea. However, Docker is not really made for this: the downsides (it took a couple of days to get it right, namely due to the file hashing invalidation and image pulling is slow in comparison with a caching-optimized storage) don't make it up for the upsides (having a CI-system-independent caching system). I think it would had taken me way less effort to use the CircleCI cache (see https://github.com/weaveworks/flux/pull/2078 ). Well, at least I learned about layer caching in Docker ...~


errordeveloper:
Awesome, thanks!

2opremio:
I've sprinkled the makefile with `time` invocations to see why the build/testing still takes so long. The plan is to remove it once we optimize it.

Makefile
```diff
@@ -22,7 +23,7 @@ install-build-deps: ## Install dependencies (packages and tools)
 
 .PHONY: build
 build: generate-bindata-assets generate-kubernetes-types  ## Build eksctl
-	CGO_ENABLED=0 go build -tags "$(GO_BUILD_TAGS)" -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
+	CGO_ENABLED=0 time go build -v -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
```
martina-if: should we remove the `time` before merging?

Makefile
```diff
@@ -22,7 +23,7 @@ install-build-deps: ## Install dependencies (packages and tools)
 
 .PHONY: build
 build: generate-bindata-assets generate-kubernetes-types  ## Build eksctl
-	CGO_ENABLED=0 go build -tags "$(GO_BUILD_TAGS)" -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
```
martina-if: What were the tags for?

Makefile
2opremio: Maybe, I would like to leave them there until I find the culprit though

Makefile
2opremio: Ancient legacy.

See the commit message at https://github.com/weaveworks/eksctl/pull/944/commits/64f63b1424d4fb6ba4e23112466851e4d37d165b

.gitignore
```diff
@@ -31,3 +31,5 @@ integration/kubeconfig-*
 # Generated header file
 .license-header
 
+# Go packages eksctl depends on, for the build cache
+go-deps.txt
```
errordeveloper: Would you mind calling it `.go-deps` or `.go-dependencies` or something? Dot-TXT is so Windows-y... ;)

Makefile
```diff
@@ -180,9 +172,33 @@ else
 EKSCTL_IMAGE_BUILD_ARGS += --build-arg=TEST_TARGET=test
 endif
 
+.PHONY: eksctl-deps-image
+eksctl-deps-image: ## Create a cache image with all the build dependencies
+	@# Pin dependency file permissions.
+	@# Docker uses the file permissions as part of the COPY hash, which can lead to cache misses
+	@# in hosts with different default file permissions (umask).
+	chmod 0600 go.mod go.sum
+	chmod 0700 install-build-deps.sh
```
errordeveloper: Alternatively, we could tar up the context and pass to docker via stdin, I've seen it done in LinuxKit. Just FYI.

Makefile
errordeveloper: Modern build system measure time for you... Just saying ;)

I'm happy with make for this project, mostly because it's what everyone is familiar with.

2opremio:
The main bottlenecks in the build currently are:
* Linting, which takes a ridiculous amount of time (2 minutes)
* Code generation
* Duplicated things. Due to buggy make rules, we do the kubernetes code generation twice and call `make make build-integration-test`.

.gitignore
2opremio: I would rather have an explicit file rather than a hidden one. But sure ....

2opremio:
**Update:**

I have updated the PR to:
1.  Use CircleCI cache for persisting the project's build cache and modules, and a Docker image cache for the project dependencies. This simplifies the code (for instance, no need to generate dependencies) and brings a big performance boost.
2. Improve the Makefiles rules. Now no code is generated if there is no need to.


We are now at a 4.5-minute build mark!

I have updated the PR description and measurements accordingly.

I know @errordeveloper is not going to like the use of CircleCI, but persisting Go's build cache and modules for changing source code can't really be done with Docker image caching (i.e. `--cache-from`) due to its layer-matching strictness (as soon as a single input source-file is off the whole cache is invalidated).

@errordeveloper @martina-if PTAL

.gitignore
2opremio: This is no longer needed (we don't generate a deps file anymore)

Makefile
2opremio: Do you have a link?

.gitignore
errordeveloper: Very good, thanks!

Makefile
```diff
@@ -20,9 +20,11 @@ install-build-deps: ## Install dependencies (packages and tools)
 
 ##@ Build
 
-.PHONY: build
-build: generate-bindata-assets generate-kubernetes-types  ## Build eksctl
-	CGO_ENABLED=0 go build -tags "$(GO_BUILD_TAGS)" -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
+godeps_cmd = go list -deps -f '{{if not .Standard}}{{ $$dep := . }}{{range .GoFiles}}{{$$dep.Dir}}/{{.}} {{end}}{{end}}' $(1) | sed "s%${PWD}/%%g"
+godeps = $(shell $(call godeps_cmd,$(1)))
```
errordeveloper: Should this be an immediate assignment, or it's intended to be evaluated each time?

errordeveloper: Never mind, I see the usage now.

Makefile
```diff
@@ -112,78 +111,83 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
...
+DEEP_COPY_HELPER := pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
+GENERATED_FILES := pkg/addons/default/assets.go \
+pkg/nodebootstrap/assets.go \
+$(DEEP_COPY_HELPER) \
+pkg/ami/static_resolver_ami.go \
+site/content/usage/20-schema.md \
+$(AWS_SDK_MOCKS)
+
+.PHONY: regenerate-sources
+# TODO: generate-ami is broken (see https://github.com/weaveworks/eksctl/issues/949 ), include it when fixed
```
errordeveloper: It take forever as it goes into each region and I don't think it need optimising, it should just ran as a periodic job.

Makefile
2opremio: Yeah, you cannot evaluate it immediately since it has a parameter.

.circleci/config.yml
```diff
@@ -2,11 +2,36 @@ version: 2
...
       - checkout
+      - restore_cache:
+            keys:
+                - go-cache-{{ .Branch }}-{{ .Revision }}
+                - go-cache-{{ .Branch }}-
+                - go-cache-
+      - run:
+            name: Install Golang (needed for dependency calculation)
+            command: |
+                curl -OL https://storage.googleapis.com/golang/go${GO_VERSION}.linux-amd64.tar.gz
```
errordeveloper: Really?

Makefile
2opremio: alright

.circleci/config.yml
2opremio: Ehm, yes? I don't know what you mean by that. Please elaborate.

.circleci/config.yml
```diff
@@ -2,11 +2,36 @@ version: 2
 jobs:
   make-eksctl-image:
     # use machine build, as otherwise we cannot use make and have to duplicate makefile
-    machine: true
+    machine:
+      image: ubuntu-1604:201903-01
```
errordeveloper: Isn't there an image that include Go toolchain?

.circleci/config.yml
2opremio: Ah, you edited the comment.

2opremio: It's only used for the dependency computation in the makefile. Nothing to do with construction or the like.

If you want I can factor out `make eksctl-image`  in a script and avoid installing it altogether.

.circleci/config.yml
2opremio: This was done for other reasons (see https://github.com/weaveworks/eksctl/pull/944/commits/51cfdfc509cf714eef400df9f27f4ec6b2102802 ).

Regardless, all the images contain old versions of Go (i.e. <1.11) which don't support go modules.

Makefile
errordeveloper: No, LinuxKit has build system written in Go no.

But see the [docs](https://docs.docker.com/engine/reference/commandline/build/), there are multiple options of how you can pass the context.

.circleci/config.yml
errordeveloper: Maybe, but don't have to do it now. Can we just cache it, or install it with `apt`?

.circleci/config.yml
errordeveloper: I see, never mind :)

.circleci/config.yml
2opremio: It takes 5 seconds to install so it's not worth caching. Apt in Ubuntu 16 doesn't have a newer version and I am not sure it's worth the pain of installing another APT repo.

errordeveloper: Yeah, I see! Anyway, let's leave it as is, the tarball is hosted in GCP, which is quite fast and reliable. I'd hope we will find a nice CI solution one day ;)

PR #954, opened 2019-06-28
Update docs in docs branch on release

### LINKED ISSUE #461: make sure we don't break docs when we merge changes to master
When #460 got merged, we broke the config file samples in the `examples` and in `README.md`. We need to get better at this and only update docs when release is ready to use.

martina-if:
closes #461

- [ ] Manually tested


errordeveloper:
Let call the branch 'docs' or 'documentation', calling it 'gh-pages' only makes sense if use GitHub pages, but we are going to use Netlify :)

PR #957, opened 2019-07-01
Improve site build

errordeveloper:
- Add Hugo as build dependency
- Update `serve-page` and `build-pages` make targets


errordeveloper:
@martina-if let's make sure Netlify uses either `make build-pages`, or we add another target for it, if it needs some special stuff.

install-build-deps.sh
```diff
@@ -12,18 +12,16 @@ else
   OSARCH="linux-amd64"
 fi
 
-
+go install github.com/gohugoio/hugo
```
errordeveloper: @martina-if this actually installs it without SASS support, which requires CGO... Seems like we aren't using SASS, but I don't really know.

Makefile
```diff
@@ -202,23 +202,15 @@ release: eksctl-build-image ## Create a new eksctl release
...
-	-docker rm -f eksctl-jekyll
-	$(JEKYLL) jekyll serve -d /_site --watch --force_polling -H 0.0.0.0 -P 4000
+	cd site/ ; $(GOBIN)/hugo serve
 
 .PHONY: build-pages
-build-pages: ## Generate the site using jekyll
-	-docker rm -f eksctl-jekyll
-	$(JEKYLL) jekyll build --verbose
+build-pages: ## Generate the site
+	cd site/ ; $(GOBIN)/hugo
```
martina-if: `hugo --gc --minify`

install-build-deps.sh
martina-if: We are using SCSS

Dockerfile.deps
```diff
@@ -8,7 +8,10 @@ RUN apk add --no-cache \
       make \
       bash \
       gcc \
-      musl-dev
+      g++ \
+      libsass-dev \
+      musl-dev \
+      && true
```
errordeveloper: This got remove, it's nice to have it as it allows package names to be sorted easily.

2opremio: I intentionally removed it. It makes the installation not to fail if something goes wrong. It happened to me and I spent quite some time debugging the reason.

install-build-deps.sh
```diff
@@ -8,22 +8,23 @@ fi
 
 if [ "$(uname)" = "Darwin" ] ; then
   OSARCH="darwin-amd64"
+  env CGO_ENABLED=1 go install -tags extended github.com/gohugoio/hugo
 else
   OSARCH="linux-amd64"
+  env CGO_ENABLED=1 go install -tags extended -ldflags "-linkmode external -extldflags -static" github.com/gohugoio/hugo
```
2opremio: Please factor out the duplicated `go install` and move it out of the `if`. You can use a variable for the differing arguments.

2opremio:
Please add the Hugo dependency to `tools.go`, and don't forget to push the build cache once we merge.

Dockerfile.deps
```diff
@@ -31,7 +34,8 @@ RUN apk add --no-cache --initdb --root /out \
     ca-certificates \
     coreutils \
     git \
-    libc6-compat
+    libc6-compat \
+    && true
```
2opremio: Same here

Makefile
```diff
@@ -202,23 +202,17 @@ release: eksctl-build-image ## Create a new eksctl release
...
 serve-pages: ## Serve the site locally
-	-docker rm -f eksctl-jekyll
-	$(JEKYLL) jekyll serve -d /_site --watch --force_polling -H 0.0.0.0 -P 4000
+	cd site/ ; $(HUGO) serve $(HUGO_ARGS)
 
 .PHONY: build-pages
-build-pages: ## Generate the site using jekyll
-	-docker rm -f eksctl-jekyll
-	$(JEKYLL) jekyll build --verbose
+build-pages: ## Generate the site
```
2opremio: I don't know how Hugo works. But can we have explicit file prerequisites (and maybe even an explicit file target(s)) for this?

Dockerfile.deps
errordeveloper: > It makes the installation to fail silently if something goes wrong 

No, because it's `&& true`, not `|| true`.

Makefile
errordeveloper: It's really fast (under 1s). I don't see a reason, and it would be too much effort - everything in `site` is a dependency, and it outputs lots of files.

install-build-deps.sh
errordeveloper: Sure, will also double check if linker flags are really needed on Linux, I copied this from somewhere.

2opremio: Ah, you are right. My bad

Issue #959, opened 2019-07-01
Create Netlify account for the team


martina-if:
Right now the site is deployed by Netlify in my private account. I will set up a team in Netlify so that everyone in the team can manage it.

martina-if:
I sent them a message and I am waiting for their reply.

PR #962, opened 2019-07-01
Improve code highlighting and copy button

martina-if:
- Improved the highlighting by switching from pygments to highlight.js
- Improved the copy button's style
- Added Netlify credit to the land page footer

highlight

PR #963, opened 2019-07-02
Add credit to netlify

martina-if:
related to #959


martina-if:
Not needed anymore since #962 was merged

Issue #903, opened 2019-06-19
get rid of latest_release tag


errordeveloper:
TIL, thanks @mhausenblas, we learned that `latest_release` is no longer necessary, as Github has a fix URL for latest release (albeit it's not what you'd guess it would, but kind of similar).

We've used `latest_release` trick for a while in order to provide a one-line command to install binaries with curl. It looks like we can get rid of it now.

A canonical URL for release binaries is e.g.:

```
https://github.com/weaveworks/eksctl/releases/download/0.1.36/eksctl_Darwin_amd64.tar.gz
```

It is evident that github supports `releases/latest`, that is an automatic redirect to whichever is the latest release. However, one cannot use `releases/download/latest/eksctl_Darwin_amd64.tar.gz`.

What we have learned today is that the following URL works:
```
https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Darwin_amd64.tar.gz
```
The order of elements in the path is unintuitive, but it's a thing that is supported out of the box.

If we get rid of `latest_release` tag, we will make release page less confusing, and our release pipeline will be almost twice as fast.

errordeveloper:
There is one downside, however, this change can be rather annoying to users who have relied on our current `latest_release` URL.

errordeveloper:
We've had an issue with 0.1.37, which concerns the complexity of `latest_release` logic. Seems plausible to just stop doing this and simplify things. I'll have to double check if somehow we started using a newer version of goreleaser, or what else broke - but seems plausible to just reduce the complexity.

errordeveloper:
I think 37062ba3af2fa7e5e2b67cc4a58918a028788392 should fix the issue with goreleaser for now.

gemagomez:
We should communicate we are going to make this change, i.e. announce we will be deprecating this feature. Contact any users that may be using the link (and raise issues so they are aware of the change). Let's then do the change on a subsequent release.

errordeveloper:
I think new URL should be in our control, so we can change it again, if we have to. Just a redirect should be sufficient to begin with.

sayboras:
Might not be completed related to this, but just wonder furture release tagging will follow the same golang standard (i.e. `v`0.18.0 instead of only 0.18.0).

Understand that currently `eksctl` should be used with binary release only, however, I am trying to re-use some of code by adding eksctl as go module dependency. Without `v` prefix, I need to use pseudo version now. Let me know if you want me to raise new issue to track this point.

PR #966, opened 2019-07-02
Enable release candidates

### LINKED ISSUE #965: release automation should handle RCs
We are planning on making an RC this week, so will need to make sure our release automation can handle this.

errordeveloper:
Fixes #965

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,18 +191,24 @@ eksctl-image: eksctl-deps-image## Create the eksctl image
 
 ##@ Release
 
+docker_run_release_script = docker run \
+  --env=GITHUB_TOKEN \
+  --env=CIRCLE_TAG \
+  --env=CIRCLE_PROJECT_USERNAME \
+  --volume=$(CURDIR):/src \
+  --workdir=/src \
+    $(EKSCTL_BUILDER_IMAGE)
+
+.PHONY: release-candidate
+release-candidate: eksctl-image ## Create a new eksctl release candidate
+	$(call docker_run_release_script) ./do-release-candidate.sh
+
 .PHONY: release
-release: eksctl-build-image ## Create a new eksctl release
-	docker run \
-	  --env=GITHUB_TOKEN \
-	  --env=CIRCLE_TAG \
-	  --env=CIRCLE_PROJECT_USERNAME \
-	  --volume=$(CURDIR):/src \
-	  --workdir=/src \
-	    $(EKSCTL_BUILDER_IMAGE) \
-	      ./do-release.sh
+release: eksctl-image ## Create a new eksctl release
+	$(call docker_run_release_script) ./do-release.sh
 
 ##@ Site
+
 HUGO := $(GOBIN)/hugo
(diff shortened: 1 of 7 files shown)


do-release-candidate.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh -ex
```
errordeveloper: I figure I'll make a copy for now, as `do-release.sh` is already complicated and I'd rather not make changes to it yet. We should be able to merge after we get rid of `latest_release` non-sense (see #903).

tag-release-candidate.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
martina-if: Why not reuse the same tag-release script? The only special thing about this one I can see is the release notes file.

errordeveloper: There would be a few more conditions. I wanted to keep them separate for now, we can merge later. In fact after we drop `latest_release`, these scripts should become nearly identical.

do-release-candidate.sh
```diff
@@ -0,0 +1,16 @@
...
+if [ -z "${CIRCLE_PULL_REQUEST}" ] && [ -n "${CIRCLE_TAG}" ] && [ "${CIRCLE_PROJECT_USERNAME}" = "weaveworks" ] ; then
+  export RELEASE_DESCRIPTION="${CIRCLE_TAG}"
+  RELEASE_NOTES_FILE="docs/release_notes/${CIRCLE_TAG/-rc.*}.md"
+
+  if [[ ! -f "${RELEASE_NOTES_FILE}" ]]; then
+    echo "Release notes ${RELEASE_NOTES_FILE} not found. Exiting..."
+    return
+  fi
+
+  goreleaser release --skip-validate --config=./.goreleaser.yml --release-notes="${RELEASE_NOTES_FILE}"
```
martina-if: Will this push the RC to brew and other platforms?

Makefile
```diff
@@ -191,18 +191,24 @@ eksctl-image: eksctl-deps-image## Create the eksctl image
 
 ##@ Release
 
+docker_run_release_script = docker run \
+  --env=GITHUB_TOKEN \
+  --env=CIRCLE_TAG \
+  --env=CIRCLE_PROJECT_USERNAME \
+  --volume=$(CURDIR):/src \
+  --workdir=/src \
+    $(EKSCTL_DEPENDENCIES_IMAGE)
```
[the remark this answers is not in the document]
errordeveloper: Maybe it should, but I wasn't clear to me and I just did what worked.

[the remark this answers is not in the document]
martina-if: @2opremio

do-release-candidate.sh
errordeveloper: No

Makefile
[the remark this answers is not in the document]
errordeveloper: Now that I understand the purpose, I'll make use of it.

PR #871, opened 2019-06-12
snap

mikeroyal:
Includes  snapcraft.yaml to build [snap](https://snapcraft.io/).

### Description

Provides a yaml file to package the CLI for Amazon EKS  as a snap. Snap makes the setup and install process easier for the end user by auto-updating to new versions or rollback to previous version if the newest version breaks.
### Checklist

- [ ] Code compiles correctly (i.e `make build`)
- [ ] Added tests that cover your change (if possible)
- [ ] All unit tests passing (i.e. `make test`)
- [ ] All integration tests passing (i.e. `make integration-test`)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [ ] Added yourself to the `humans.txt` file

PR #969, opened 2019-07-03
Enable building snaps from source

dholbach:
Build a snap from eksctl source. @mikeroyal and I worked on this (https://github.com/weaveworks/eksctl/pull/871 and https://github.com/mikeroyal/eksctl/pull/1 are previous attempts at getting this in).

### Description

To build the snap, run

```sh
# Prep
sudo snap install snapcraft --classic
sudo snap install multipass --beta --classic

# Actual build
snapcraft
```

The resulting snap works nicely for me.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -228,6 +228,60 @@ This allows the message to be easier to read on GitHub as well as in various git
 7. Ensure the release was successfully [published in Github](https://github.com/weaveworks/eksctl/releases).
 8. Download the binary just released, verify its checksum, and perform any relevant manual testing.
 
+### Releasing snaps of eksctl
+
+Snaps are software packages which run on a variety of Linux flavours.
+
+*Note:* This is still somewhat [TBD](https://github.com/weaveworks/eksctl/issues/215).
+
+Setting up the environment to build snaps on e.g. Ubuntu:
+
+```sh
+sudo apt install snapd
+sudo snap install snapcraft --classic
+sudo snap install multipass --classic
+```
+
+Building the snap can be done by running this command in the top-level directory:
+
+```sh
+snapcraft
+```
+
+*Note:* By default the `grade` of the snap is automatically set to `devel` and thus cannot be released to the e.g. `stable` channel of the snap. If you want to release a stable version, you want to check out the tag first, so e.g. `git checkout 0.12.0` and then run `snapcraft`.
+
+Testing the resulting package can be done via:
+
+```sh
+sudo snap install eksctl_<version>_amd64.snap --dangerous
+```
+
+The `--dangerous` flag is required as it's a locally built snap and doesn't come from the store.
+
+Publishing the snap can be done by following these steps.
+
+Login to the Snap Store:
+
+```sh
+snapcraft login
+```
+
+Or
+
+```sh
+snapcraft login --with <login-file>
+```
+
+Where `<login-file>` was generated via `snapcraft export-login`.
+
+Then publish via:
+
+```sh
+snapcraft push eksctl_<version>_amd64.snap --release [stable,beta,candidate,edge]
+```
+
+TODO: Further automate these steps in CircleCI, etc.
+
 ### Notes on Integration Tests
 
 It's recommended to run containerised tests with `make integration-test-container TEST_V=1 AWS_PROFILE="<AWS profile name>"`. The tests require:
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ integration/kubeconfig-*
 
 # direnv's files:
 .envrc
+
+# Snaps
+eksctl_*_*.snap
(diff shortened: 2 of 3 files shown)


dholbach:
This would need to be hooked up with CI and all, but this should be first steps for allowing us to 

- push directly to the snap store (without gatekeepers)
- get packages out to all these distros: https://docs.snapcraft.io/installing-snapd
- keep users automatically up to date
- have different [release channels](https://docs.snapcraft.io/channels)

errordeveloper:
We should integrate this as part of our release process, goreleaser supports snaps. I believe this simply implies we need make `snap/snapcraft.yaml` part of `.goreleaser.yml`,  probably in a manner similar to how `.goreleaser.brew.yml` is done.

dholbach:
Thanks Ilya for the feedback. I'm not sure what the best way is to hook this up with goreleaser. :man_shrugging:

errordeveloper:
My comment above was indeed to explain the direction, did you replied
before I added extra details to it?

dholbach:
Ah ok. Shall we do two separate PRs for this?

errordeveloper:
@dholbach are you interested in finishing this?

dholbach:
I was thinking we could land this piece of work first, then move on to `goreleaser` et al in a second PR. It's what I'm unfamiliar with.

dholbach:
Can we maybe just merge this and then do second PR with whatever's required to enable the build and publication of the snap? Goreleaser has support for `snapcraft`, but it's [not complete](https://github.com/goreleaser/goreleaser/issues/914).

dholbach:
Either we use build.snapcraft.io for this (the GH account to set this up will [need to be admin](https://github.com/canonical-web-and-design/build.snapcraft.io/issues/1191) unfortunately - to be able to set up the webhook).

If that's not an option, we could try to roll our own and do a:

```console
# Prep
sudo apt install snapd
sudo snap install snapcraft --classic
sudo snap install multipass --beta --classic

# Actual build
snapcraft
snapcraft push eksctl_<version>_amd64.snap
````

For the last step we'd need to login to the store, etc.

**Comment by dholbach** [comment]
@martina-if Thanks for assigning.

On this one I think we're blocked because there was the idea to use `goreleaser` which I don't know anything about.

What I'd suggest:

1. I update this PR
1. we merge it
1. we create a shared account on https://snapcraft.io/
1. I talk to the Snap Store maintainers, so we can take over https://snapcraft.io/eksctl from the inactive developer
1. we figure out a way to do this in CI later (using build.snapcraft.io for this unfortunately requires pretty wide access to the Github repo to set up a webhook for us and there's no manual way to do it: https://github.com/canonical-web-and-design/build.snapcraft.io/issues/1191)

The last point we haven't figured out for `fluxctl` either yet. I push snaps to the store manually every now and then. The workflow generally is:

```
# Prep
sudo snap install snapcraft --classic
sudo snap install multipass --classic

# Actual build
snapcraft

# Publishing
snapcraft push eksctl_<version>_amd64.snap --release [stable|beta|edge}
```

Let me know if it's unclear and if you have different thoughts.

**Comment by dholbach** [comment]
Updated. The PR is good to go AFAICS. Let's decouple the snap build bits from the publication in the store.

**Comment by martina-if** [comment]
Hi @dholbach thanks for updating this PR! If you think merging this is the best first step let's do that. Before that, can you please add those instructions here? https://github.com/weaveworks/eksctl/blob/master/CONTRIBUTING.md#release-process

**Comment by dholbach** [comment]
@martina-if I documented things TTBOMK. Please review.

Our options for publishing (once we took over maintenance of the snap):

- publish manually whenever we please
- publish through CI through a mechanism we can figure out ourselves
- publish using build.snapcraft.io (which would require us to at least temporarily give admin privs of the GH repo to the build service, so it can set up a webhook - not sure if that's acceptable for us)

File: CONTRIBUTING.md [hunk 1]
```diff
@@ -228,6 +228,60 @@ This allows the message to be easier to read on GitHub as well as in various git
...
+sudo snap install multipass --classic
+```
+
+Building the snap can be done by running this command in the top-level directory:
+
+```sh
+snapcraft
+```
+
+*Note:* By default the `grade` of the snap is automatically set to `devel` and thus cannot be released to the e.g. `stable` channel of the snap. If you want to release a stable version, you want to check out the tag first, so e.g. `git checkout 0.12.` and then run `snapcraft`.
```
martina-if: nit: `git checkout 0.12.` -> `git checkout 0.12.0`

CONTRIBUTING.md
```diff
@@ -228,6 +228,60 @@ This allows the message to be easier to read on GitHub as well as in various git
...
+
+Where `<login-file>` was generated via `snapcraft export-login`.
+
+Then publish via:
+
+```sh
+snapcraft push eksctl_<version>_amd64.snap --release [stable,beta,candidate,edge]
+```
+
+We might want to automate some of these steps obviously. This section will likely be shorter at some stage. :-)
```
martina-if: I would remove this line or say something like:

```
TODO: automate these steps on circleci
```

PR #981, opened 2019-07-04
Pin base dependencies image

2opremio:
This is needed to achieve reproducible builds and avoid unexpected build cache misses.

Unfortunately the `golang` dockerhub image doesn't tag minor versions of Alpine
(e.g. it provides golang:1.12.6-alpine3.9 but not golang:1.12.6-alpine3.9.4
making all the images mutable when patches are released).

Thus, we ping the image digest instead of a tag.

As a side note: there is no need to bump the version of the dependencies image
(`EKSCTL_DEPENDENCIES_IMAGE` in the Makefile) since the change doesn't affect
layering.

Followup on #920


errordeveloper:
Merging, as I'm about to tag 0.1.39-rc.1 and should include this.

PR #988, opened 2019-07-05
Force latest version of github.com/docker/distribution

errordeveloper:
### Description

We previously used github.com/docker/distribution@edc3ab29cdff,
and the build broke when goreleaser was updated, as somehow
go mod picked github.com/docker/distribution@v2.6.2.
We force latest version as it's not too different from one we
used before, but it fully supports go mod.

### Checklist

- [x] Manually tested


errordeveloper:
Also need to bump dependencies image.

go.mod
```diff
@@ -92,4 +89,9 @@ require (
 	sigs.k8s.io/yaml v1.1.0
 )
 
-replace github.com/awslabs/goformation => github.com/errordeveloper/goformation v0.0.0-20190507151947-a31eae35e596
+replace (
+	github.com/awslabs/goformation => github.com/errordeveloper/goformation v0.0.0-20190507151947-a31eae35e596
+	// it appears that we pick wrong version of github.com/docker/distribution, and it breaks k8s.io/kubernetes
```
2opremio: Can you be more specific? The problem is that the version of go releaser we depend on uses a version of docker distribution which is in compatible with the version of k8s.io/kubernetes we use.

errordeveloper: That's what I said, but sure.

errordeveloper: To be honest, I'd rather move along and release out, changing this comment will take several minutes as I'll have to rebuild the dependencies image and push it (which takes 10m at least)...

errordeveloper:
@2opremio I've clarified the comment with respect to exact version of `k8s.io/kubernetes` that breaks, I'll merge now as we should make this release happen before it's Friday afternoon. Please feel free to amend the comment later if you think it's not as clear, I understand that for maintenance purpose documenting non-trivial dependency declarations is important, but it's clear enough as far as I can tell.

2opremio:
It's ok, thanks for updating the comment

PR #993, opened 2019-07-05
 Add support for `aws eks get-token` authenticator and IAM role ARN

### LINKED ISSUE #749: Feature request: Role option with eksctl utils write-kubeconfig
`aws-iam-authenticator`has a [role option on the token command](https://github.com/kubernetes-sigs/aws-iam-authenticator#4-set-up-kubectl-to-use-authentication-tokens-provided-by-aws-iam-authenticator-for-kubernetes) that's used for authentication in the kubeconfigs produced, and it would be useful if `eksctl utils write-kubeconfig` would allow for a role to be used in the produced config. `aws-iam-authenticator` assumes the given role before grabbing a token, which has the benefit of eschewing individual user assignments in the `aws-auth` configmap (see https://github.com/weaveworks/eksctl/issues/625). This would make the `write-kubeconfig` subcommand more useful and allow eksctl to support this approach to authentication.

A new option flag added to `eksctl utils write-kubeconfig` of a role arn to use for authentication, e.g. `-R --role-arn <ROLE_ARN>`. If included, the `aws-iam-authenticator token` calls in kubeconfigs should add `-r <ROLE_ARN>`.

### LINKED ISSUE #788: Update KubeConfig Generation to use aws cli get-token
This will remove some dependencies for you to install when using `eks` and `eksctl`.

Change the Kubeconfig generation to use the `get-token` command once it's merged and released - https://github.com/aws/aws-cli/pull/4141

cPu1:
### Description

This PR adds support for:
1. Using `aws eks get-token` as the authenticator for signing requests if neither `aws-iam-authenticator` nor `heptio-authenticator-aws` is available in `PATH`.
2. Specifying a role ARN to use for `aws-iam-authenticator` and `aws eks get-token`, to `eksctl create cluster` and `eksctl utils write-kubeconfig` via flag `--authenticator-role-arn`.

Support for `aws eks get-token` was added recently, so presence of `aws` in `PATH` does not imply `aws eks get-token` would work. Instead of inspecting the version or checking if it supports `get-token`, we simply use it as a fall back if other authenticators aren't available in `PATH` and `aws` is available (which is also the motivation for using `aws eks get-token` since most users would already have `aws` installed).

- Closes #788
- Closes #749 

### Checklist

- [x] Manually tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ or [environment variables][awsenv]. For more information read [AWS documentation
 [awsenv]: https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html
 [awsconfig]: https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html
 
-You will also need [AWS IAM Authenticator for Kubernetes](https://github.com/kubernetes-sigs/aws-iam-authenticator) command (either `aws-iam-authenticator` or `heptio-authenticator-aws`) in your `PATH`.
+You will also need [AWS IAM Authenticator for Kubernetes](https://github.com/kubernetes-sigs/aws-iam-authenticator) command (either `aws-iam-authenticator` or `aws eks get-token` (available in version 1.16.156 or greater of AWS CLI) in your `PATH`.
 
 ## Basic usage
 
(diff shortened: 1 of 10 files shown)


errordeveloper:
Thanks @cPu1! I'll review in detail next week, I think overall this is in the right direction :)

martina-if:
LGTM

PR #995, opened 2019-07-08
Document NAT Gateways

### LINKED ISSUE #972: Test NAT Gateway mode upgrade
At the very least we should document current behaviour and avoid breaking clusters (which seems possible in theory).

martina-if:
closes #972

### DIFF
--- a/site/content/usage/06-vpc-networking.md
+++ b/site/content/usage/06-vpc-networking.md
@@ -143,3 +143,20 @@ nodeGroups:
     kubeletExtraConfig:
         clusterDNS: ["169.254.20.10","172.20.0.10"] 
 ```
+
+### NAT Gateway
+
+The NAT Gateway for a cluster can be configured to be `Disabled`, `Single` (default) or `HighlyAvailable`. It can be 
+specified through the `--vpc-nat-mode` CLI flag or in the cluster config file like the example below:
+
+
+```yaml
+vpc:
+  nat:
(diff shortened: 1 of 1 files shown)


site/content/usage/06-vpc-networking.md
```diff
@@ -143,3 +143,20 @@ nodeGroups:
     kubeletExtraConfig:
         clusterDNS: ["169.254.20.10","172.20.0.10"] 
 ```
+
+### NAT Gateway
+
+The NAT Gateway for a cluster can be configured through config files. The available options are: disabled, single
```
gemagomez: Can we follow the same casing as in the yaml? `Disabled, Single`

cPu1: I think we should also mention that this is configurable via a CLI arg as well: `--vpc-nat-mode`.

site/content/usage/06-vpc-networking.md
```diff
@@ -143,3 +143,20 @@ nodeGroups:
     kubeletExtraConfig:
         clusterDNS: ["169.254.20.10","172.20.0.10"] 
 ```
+
+### NAT Gateway
+
+The NAT Gateway for a cluster can be configured through config files. The available options are: `Disabled`, `Single`
```
cPu1: I think we should also mention that this is configurable via a CLI arg as well: `vpc-nat-mode`.

PR #211, opened 2018-09-14
EKS authentication docs

tedmiston:
### Description

This PR adds and improves documentation around EKS authentication, particularly stock kubectl vs EKS-vended kubectl, the AWS IAM Authenticator for Kubernetes, and aws-iam-authenticator vs heptio-authenticator-aws.

It also adds headers to gather setup info ahead of usage info.

This is based on the discussion in https://github.com/weaveworks/eksctl/issues/209.

### Checklist
- [x] Code compiles correctly (i.e `make build`)
- [x] Added tests that cover your change (if possible)
- [x] All tests passing (i.e. `make test`)
- [x] Added/modified documentation as required (such as the README)
- [x] Added yourself to the `humans.txt` file

Issue #209, opened 2018-09-13
Are aws-iam-authenticator / heptio-authenticator-aws optionally required locally?


tedmiston:
**What happened?**

When using the eksctl create cluster command, a warning is output about "neither aws-iam-authenticator nor heptio-authenticator-aws are installed".

(Note: some not relevant flags are omitted and the cluster name is changed for the purpose of this post.)

```
$ eksctl create cluster --name=Foo --node-type=m5.large ...
2018-09-13T15:29:55-04:00 [ℹ]  using "ami-0b2ae3c6bda8b5c06" for nodes
2018-09-13T15:29:55-04:00 [ℹ]  creating EKS cluster "Foo" in "us-east-1" region
2018-09-13T15:29:55-04:00 [ℹ]  will create 2 separate CloudFormation stacks for cluster itself and the initial nodegroup
2018-09-13T15:29:55-04:00 [ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=us-east-1 --name=Foo'
2018-09-13T15:29:55-04:00 [ℹ]  creating cluster stack "eksctl-Foo-cluster"
2018-09-13T15:43:52-04:00 [ℹ]  creating nodegroup stack "eksctl-Foo-nodegroup-0"
2018-09-13T15:47:30-04:00 [✔]  all EKS cluster resource for "Foo" had been created
2018-09-13T15:47:30-04:00 [✔]  saved kubeconfig as "/Users/taylor/.kube/config"
2018-09-13T15:47:34-04:00 [ℹ]  the cluster has 0 nodes
2018-09-13T15:47:34-04:00 [ℹ]  waiting for at least 1 nodes to become ready
2018-09-13T15:48:00-04:00 [ℹ]  the cluster has 3 nodes
2018-09-13T15:48:00-04:00 [ℹ]  node "ip-192-168-109-14.ec2.internal" is ready
2018-09-13T15:48:00-04:00 [ℹ]  node "ip-192-168-143-134.ec2.internal" is ready
2018-09-13T15:48:00-04:00 [ℹ]  node "ip-192-168-238-209.ec2.internal" is ready
2018-09-13T15:48:00-04:00 [✖]  neither aws-iam-authenticator nor heptio-authenticator-aws are installed
2018-09-13T15:48:00-04:00 [ℹ]  cluster should be functional despite missing (or misconfigured) client binaries
[...]

PR #999, opened 2019-07-08
Docs: add requirement for aws-iam-authenticator

bboreham:
### Description

Mention that the tool `aws-iam-authenticator` tool is needed, in the basic installation instructions.

I hit the error message described at #209, and since #211 was never finished I couldn't see any hints what to do.

The text in this PR is adapted from #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ or [environment variables][awsenv]. For more information read [AWS documentation
 [awsenv]: https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html
 [awsconfig]: https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html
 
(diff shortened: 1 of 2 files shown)


2opremio:
I also experienced this problem on my first use

errordeveloper:
LGTM, but I am sure this was documented previously, could we double-check please?

bboreham:
Check away: https://github.com/weaveworks/eksctl/search?l=Markdown&q=aws-iam-authenticator

errordeveloper:
@bboreham I see, so we only say:

> If you have `kubectl` v1.10.x as well as `aws-iam-authenticator` commands in your PATH, you should be
able to use `kubectl`.

Which is very much buried right now... it probably used to be sufficient when the docs had much less to say.

PR #945, opened 2019-06-26
Upgrade AWS VPC CNI plugin to 1.5

martina-if:
closes #927 

- [x] Code compiles correctly (i.e `make build`)
- [x] Added tests that cover your change (if possible)
- [x] All unit tests passing (i.e. `make test`)
- [x] Manually tested updating aws-node addon
- [x] Manually tested creating cluster
- [x]  Wait for green light from AWS

Questions:
- When I create a cluster, the cni version is still 1.4.1. I am assuming that we don't control the version of the addon when we create a cluster; only when we update it. Is this correct?

PR #1001, opened 2019-07-08
add support for hongkong

M00nF1sh:
### Description
Add support for new region

### Checklist

- [x] Manually tested


pkg/nodebootstrap/assets/10-eksclt.al2.conf
```diff
@@ -20,6 +20,6 @@ ExecStart=/usr/bin/kubelet \
   --network-plugin=cni \
   --cni-bin-dir=/opt/cni/bin \
   --cni-conf-dir=/etc/cni/net.d \
-  --pod-infra-container-image=602401143452.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com/eks/pause-amd64:3.1 \
+  --pod-infra-container-image=${AWS_EKS_ECR_ACCOUNT}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com/eks/pause-amd64:3.1 \
```
errordeveloper: please also see `pkg/nodebootstrap/assets/bootstrap.ubuntu.sh`

pkg/nodebootstrap/userdata.go
```diff
@@ -24,6 +24,9 @@ import (
 const (
 	configDir            = "/etc/eksctl/"
 	kubeletDropInUnitDir = "/etc/systemd/system/kubelet.service.d/"
+
+	eksECRAccountStandard = "602401143452"
```
errordeveloper: We use this account ID for searching AMIs also, should we parametrised it everywhere?

https://github.com/weaveworks/eksctl/blob/e9fc98eef14a8bbe86f417ba10b56b9bb438964c/pkg/ami/auto_resolver.go#L53

errordeveloper:
Also, AMIs would need to be added also. If `602401143452` owns the AMIs, `make generate-ami` should be sufficient.

pkg/nodebootstrap/userdata.go
M00nF1sh: Thanks for the info :D. I'll update here as well. For opt-in regions, we have to use another account to host these resources. ami/ecr..

pkg/nodebootstrap/assets/10-eksclt.al2.conf
M00nF1sh: updated :D

M00nF1sh:
@errordeveloper Hi, do you know how can i trigger the addons updates? would like to manual test these code branches as well :D

M00nF1sh:
The amiIDs will only be public at the time of actual launch.

errordeveloper:
@M00nF1sh did you see https://eksctl.io/usage/cluster-upgrades/?

martina-if:
Hi @M00nF1sh I think maybe if you create a cluster with version 1.12 then you might be able to test add-ons upgrades?

By the way, do you have time to rebase and finish this PR this week?

pkg/ami/auto_resolver.go
```diff
@@ -48,10 +49,13 @@ var ImageSearchPatterns = map[string]map[string]map[int]string{
...
 
-// ImageFamilyToAccountID is a map of image families to account Ids
-var ImageFamilyToAccountID = map[string]string{
-	ImageFamilyAmazonLinux2: "602401143452",
-	ImageFamilyUbuntu1804:   "099720109477",
+// accountIDResolverFunc resolves ami provider accountID by region
+type accountIDResolverFunc func(region string) string
+
+// ImageFamilyToAccountID is a map of image families to accountIDResolverFunc
+var ImageFamilyToAccountID = map[string]accountIDResolverFunc{
```
martina-if: Perhaps now that we need a function anyway it makes more sense to get rid of the map and make it a function:

```
accountId(imageFamily, region) (string, error) {
  switch imageFamily {
    case ImageFamilyUbuntu1804:
      return "099720109477", nil
    case ImageFamilyAmazonLinux2:
      return api.EKSResourceAccountID(region), nil
    default:
      return nil, new error....
  }
}
```

martina-if: Then we can unit test it

M00nF1sh: make sense :D, modified

M00nF1sh:
@martina-if 
I have rebased and test the upgrade addons, it works :D.
Actually the upgrade addons have other problems(but not in the scope of this PR):
1. the kube-proxy minor version may not be same as CP version. (we don't always upgrade it for new k8s minor versions).
1. the awsNode version is not tied to specific k8s major version. e.g. the newest awsNode in 1.12 have 1.5.0 as CNI version, eksctl upgrade will actually downgrade it to 1.4.1 currently.

martina-if:
Hi @M00nF1sh , thanks for the fixes and for pointing those issues out . We already have a PR for nr 2 :)

errordeveloper:
> the upgrade addons have some problems

Please open an issue for each of these items, and we will discuss there.

gemagomez:
2) has a PR already that we are about to land: https://github.com/weaveworks/eksctl/pull/945
1) needs a new issue.

martina-if:
closes https://github.com/weaveworks/aws-workstream/issues/43

PR #1002, opened 2019-07-09
Clarify cluster vs nodegroup upgrade drift

### LINKED ISSUE #1000: Website upgrading page possible misleading about control plane upgrades
Yay for the new website! 🎉 ❤️

This comment on the upgrade page on the website reads to me like you can upgrade the control panel 2 minor versions at a time, which I don't think was the intention.

https://eksctl.io/usage/cluster-upgrades/
> NOTE: Kubernetes supports version drift of up-to 2 minor versions during upgrade process.

We probably need to be clearer?

> NOTE: Kubernetes supports version drift of up-to two minor versions during upgrade process. So nodes can be up to two minor versions ahead or behind the control plane version. You can only upgrade the control plane one minor version at a time, but nodes can be upgraded more than one minor version at a time, provided the nodes stay within two minor versions of the control plane.

martina-if:
closes #1000

### DIFF
--- a/site/content/usage/03-cluster-upgrade.md
+++ b/site/content/usage/03-cluster-upgrade.md
@@ -17,7 +17,11 @@ An _`eksctl`-managed_ cluster can be upgraded in 3 easy steps:
 
 Please make sure to read this section in full before you proceed.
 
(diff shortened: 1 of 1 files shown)


errordeveloper:
@whereisaaron would you mind to have a look at this?

PR #1005, opened 2019-07-09
[create cluster] Fix panic when using a config file without VPC.NAT set

### LINKED ISSUE #1003: Create cluster: panic: runtime error with eksctl version 0.1.39
**What happened?**
Panic: runtime error: invalid memory address or nil pointer dereference

After switching from 0.1.38 to 0.1.39 `eksctl create cluster -f cluster.yaml` is broken for me

**How to reproduce it?**
Used config (cluster.yaml):
```
 kind: ClusterConfig

 metadata:
   name: dev-eks
   region: eu-central-1

 vpc:
   subnets:
     private:
       eu-central-1a: {id: subnet-00000000000000000}
       eu-central-1b: {id: subnet-00000000000000001}
       eu-central-1c: {id: subnet-00000000000000002}

 nodeGroups:
   - name: ng-1a
     ami: ami-0b7127e7a2a38802a
     instanceType: t3.medium
     desiredCapacity: 1
     availabilityZones: ["eu-central-1a"]
     privateNetworking: true
     ssh:
       publicKeyName: bla.bla
     iam:
       attachPolicyARNs:
         - arn:aws:iam::000000000000:policy/ManageRoute53HostedZone
```
**Versions**
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.39"}
```

**Logs**
```
eksctl create cluster -f cluster.yaml
[ℹ]  using region eu-central-1
[✔]  using existing VPC (vpc-xxx) and subnets (private:[subnet-0... subnet-1... subnet-2...] public:[])
[!]  custom VPC/subnets will be used; if resulting cluster doesn't function as expected, make sure to review the configuration of VPC/subnets
[ℹ]  nodegroup "ng-1a" will use "ami-0b7127e7a2a38802a" [AmazonLinux2/1.13]
[ℹ]  using EC2 key pair "bla.bla"
[ℹ]  using Kubernetes version 1.13
[ℹ]  creating EKS cluster "dev-eks" in "eu-central-1" region
[ℹ]  1 nodegroup (ng-1a) was included
[ℹ]  will create a CloudFormation stack for cluster itself and 1 nodegroup stack(s)
[ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=eu-central-1 --name=dev-eks'
[ℹ]  2 sequential tasks: { create cluster control plane "dev-eks", create nodegroup "ng-1a" }
[ℹ]  building cluster stack "eksctl-dev-eks-cluster"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x20ac393]

goroutine 51 [running]:
github.com/weaveworks/eksctl/pkg/cfn/builder.(*ClusterResourceSet).addOutputsForVPC(0xc00011a280)
	/src/pkg/cfn/builder/vpc.go:160 +0x103
github.com/weaveworks/eksctl/pkg/cfn/builder.(*ClusterResourceSet).AddAllResources(0xc00011a280, 0xc00011a280, 0xc000550e80)
	/src/pkg/cfn/builder/cluster.go:49 +0xac
github.com/weaveworks/eksctl/pkg/cfn/manager.(*StackCollection).createClusterTask(0xc000354f60, 0xc0000407e0, 0x60, 0x0)
	/src/pkg/cfn/manager/cluster.go:33 +0x329
github.com/weaveworks/eksctl/pkg/cfn/manager.(*taskWithoutParams).Do(0xc00010f100, 0xc0000407e0, 0xc0000407e0, 0x1)
	/src/pkg/cfn/manager/tasks.go:132 +0x34
github.com/weaveworks/eksctl/pkg/cfn/manager.doSingleTask(0xc000040780, 0x3a86800, 0xc00010f100, 0x0)
	/src/pkg/cfn/manager/tasks.go:185 +0xec
github.com/weaveworks/eksctl/pkg/cfn/manager.doSequentialTasks(0xc000040780, 0xc00010f160, 0x2, 0x2)
	/src/pkg/cfn/manager/tasks.go:215 +0x70
created by github.com/weaveworks/eksctl/pkg/cfn/manager.(*TaskTree).DoAllSync
	/src/pkg/cfn/manager/tasks.go:115 +0x246
```

cPu1:
### Description

Fixes #1003 

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -56,7 +56,9 @@ func (l *commonClusterConfigLoader) Load() error {
 		return l.validateWithoutConfigFile()
 	}
 
-	if err := eks.LoadConfigFromFile(l.ClusterConfigFile, l.ClusterConfig); err != nil {
+	var err error
+
+	if l.ClusterConfig, err = eks.LoadConfigFromFile(l.ClusterConfigFile); err != nil {
 		return err
 	}
 	meta := l.ClusterConfig.Metadata
@@ -150,6 +152,14 @@ func NewCreateClusterLoader(rc *ResourceCmd, ngFilter *NodeGroupFilter) ClusterC
 	l.validateWithConfigFile = func() error {
 		if l.ClusterConfig.VPC == nil {
 			l.ClusterConfig.VPC = api.NewClusterVPC()
+		}
+
+		if l.ClusterConfig.VPC.NAT == nil {
+			l.ClusterConfig.VPC.NAT = api.DefaultClusterNAT()
+		}
+
+		if !api.IsSetAndNonEmptyString(l.ClusterConfig.VPC.NAT.Gateway) {
+			*l.ClusterConfig.VPC.NAT.Gateway = api.ClusterSingleNAT
 		}
 
 		if l.ClusterConfig.HasAnySubnets() && len(l.ClusterConfig.AvailabilityZones) != 0 {
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -58,6 +58,9 @@ func (l *commonClusterConfigLoader) Load() error {
 
 	var err error
 
+	// The reference to ResourceCmd.ClusterConfig should only be reassigned if ClusterConfigFile is specified
+	// because other parts of the code store the pointer locally and access it directly instead of via
+	// the ResourceCmd reference
 	if l.ClusterConfig, err = eks.LoadConfigFromFile(l.ClusterConfigFile); err != nil {
 		return err
 	}
--- a/pkg/ctl/cmdutils/configfile_test.go
+++ b/pkg/ctl/cmdutils/configfile_test.go
@@ -99,5 +99,35 @@ var _ = Describe("cmdutils configfile", func() {
 				Expect(cfg.Metadata.Version).To(BeEmpty())
 			}
 		})
+
+		It("should set VPC.NAT.Gateway with the correct value", func() {
+			natTests := []struct {
+				configFile      string
+				expectedGateway string
+			}{
+				// No VPC set
+				{"01-simple-cluster.yaml", api.ClusterSingleNAT},
+				// VPC set but not NAT
+				{"02-custom-vpc-cidr-no-nodes.yaml", api.ClusterSingleNAT},
+				// VPC and subnets set but not NAT
+				{"04-existing-vpc.yaml", api.ClusterSingleNAT},
+				// NAT set
+				{"09-nat-gateways.yaml", api.ClusterHighlyAvailableNAT},
+			}
+
+			for _, natTest := range natTests {
+				rc := &ResourceCmd{
+					Command:           newCmd(),
+					ClusterConfigFile: filepath.Join(examplesDir, natTest.configFile),
+					ClusterConfig:     api.NewClusterConfig(),
+					ProviderConfig:    &api.ProviderConfig{},
+				}
+
+				Expect(NewCreateClusterLoader(rc, nil).Load()).To(Succeed())
+				cfg := rc.ClusterConfig
+				Expect(cfg.VPC.NAT.Gateway).To(Not(BeNil()))
+				Expect(*cfg.VPC.NAT.Gateway).To(Equal(natTest.expectedGateway))
+			}
+		})
 	})
 })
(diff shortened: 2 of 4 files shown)


pkg/eks/api.go
```diff
@@ -163,10 +164,11 @@ func New(spec *api.ProviderConfig, clusterSpec *api.ClusterConfig) *ClusterProvi
 }
 
 // LoadConfigFromFile populates cfg based on contents of configFile
-func LoadConfigFromFile(configFile string, cfg *api.ClusterConfig) error {
+func LoadConfigFromFile(configFile string) (*api.ClusterConfig, error) {
```
martina-if: great!

pkg/ctl/cmdutils/configfile.go
```diff
@@ -56,7 +56,9 @@ func (l *commonClusterConfigLoader) Load() error {
 		return l.validateWithoutConfigFile()
 	}
 
-	if err := eks.LoadConfigFromFile(l.ClusterConfigFile, l.ClusterConfig); err != nil {
+	var err error
+
+	if l.ClusterConfig, err = eks.LoadConfigFromFile(l.ClusterConfigFile); err != nil {
```
errordeveloper: To be clear, from now on `l.ClusterConfig` will be initialised by explicit assignment, right? In that case, we must make sure that `l.ClusterConfigFile == ""` always returns before we get here, so that `l.ClusterConfig` doesn't get reset. Perhaps it would help to add a comment above to clarify this intention, or restructure the code a little to signify how it's supposed to works, what do you think?

errordeveloper: Or we can add a test that ensure we don't reset the reference that `l.ClusterConfig` holds when `l.ClusterConfigFile == ""`... perhaps that's the best thing we can do here?

cPu1: >To be clear, from now on l.ClusterConfig will be initialised by explicit assignment, right?

Yes, but only if `ClusterConfigFile` is set.

> In that case, we must make sure that l.ClusterConfigFile == "" always returns before we get here, so that l.ClusterConfig doesn't get reset.

Yes, we do make sure of that.

>Perhaps it would help to add a comment above to clarify this intention, or restructure the code a little to signify how it's supposed to works, what do you think?

Agreed, I'll add a comment.

The change here is that instead of dereferencing the pointer and assigning it the new config in `eks.LoadConfigFromFile`, we have it return the new config and overwrite it at the call site. This would affect code that assigned the ClusterConfig locally instead of accessing it via the `ResourceCmd` reference (`l.ClusterConfig` here) before this call is made, which is only done for setting flags in `createClusterCmd`.


>Or we can add a test that ensures we don't reset the reference that l.ClusterConfig holds when l.ClusterConfigFile == ""... perhaps that's the best thing we can do here?

Should be easy to add a test for this but I believe this isn't something a unit test should test. As long as we test both forms (with and without ClusterConfigFile) and assert the values are set properly, it should cover this change. 

Also, relying on having the contents of the pointer mutated from elsewhere can be error-prone.

errordeveloper: > Also, relying on having the contents of a pointer mutated from elsewhere can be error-prone.

Yes, but that's currently something we do throughout. Personally, I just think that Go is unhelpful as it doesn't provide ways to prevent mutation and often encourages it. So I just embrace mutation, because it's often the easiest way. Something that would be good to discuss, but I wouldn't make it a priority just yet, it will a very big change if we decide to do something about it.

errordeveloper: > Agreed, I'll add a comment.

Please do that, and we can merge!

pkg/ctl/cmdutils/configfile_test.go
```diff
@@ -82,23 +82,52 @@ var _ = Describe("cmdutils configfile", func() {
...
+				{"01-simple-cluster.yaml", api.ClusterSingleNAT},
+				// VPC set but not NAT
+				{"02-custom-vpc-cidr-no-nodes.yaml", api.ClusterSingleNAT},
+				// VPC and subnets set but not NAT
+				{"04-existing-vpc.yaml", api.ClusterSingleNAT},
+				// NAT set
+				{"09-nat-gateways.yaml", api.ClusterHighlyAvailableNAT},
+			}
+
+			for _, natTest := range natTests {
```
errordeveloper: Nice!

pkg/ctl/cmdutils/configfile.go
martina-if: We indeed need to do something about this. Mutating parameters is not good practice. Even if the code has other instances of this pattern that doesn't make it a good reason to keep following it. Little by little we have to get rid of that.

PR #1008, opened 2019-07-09
New release cadence documentation

### LINKED ISSUE #977: Document new release cadence
Some of the aspects we should formalise in a document:

- compatibility promisses
  - general aspects of compatibility and semver
  - for the config files it should be based on what Kubernetes dictates 
- feature deprecation policy (probably based on Kubernetes, but could be more relaxed pre-1.0)
- change in default Kubernetes version we ship should trigger minor version change
- how does change in node AMIs affect version number (?)

gemagomez:
Fixes #977

### Description

New release process documentation added to README.md.

### Checklist

### DIFF
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -12,7 +12,7 @@ import (
 
 // AddConfigFileFlag adds common --config-file flag
 func AddConfigFileFlag(fs *pflag.FlagSet, path *string) {
-	fs.StringVarP(path, "config-file", "f", "", "load configuration from a file")
+	fs.StringVarP(path, "config-file", "f", "", "load configuration from a file (or stdin if set to '-')")
 }
 
 // ClusterConfigLoader is an inteface that loaders should implement
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -164,7 +164,7 @@ func New(spec *api.ProviderConfig, clusterSpec *api.ClusterConfig) *ClusterProvi
 
 // LoadConfigFromFile populates cfg based on contents of configFile
 func LoadConfigFromFile(configFile string, cfg *api.ClusterConfig) error {
-	data, err := ioutil.ReadFile(configFile)
+	data, err := readConfig(configFile)
 	if err != nil {
 		return errors.Wrapf(err, "reading config file %q", configFile)
 	}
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -104,6 +104,17 @@ To learn more abut what `eksctl` can do check [eksctl.io](https://eksctl.io). A
 
 Code contributions are very welcome. If you are interested in helping make `eksctl` great then see our [contributing guide](CONTRIBUTING.md).
 
+## Releases
+
+`eksctl` will be releasing on a 2 week cadence (minor releases) with extra patch releases in between as needed.
```
errordeveloper: I think we can avoid having to change it after 0.2.0 is out... How about something like this:
```
As of eksctl 0.2.0, minor releases should be expected every two weeks and patch release will be made available as soon as bugs get fixed. One or more release candidate(s) will be made available prior to each minor release.
```

errordeveloper: I think one paragraph should be sufficient here.

README.md
```diff
@@ -104,6 +104,17 @@ To learn more abut what `eksctl` can do check [eksctl.io](https://eksctl.io). A
...
+
+This versioning scheme will be effective from version 0.2.0 onwards.
+
+RC releases will be published for testing purposes 2 working days prior to every minor release. At the end of this testing period, and after any regressions have been found and fixed, the release will take place.
+
+It is highly discouraged to use RC releases in any production environment. They are for testing purposes only.
```
errordeveloper: We don't have any other recommendations on what one should or shouldn't use in production, let's remove this statement.

README.md
```diff
@@ -104,6 +104,17 @@ To learn more abut what `eksctl` can do check [eksctl.io](https://eksctl.io). A
...
+
+
```
errordeveloper: Let's keep it single blank line please...

README.md
gemagomez: Agreed.

README.md
gemagomez: I feel it is important for users to know this is a test build, not a production one. We don't have a disclaimer?

README.md
gemagomez: +1

README.md
errordeveloper: True. As you suggested on slack, let's just state RCs are only for testing.

errordeveloper:
Will re-open from another branch, turns out that I cannot push to your master branch...

PR #1010, opened 2019-07-09
Delete `LoadBalancer` Services before the cluster

### LINKED ISSUE #103: delete VPC resources
Currently deleting VPC stack will fail when there resources such as ELBs. We should be able to delete these, the questions is whether we should do it by default when we own the VPC or not?

2opremio:
### Description

`type: LoadBalancer` Kubernetes services are backed by ELBs in AWS.

If we simply try to remove the cluster templates directly when deleting the cluster, these ELBs are going to dangle (since they are not part of the templates). This in turn causes makes the cluster's VPC deletion to fail.

Fixes #103

### Checklist
- [-] ~Added tests that cover your change (if possible)~ It's mostly AWS API calls, which makes it a bit hard to test the logic with unit tests. Happy to add an integration test if needed (once I learn how to do it).
- [x] Manually tested. For the record, I created 16 `type: LoadBalancer` services before deleting the cluster (8 backed by classic ELBs ones and 8 backed by v2 ELBS) after I proceeded to delete the cluster, which happened cleanly. I did the following:

```
$ for I in `seq 8`; do cat memcache-svc.yaml | I=$I envsubst  | kubectl apply -f - ; done
$ for I in `seq 8`; do cat memcache-svc-v2.yaml | I=$I envsubst  | kubectl apply -f - ; done
```

Where:

`memcache-svc.yaml`:
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: service$I
spec:
  ports:
    - name: memcached
      port: 11211
  type: LoadBalancer
  selector:
    name: memcached
``` 

`memcache-svc-v2.yaml`:
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: servicev2$I
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
spec:
  ports:
    - name: memcached
      port: 11211
  type: LoadBalancer
  selector:
    name: memcached
```


2opremio:
I need bump the build cache. Will do it later

pkg/vpc/cleanup.go
```diff
@@ -79,3 +79,4 @@ func CleanupNetworkInterfaces(ec2API ec2iface.EC2API, spec *api.ClusterConfig) e
 	}
 	return nil
 }
+
```
errordeveloper: blank line

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
+	corev1 "k8s.io/api/core/v1"
```
errordeveloper: Would you mind to separate AWS SDK and Kubernetes dependencies please?

pkg/eks/api.go
```diff
@@ -34,10 +22,25 @@ import (
...
 	"github.com/aws/aws-sdk-go/service/eks/eksiface"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/iam/iamiface"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+	"github.com/kris-nova/logger"
```
errordeveloper: We tend to have the following kind grouping:

- standard library
- anything from github (includes: logger, errors)
- AWS SDK
- Kubernetes
- Weaveworks

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -4,16 +4,16 @@ import (
...
 	"github.com/aws/aws-sdk-go/service/cloudformation/cloudformationiface"
 	"github.com/aws/aws-sdk-go/service/cloudtrail/cloudtrailiface"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/aws/aws-sdk-go/service/eks/eksiface"
+	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
 	"github.com/aws/aws-sdk-go/service/iam/iamiface"
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
errordeveloper: Also here...

pkg/eks/api.go
2opremio: The problem with that is that:
1. It's not really respected in general (I've run into a few examples in which the `eksctl` dependencies were not at the end)
2. You cannot automate it with an IDE (at least not easily). At least GoLand only lets you organize by:
```
Stdlib

external dependencies

local dependencies
```

I am pretty sure (1) is most probably due to (2)

Also (and this is just a personal opinion) I don't see the use of it. How is this helpful to you?

martina-if: Also remember about our external contributors. I think whichever we decide it should be as standard as possible.

2opremio:
I have just found out that the aws cloud controller sometimes leaves security groups associated to the load balancer, preventing the VPC from being deleted. I can fix this in a separate PR.

errordeveloper:
> I have just found out that the aws cloud controller sometimes leaves security groups associated to the load balancers, preventing the VPC from being deleted. I can fix this in a separate PR.

I suppose it leaves them in a hope to re-use? Is there a GC that is supposed to trigger eventually?

Separate PR sounds good to me.

errordeveloper:
So to summaries for an external observer who may not read the code, what we are doing is this:

- using Kubernetes API: find all services that carry a load balancer and delete them
- using ELB APIs: wait with a timeout for load balancer to get deleted, if timeout occurs, we produce an error

And we don't attempt forcing deletion ourselves just yet, correct?

It maybe a good idea to document this, opening a follow-up issue would a good start.

errordeveloper:
Another follow-up issue would be to review what popular ingress controllers bring into the picture, they might also create ELBs. Of course, ingress controllers are not built-in, but Ingress objects are part of Kubernetes. The answer may be that we just leave it until we support certain ingress controller add-ons, but let's open an issue to track it.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/weaveworks/eksctl/pkg/eks"
+)
+
+const (
+	elbClassic = iota
+	elbV2      = iota
+)
+
+// CleanupLoadBalancers finds and deletes any dangling ELBs attached to a service
```
errordeveloper: nit: We could call it `Cleanup`, as the package is dedicated to ELBs, so it would be `elb.Cleanup` and `elb.CleanupLoadBalancers` read a little redundant....

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+	services, err := kubernetesCS.CoreV1().Services("").List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Delete Services of type 'LoadBalancer'
+	elbs := map[string]int{}
```
errordeveloper: I'd use [`k8s.io/apimachinery/pkg/util/sets`](https://godoc.org/k8s.io/apimachinery/pkg/util/sets), we use it in other places already for this purpose. I can see that you store them by type, but seems like using separate sets would be just a little more readable. Or a `map[int]sets.String`... Up to you, just a suggestion.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+	elbV2      = iota
+)
+
+// CleanupLoadBalancers finds and deletes any dangling ELBs attached to a service
+func CleanupLoadBalancers(elbapi elbiface.ELBAPI, elbv2api elbv2iface.ELBV2API, client *eks.Client) error {
+	kubernetesCS, err := client.NewClientSet()
+	if err != nil {
+		return err
+	}
+	services, err := kubernetesCS.CoreV1().Services("").List(metav1.ListOptions{})
```
errordeveloper:
```suggestion
	services, err := kubernetesCS.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
```

pkg/ctl/delete/cluster.go
```diff
@@ -103,11 +104,23 @@ func doDeleteCluster(rc *cmdutils.ResourceCmd) error {
 	}
 
 	{
+
+		logger.Info("cleaning up LoadBalancer services")
+		if err := ctl.GetCredentials(cfg); err != nil {
+			return err
+		}
+		client, err := ctl.NewClient(cfg, false)
```
errordeveloper:
```suggestion
		clientSet, err := ctl.NewClientSet(cfg, false)
```

pkg/ctl/delete/cluster.go
```diff
@@ -103,11 +104,23 @@ func doDeleteCluster(rc *cmdutils.ResourceCmd) error {
...
+		if err != nil {
+			return err
+		}
+		elb.CleanupLoadBalancers(ctl.Provider.ELB(), ctl.Provider.ELBV2(), client)
```
errordeveloper: We should prepend it to tasks actually, but you will need some of the changes from #847 for that, I can pick those onto another PR if you like, or we refactor later on.

pkg/ctl/delete/cluster.go
```diff
@@ -103,11 +104,23 @@ func doDeleteCluster(rc *cmdutils.ResourceCmd) error {
...
+
 		tasks, err := stackManager.NewTasksToDeleteClusterWithNodeGroups(rc.Wait, func(errs chan error, _ string) error {
 			logger.Info("trying to cleanup dangling network interfaces")
 			if err := ctl.GetClusterVPC(cfg); err != nil {
 				return errors.Wrapf(err, "getting VPC configuration for cluster %q", cfg.Metadata.Name)
 			}
+
 			go func() {
 				errs <- vpc.CleanupNetworkInterfaces(ctl.Provider.EC2(), cfg)
```
errordeveloper: This ENI callback needs some refactoring, will need to be addressed as part of #536.

go.mod
```diff
@@ -84,7 +84,7 @@ require (
 	k8s.io/kops v0.0.0-20190222135932-278e6606534e
 	k8s.io/kube-openapi v0.0.0-20181018171734-e494cc581111 // indirect
 	k8s.io/kubelet v0.0.0-20190313123811-3556bcde9670
-	k8s.io/kubernetes v1.12.6 // indirect
+	k8s.io/kubernetes v1.12.6
```
errordeveloper: I don't understand, why was it indirect and not any more?

errordeveloper: Ok, so this is no longer indirect because we added `k8s.io/kubernetes/pkg/cloudprovider`.

go.sum
```diff
@@ -159,11 +159,13 @@ github.com/coreos/bbolt v1.3.1-coreos.6/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE55
 github.com/coreos/bbolt v1.3.3 h1:n6AiVyVRKQFNb6mJlwESEvvLoDyiTzXX7ORAUlkeBdY=
 github.com/coreos/bbolt v1.3.3/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/coreos/etcd v3.3.10+incompatible/go.mod h1:uF7uidLiAD3TWHmW31ZFd/JWoc32PjwdhPthX9715RE=
+github.com/coreos/etcd v3.3.11+incompatible h1:0gCnqKsq7XxMi69JsnbmMc1o+RJH3XH64sV9aiTTYko=
```
errordeveloper: I'll assume that this file is not intended to be read by humans...

2opremio: No, it's not

go.mod
2opremio: I don't think that's it. But you can ask `go mod` with `go mod why` (I am AFKbut I will do it later)

pkg/ctl/delete/cluster.go
2opremio: I guess you meant `ctl.NewStdClientSet`, correct?

2opremio: No, that one goes with embedded token. Please clarify.

pkg/elb/cleanup.go
2opremio: I would rather not depend on kubernetes for data structures we can obtain directly from the standard library (even at a readability cost)

2opremio:
> And we don't attempt forcing deletion ourselves just yet, correct?

Correct

2opremio:
> I have just found out that the aws cloud controller sometimes leaves security groups associated to the load balancers, preventing the VPC from being deleted. I can fix this in a separate PR.

I ended up fixing it in this PR. It was a pain in the ass because I ended up running into https://github.com/kubernetes/kubernetes/issues/79994 which I worked around by deleting orphan security groups as a last step.

I also rebased it on `master`.

@errordeveloper PTAL

pkg/ctl/delete/cluster.go
errordeveloper: Yes, that's right, it'd be `ctl.NewStdClientSet`.

2opremio: What's the difference between embedding a token or not? Right now the client is created without one, following what `eksctl utils write-kubecobfig` does, whereas `NewNewStdClientSet()` doesn't.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,299 @@
...
+	// Delete Services of type 'LoadBalancer', collecting their ELBs to wait for them to be deleted later on
+	loadBalancers := map[string]loadBalancer{}
+	for _, s := range services.Items {
+		if s.Spec.Type == corev1.ServiceTypeLoadBalancer {
+			name := cloudprovider.DefaultLoadBalancerName(&s)
+			kind := getLoadBalancerKind(&s)
+			ctx, cleanup := context.WithTimeout(context.Background(), 30*time.Second)
+			securityGroupIDs, err := getSecurityGroupsOwnedByLoadBalancer(ctx, ec2API, elbAPI,
+				clusterConfig.Metadata.Name, name, kind)
+			cleanup()
```
martina-if: What is this cleanup function?

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,299 @@
...
+			logger.Debug("Deleting 'type: LoadBalancer' service %s/%s", s.Namespace, s.Name)
+			err = kubernetesCS.CoreV1().Services(s.Namespace).Delete(s.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Wait for all the load balancers backing the LoadBalancer services to disappear, for a maximum of 10 minutes
+	pollInterval := 1 * time.Second
```
martina-if: Can we split lines 84-112 into a function like "waitForDeletion()" to make `Cleanup()` smaller?

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,299 @@
...
+	services, err := kubernetesCS.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Delete Services of type 'LoadBalancer', collecting their ELBs to wait for them to be deleted later on
+	loadBalancers := map[string]loadBalancer{}
+	for _, s := range services.Items {
+		if s.Spec.Type == corev1.ServiceTypeLoadBalancer {
+			name := cloudprovider.DefaultLoadBalancerName(&s)
```
martina-if: Maybe splitting lines 59-79 into their own function we can make testing easier

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,299 @@
...
+		if awsError, ok := err.(awserr.Error); ok {
+			if awsError.Code() == "LoadBalancerNotFound" {
+				return nil, nil
+			}
+		}
+		return nil, err
+	}
+
+	var ret *elb.LoadBalancerDescription
+	for _, loadBalancer := range response.LoadBalancerDescriptions {
```
martina-if: why not just check the size of the array for the warning?

2opremio:
> Can you add tests?
That will be covered by #1031 , which I will work on for 0.4.0

pkg/elb/cleanup.go
2opremio: It cancels the context, see https://golang.org/pkg/context/#WithTimeout :

> Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete:
>```golang
>func slowOperationWithTimeout(ctx context.Context) (Result, error) {
>	  ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
>	  defer cancel()  // releases resources if slowOperation completes before timeout elapses
>	  return slowOperation(ctx)
>}
> ```

pkg/elb/cleanup.go
2opremio: I considered it but decided against it because it would make the ownership of `loadBalancers` unclear (the loop deletes them for book-keeping).

pkg/elb/cleanup.go
2opremio: Yeah, that makes more sense. I just copy-pasted from https://github.com/kubernetes/legacy-cloud-providers/blob/release-1.15/aws/aws.go#L2728-L2734

pkg/ctl/delete/cluster.go
errordeveloper: Good question!

The client with embedded token is most desirable for internal use. We don't need to execute the authenticator plugin and we instead call the authenticator as a library and store the token in the client config structure.

When we want to write the `kubeconfig` file out for users to consume, we want it to use the authenticator binary every time user calls `kubectl`. If we were to use embedded token there, we'd have to provide a token refresh mechanism also.

2opremio: Ah, great, thanks a lot for the clarification

pkg/elb/cleanup.go
2opremio: I followed a different approach (factoring out well-defined sub-operations) but made the function smaller

pkg/elb/cleanup.go
2opremio: I will address that in #1031

pkg/eks/api.go
```diff
@@ -113,6 +122,8 @@ func New(spec *api.ProviderConfig, clusterSpec *api.ClusterConfig) *ClusterProvi
 	provider.cfn = cloudformation.New(s)
 	provider.eks = awseks.New(s)
 	provider.ec2 = ec2.New(s)
+	provider.elb = elb.New(s)
+	provider.elbv2 = elbv2.New(s)
```
errordeveloper: Please also see custom endpoint logic below, we need to add `AWS_ELB_ENDPOINT` and `AWS_ELBV2_ENDPOINT` for consistency.

https://github.com/weaveworks/eksctl/pull/1010/files#diff-eb5619b1d325aef63a641b639f6dde9fL132

2opremio:
I've addressed all the comments and tested it manually again

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
```diff
@@ -353,6 +353,11 @@ func (in *NodeGroup) DeepCopyInto(out *NodeGroup) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.VolumeIOPS != nil {
```
errordeveloper: This should have landed in master, probably my fault when I was rebasing @cPu1's PR...

site/content/usage/20-schema.md
```diff
@@ -215,6 +215,8 @@ NodeGroup:
       type: array
     volumeEncrypted:
       type: boolean
+    volumeIOPS:
```
errordeveloper: Same as earlier, this should had been on master already.

pkg/ctl/delete/cluster.go
```diff
@@ -103,11 +106,27 @@ func doDeleteCluster(rc *cmdutils.ResourceCmd) error {
...
+
+		logger.Info("cleaning up LoadBalancer services")
+		if err := ctl.GetCredentials(cfg); err != nil {
+			return err
+		}
+		cs, err := ctl.NewStdClientSet(cfg)
+		if err != nil {
+			return err
+		}
+		ctx, cleanup := context.WithTimeout(context.Background(), 10*time.Minute)
```
errordeveloper: We have a global `--timeout` flag, it's defined as 'wait time in any polling operations (default 25m0s)'.
I think we should use it here, but it would a good idea to review how we do polling in general, as what's here is different from other places. I don't think we have to address it now, I'll open an issue.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func getServiceLoadBalancer(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI, clusterName string,
+	service *corev1.Service) (*loadBalancer, error) {
+	if service.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		return nil, nil
+	}
+	name := cloudprovider.DefaultLoadBalancerName(service)
+	kind := getLoadBalancerKind(service)
+	ctx, cleanup := context.WithTimeout(context.Background(), 30*time.Second)
```
errordeveloper: Same concern as https://github.com/weaveworks/eksctl/pull/1010/files#r305305236

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,314 @@
...
+	result := make([]string, 0, len(set))
+	for value := range set {
+		result = append(result, value)
+	}
+	return result
+}
+
+func deleteOrphanLoadBalancerSecurityGroups(ctx context.Context, ec2API ec2iface.EC2API, clusterConfig *api.ClusterConfig) error {
+	clusterName := clusterConfig.Metadata.Name
+	describeRequest := &ec2.DescribeSecurityGroupsInput{
```
errordeveloper: Our current convention is just to call these `input`, and whatever we get back `output`, and avoid having more than one API call in one function.

errordeveloper: Just to say that there is a convention, don't need to worry about it right now.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,314 @@
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag-key"),
+				Values: []*string{aws.String(awsprovider.TagNameKubernetesClusterPrefix + clusterName)},
+			},
+		},
+	}
+	for {
+		result, err := ec2API.DescribeSecurityGroupsWithContext(ctx, describeRequest)
```
errordeveloper: I am not very clear on the benefits of `*WithContext` calls, another discussion topic to follow-up on.

pkg/ctl/delete/cluster.go
errordeveloper: I've opened #1036.

pkg/elb/cleanup.go
errordeveloper: See #1036.

pkg/elb/cleanup.go
errordeveloper: See #1035.

pkg/elb/cleanup.go
errordeveloper: Also see #1035.

pkg/elb/cleanup.go
```diff
@@ -0,0 +1,314 @@
...
+	name string) (*elb.LoadBalancerDescription, error) {
+
+	request := &elb.DescribeLoadBalancersInput{
+		LoadBalancerNames: []*string{&name},
+	}
+
+	response, err := elbAPI.DescribeLoadBalancersWithContext(ctx, request)
+	if err != nil {
+		if awsError, ok := err.(awserr.Error); ok {
+			if awsError.Code() == "LoadBalancerNotFound" {
```
errordeveloper: They have `elbAPI.ErrCodeAccessPointNotFoundException`, but that reads rather weird, so indeed let's use a string.

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
2opremio: Yep. Maybe the file generation check isn't working as it should?

errordeveloper: Yes, and it's been fixed in #1034, I think it was to do with the makefile bug you and Gema spotted earlier. Could you please rebase?

errordeveloper:
@2opremio thanks! I'm happy to merge this once we rebase, let's make sure we all test it one RC is out and understand what kind of pain users have had until now!

pkg/elb/cleanup.go
2opremio: I got this from https://github.com/kubernetes/legacy-cloud-providers/blob/release-1.15/aws/aws.go#L2720-L2723

`ErrCodeAccessPointNotFoundException` doesn't exist in the SDK https://github.com/aws/aws-sdk-go/search?q=ErrCodeAccessPointNotFoundException&unscoped_q=ErrCodeAccessPointNotFoundException

2opremio: Actually, it does exist, I will use that

pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
2opremio: done

2opremio:
>  I'm happy to merge this once we rebase

Done

pkg/elb/cleanup.go
martina-if: @2opremio can we rename it to `cancelContext()` (in the next PR) so it doesn't get confused with the cleanup happening in this package?

2opremio:
So, please go ahead

Issue #1012, opened 2019-07-10
Cluster autoscaler cannot scale mixed nodegroup on AWS


alexei-led:
**What happened?**
Created a nodegroup with mixed spot instances, tried to set `instanceType` to `mixed` and leave it empty.
Cluster autoscaler reports the following error

```
...
Unable to build proper template node for eksctl-moon-kube-nodegroup-gpu-spot-ng-a-NodeGroup-17OCXU2R4F8ES: unable to find instance type within launch template
...
```

Looking at cluster autoscaler code, I see that it uses non-empty and valid `instanceType` in `LaunchTemplate` and fails if it's not specified (see error above)

Cluster autoscaler [code](https://github.com/kubernetes/autoscaler/blob/6409124ae38c8fe6cad0daffafffca467a522bd3/cluster-autoscaler/cloudprovider/aws/ec2.go#L52)

**What you expected to happen?**

I expect that a mixed nodegroup created with `eksctl` will support working with cluster autoscaler

**How to reproduce it?**
Create a mixed NG with spot instances. Deploy a workload that forces CA to scale up, see above error.

**Anything else we need to know?**
MacOS
EKS cluster
Cluster autoscaler `k8s.gcr.io/cluster-autoscaler:v1.15.0`

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.39"}

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-20T04:49:16Z", GoVersion:"go1.12.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-eks-c57ff8", GitCommit:"c57ff8e35590932c652433fab07988da79265d5b", GitTreeState:"clean", BuildDate:"2019-06-07T20:43:03Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

```

**Logs**
Include the output of the command line when running eksctl. If possible, eksctl should be run with debug logs. For example:
`eksctl get clusters -v 4`
Make sure you redact any sensitive information before posting.
If the output is long, please consider a Gist.

scottyhq:
[...]

PR #1013, opened 2019-07-10
fixing #1012, cluster autoscaler requres a valid LT intance type

alexei-led:
Kubernetes cluster autoscaler (AWS) requres a valid LT intance type to scale up/down ASG. See #1012 

### Description

Adding first instance type from distribution instance types as a default `LaunchTemplate` instance type.

### Checklist
- [x] Updated tests that cover your change 
- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/api_test.go
+++ b/pkg/cfn/builder/api_test.go
@@ -2499,7 +2499,7 @@ var _ = Describe("CloudFormation template builder API", func() {
 			Expect(ngTemplate.Resources).To(HaveKey("NodeGroupLaunchTemplate"))
 
 			launchTemplateData := getLaunchTemplateData(ngTemplate)
-			Expect(launchTemplateData.InstanceType).To(Equal(""))
+			Expect(launchTemplateData.InstanceType).To(Equal("m5.large"))
 			Expect(launchTemplateData.InstanceMarketOptions).To(BeNil())
 
 			nodeGroupProperties := getNodeGroupProperties(ngTemplate)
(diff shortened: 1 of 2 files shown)


errordeveloper:
@alexei-led thanks for this PR! This looks good overall, but I'm just not sure if there are any implications on functionality, would you mind to provide a reference to AWS docs that suggests what does it mean when mixed instances are set along with traditional instance type?

alexei-led:
@errordeveloper Ilya take a look at [AWS LaunchTemplate API docs](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LaunchTemplate.html)

>> The overrides are used to override the instance type specified by the launch template with multiple instance types that can be used to launch On-Demand Instances and Spot Instances. 

i.e. when LT has `Overrides` section the `instanceType` in LT is ignored. I've also checked this.

martina-if:
:+1: Thanks @alexei-led . I tested this creating 3 nodegroups: all spot instances, all on demand and mixed instances and it seems to work fine.

errordeveloper:
Let's merge it for 0.2.0. We have 0.1.40 underway, and I'd prefer that to fix exactly one bug (#1003).

whereisaaron:
@errordeveloper is `m5` a good choice for this? Remember that some AZ's don't have `m5` instances like `ap-southeast-2b`. Will that affect being able to use this change in that zone, or will the override prevent any errors.

alexei-led:
@whereisaaron any non-empty string is a good choice - cluster autoscaler validates only that `instanceType` is not empty, no actual validation is done :)

PR #1025, opened 2019-07-17
Support EBS optimized instance.

d-kuro:
### Description

Added support for EBS Optimized instance.

ref: 

* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/nodegroup.go
+++ b/pkg/cfn/builder/nodegroup.go
@@ -216,8 +216,9 @@ func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_La
 		IamInstanceProfile: &gfn.AWSEC2LaunchTemplate_IamInstanceProfile{
 			Arn: n.instanceProfileARN,
 		},
-		ImageId:  gfn.NewString(n.spec.AMI),
-		UserData: n.userData,
+		ImageId:      gfn.NewString(n.spec.AMI),
+		EbsOptimized: gfn.True(),
+		UserData:     n.userData,
 		NetworkInterfaces: []gfn.AWSEC2LaunchTemplate_NetworkInterface{{
 			AssociatePublicIpAddress: gfn.NewBoolean(!n.spec.PrivateNetworking),
 			DeviceIndex:              gfn.NewInteger(0),
--- a/pkg/cfn/builder/api_test.go
+++ b/pkg/cfn/builder/api_test.go
@@ -112,6 +112,7 @@ type LaunchTemplateData struct {
 	IamInstanceProfile              struct{ Arn interface{} }
 	UserData, InstanceType, ImageId string
 	BlockDeviceMappings             []interface{}
+	EbsOptimized                    bool
 	NetworkInterfaces               []struct {
 		DeviceIndex              int
 		AssociatePublicIpAddress bool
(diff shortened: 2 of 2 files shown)


errordeveloper:
Hi! Thanks for contributing this, but we don't have an issue for it, could you please open one and outline the exact difference that this would make? It is not very clear from the docs, my understanding was this actually gets determined automatically.

d-kuro:
@errordeveloper 
Thank you for your reply 😄
The Kubernetes Pod's logs are written to a file, causing I/O to EBS.
I/O to EBS conflicts with network bandwidth, but enabling EBS-optimized delivers dedicated bandwidth to EBS.
This is expected to improve performance.

AWS already has EBS-optimized enabled by default for many instance types, but you need to explicitly enable it when creating an instance with CloudFormation.
I'm want to support EBS-optimized with eksctl.

ref:
* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-ebsoptimized

errordeveloper:
@d-kuro from reading the documentation, I understand that wast majority of instance types are EBS-optimised by default. Could you please let us know which instance type are you using?

d-kuro:
@errordeveloper 
EbsOptimized is disabled by default in CloudFormation.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-ebsoptimized

スクリーンショット 2019-07-26 14 33 47

errordeveloper:
@d-kuro I don't see an issue for this, please open one first and we will discuss there how this needs to be exposed, and if it's indeed needed or not.

mreferre:
@d-kuro there are three possible scenarios here:
1- the (old) EC2 instance doesn't support EBS-optimization
2- the (prev gen) EC2 instance supports EBS-optimization and can be opted in (at an additional cost
3- the (current gen) EC2 instance supports EBS-optimization and is enabled by default

Actually, for #3 [the flag is not important because the behavior remains the same regardless of the status of the flag itself:](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html)

`"For instance types that are EBS–optimized by default, there is no need to enable EBS optimization and no effect if you disable EBS optimization. For instances that are not EBS–optimized by default, you can enable EBS optimization when you launch the instances, or enable EBS optimization after the instances are running..... When you enable EBS optimization for an instance that is not EBS-optimized by default, you pay an additional low, hourly fee for the dedicated capacity."`

Given that [the number of instances that fall in #2 is limited, the perf advantages are valuable and the cost associated to enable it is minimal](https://aws.amazon.com/ec2/previous-generation/), we should consider enabling this for all instances by default thus minimizing the number of switches we expose to the user. In the (rare?) event a customer would NOT want EBS-optimization for their previous generation instances they can always turn it off manually. Introducing a switch and exposing it to the customer when we expect this event to be so unlikely, it would mean adding complexity to the user for no practical reasons.

Do you believe this would be a proper course of action?

d-kuro:
@mreferre Thank you for your reply.

>`"For instance types that are EBS–optimized by default, there is no need to enable EBS optimization and no effect if you disable EBS optimization. For instances that are not EBS–optimized by default, you can enable EBS optimization when you launch the instances, or enable EBS optimization after the instances are running..... When you enable EBS optimization for an instance that is not EBS-optimized by default, you pay an additional low, hourly fee for the dedicated capacity."`

Sorry I didn't know that.
Since EBS optimization is already enabled by default for many instance types.
I think it is okay to close this Pull Request.
What do you think?

errordeveloper:
>  ...we should consider enabling this for all instances by default thus minimizing the number of switches we expose to the user.
> Do you believe this would be a proper course of action?

@mreferre Yes, indeed!

@d-kuro instead of just closing it, could you enabled `EbsOptimized` unconditionally in `pkg/cfn/builder/nodegroup.go`?

d-kuro:
@errordeveloper 
Fixed to enable EbsOptimized by default 😀

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -452,6 +452,8 @@ type NodeGroup struct {
 	Tags map[string]string `json:"tags,omitempty"`
 	// +optional
 	PrivateNetworking bool `json:"privateNetworking"`
+	// +optional
+	EbsOptimized *bool `json:"ebsOptimized,omitempty"`
```
errordeveloper: I don't see a good reason for adding the field actually.

pkg/cfn/builder/nodegroup.go
```diff
@@ -216,8 +216,9 @@ func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_La
 		IamInstanceProfile: &gfn.AWSEC2LaunchTemplate_IamInstanceProfile{
 			Arn: n.instanceProfileARN,
 		},
-		ImageId:  gfn.NewString(n.spec.AMI),
-		UserData: n.userData,
+		ImageId:      gfn.NewString(n.spec.AMI),
+		EbsOptimized: gfn.NewBoolean(api.IsEnabled(n.spec.EbsOptimized)),
```
errordeveloper: So I think this should be just `EbsOptimized: gfn.True()`.

d-kuro:
@errordeveloper 
Thanks review!
Removed config field.
netlify CI failed, what should I do?

errordeveloper:
@d-kuro thanks for following up, I appreciate that! I'll mere it and fix Netlify separately, it's not critical for this and there are no documentation changes.

PR #1028, opened 2019-07-17
Fix dependency file construction on makefile

gemagomez:
This problem was introduced in PR #1014


errordeveloper:
I'll cherry-pick this.

PR #992, opened 2019-07-05
Second minor release (0.2.0)

errordeveloper:
### Description

Add release notes for 0.2.0.

Close #976.

Issue #549, opened 2019-02-17
allow user to patch or delete default storage class


whereisaaron:
**What happened?**
I create a cluster with [`--storage-class=false`](https://github.com/weaveworks/eksctl/issues/175), yet the created cluster still had a `gp2` storage class created.
```
$ kubectl get storageclasses
NAME            PROVISIONER             AGE
gp2 (default)   kubernetes.io/aws-ebs   21h
```
I created this cluster with no nodegroup and added and removed nodegroups, could the `StorageClass` have been created during those changes. `--storage-class` is only a cluster argument right, I don't need to add it for nodegroups too?

**What you expected to happen?**
No `StorageClass`

**How to reproduce it?**
Include the steps to reproduce the bug

```
eksctl create cluster \
  --name=$CLUSTER_NAME \
  --region=$AWS_DEFAULT_REGION \
  --profile=$AWS_PROFILE \
  --node-private-networking \
  --vpc-private-subnets=$SUBNETS \
  --without-nodegroup \
  --storage-class=false \
  --asg-access \
  --timeout=30m0s
```

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.21"}
$ uname -a
Linux 4.4.0-17763-Microsoft #253-Microsoft Mon Dec 31 17:49:00 PST 2018 x86_64 x86_64 x86_64 GNU/Linux
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T07:10:00Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.5-eks-6bad6d", GitCommit:"6bad6d9c768dc0864dab48a11653aa53b5a47043", GitTreeState:"clean", BuildDate:"2018-12-06T23:13:14Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

#175 #224

errordeveloper:
The default storage class is only optional in 1.10, as of 1.11 it's present out of the box.

[...]

PR #1029, opened 2019-07-17
Removing version 1.10

### LINKED ISSUE #1026: Deprecate Kubernetes v1.10 in ekcstl
AWS are [deprecating support](https://github.com/aws/containers-roadmap/issues/300) for Kubernetes 1.10 in EKS on 22nd July. From that date on users should not be able to create a 1.10 cluster with eksctl.

Version 1.10 is not the default, so it has to be supplied by the user from a flag. So we should give an error message when someone provides that flag. At the moment for an unsupported version we do:

```
$ eksctl create cluster --version=1.8
[ℹ]  using region us-west-2
[✖]  invalid version, supported values: 1.10, 1.11, 1.12
```

I think given that this was support and has been deprecated, we should do something like the following:

```
$ eksctl create cluster --version=1.10
[✖] From 22nd July 2019 EKS no longer supports Kubernetes version 1.10. See https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html for more information. 
```

gemagomez:
### Description

Remove version 1.10 from the codeline. 

Fixes #1026

### Checklist
This was not done due to the change being a removal of code that didn't have explicit coverage. Tests have not been removed either as part of this change. 
- [ ] Manually tested


pkg/ctl/utils/install_corends.go
```diff
@@ -57,13 +55,6 @@ func doInstallCoreDNS(rc *cmdutils.ResourceCmd) error {
 		return errors.Wrapf(err, "getting credentials for cluster %q", meta.Name)
 	}
 
-	switch ctl.ControlPlaneVersion() {
-	case "":
```
errordeveloper: Are we sure about removing this?

errordeveloper:
You'd also want to remove `eksctl utils install-coredns` entirely, but I'm not very clear how upgrades will work, as it's needed on 1.10 -> 1.11 upgrade.

errordeveloper:
Also, this whole block needs to be removed:
https://github.com/weaveworks/eksctl/blob/6461e6db97f02b1cacd4456f7017320f6bc2a602/pkg/ctl/create/cluster.go#L355-L365

gemagomez:
> You'd also want to remove `eksctl utils install-coredns` entirely, but I'm not very clear how upgrades will work, as it's needed on 1.10 -> 1.11 upgrade.

According to the issue any users on 1.10 will be automatically upgraded to 1.11. That is not something we need to do.

pkg/ctl/utils/install_corends.go
gemagomez: @2opremio said it was ok to do, we wanted you to tell us whether it was actually ok. Shall we still leave the case for "" ?

gemagomez:
> Also, this whole block needs to be removed, along with the `--config-file` flag:
> https://github.com/weaveworks/eksctl/blob/6461e6db97f02b1cacd4456f7017320f6bc2a602/pkg/ctl/create/cluster.go#L355-L365

done!

errordeveloper:
> What do you mean with "along with the --config-file flag?

I mean to say the definition of that flag, but copy-pasted the wrong one.

It's the `--storage-class` that needs to be removed, which is defined in:

https://github.com/weaveworks/eksctl/blob/bef6c33908a40343a939d20088589177f986f19b/pkg/ctl/create/cluster.go#L68

and there are other references too, as well as struct field... 

That said assuming we want to do what was originally intended to happen per comment in that block of code. However, we have https://github.com/weaveworks/eksctl/issues/549 that needs to be decided on.

Another special case (pretty sure the last one), is this:
https://github.com/weaveworks/eksctl/blob/6461e6db97f02b1cacd4456f7017320f6bc2a602/pkg/addons/default/aws_node.go#L38-L40

pkg/ctl/utils/install_corends.go
errordeveloper: I'll need to review what the function does.

marccarre:
I'll remove these now:
```
$ git grep '"1.10."'
pkg/addons/default/aws_node.go: if strings.HasPrefix(controlPlaneVersion, "1.10.") {
pkg/addons/default/coredns.go:  if strings.HasPrefix(controlPlaneVersion, "1.10.") {
```
1. is what @errordeveloper was referring to [here](https://github.com/weaveworks/eksctl/pull/1029#issuecomment-512479332) ;
2. is `eksctl utils install-coredns` which @errordeveloper was referring to [here](https://github.com/weaveworks/eksctl/pull/1029#issuecomment-512337610).

marccarre:
I _believe_ [1] that the cleanup is now complete:

- **a.** what was mentioned above has been removed (dead code, CLI arguments, etc.) ;
- **b.** some tests having dependencies on production code previously removed by @gemagomez were also removed (also, their names implied there were related to upgrading from 1.10 which, if I understood correctly, is no longer something required) ;
- **c.** some documentation (`site/`) was also updated, as it was referring to the `storage-class` CLI argument, which `eksctl` will no longer support after this PR is merged ;
- **d.** CI seems to be pleased. 

@errordeveloper, PTAL 🙏🙇 

[1]: based on the aforementioned hints & pointers, basic `git grep`ping, and basic understanding of English in tests, comments, etc. Disclaimer: I'm _not_ familiar with `eksctl`, so I mostly don't know what I'm doing and just used "common sense" here.

pkg/addons/default/aws_node_test.go
```diff
@@ -52,59 +51,6 @@ var _ = Describe("default addons - aws-node", func() {
 			Expect(awsNode.Spec.Template.Spec.Containers[0].Image).To(
 				Equal("602401143452.dkr.ecr.eu-west-2.amazonaws.com/amazon-k8s-cni:v1.3.2"),
 			)
-
-		})
-
-		It("can update 1.10 sample to latest", func() {
```
errordeveloper: It looks like there is no similar test based on 1.11 or another version, we should migrate this code.

pkg/ctl/create/cluster.go
```diff
@@ -66,7 +65,6 @@ func createClusterCmd(rc *cmdutils.ResourceCmd) {
 
 	rc.FlagSetGroup.InFlagSet("Cluster and nodegroup add-ons", func(fs *pflag.FlagSet) {
 		cmdutils.AddCommonCreateNodeGroupIAMAddonsFlags(fs, ng)
-		fs.BoolVar(&params.addonsStorageClass, "storage-class", true, "if true (default) then a default StorageClass of type gp2 provisioned by EBS will be created")
```
errordeveloper: Let's make sure we mention this in the changelog

errordeveloper: Added a note https://github.com/weaveworks/eksctl/pull/992/files#r305223547

pkg/ctl/create/cluster.go
```diff
@@ -350,18 +348,6 @@ func doCreateCluster(rc *cmdutils.ResourceCmd, params *createClusterCmdParams) e
 			return err
 		}
 
-		// add default storage class only for version 1.10 clusters
-		if meta.Version == "1.10" {
-			// --storage-class flag is only for backwards compatibility,
-			// we always create the storage class when --config-file is
-			// used, as this is 1.10-only
-			if params.addonsStorageClass || rc.ClusterConfigFile != "" {
-				if err = ctl.AddDefaultStorageClass(clientSet); err != nil {
```
errordeveloper: Actually we should also delete the function itself :)

pkg/addons/default/coredns_test.go
```diff
@@ -1,254 +0,0 @@
...
-			Expect(*coreDNS.Spec.Replicas == 2).To(BeTrue())
-			Expect(coreDNS.Spec.Template.Spec.Containers).To(HaveLen(1))
-			Expect(coreDNS.Spec.Template.Spec.Containers[0].Image).To(
-				Equal("602401143452.dkr.ecr.eu-west-1.amazonaws.com/eks/coredns:v1.1.3"),
-			)
-		})
-
-	})
-
-	loadSampleAndCheck("1.11", "1.1.3")
```
errordeveloper: We definitely need to keep this test and all of the ones below.

pkg/addons/default/coredns.go
```diff
@@ -1,245 +0,0 @@
...
-	}
-
-	logger.Info("deleted %q", KubeDNS)
-
-	logger.Info("%q is now up-to-date", CoreDNS)
-	return false, nil
-}
-
-// UpdateCoreDNS will update the `coredns` add-on
-func UpdateCoreDNS(rawClient kubernetes.RawClientInterface, region, controlPlaneVersion string, plan bool) (bool, error) {
```
errordeveloper: We want to keep this function.

pkg/addons/default/coredns.go
```diff
@@ -1,245 +0,0 @@
...
-	KubeDNS = "kube-dns"
-
-	componentLabel = "eks.amazonaws.com/component"
-
-	coreDNSImagePrefix = "602401143452.dkr.ecr."
-	coreDNSImageSuffix = ".amazonaws.com/eks/coredns"
-)
-
-// InstallCoreDNS will install the `coredns` add-on in place of `kube-dns`
-func InstallCoreDNS(rawClient kubernetes.RawClientInterface, region, controlPlaneVersion string, waitTimeout *time.Duration, plan bool) (bool, error) {
```
errordeveloper: We only want to get rid of this function.

pkg/ctl/utils/utils.go
```diff
@@ -15,8 +15,6 @@ func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
 	cmdutils.AddResourceCmd(flagGrouping, verbCmd, updateClusterStackCmd)
 	cmdutils.AddResourceCmd(flagGrouping, verbCmd, updateKubeProxyCmd)
 	cmdutils.AddResourceCmd(flagGrouping, verbCmd, updateAWSNodeCmd)
-	cmdutils.AddResourceCmd(flagGrouping, verbCmd, updateCoreDNSCmd)
```
errordeveloper: `eksctl utils update-coredns` is to stay, only `eksctl utils install-coredns` needs to go, as that handles flipping from kube-dns to CoreDNS, which is needed during 1.10->1.11 upgrade.

errordeveloper:
> c. some documentation (`site/`) was also updated, as it was referring to the `storage-class` CLI argument, which `eksctl` will no longer support after this PR is merged ;

Thanks, that looks good. Also take a look at the upgrade doc, in particular the 1.10->1.11 special case:
```
site/content/usage/03-cluster-upgrade.md:146:eksctl utils update-coredns
```

pkg/addons/default/aws_node_test.go
errordeveloper: See https://github.com/weaveworks/eksctl/pull/1029/files#diff-6ddb8dc16176800c04be724aaff37219L24

I'd start by changing it to:
```
sampleAddons := testutils.LoadSamples("testdata/sample-1.12.json")
```

And then make the rest of the test relevant.

pkg/ctl/create/cluster.go
marccarre: Thanks ✨

pkg/ctl/create/cluster.go
marccarre: Thanks, had missed this. Now removed.

pkg/addons/default/coredns.go
marccarre: Change reverted, and `UpdateCoreDNS` kept.

pkg/addons/default/coredns.go
marccarre: Change reverted, `InstallCoreDNS` removed and `UpdateCoreDNS` kept.

pkg/ctl/utils/utils.go
marccarre: Reverted.

pkg/addons/default/aws_node_test.go
marccarre: > make the rest of the test relevant

This is the hard bit, given I know little about this logic and the background of this change, but I've had a go.

pkg/ctl/utils/install_corends.go
errordeveloper: So the function currently relies on cache, it's possible for it to return an empty string.

https://github.com/weaveworks/eksctl/blob/079c5b71ee90b9fcfa6d4c10f07e81a2c8fc0ac0/pkg/eks/eks.go#L77-L83

The cache is populated when we make `ctl.GetCredentials` call, which is the one that we must do whenever we want to do anything with a cluster. Cache would be empty only if someone messes up the order of calls, or does something else accidentally.

https://github.com/weaveworks/eksctl/blob/079c5b71ee90b9fcfa6d4c10f07e81a2c8fc0ac0/pkg/eks/eks.go#L72

I think it's a minor detail that can improved, and we can open an issue for it if we'd like to discuss this, but thinking about it right now would be a distraction.

Let's put the empty string case back.

marccarre:
Some manual testing along the lines of what was done [here](https://github.com/weaveworks/eksctl/issues/1026#issue-469180154):
```console
$ ./eksctl create cluster --version=1.8
[ℹ]  using region us-west-2
[✖]  invalid version, supported values: 1.11, 1.12, 1.13
```
```console
$ ./eksctl create cluster --version=1.10
[ℹ]  using region us-west-2
[✖]  invalid version, supported values: 1.11, 1.12, 1.13
```
Do we want the message suggested by Kalbir?
It feels like this would create unnecessary maintenance, and that a similar comment in the release notes would be good enough, and more permanent. Thoughts?

pkg/addons/default/aws_node_test.go
```diff
@@ -50,18 +48,18 @@ var _ = Describe("default addons - aws-node", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(awsNode.Spec.Template.Spec.Containers).To(HaveLen(1))
 			Expect(awsNode.Spec.Template.Spec.Containers[0].Image).To(
-				Equal("602401143452.dkr.ecr.eu-west-2.amazonaws.com/amazon-k8s-cni:v1.3.2"),
+				Equal("602401143452.dkr.ecr.eu-west-1.amazonaws.com/amazon-k8s-cni:v1.4.1"),
```
errordeveloper: Now that I think of it, new test doesn't actually trigger any changes, as `sample-1.12.json` comes with 1.4.1 already... let me see what can be done here.

marccarre: Yes, shouldn't we use 1.11 instead? (That's what I thought when I tried to make the tests relevant.)

errordeveloper: Ok, let's not worry about that, the test would just look kind of weird. I'll remove `sample-1.10.json`.

errordeveloper:
@marccarre thanks for bringing up the error message, I can see the benefit of it, but it's kind of minor. It would be easy to do in ad-hoc way, but to do it properly we will need to maintain additional functionality that keeps track of unsupported versions and when exactly they were deprecated.

errordeveloper:
I'm going to update `pkg/ami`, and I think we are all set after that.

PR #1039, opened 2019-07-19
[WIP] Experimental command: export-templates

errordeveloper:
### Description

This replaces #718.

### Checklist

- [x] Manually tested

### DIFF
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/weaveworks/eksctl/pkg/ctl/create"
 	"github.com/weaveworks/eksctl/pkg/ctl/delete"
 	"github.com/weaveworks/eksctl/pkg/ctl/drain"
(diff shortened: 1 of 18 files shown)


errordeveloper:
@pawelprazak the idea of putting this under `eksctl experimental` is so that we could ship this sooner, and improve before moving under utils or some other place.

gemagomez:
After discussion with Ilya and the team, we decided to not have an experimental flag, and have an environment variable that enables experimental features instead. #1055 has been created to deal with experimental features.

errordeveloper:
There is a new challenge with this, we will need to ignore `cloudWatch.clusterLogging`  and warn the user, that is because this is currently handled outside if CloudFormation. There are likely to be other features like this, as CloudFormation tends to be behind EKS.

pawelprazak:
Yes, one that I know of is private API access.

martina-if:
Closing until this comes back into our planning

PR #1050, opened 2019-07-23
eksctl now prints "invalid version, X.Y is now deprecated" on deprecated Kubernetes versions

### LINKED ISSUE #1049: Print deprecation message when trying to create an EKS cluster with Kubernetes v1.10

When trying to create a cluster with Kubernetes v1.10, from 0.2.0 onwards (post #1026 and #1029), the following is printed:
```console
$ ./eksctl-0.2.0 create cluster --version=1.10
[ℹ]  using region us-west-2
[✖]  invalid version, supported values: 1.11, 1.12, 1.13
```
This may be confusing for end-users who used to use version 1.10 and may have missed [0.2.0's release notes](https://github.com/weaveworks/eksctl/blob/master/docs/release_notes/0.2.0.md#bug-fixes-since-0140). 

We should instead have a message making it explicit that Kubernetes v1.10 was deprecated by EKS, e.g.:
```console
$ ./eksctl-0.2.0 create cluster --version=1.10
[ℹ]  using region us-west-2
[✖]  invalid version, 1.10 is now deprecated ; supported values: 1.11, 1.12, 1.13
see also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
```

marccarre:
### Description

Fixes #1049.

When provided with `--version=1.10`, `eksctl` now prints:
```console
(issues/1049-deprecated-msg)$ ./eksctl create cluster --version=1.10
[ℹ]  using region us-east-1
[✖]  invalid version, 1.10 is now deprecated, supported values: 1.11, 1.12, 1.13
see also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
```

**Changelog**:
- Extracted version validation logic to method in order to remove duplication.
- Added custom error message for deprecated Kubernetes versions.
- Added release notes for 0.2.1 given this is the first PR targeting this version. N.B.: I duplicated the "_Bug Fixes Since 0.1.40_" section from `0.2.0.md` since this may still be relevant as the two versions are only going to be one day apart, and we're still transitioning from older versioning/release process. Let me know if appropriate. I can make it leaner otherwise.

### Checklist

- [x] Code compiles correctly (i.e `make build`).
- [x] ~Added tests that cover your change (if possible).~ (Unit tests for this would not add much value.)
- [x] All unit tests passing (i.e. `make test`).
- [x] Added/modified documentation as required: `docs/release_notes/0.2.1.md`.
- [x] Manually tested (see above).

### DIFF
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -113,6 +113,9 @@ func doCreateCluster(rc *cmdutils.ResourceCmd, params *createClusterCmdParams) e
 	}
 	if cfg.Metadata.Version != api.DefaultVersion {
 		if !isValidVersion(cfg.Metadata.Version) {
+			if isDeprecatedVersion(cfg.Metadata.Version) {
+				return fmt.Errorf("invalid version, %s is now deprecated, supported values: %s\nsee also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html", cfg.Metadata.Version, strings.Join(api.SupportedVersions(), ", "))
+			}
 			return fmt.Errorf("invalid version, supported values: %s", strings.Join(api.SupportedVersions(), ", "))
 		}
 	}
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -112,14 +112,7 @@ func doCreateCluster(rc *cmdutils.ResourceCmd, params *createClusterCmdParams) e
 		cfg.Metadata.Version = api.DefaultVersion
 	}
 	if cfg.Metadata.Version != api.DefaultVersion {
-		validVersion := false
-		for _, v := range api.SupportedVersions() {
-			if cfg.Metadata.Version == v {
-				validVersion = true
-				break
-			}
-		}
-		if !validVersion {
+		if !isValidVersion(cfg.Metadata.Version) {
 			return fmt.Errorf("invalid version, supported values: %s", strings.Join(api.SupportedVersions(), ", "))
 		}
 	}
--- a/docs/release_notes/0.2.1.md
+++ b/docs/release_notes/0.2.1.md
@@ -1,0 +1,167 @@
+# Release 0.2.1
+
+## Bug Fixes Since 0.2.0
+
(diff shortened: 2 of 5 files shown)


kalbir:
Yes that's exactly what we were looking for. Thanks @marccarre

pkg/ctl/create/cluster.go
```diff
@@ -112,14 +112,10 @@ func doCreateCluster(rc *cmdutils.ResourceCmd, params *createClusterCmdParams) e
...
 	}
 	if cfg.Metadata.Version != api.DefaultVersion {
-		validVersion := false
-		for _, v := range api.SupportedVersions() {
-			if cfg.Metadata.Version == v {
-				validVersion = true
-				break
+		if !isValidVersion(cfg.Metadata.Version) {
+			if isDeprecatedVersion(cfg.Metadata.Version) {
+				return fmt.Errorf("invalid version, %s is now deprecated, supported values: %s\nsee also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html", cfg.Metadata.Version, strings.Join(api.SupportedVersions(), ", "))
```
cPu1: What do you think about starting this error message with "unsupported version" instead of "invalid version"?
```suggestion
				return fmt.Errorf("unsupported version, %s is now deprecated, supported values: %s\nsee also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html", cfg.Metadata.Version, strings.Join(api.SupportedVersions(), ", "))
```

marccarre: I don't really mind. `invalid version, supported values: 1.11, 1.12, 1.13` was the generic error message up until now, hence I used `invalid`. 

Related to your suggestion, `invalid version, X.Y is no longer supported` was also suggested (and is being discussed in #1051), instead of `invalid version, X.Y is now deprecated`.

Unless a consensus can be found quickly and we can also close #1051 as part of this PR, I'd rather keep things as-is for now, and refine later after this has been bikeshedded properly 🙂.

cPu1: Sounds good :+1:

Issue #784, opened 2019-05-08
preBootstrapCommands 


arielvinas:
**What happened?**
I'm trying to allocate resourses for the kubelet and system daemons, and also add a custom eviction force to the node, but preBootstrapCommands aren't taking effect on the node.

**What you expected to happen?**
This is my preBootstrapCommands in the yaml:
```
preBootstrapCommands:
      - "#!/bin/bash -xe"
      - "set -o xtrace"
      - "yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm"
      - "systemctl enable amazon-ssm-agent"
      - "systemctl start amazon-ssm-agent"
      - "set -a"
      - "source /etc/environment"
      - "/etc/eks/bootstrap.sh clustername \ --kubelet-extra-args '--kube-reserved=cpu=200m,memory=200Mi,ephemeral-storage=5Gi\
        --system-reserved=cpu=300m,memory=300Mi,ephemeral-storage=5Gi \
        --eviction-hard=memory.available<100Mi,nodefs.available<5%'"
```

**How to reproduce it?**
Create a nodegroup with that preBootstrapCommands and check kubelet, system and evition config and won't be the one defined in the yaml

**Anything else we need to know?**
I'm running on macOS mojave 10.14.4 with eksctl downloaded from brew as explained in the doc

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.1.31"}
$ uname -a
Darwin Macbook-air-MoniOnline 18.5.0 Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64
$ kubectl version 
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-19T22:12:47Z", GoVersion:"go1.12.4", Compiler:"gc", Platform:"darwin/amd64"}
```
Also include your version of `heptio-authenticator-aws`
{"Version":"0.4.0-alpha.1","Commit":"82544ec86140b7027a07890b09c2f03fca97228a"}
**Logs**
[...]

PR #994, opened 2019-07-05
[Ubuntu] Use same kubeconfig.yaml as Amazon Linux

davidhao3300:
### Description

Switch flag usage in Ubuntu node bootstrapping to usage of kubelet.yaml, similar to how Amazon Linux does it.

Discovered this discrepancy after encountering an issue similar to https://github.com/weaveworks/eksctl/issues/350, and examining the PR addressing that issue showed that the fix was only done for Amazon Linux, but not Ubuntu.

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] Added tests that cover your change (if possible)
- [x] All unit tests passing (i.e. `make test`)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [x] Manually tested

closes #355

PR #1057, opened 2019-07-23
Allow customizing kubelet featureGates

martina-if:
https://github.com/weaveworks/eksctl/issues/784#issuecomment-513704672

- [x] Manually tested

### DIFF
--- a/site/content/usage/11-customizing-the-kubelet.md
+++ b/site/content/usage/11-customizing-the-kubelet.md
@@ -13,7 +13,7 @@ of resource starvation the kubelet might not be able to evict pods and eventuall
 
  
 Some fields in the `kubelet.yaml` are set by eksctl and therefore are not overwritable, such as the `address`, 
-`clusterDomain`, `authentication`, `authorization`, `serverTLSBootstrap` or `featureGates`. 
+`clusterDomain`, `authentication`, `authorization`, or `serverTLSBootstrap`.
 
 The following example config file creates a nodegroup that reserves `300m` vCPU, `300Mi` of memory and `1Gi` of 
 ephemeral-storage for the kubelet; `300m` vCPU, `300Mi` of memory and `1Gi`of ephemeral storage for OS system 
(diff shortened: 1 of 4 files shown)


errordeveloper:
@martina-if could you please clarify - do we merge what user provided with what we have built-in? If so, perhaps there should be a test for that?

errordeveloper:
@martina-if maybe we could cherry-pick #994 here?

gemagomez:
The code looks good. It'd be useful to add to the documentation how to customize featureGates, not just to the tests. It would make sense to have an example of customization in the docs and a link to any upstream related document that users may find useful.

errordeveloper:
@martina-if let me know when you want me to have another look :)

site/content/usage/11-customizing-the-kubelet.md
```diff
@@ -45,7 +45,14 @@ nodeGroups:
         evictionHard:
             memory.available:  "200Mi"
             nodefs.available: "10%"
+        featureGates:
+            DynamicKubeletConfig: true
```
errordeveloper: Could you please add `RotateKubeletServerCertificate` here as well, and make sure to add a comment, e.g.

```
                         RotateKubeletServerCertificate: true # has to be enabled, otherwise it will be disabled
```

site/content/usage/11-customizing-the-kubelet.md
```diff
@@ -45,7 +45,14 @@ nodeGroups:
...
 ```
 
 In this example, given instances of type `m5a.xlarge` which have 4 vCPUs and 16GiB of memory, the `Allocatable` amount
- of CPUs would be 3.4 and 15.4 GiB of memory. 
+of CPUs would be 3.4 and 15.4 GiB of memory. In addition, the `DynamicKubeletConfig` feature gate is also enabled. It is
+important to know that the values specified in the config file for the `featureGates` will completely overwrite the
+default values specified by eksctl, like `RotateKubeletServerCertificate`, so if you still want to keep it you need to
```
errordeveloper: I'd make this a little more concrete, e.g.:

```markdown
IMPORTANT: by default `eksctl` sets  **`featureGates.RotateKubeletServerCertificate`** to `true`, when custom **`featureGates`** are provide, it will be unset. You should always include **`featureGates.RotateKubeletServerCertificate`**, unless you have to disable it.
```

PR #1038, opened 2019-07-19
Add `eksctl experimental` without any commands

errordeveloper:
### Description

This ~will~ can be the landing place for `install-flux` and other commands.

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] All unit tests passing (i.e. `make test`)
- [x] Manually tested

PR #1061, opened 2019-07-25
Refactor and cleanup main package

errordeveloper:
### Description

This is a cherry-pick from #1038.
 
This mostly gets rid of global variables, and increases the spacing in help text.

### Checklist

- [x] Manually tested

### DIFF
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -19,15 +19,15 @@ import (
 )
 
 func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
-	rootCmd.AddCommand(versionCmd(flagGrouping))
 	rootCmd.AddCommand(create.Command(flagGrouping))
-	rootCmd.AddCommand(delete.Command(flagGrouping))
 	rootCmd.AddCommand(get.Command(flagGrouping))
 	rootCmd.AddCommand(update.Command(flagGrouping))
+	rootCmd.AddCommand(delete.Command(flagGrouping))
 	rootCmd.AddCommand(scale.Command(flagGrouping))
 	rootCmd.AddCommand(drain.Command(flagGrouping))
 	rootCmd.AddCommand(utils.Command(flagGrouping))
 	rootCmd.AddCommand(completion.Command(rootCmd))
+	rootCmd.AddCommand(versionCmd(flagGrouping))
 }
 
 func main() {
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -18,60 +18,7 @@ import (
 	"github.com/weaveworks/eksctl/pkg/ctl/utils"
 )
 
-var rootCmd = &cobra.Command{
-	Use:   "eksctl [command]",
-	Short: "a CLI for Amazon EKS",
-	Run: func(c *cobra.Command, _ []string) {
-		if err := c.Help(); err != nil {
-			logger.Debug("ignoring error %q", err.Error())
-		}
-	},
-}
-
-func init() {
-
-	var colorValue string
-
-	flagGrouping := cmdutils.NewGrouping()
-
-	addCommands(flagGrouping)
-
-	rootCmd.PersistentFlags().BoolP("help", "h", false, "help for this command")
-	rootCmd.PersistentFlags().StringVarP(&colorValue, "color", "C", "true", "toggle colorized logs (true,false,fabulous)")
-	rootCmd.PersistentFlags().IntVarP(&logger.Level, "verbose", "v", 3, "set log level, use 0 to silence, 4 for debugging and 5 for debugging with AWS debug logging")
-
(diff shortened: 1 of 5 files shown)


cmd/eksctl/main.go
```diff
@@ -18,67 +18,67 @@ import (
...
 }
 
-func init() {
+func main() {
+	cobra.EnableCommandSorting = false
 
-	var colorValue string
+	rootCmd := &cobra.Command{
+		Use:   "eksctl [command]",
+		Short: "a CLI for Amazon EKS",
```
cPu1: Now that we've changed the repo's description to "The official CLI for AWS EKS", what do you think about mentioning "official" here also?

cmd/eksctl/main.go
```diff
@@ -18,67 +18,67 @@ import (
...
+		color = false
+	case "fabulous":
+		color = false
+		fabulous = true
+	}
+	logger.Color = color
+	logger.Fabulous = fabulous
+
+	// Add timestamps for debugging
+	logger.Timestamps = false
```
cPu1: https://github.com/weaveworks/eksctl/blob/5f340977c12398db57ad922d9de4faa43fd64e92/cmd/eksctl/main.go#L80-L83

Very minor nit: I know this part was just extracted from `main`, but this could also be:

```suggestion
	logger.Timestamps = logger.Level >= 4
```

gemagomez:
- Does this change break compatibility of the CLI?

- Could you describe what manual testing was done?

cmd/eksctl/main.go
errordeveloper: good idea, thanks

cmd/eksctl/main.go
errordeveloper: Yes, I'm all for consistency :)

cmd/eksctl/main.go
```diff
@@ -18,67 +18,67 @@ import (
...
+	rootCmd.AddCommand(update.Command(flagGrouping))
+	rootCmd.AddCommand(scale.Command(flagGrouping))
+	rootCmd.AddCommand(drain.Command(flagGrouping))
+	rootCmd.AddCommand(utils.Command(flagGrouping))
+	rootCmd.AddCommand(completion.Command(rootCmd))
 }
 
-func init() {
+func main() {
+	cobra.EnableCommandSorting = false
```
cPu1: I think we should note this and mention in the release notes that the command list would now be in the order of importance:

```Commands:
  eksctl version          Output the version of eksctl
  eksctl create           Create resource(s)
  eksctl delete           Delete resource(s)
  eksctl get              Get resource(s)
  eksctl update           Update resource(s)
  eksctl scale            Scale resources(s)
  eksctl drain            Drain resource(s)
  eksctl utils            Various utils
  eksctl completion       Generates shell completion scripts
  eksctl help             Help about any command
```

Also, what do you think about moving `eksctl version` below in the list? `kubectl` for e.g., lists version at the end.

gemagomez: What do you mean in order of importance? The version is the most important command?

errordeveloper:
> Does this change break compatibility of the CLI?

No, it only make cosmetic changes to the help output.

> Could you describe what manual testing was done?

I've tested the flags where logic was simplified:

1. coloured logging still works as expected

- `eksctl delete cluster non-existing`
- `eksctl delete cluster non-existing --color=fabulous`
- `eksctl delete cluster non-existing --color=false`
- `eksctl delete cluster non-existing --color=true`

2. timestamps appear in logs a level 4 and above

- `eksctl delete cluster non-existing --verbose=4`

cmd/eksctl/main.go
cPu1: `version` is not important, which is why I suggested moving it to the end.

errordeveloper: What I was mostly trying to address is disable alphabetical sorting and make it sort in order of declaration. But oder of declaration should be the oder importance, so I'll move version command to the bottom.

errordeveloper: New look:
```
The official CLI for Amazon EKS

Usage: eksctl [command] [flags]

Commands:
  eksctl create           Create resource(s)
  eksctl get              Get resource(s)
  eksctl update           Update resource(s)
  eksctl delete           Delete resource(s)
  eksctl scale            Scale resources(s)
  eksctl drain            Drain resource(s)
  eksctl utils            Various utils
  eksctl completion       Generates shell completion scripts
  eksctl version          Output the version of eksctl
  eksctl help             Help about any command

Common flags:
  -C, --color string   toggle colorized logs (valid options: true, false, fabulous) (default "true")
  -h, --help           help for this command
  -v, --verbose int    set log level, use 0 to silence, 4 for debugging and 5 for debugging with AWS debug logging (default 3)

Use 'eksctl [command] --help' for more information about a command.
```

gemagomez: Looks good!

PR #1064, opened 2019-07-25
Add `eksctl install flux` command

2opremio:
### Description

This PR adds a new command: `eksctl install flux`, which, given a Git repository and an EKS cluster, it will Bootstrap [Flux](https://github.com/fluxcd/flux), installing it in the cluster and initializing its manifests in the Git repository.

Depends on https://github.com/fluxcd/flux/pull/2287 , Fixes #1027 ,  Fixes #1055 

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] Added _unit_ tests that cover ~your change (if possible)~ utilities and trivially testable code
- [x] All unit tests passing (i.e. `make test`)
- [x] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~ Experimental: let's write the documentation once this has matured a bit.
- [x] Manually tested
- [x] ~Build and push Docker image `weaveworks/eksctl-build:0.9` (after the review is done)~ (Irrelevant?)

PR #1067, opened 2019-07-27
Log internal kubernetes errors uniformly

2opremio:
### Description

Kubernetes has a way to log errors without exposing them (sigh): `k8s.io/apimachinery/pkg/util/runtime.HandleError`. This results in inconsistently-formatted logs. For instance, see the `ERROR: logging before flag.Parse` line below:

```
[ℹ]  Waiting for Flux to start
ERROR: logging before flag.Parse: E0726 19:56:16.160725    1598 portforward.go:331] an error occurred forwarding 49999 -> 3030: error forwarding port 3030 to pod f5bb5827e0fc76ffa136007d94300bab4da5d1ee87dc577a80444a4d3cdc9780, uid : exit status 1: 2019/07/26 17:56:16 socat[5745] E connect(5, AF=2 127.0.0.1:3030, 16): Connection refused
[!]  Flux is not ready yet (executing HTTP request: executing HTTP request: Post http://127.0.0.1:49999/api/flux/v9/git-repo-config: EOF), retrying ...
[ℹ]  Flux started successfully
[ℹ]  Commiting and pushing Flux manifests to git@github.com:2opremio/my-gitops-repo
```

To fix this, we override `k8s.io/apimachinery/pkg/util/runtime.ErrorHandlers`.

Found while testing #1064 . _Inspired_ by https://github.com/fluxcd/flux/commit/6bc1e8931074c819e0b9ef8a7c0c34dcfd7d247d#diff-350e484490d6241fa81c13942d160480 

### Checklist

- [x] Manually tested

### DIFF
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/kris-nova/logger"
 	"github.com/spf13/cobra"
-	k8sruntime "k8s.io/apimachinery/pkg/util/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 
 	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
 	"github.com/weaveworks/eksctl/pkg/ctl/completion"
@@ -43,14 +43,14 @@ func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
 func main() {
 
 	// Log internally-logged kubernetes errors uniformly
-	logK8sError := func(err error) {
+	logKubeError := func(err error) {
 		// we need to go one level deeper than k8sruntime.HandleError
 		_, file, line, _ := runtime.Caller(1)
 		idx := strings.Index(file, "/k8s.io/")
 		finalFileLine := file[idx+1:] + ":" + strconv.Itoa(line)
-		logger.Warning("Internal Kubernetes error at %s: %s", finalFileLine, err)
+		logger.Warning("Internal Kubernetes client error at %s: %s", finalFileLine, err)
 	}
-	k8sruntime.ErrorHandlers = []func(error){logK8sError}
+	utilruntime.ErrorHandlers = []func(error){logKubeError}
 
 	cobra.EnableCommandSorting = false
 
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -4,9 +4,13 @@ import (
 	"fmt"
 	"github.com/weaveworks/eksctl/pkg/ctl/generate"
 	"os"
+	"runtime"
+	"strconv"
+	"strings"
 
 	"github.com/kris-nova/logger"
 	"github.com/spf13/cobra"
+	k8sruntime "k8s.io/apimachinery/pkg/util/runtime"
 
 	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
 	"github.com/weaveworks/eksctl/pkg/ctl/completion"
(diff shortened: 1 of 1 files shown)


cmd/eksctl/main.go
```diff
@@ -31,6 +35,17 @@ func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
...
 
 func main() {
+
+	// Log internally-logged kubernetes errors uniformly
+	logK8sError := func(err error) {
+		// we need to go one level deeper than k8sruntime.HandleError
+		_, file, line, _ := runtime.Caller(1)
+		idx := strings.Index(file, "/k8s.io/")
+		finalFileLine := file[idx+1:] + ":" + strconv.Itoa(line)
+		logger.Warning("Internal Kubernetes error at %s: %s", finalFileLine, err)
```
errordeveloper: As I mentioned in your other PR, let's not start new lines with capital letters.

Also, it's not internal to Kubernetes itself, it's a client error, I'd make it 'internal Kubernetes client error'.

errordeveloper:
I'm not very clear that we need definitely need this. Port-forwarding appears to be an edge-case, all other client calls simply return errors right away. If we must use port-forwarding in #1064, let's just add it to that one command for now.

cmd/eksctl/main.go
```diff
@@ -3,9 +3,13 @@ package main
 import (
 	"fmt"
 	"os"
+	"runtime"
+	"strconv"
+	"strings"
 
 	"github.com/kris-nova/logger"
 	"github.com/spf13/cobra"
+	k8sruntime "k8s.io/apimachinery/pkg/util/runtime"
```
errordeveloper: We don't use `k8s` as import prefix, import it as `utilsruntime`, per Kubernetes convention.

cmd/eksctl/main.go
```diff
@@ -31,6 +35,17 @@ func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
 }
 
 func main() {
+
+	// Log internally-logged kubernetes errors uniformly
+	logK8sError := func(err error) {
```
errordeveloper: We don't use `k8s` in variable names, use `kube` if you must have something short.

cmd/eksctl/main.go
```diff
@@ -31,6 +35,17 @@ func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
...
+		// we need to go one level deeper than k8sruntime.HandleError
+		_, file, line, _ := runtime.Caller(1)
+		idx := strings.Index(file, "/k8s.io/")
+		finalFileLine := file[idx+1:] + ":" + strconv.Itoa(line)
```
errordeveloper: We don't log file paths in our own logs, let's drop it from here. We can discuss logging file paths separately.

2opremio:
> I'm not very clear that we need definitely need this. Port-forwarding appears to be an edge-case, all other client calls simply return errors right away. If we must use port-forwarding in #1064, let's just add it to that one command for now.

That was just an example. The Kubernetes libraries logging internal errors are not edge cases. It hspoens often in flux, for instance.

cmd/eksctl/main.go
2opremio: I disagree. Without the line you won't easily know where the internal Kubernetes error was logged. They are often cryptic and lack context

errordeveloper:
> That was just an example. The Kubernetes libraries logging internal errors are not edge cases. It hspoens often in flux, for instance.

Please provide 2-3 examples other then port-forwarding.

2opremio:
Just grep the Kubernetes codebase for `runtime.HandleError()`

2opremio:
To be honest, I don't have the energy to discuss this minor change. If you are happy with internal errors being printed out in an inconsistent way, just close this.

gemagomez:
This can wait until we have the time to deal with it properly.

cmd/eksctl/main.go
martina-if: What if we added this as a debug line? Would that be ok?

errordeveloper:
Close due to inactivity. I would rather see the use of port forwarding library being eliminated (#1228).

PR #1071, opened 2019-07-30
Rename `ctl.GetCredentials` to `ctl.GetActiveCluster`

errordeveloper:
# Description

This function was already doing more then just collecting
credetials and was due to be renamed for a while.

### Checklist

### DIFF
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -47,12 +47,12 @@ func (c *ClusterProvider) DescribeControlPlaneMustBeActive(cl *api.ClusterMeta) 
 	return cluster, nil
 }
 
-// GetActiveCluster calls c.DescribeControlPlaneMustBeActive and caches the results;
+// RefreshClusterConfig calls c.DescribeControlPlaneMustBeActive and caches the results;
 // it parses the credentials (endpoint, CA certificate) and stores them in spec.Status,
 // so that a Kubernetes client can be constructed; additionally it caches Kubernetes
 // version (use ctl.ControlPlaneVersion to retrieve it) and other properties in
 // c.Status.cachedClusterInfo
-func (c *ClusterProvider) GetActiveCluster(spec *api.ClusterConfig) error {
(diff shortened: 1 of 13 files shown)


pkg/eks/eks.go
```diff
@@ -47,8 +47,12 @@ func (c *ClusterProvider) DescribeControlPlaneMustBeActive(cl *api.ClusterMeta)
 	return cluster, nil
 }
 
-// GetCredentials retrieves cluster endpoint and the certificate authority data
-func (c *ClusterProvider) GetCredentials(spec *api.ClusterConfig) error {
+// GetActiveCluster calls c.DescribeControlPlaneMustBeActive and caches the results;
+// it parses the credentials (endpoint, CA certificate) and stores them in spec.Status,
+// so that a Kubernetes client can be constructed; additionally it caches Kubernetes
+// version (use ctl.ControlPlaneVersion to retrieve it) and other properties in
+// c.Status.cachedClusterInfo
```
cPu1: That's very well documented! :sparkles:

pkg/eks/eks.go
```diff
@@ -47,8 +47,12 @@ func (c *ClusterProvider) DescribeControlPlaneMustBeActive(cl *api.ClusterMeta)
...
+func (c *ClusterProvider) GetActiveCluster(spec *api.ClusterConfig) error {
```
cPu1: Minor nit: `GetSomething` sounds like it should return `Something`, but `GetActiveCluster`'s signature returns only an error. Since this method sets/populates fields on its receiver, what do you think about naming it `PopulateActiveCluster` or something else that better describes it?

D3nn: Looking at this a bit further, it seems to me that the main thrust of this is to update the passed in ClusterConfig and (incidentally?) also keeping a cached version of the cluster in the receiver.  If the main purpose is to update the passed in spec (ClusterConfig), WDYT about renaming to RefreshClusterConfig or something similar?

PR #1074, opened 2019-07-31
Implement `eksctl generate profile`

cPu1:
### Description
Implements `eksctl generate profile`

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/gitops/profile.go
+++ b/pkg/ctl/gitops/profile.go
@@ -1,0 +1,74 @@
+package gitops
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/weaveworks/eksctl/pkg/gitops"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+// Command creates `generate` commands
+func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
+	verbCmd := cmdutils.NewVerbCmd("generate", "Generate GitOps manifests", "")
+	cmdutils.AddResourceCmd(flagGrouping, verbCmd, generateProfileCmd)
+	return verbCmd
+}
+
+type options struct {
+	gitops.GitOptions
+	ProfilePath string
+}
+
+func generateProfileCmd(rc *cmdutils.ResourceCmd) {
+	cfg := api.NewClusterConfig()
+	rc.ClusterConfig = cfg
+
+	rc.SetDescription("profile", "Generate a GitOps profile", "")
+
+	var o options
+
+	rc.SetRunFuncWithNameArg(func() error {
+		return doGenerateProfile(rc, o)
+	})
+
+	rc.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&o.URL, "git-url", "", "", "Git repository URL")
+		fs.StringVarP(&o.Branch, "git-branch", "", "master", "Git branch; defaults to master")
+		fs.StringVarP(&o.ProfilePath, "profile-path", "", "", "Path to generate the profile in; defaults to CWD")
+		cobra.MarkFlagRequired(fs, "git-url")
+
+		cmdutils.AddNameFlag(fs, cfg.Metadata)
+		cmdutils.AddRegionFlag(fs, rc.ProviderConfig)
+		cmdutils.AddConfigFileFlag(fs, &rc.ClusterConfigFile)
+	})
+
+	cmdutils.AddCommonFlagsForAWS(rc.FlagSetGroup, rc.ProviderConfig, false)
+}
+
+func doGenerateProfile(rc *cmdutils.ResourceCmd, o options) error {
+	if err := cmdutils.NewMetadataLoader(rc).Load(); err != nil {
+		return err
+	}
+
+	profile := &gitops.Profile{
+		ClusterName: rc.ClusterConfig.Metadata.Name,
+		Path:        o.ProfilePath,
+		CloneFunc:   gitops.CloneContext,
+		Fs:          afero.NewOsFs(),
+		IO:          afero.Afero{Fs: afero.NewOsFs()},
+	}
+
+	err := profile.Generate(context.Background(), o.GitOptions)
+
+	if err != nil {
+		return errors.Wrap(err, "error generating profile")
+	}
+
+	return nil
+}
--- a/pkg/gitops/profile.go
+++ b/pkg/gitops/profile.go
@@ -74,15 +74,19 @@ func (p *Profile) Generate(ctx context.Context, o GitOptions) error {
 		return nil
 	}
 
-	manifestFiles, err := p.MakeOverlays()
-	if err != nil {
-		return err
-	}
-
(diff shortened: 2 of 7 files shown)


pkg/ctl/gitops/profile.go
```diff
@@ -0,0 +1,74 @@
...
+
+func doGenerateProfile(rc *cmdutils.ResourceCmd, o options) error {
+	if err := cmdutils.NewMetadataLoader(rc).Load(); err != nil {
+		return err
+	}
+
+	profile := &gitops.Profile{
+		ClusterName: rc.ClusterConfig.Metadata.Name,
+		Path:        o.ProfilePath,
+		CloneFunc:   gitops.CloneContext,
```
martina-if: I am separating all the git operations into it's own tested package perhaps we can inject that here. I'll probably put the PR up on monday.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,196 @@
...
+
+// CloneFunc is a function that clones a Git repository
+type CloneFunc func(ctx context.Context, path string, options *git.CloneOptions) error
+
+// Profile represents a GitOps profile
+type Profile struct {
+	ClusterName string
+	Path        string
+	CloneFunc   CloneFunc
+	Fs          afero.Fs
```
martina-if: Nice with the dependency injection! Perhaps we can encapsulate the Io operations into a package with an interface so firstly we share that code and we always use the same libraries and secondly, we then mock our own code instead of a third party library.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,196 @@
...
+
+// ManifestFile represents a manifest file with data
+type ManifestFile struct {
+	Name string
+	Data []byte
+}
+
+// Generate clones the specified Git repo in a base directory and generates overlays if the Git repo
+// points to a profile repo
+func (p *Profile) Generate(ctx context.Context, o GitOptions) error {
```
martina-if: Very nice separating this into a different package without dependencies to cobraCmd things

pkg/gitops/profile.go
cPu1: afero also provides a memory-backed filesystem that I have used in my tests. If we don't use a library like afero, we'd have to rewrite the same interfaces and a wrapper implementation that delegates to filesystem operations provided by the standard library. The mock code would also require checking if write operations were called with the right arguments.

pkg/ctl/gitops/profile.go
cPu1: I saw that code is shelling out to git for Git operations (like Flux does), but I'm using a pure Go library `go-git` for clone.

martina-if: Aha, I though there wasn't any library for git operations. In that case, why don't we use the git library for flux? Then we don't need to test the git code, only the integration part.

pkg/gitops/profile.go
martina-if: Ah! But then if we have the possibility to switch to in-memory calls with no side effects for the tests we don't need to encapsulate it. My concern was having to mock it during tests.

pkg/ctl/gitops/profile.go
cPu1: The `go-git` library has not implemented all features that `git` provides (https://github.com/src-d/go-git#comparison-with-git). If all the git operations we're using as part of `install flux` (commit and push) are supported by go-git (I believe it does) then I think we should use it. 
There's also a Go binding for libgit2 but I avoided it because it requires building the libgit2 library and modifying our build process.

pkg/gitops/profile.go
cPu1: For tests, I'm injecting the memory-based filesystem, so no mocking is required.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,230 @@
...
+	filename    string
+	patches     []jsonPatch
+	patchTarget *patch.Target
+}
+
+// MakeOverlays generates overlays for the default profile
+func (p *Profile) MakeOverlays() ([]ManifestFile, error) {
+	kustomizePatches := []kustomizePatch{
+		{
+			filename: "alb-patch.yaml",
```
martina-if: Can't we put this information in the repo somehow? It seems a bit strange that we try to keep the repo as the source of truth but then the code has these things hardcoded

pkg/gitops/profile.go
```diff
@@ -0,0 +1,230 @@
...
+	data, err := yaml.Marshal(fluxConfig)
+	if err != nil {
+		return errors.Wrap(err, "error marshaling Flux config")
+	}
+
+	return p.IO.WriteFile(filepath.Join(p.Path, ".flux.yaml"), data, 0644)
+}
+
+func matchesDefaultProfile(e *transport.Endpoint) bool {
+	return e.Host == defaultProfileGitHost && ((e.Protocol == "ssh" && e.Path == defaultProfileGitPath) ||
```
martina-if: don't we need to also check that it is a specified revision? Otherwise we will need to make every change in the repo backwards compatible.

cPu1: That is out of scope. Right now, we're assuming a recent version of `eksctl` will work with the latest revision. If we hard-code the revision and check for equality here, we'd have to update the client `eksctl` every time we push a change to the base repo, even if it doesn't necessitate changes to manifest generation. There's also a proposal to off-load manifest generation to an HTTP server that'd avoid the need to keep eksctl up-to-date with the base repo: https://docs.google.com/document/d/1riPEgKQY-Yei0-nG3Vl7zPrK6Wln4iGORUECsrcC50Q/edit#heading=h.uqo7rv6ekotf

pkg/gitops/profile.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+type kustomizePatch struct {
+	filename    string
+	patches     []jsonPatch
+	patchTarget *patch.Target
+}
+
+// MakeOverlays generates overlays for the default profile
+func (p *Profile) MakeOverlays() ([]ManifestFile, error) {
```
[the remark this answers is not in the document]
cPu1: I don't think of it as internal since it doesn't depend on any intermediate state and has no side effects. I prefer keeping tests in a separate `${package}_test`  package, where possible, to avoid having the tests rely too much on implementation details like unexported fields.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error cloning repository")
+	}
+
+	gitEndpoint, err := transport.NewEndpoint(o.URL)
+	if err != nil {
+		return errors.Wrap(err, "error parsing Git URL")
+	}
+
+	if !matchesDefaultProfile(gitEndpoint) {
```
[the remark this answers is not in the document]
cPu1: Yes, I still need to add logging. This is not an exceptional condition though but merely a non-default profile case for which no overlays are generated.  But you're right, we should at least add some logging info here.

pkg/gitops/profile.go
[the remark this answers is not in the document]
martina-if: Internal vs exported is based on whether a method is part of an API or not, and in this case it is not needed externally. 

Well, basically what we are doing here is to expose a method to the external API of a package just to be able to test it, so I disagree completely with this strategy of putting the tests in a different package (I know that this is how it's done through out the whole project but I think it's wrong).

What happens here is that we have a method that is not needed in the API but it's forcefully exposed to be able to test it. Instead is better to have the tests in the same pacakge and then we, as the maintainers of the code, decide when it is worth testing internal details of a package or not, but we can chose to do that without polluting an API.

errordeveloper:
@cPu1 I'm going to close this, will keep the branch for now.

PR #1095, opened 2019-08-05
Refactor: extract git logic from install flux

martina-if:
Extract git logic from install flux to another package to make code testable.

- [x] Manually tested


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/coreos/go-systemd v0.0.0-20190620071333-e64a0ec8b42a // indirect
 	github.com/dave/jennifer v1.3.0
 	github.com/dlespiau/kube-test-harness v0.0.0-20190110151726-c51c87635b61
-	github.com/docker/docker v1.13.1 // indirect
+	github.com/docker/docker v1.13.1
```
errordeveloper: Did you figure out why did this change?

martina-if: no :woman_shrugging:

errordeveloper:
@martina-if is this complete or you want to change something?

martina-if:
@errordeveloper this step is complete so I think the PR is good to go

pkg/ctl/install/install.go
```diff
@@ -138,12 +140,15 @@ func newFluxInstaller(rc *cmdutils.ResourceCmd, opts *installFluxOpts) (*fluxIns
 	if err != nil {
 		return nil, errors.Errorf("cannot create Kubernetes client set: %s", err)
 	}
+
+	gitClient := git.NewGitClient(ctx, opts.templateParams.GitUser, opts.templateParams.GitEmail, opts.timeout)
```
marccarre: Nit: we could pass a `GitClientParams` (or so) `struct` to group all git-related configuration & options.

pkg/ctl/install/install.go.orig
```diff
@@ -0,0 +1,657 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+<<<<<<< HEAD
```
marccarre: Is this `.orig` file an accidental addition? This looks like the output of git during some conflict resolution, so I'm guessing we don't want it here, do we?

martina-if: upssss thanks!

pkg/ctl/install/install.go
martina-if: since they are passed only once to a constructor of the git.Client struct I will leave them for now

martina-if:
Approved by @marccarre via slack

PR #1103, opened 2019-08-06
Clarify the need of write-access in `eksctl install flux`

2opremio:


pkg/ctl/install/install.go
```diff
@@ -206,8 +206,9 @@ func (fi *fluxInstaller) run(ctx context.Context) error {
 	}
 	cleanCloneDir = true
 
-	logger.Info("Flux will operate properly only once it has SSH access to: %s", fi.opts.templateParams.GitURL)
-	logger.Info("please add the following Flux public SSH key to your repository:\n%s", fluxSSHKey.Key)
+	logger.Info("Flux will operate properly only once it has SSH write-access")
```
martina-if: Could this be misunderstood? I would drop the `SSH` word here and leave it like: "Flux will operate properly only once it has SSH write-access" 
What do you think?

2opremio: > I would drop the `SSH` word here and leave it like: "Flux will operate properly only once it has SSH write-access"

Do you mean "Flux will operate properly only once it has write-access"?

martina-if: hahaha, sorry, yes, that's what I mean, but I don't have a strong opinion

2opremio: Makes sense, I will change it

PR #1106, opened 2019-08-07
Fix a number of links in documentation

dholbach:
linkchecker now says:

	That's it. 98 links in 98 URLs checked. 0 warnings found. 0 errors found.

### DIFF
--- a/site/content/introduction/01-getting-started.md
+++ b/site/content/introduction/01-getting-started.md
@@ -152,4 +152,4 @@ eksctl delete cluster --name=<name> [--region=<region>]
 
 ### Contributions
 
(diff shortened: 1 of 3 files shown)


dholbach:
In Flux we use something like this to check links (same in ignite):

- https://github.com/fluxcd/flux/blob/master/.circleci/config.yml#L95
- https://github.com/fluxcd/flux/blob/master/Makefile#L140-L141
- https://github.com/fluxcd/flux/blob/master/docs/Dockerfile#L5

Maybe move the site generation into a container too? Food for thought.

martina-if:
Thank you!

PR #1117, opened 2019-08-08
Rename `api.NodeGroupKubeletConfig` to `api.InlineDocument`

errordeveloper:
### Description

This simply renames the extension of `map[string]interface{}`, as it's of some general use.

### Checklist


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -594,16 +594,16 @@ type (
 	}
 )
 
-// NodeGroupKubeletConfig contains extra config parameters for the kubelet.yaml
-type NodeGroupKubeletConfig map[string]interface{}
+// InlineDocument contains extra config parameters for the kubelet.yaml
```
cPu1: Since you mentioned elsewhere that `InlineDocument` will get used to enable inline IAM policies, I think we should modify this doc comment to reflect that.

errordeveloper: Ah, yes, I'll generalise the comment, thanks!

PR #1119, opened 2019-08-08
Kubernetes client improvements

### LINKED ISSUE #1118: eksctl should not enforce namespace labels without a functional requirement
At the moment namespace helpers in `pkg/kubernetes` set `name` label automatically, it is not at all necessary and shouldn't need to be set.

errordeveloper:
### Description

- Improve Kubernetes client wrapper interface
  - add an alias for `"k8s.io/client-go/kubernetes".Interface`
  - define clientset getter interface
- Improve namespace object helpers
  - add client helpers
  - remove `name` label (close #1118)
  - add tests
- Add serviceaccount object helpers

### Checklist

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -25,16 +25,58 @@ import (
 	"k8s.io/client-go/restmapper"
 )
 
+// Interface is an alias to avoid having to import k8s.io/client-go/kubernetes
+// along with this package, so that most of our packages only care to import
+// our kubernetes package
+type Interface = kubeclient.Interface
+
+// ClientSetGetter is an interface used for anything that requires
+// to obtain Kubernetes client whe it's not possible to pass the
+// client directly
+type ClientSetGetter interface {
+	ClientSet() (Interface, error)
+}
+
+// CachedClientSet provides a basic implementation of ClientSetGetter
+// where the client is a field of a struct
+type CachedClientSet struct {
+	CachedClientSet Interface
+}
+
+// NewCachedClientSet costructs a new CachedClientSets
+func NewCachedClientSet(clientSet Interface) *CachedClientSet {
+	return &CachedClientSet{CachedClientSet: clientSet}
+}
+
+// ClientSet returns g.CachedClientSet or an error it is nil
+func (g *CachedClientSet) ClientSet() (Interface, error) {
+	if g.CachedClientSet == nil {
+		return nil, fmt.Errorf("no client instance provided")
+	}
+	return g.CachedClientSet, nil
+}
+
+// CallbackClientSet provides an implementation of ClientSetGetter
+// where the client is provided via a callback
+type CallbackClientSet struct {
+	Callback func() (Interface, error)
+}
+
+// ClientSet returns g.ClientSet or an error it is nil
+func (g *CallbackClientSet) ClientSet() (Interface, error) {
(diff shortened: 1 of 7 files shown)


pkg/kubernetes/namespace.go
```diff
@@ -2,39 +2,71 @@ package kubernetes
...
+	nsFmt := strings.Join(
+		[]string{
+			"---",
+			"apiVersion: v1",
+			"kind: Namespace",
+			"metadata: {name: %s}",
+		},
+		"\n")
+
+	return []byte(fmt.Sprintf(nsFmt, name))
```
cPu1: Any reason for not using `yaml.Marshal` from `sigs.k8s.io/yaml` here? Then we could do:

```golang
func NewNamespaceYAML(name string) ([]byte, error) {
    return yaml.Marshal(NewNamespace(name))
}
```

This would be more type-safe.

pkg/kubernetes/serviceaccount.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if current.Labels == nil {
+		current.Labels = make(map[string]string)
+	}
+	for k, v := range meta.Labels {
+		cv, ok := current.Labels[k]
+		updateRequired = !ok || ok && cv != v
+		current.Labels[k] = v
+	}
```
cPu1: WDYT about extracting the two loops out into an anonymous function?

pkg/kubernetes/client.go
```diff
@@ -25,16 +25,58 @@ import (
 	"k8s.io/client-go/restmapper"
 )
 
+// Interface is an alias to avoid having to import k8s.io/client-go/kubernetes
+// along with this package, so that most of our packages only care to import
+// our kubernetes package
+type Interface = kubeclient.Interface
```
cPu1: Nice use of a type alias!

pkg/kubernetes/client.go
```diff
@@ -25,16 +25,58 @@ import (
...
+
+// NewCachedClientSet costructs a new CachedClientSets
+func NewCachedClientSet(clientSet Interface) *CachedClientSet {
+	return &CachedClientSet{CachedClientSet: clientSet}
+}
+
+// ClientSet returns g.CachedClientSet or an error it is nil
+func (g *CachedClientSet) ClientSet() (Interface, error) {
+	if g.CachedClientSet == nil {
+		return nil, fmt.Errorf("no client instance provided")
```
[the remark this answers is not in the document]
errordeveloper: I guess that was an extra check I though I should provide, to avoid a potential panic as the field needs to be exported.

pkg/kubernetes/namespace.go
errordeveloper: It's in the comment - we want to output a clean object without `creationTimestap: nil`, as it will be written out to user's git repo and they might end up editing it or at least looking at it.

pkg/kubernetes/serviceaccount.go
errordeveloper: Yeah, sound like a good idea, thanks!

pkg/kubernetes/namespace.go
cPu1: Ah, I missed that comment since it wasn't introduced in this PR. Thanks for clarifying.

PR #1121, opened 2019-08-08
Start new arelease notes file for 0.4.0

errordeveloper:


docs/release_notes/0.4.0.md
```diff
@@ -0,0 +1,9 @@
+# Release 0.4.0
+
+## Features
+- ...
```
cPu1: Should we leave this line out until we have a feature to add here?

errordeveloper: We will need to review the notes before the release goes out anyway...

whereisaaron:
```
## Features

Includes all features of the previous version!
```

PR #1124, opened 2019-08-08
Improve consistency of command functions

errordeveloper:
### Description

A follow up to #1116.

### Checklist

### DIFF
--- a/pkg/ctl/cmdutils/resource.go
+++ b/pkg/ctl/cmdutils/resource.go
@@ -7,6 +7,7 @@ import (
 	"github.com/spf13/cobra"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/eks"
 )
 
 // ResourceCmd holds attributes that most of the commands use
@@ -14,7 +15,7 @@ type ResourceCmd struct {
 	Command      *cobra.Command
 	FlagSetGroup *NamedFlagSetGroup
 
-	Plan, Wait bool
+	Plan, Wait, Validate bool
 
 	NameArg string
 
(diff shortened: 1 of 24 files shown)


pkg/ctl/scale/nodegroup.go
```diff
@@ -44,6 +45,10 @@ func doScaleNodeGroup(rc *cmdutils.ResourceCmd, ng *api.NodeGroup) error {
 
 	ctl := eks.New(rc.ProviderConfig, cfg)
```
errordeveloper: This command doesn't support `--config-file`, so we don't do defaulting/validation here.

rndstr: maybe add that as a comment?

rndstr: maybe add this as a comment?

errordeveloper:
@rndstr thanks, indeed this is getting a little mad. I've added a method that wraps it up.

pkg/ctl/create/cluster.go
```diff
@@ -97,21 +96,11 @@ func doCreateCluster(rc *cmdutils.ResourceCmd, params *createClusterCmdParams) e
...
-	if err := ngFilter.ValidateNodeGroupsAndSetDefaults(cfg.NodeGroups); err != nil {
-		return err
-	}
-
 	printer := printers.NewJSONPrinter()
-	ctl := eks.New(rc.ProviderConfig, cfg)
 
-	if !ctl.IsSupportedRegion() {
-		return cmdutils.ErrUnsupportedRegion(rc.ProviderConfig)
+	ctl, err := rc.NewCtl()
```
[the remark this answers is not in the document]
errordeveloper: Indeed, but let's make that change separately. I'm just trying to take care of consistency here. I think that change will be easier to make now.

Actually, we could just hide that part inside of `NewCtl`, i.e. make it possible to set the provider implementation in `ResourceCmd` (which gets constructed before the handler function runs).

pkg/ctl/cmdutils/resource.go
```diff
@@ -26,14 +27,49 @@ type ResourceCmd struct {
...
+				return nil, err
+			}
+			logger.Warning("ignoring validation error: %s", err.Error())
+		}
+		// defaulting of nodegroup currently depends on validation;
+		// that may change, but at present that's how it's meant to work
+		api.SetNodeGroupDefaults(i, ng)
+	}
+
+	ctl := eks.New(rc.ProviderConfig, rc.ClusterConfig)
```
errordeveloper: This could be something like:
```
if rc.MockProvider != nil {
  return rc.MockProvider, nil
}

ctl := eks.New(rc.ProviderConfig, rc.ClusterConfig)
```

errordeveloper: But let's leave until next iteration, I also have a PR to rename `rc` to simply `cmd`, which I promised to Martina a long time ago.

pkg/ctl/create/cluster.go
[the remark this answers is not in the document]
cPu1: Sounds good to do it later in a separate PR.

PR #1129, opened 2019-08-09
Refactor how additional cluster config tasks are handled

errordeveloper:
### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/eks/tasks.go
+++ b/pkg/eks/tasks.go
@@ -1,0 +1,44 @@
+package eks
+
+import (
+	"github.com/kris-nova/logger"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/cfn/manager"
+)
+
+type clusterConfigTask struct {
+	info string
+	spec *api.ClusterConfig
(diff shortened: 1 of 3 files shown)


pkg/eks/tasks.go
```diff
@@ -0,0 +1,44 @@
...
+func (t *clusterConfigTask) Describe() string { return t.info }
+
+func (t *clusterConfigTask) Do(errs chan error) error {
+	err := t.call(t.spec)
+	close(errs)
+	return err
+}
+
+// AppendExtraClusterConfigTasks returns all tasks for updating cluster configuration or nil if there are no tasks
+func (c *ClusterProvider) AppendExtraClusterConfigTasks(cfg *api.ClusterConfig, tasks *manager.TaskTree) {
```
cPu1: Could we leave `Extra` out of the method's name? I think `Append` conveys that information.

```suggestion
func (c *ClusterProvider) AppendClusterConfigTasks(cfg *api.ClusterConfig, tasks *manager.TaskTree) {
```

errordeveloper: Sort of, but they are all optional, so I'd rather stick to a more explicit name...

errordeveloper:
Thanks for the review, Chetan!

PR #1130, opened 2019-08-09
WIP: Testing out an initial MD pattern for reference docs

sbernheim:
Submitting as a WIP PR for feedback and possible testing against the docs site build pipeline tooling.

**DO NOT MERGE THIS PR!**

### DIFF
--- a/site/content/reference/01-commands.md
+++ b/site/content/reference/01-commands.md
@@ -5,27 +5,34 @@ weight: 10
 
 # Commands reference
 
-```
 a CLI for Amazon EKS
 
-Usage: eksctl [command] [flags]
+Usage: `eksctl [command] [flags]`
 
 Commands:
-  eksctl completion  Generates shell completion scripts
-  eksctl create      Create resource(s)
-  eksctl delete      Delete resource(s)
-  eksctl drain       drain resources(s)
-  eksctl get         Get resource(s)
-  eksctl help        Help about any command
-  eksctl scale       Scale resources(s)
-  eksctl update      Update resource(s)
-  eksctl utils       Various utils
-  eksctl version     Output the version of eksctl
 
-Use 'eksctl [command] --help' for more information about a command.
-```
+  `eksctl completion`  Generates shell completion scripts
 
-`Usage: eksctl [command] [flags]`
+  `eksctl create`      Create resource(s)
+
+  `eksctl delete`      Delete resource(s)
+
+  `eksctl drain`       drain resources(s)
+
+  `eksctl get`         Get resource(s)
+
+  `eksctl help`        Help about any command
+
+  `eksctl scale`       Scale resources(s)
+
+  `eksctl update`      Update resource(s)
+
+  `eksctl utils`       Various utils
+
+  `eksctl version`     Output the version of eksctl
+
+Use `eksctl [command] --help` for more information about a command.
+
 
 ## common flags<a name="common-flags"></a>
 
@@ -40,7 +47,7 @@ These flags can be applied to all eksctl commands and behave consistently betwee
 
 Generates shell completion scripts for various Unix command-line shells.
 
-`Usage: eksctl completion [command]`
+Usage: `eksctl completion [command]`
 
 The `completion` commands do not support any additional flags.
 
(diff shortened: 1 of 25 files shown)


site/content/reference/01-commands.md
```diff
@@ -0,0 +1,233 @@
...
+  eksctl update      Update resource(s)
+  eksctl utils       Various utils
+  eksctl version     Output the version of eksctl
+
+Use 'eksctl [command] --help' for more information about a command.
+```
+
+`Usage: eksctl [command] [flags]`
+
+##### common flags<a name="common-flags"></a>
```
bia: I would not use `h4` or `h5` since they are small enough to look confusing in the context of our base font size.

bia:
Nice one!

The syntax highlighting is behaving strangely here. We could also go for line-by-line code snippets?
Screenshot 2019-08-09 at 17 50 10

site/content/reference/01-commands.md
sbernheim: OK - let's connect Monday morning and figure out how to amend the tags.  I want it to look good as rendered, but it should remain a structured doc like an outline format somehow.

sbernheim:
@bia Please take a look at `03-config-yaml.md` and related links between flags and config yaml attributes when you get a chance.

We can substitute pre-formatted blocks with something else if that works better.  The example for the top-level in the `01-commands.md` file is just a placeholder.  We can swap in whatever would work better for Hugo.

site/content/reference/01-commands.md
```diff
@@ -0,0 +1,235 @@
...
+
+These flags can be applied to all eksctl commands and behave consistently between them.
+
+- [C, color](02-flags.md#color)
+- [h, help](02-flags.md#help)
+- [v, verbose](02-flags.md#verbose)
+
+### completion <a name="completion"></a>
```
gemagomez: This is very linux centric. Do we really need to document this behaviour here?

site/content/reference/01-commands.md
```diff
@@ -0,0 +1,235 @@
...
+
+Generates shell completion scripts for various Unix command-line shells.
+
+`Usage: eksctl completion [command]`
+
+The `completion` commands do not support any additional flags.
+
+Without an invoking a sub-command, the completion command simply prints the help screen.
+
+#### bash <a name="completion-bash"></a>
```
gemagomez: Same as before.

site/content/reference/01-commands.md
```diff
@@ -0,0 +1,235 @@
...
+- [authenticator-role-arn](02-flags.md#authenticator-role-arn)
+- [set-kubeconfig-context](02-flags.md#set-kubeconfig-context)
+- [auto-kubeconfig](02-flags.md#auto-kubeconfig)
+- [write-kubeconfig](02-flags.md#write-kubeconfig)
+
+#### iamidentitymapping <a name="create-iamidentitymapping"></a>
+
+Creates a mapping from IAM role to Kubernetes user and groups.
+
+*Note:* aws-iam-authenticator only considers the last entry for any given
```
gemagomez: \`aws-iam-authenticator\`

site/content/reference/01-commands.md
sbernheim: I think Linux users would appreciate some documentation, yes. 😄 

It's also supposed to cover MacOS usage, but I think it might be inaccurate.  😐 

Reference documentation is supposed to be comprehensive, so if the code supports a feature you'll want to document that feature - no matter how trivial it may seem.  If it supports tab completion for a Unix/Linux shell, it should really be documented.

This is literally cut-and-paste from the help screens, so it was easy to include.  I don't know why you wouldn't.

site/content/reference/01-commands.md
sbernheim: `bash` users are people too!!!  😄 

`zsh` is an even more obscure shell, but someone liked it enough to write completion support for it, so who am I to judge?

errordeveloper:
@sbernheim are you planning to continue working on this?

martina-if:
@sbernheim given that it seems we can't put much time on this right now I will close this but please feel free to reopen it if you have time again!

PR #1131, opened 2019-08-12
Add generation of cluster autoscaler addon's manifest

marccarre:
### Description

This PR essentially adds:
```go
func GenerateManifests(params TemplateParameters) ([]byte, error)
```
under `pkg/addons/clusterautoscaler`, which can be used to generate the manifests required to easily [install the cluster autoscaler on EKS](https://aws.amazon.com/premiumsupport/knowledge-center/eks-cluster-autoscaler-setup/).

Note that, it relies on a YAML Go template which is embedded in the `eksctl` binary using `vfsgen`, added to the dependencies via:
```bash
$ GO111MODULE=on go get -u github.com/shurcooL/vfsgen
$ GO111MODULE=on go mod vendor
```

### Checklist

- [x] ~Manually tested~

### DIFF
--- a/pkg/addons/clusterautoscaler/cluster_autoscaler.go
+++ b/pkg/addons/clusterautoscaler/cluster_autoscaler.go
@@ -1,0 +1,82 @@
+package clusterautoscaler
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"html/template"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/shurcooL/httpfs/vfsutil"
+	"github.com/weaveworks/eksctl/pkg/kubernetes"
+)
+
+const (
+	defaultNamespace    = "kube-system"
+	defaultImageVersion = "v1.12.3"
+)
+
+// TemplateParameters groups the parameters required to configure cluster
+// autoscaling.
+// N.B.: these parameters are injected into the cluster autoscaler's YAML
+// template, hence their names need to be consistent both here and in the
+// template.
+type TemplateParameters struct {
+	Namespace    string // Namespace under which cluster autoscaler resources will be created.
+	ClusterName  string // Name of the EKS cluster, as configured in eksctl's manifest or via CLI flags.
+	ImageVersion string // Version of the k8s.gcr.io/cluster-autoscaler image.
+}
+
+// Validate validates the values currently set and defaults the ones missing.
+func (p *TemplateParameters) Validate() error {
+	if p.ClusterName == "" {
+		return errors.New("blank cluster name")
+	}
+	if p.Namespace == "" {
+		p.Namespace = defaultNamespace
+	}
+	if p.ImageVersion == "" {
+		p.ImageVersion = defaultImageVersion
+	}
+	return nil
+}
+
+// GenerateManifests fills the cluster autoscaler's YAML templates with the
+// provided template parameters to generate actionable YAML manifests.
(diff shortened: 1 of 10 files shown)


marccarre:
N.B.: given we want to use an abstraction for templates' processing, this PR now depends on #1132. Once that PR is merged, I'll update _this one_ to use that abstraction.

pkg/addons/clusterautoscaler/cluster_autoscaler.go
```diff
@@ -0,0 +1,82 @@
...
+// GenerateManifests fills the cluster autoscaler's YAML templates with the
+// provided template parameters to generate actionable YAML manifests.
+func GenerateManifests(params TemplateParameters) ([]byte, error) {
+	if err := params.Validate(); err != nil {
+		return nil, err
+	}
+	manifests := [][]byte{}
+	err := vfsutil.WalkFiles(Templates, "/", func(path string, info os.FileInfo, rs io.ReadSeeker, err error) error {
+		if err != nil {
+			return fmt.Errorf("cannot walk embedded files: %s", err)
```
rndstr: nit: may also use `errors.Wrap()`

pkg/addons/clusterautoscaler/cluster_autoscaler_test.go
```diff
@@ -0,0 +1,49 @@
+package clusterautoscaler_test
+
+import (
+	"testing"
```
errordeveloper: Did you consider using ginko/gomega? That's what all other tests use...

pkg/addons/clusterautoscaler/templates_generate.go
```diff
@@ -0,0 +1,21 @@
+// +build ignore
+
+package main
+
+import (
+	"log"
+
+	"github.com/shurcooL/vfsgen"
```
errordeveloper: We use bindata in other place, is there any particular reason to use this package here?

marccarre: Didn't know about `bindata`. I used `vfsgen` as that's what is used in Flux and in WKS.

pkg/addons/clusterautoscaler/cluster_autoscaler_test.go
marccarre: Sure, will re-write... 😓

pkg/addons/clusterautoscaler/cluster_autoscaler.go
marccarre: Good point. (FWIW, I took this from Flux, and didn't pay too much attention to such details.)

martina-if:
I think that if the other components are in a base repo (today "eks-gitops-example") this ones should also be in that quickstart repo as yaml templates. See https://github.com/weaveworks/eks-gitops-example/pull/3#event-2545776922

I don't understand why tehse are special and have to be  embedded in eksctl. Is it because it needs different variables? We can incorporate those in the TemplateParameters.

Otherwise it's going to be tricky to have a user created quickstart repo that also has autoscaler, for example.

marccarre:
@martina-if, they don't have to be embedded. That's just the result of me misunderstanding what we wanted, as:

- WKS does addons this way ;
- it had been mentioned in a call that we wanted something similar to fluxcd/flux/pull/2287, also similar to WKS ;
- it had been mentioned in some user stories, that we could also have `eksctl` generate these.

We may want this in the future, but apparently not now, so I'm gonna close this PR.

martina-if:
Sorry for the misunderstanding marc!

PR #1132, opened 2019-08-12
eksctl generate profile with Go templates

martina-if:
In this first iteration we add support for go templates for profiles.
Following will be a PR with support for JK instead

- [x] Manually tested


pkg/gitops/fileprocessor/gotemplates.go
```diff
@@ -0,0 +1,55 @@
...
+	"bytes"
+	"github.com/pkg/errors"
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"path/filepath"
+	"strings"
+	"text/template"
+)
+
+const (
+	templateExtension = ".templ"
```
marccarre: Über nit: we use `.tmpl` in Flux and other projects, so might as well be consistent?

pkg/gitops/fileprocessor/gotemplates.go
```diff
@@ -0,0 +1,55 @@
...
+// NewTemplateParameters creates a set of variables for templating given a ClusterConfig object
+func NewTemplateParameters(clusterConfig *api.ClusterConfig) TemplateParameters {
+	return TemplateParameters{
+		ClusterName: clusterConfig.Metadata.Name,
+	}
+}
+
+// GoTemplateProcessor is a FileProcessor that executes Go Templates
+type GoTemplateProcessor struct {
+	Params TemplateParameters
```
marccarre: Is there a reason why we "pass" the `File` as an argument, but the `TemplateParameters` as a field?
If feels a bit strange to me, given they the two operands of a "generate manifest" operation. So I'd rather have them both passe to `ProcessFile`.

martina-if: The reason is that the templates are stable for all the processing since they are our API so you only need one value of it. Then `processFile()` is called once per file with the same contents of the TemplateParameters. Does that make more sense?

martina-if:
I would like to tackle the documentation in a different PR

go.mod
```diff
@@ -84,6 +85,7 @@ require (
 	k8s.io/kops v0.0.0-20190222135932-278e6606534e
 	k8s.io/kubelet v0.0.0-20190313123811-3556bcde9670
 	k8s.io/kubernetes v1.12.6
+	sigs.k8s.io/kustomize v2.0.3+incompatible // indirect
```
2opremio: I guess we don't need this anymore. Did you run `go mod tidy` ?

martina-if: No, I didn't know about it, thanks :+1:

pkg/gitops/fileprocessor/gotemplates.go
marccarre: Fair enough 👍

integration/gitopscreateprofile_test.go
```diff
@@ -0,0 +1,90 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/spf13/afero"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+var _ = Describe("(Integration) generate profile", func() {
```
errordeveloper: Should we make it easy to run this in isolation from the rest of the suite?
Long-term, it'd be very nice to have parallel suites also :)

integration/gitopscreateprofile_test.go
```diff
@@ -0,0 +1,90 @@
...
+				"--profile-path", testDirectory,
+			)
+
+			fs := afero.Afero{
+				Fs: afero.NewOsFs(),
+			}
+
+			contents, err := fs.ReadFile(filepath.Join(testDirectory, "workloads/namespace.yaml"))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(string(contents)).To(Equal(
```
errordeveloper: Actually, you want to use `MatchYAML` here.

```
Expect(contents).To(MatchYAML(expected))
```
Where `contents` and `expected` can be `string`, or `[]byte`, as long as both are the same type.

integration/gitopscreateprofile_test.go
```diff
@@ -0,0 +1,90 @@
...
+metadata:
+  labels:
+    name: amazing-testing-gopher
+  name: amazing-testing-gopher
+`))
+
+			contents, err = fs.ReadFile(filepath.Join(testDirectory, "workloads/services/service.yaml"))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(string(contents)).To(Equal(
+				`apiVersion: v1
```
errordeveloper: I must admit, I do find inline YAML very irritating as it messes with indentation. But I can tolerate it in tests.

Declaring it with `---` on the first line does make it a little more readable.

pkg/gitops/gitops_test.go
```diff
@@ -0,0 +1,194 @@
+package gitops
+
+import (
```
errordeveloper: nit: grouping of imports

pkg/gitops/gitops_test.go
```diff
@@ -0,0 +1,194 @@
...
+			io.RemoveAll(outputDir)
+		})
+
+		It("processes go templates and writes them to the output directory", func() {
+			err := profile.Generate(context.Background())
+
+			Expect(err).ToNot(HaveOccurred())
+			template1, err := io.ReadFile(filepath.Join(outputDir, "a/good-template1.yaml"))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(template1).To(Equal([]byte("cluster: test-cluster")))
```
errordeveloper: I guess less important for single line, but you could use `MatchYAML` here also.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+	cloneDirPrefix = "quickstart-"
+)
+
+// Profile represents a GitOps profile
+type Profile struct {
+	Processor fileprocessor.FileProcessor
+	Path      string
+	GitOpts   GitOptions
+	GitCloner git.Cloner
+	Fs        afero.Fs
```
errordeveloper: Should be either `FS`, or `Filesystem`...

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+	IO        afero.Afero
```
errordeveloper: Why not call it `Aero`?

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
+package gitops
+
+import (
```
errordeveloper: nit: import grouping

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+	"context"
+	"github.com/kris-nova/logger"
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
```
errordeveloper: Do you think it's the right kind of thing? If so, perhaps it would be good to open an issue and where we will enumerate which package have filesystem operation that need better testing and could benefit from something like aero?

integration/gitopscreateprofile_test.go
martina-if: Yes, How can I do that?

integration/gitopscreateprofile_test.go
martina-if: wooooo Nice! thanks, I will change it

integration/gitopscreateprofile_test.go
martina-if: I agree, Is there a better way? I can for example move it to a constant somewhere else, would that be better?

pkg/gitops/profile.go
martina-if: I just called it IO because that's what we had in the previous version of this PR. I can rename it.

martina-if: Actually I thought about it makes more sense that it has a more generic name related to what it does rather than the library it uses. so unless there are strong opinions I would like to keep it as `IO`.

pkg/gitops/profile.go
martina-if: I learned from the other PR that this package helps us make the code testable by using an inMemory filesystem. Yes, I can definitely open a ticket about it.

pkg/gitops/fileprocessor/gotemplates.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// ProcessFile takes a template file and executes the template applying the TemplateParameters
+func (p *GoTemplateProcessor) ProcessFile(file File) (*File, error) {
```
cPu1: I may be missing some context, but I see no reason why the return type is a pointer instead of a value.

pkg/gitops/fileprocessor/gotemplates.go
```diff
@@ -0,0 +1,60 @@
...
+	if !isGoTemplate(file.Path) {
+		logger.Debug("copying non template file %q", file.Path)
```
cPu1: I know this is just a logging statement, but `GoTemplateProcessor` isn't aware that the files are being copied. Maybe this should read something along these lines:
```suggestion
		logger.Debug("leaving non-template file unmodified %q", file.Path)
```

pkg/git/git.go
```diff
@@ -44,10 +44,15 @@ func (git *Client) CloneRepo(cloneDirPrefix string, branch string, gitURL string
...
+	return cloneDir, git.CloneRepoInPath(cloneDir, branch, gitURL)
+}
+
+// CloneRepoInPath clones a repo to the specified directory
+func (git *Client) CloneRepoInPath(clonePath string, branch string, gitURL string) error {
+	git.dir = clonePath
 	args := []string{"clone", "-b", branch, gitURL, git.dir}
-	err = git.runGitCmd(args...)
-	return git.dir, err
+	err := git.runGitCmd(args...)
```
cPu1: Very minor nit:
```suggestion
	return git.runGitCmd(args...)
```

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	github.com/riywo/loginshell v0.0.0-20190610082906-2ed199a032f6
 	github.com/sanathkr/yaml v1.0.0 // indirect
 	github.com/sirupsen/logrus v1.4.2 // indirect
+	github.com/spf13/afero v1.2.2
```
2opremio: You will need to bump the dependency image version in the image and push it to the registry before merging

2opremio: PS: we need to automate this but I haven't had the time yet

cmd/eksctl/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/weaveworks/eksctl/pkg/ctl/generate"
```
2opremio: This should go at the end. As a suggestion, can you configure your IDE to do it automatically (in Goland, you can)

pkg/ctl/generate/profile.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/git"
+	"github.com/weaveworks/eksctl/pkg/gitops"
+	"github.com/weaveworks/eksctl/pkg/gitops/fileprocessor"
+)
+
+const (
```
2opremio: You don't need parenthesis for a single constant

pkg/ctl/generate/profile.go
```diff
@@ -0,0 +1,84 @@
...
+
+	var o options
+
+	rc.SetRunFuncWithNameArg(func() error {
+		return doGenerateProfile(rc, o)
+	})
+
+	rc.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&o.URL, "git-url", "", "", "Git repository URL")
+		fs.StringVarP(&o.Branch, "git-branch", "", "master", "Git branch; defaults to master")
```
2opremio: AFAIK the defaulting is printed in the help, no need to made it explicit

pkg/ctl/generate/profile.go
```diff
@@ -0,0 +1,84 @@
...
+	rc.SetDescription("profile", "Generate a GitOps profile", "")
+
+	var o options
+
+	rc.SetRunFuncWithNameArg(func() error {
+		return doGenerateProfile(rc, o)
+	})
+
+	rc.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&o.URL, "git-url", "", "", "Git repository URL")
```
2opremio: What repository, what for?

pkg/ctl/generate/profile.go
```diff
@@ -0,0 +1,84 @@
...
+		fs.StringVarP(&o.Branch, "git-branch", "", "master", "Git branch; defaults to master")
+		fs.StringVarP(&o.ProfilePath, "profile-path", "", "", "Path to generate the profile in; defaults to CWD")
```
2opremio: Shouldn't the default be `.` and avoid saying it explicitly?

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+	logger.Info("processing template files in repository")
+	outputFiles, err := p.processFiles(allManifests, clonedDir)
+	if err != nil {
+		return errors.Wrapf(err, "error processing manifests from repository %s", p.GitOpts.URL)
+	}
+
+	if len(outputFiles) > 0 {
+		logger.Info("writing new manifests to %q", p.Path)
+	} else {
+		logger.Info("no template files found, nothing to write")
```
cPu1:
```suggestion
		logger.Info("no template files found, nothing to write")
		return nil
```

pkg/git/git.go
```diff
@@ -3,37 +3,37 @@ package git
...
 type Client struct {
 	executor executor.Executor
 	dir      string
-	user     string
-	email    string
 }
 
 // NewGitClient returns a client that can perform git operations
-func NewGitClient(ctx context.Context, user string, email string, timeout time.Duration) *Client {
+func NewGitClient(ctx context.Context, timeout time.Duration) *Client {
```
2opremio: Unrelated to this PR, but contexts shouldn't be stored, but passed to the operations.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+type GitOptions struct {
+	URL    string
+	Branch string
+}
+
+// Generate clones the specified Git repo in a base directory and generates overlays if the Git repo
+// points to a profile repo
+func (p *Profile) Generate(ctx context.Context) error {
+	logger.Info("cloning repository %q:%s", p.GitOpts.URL, p.GitOpts.Branch)
+	clonedDir, err := p.GitCloner.CloneRepo(cloneDirPrefix, p.GitOpts.Branch, p.GitOpts.URL)
```
2opremio: Where is `clonedDir` deleted? I can't find it.

2opremio: (Maybe the cloner should output a cleanup function to delete it?)

integration/gitopscreateprofile_test.go
errordeveloper: I don't know, something that we will need to determine, we might want to play with focus options or tweak the main function of the integration suite... Perhaps best to get this in first?

integration/gitopscreateprofile_test.go
errordeveloper: A constant at the bottom of the file would be nice :)

errordeveloper: Actually, for tests we could just have a file in 'testdata/', but these are fairly short snippets, so it is kind of more convenient to have theme inline...

go.mod
errordeveloper: Can we make the image tag a hash of dependencies? There is a way to make a git hash but I don't recall what it takes to do, otherwise just a sha sum of the content would seem sufficient?

pkg/ctl/generate/profile.go
errordeveloper: Personally, I prefer to always use them outside functions, same with imports, it is just easier to read and makes package constants standout.

pkg/ctl/generate/profile.go
errordeveloper: If so, I'd rather see it as './', a dot can be hard to notice.

cmd/eksctl/main.go
martina-if: Got it, I think I didn't have my IDE correctly configured.

pkg/gitops/fileprocessor/gotemplates.go
martina-if: Good point, changing

pkg/gitops/profile.go
martina-if: good catch!

pkg/gitops/profile.go
martina-if: Good point. Done.

pkg/gitops/profile.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (p *Profile) processFiles(files []fileprocessor.File, baseDir string) ([]fileprocessor.File, error) {
+	outputFiles := make([]fileprocessor.File, 0, len(files))
+	for _, file := range files {
+		outputFile, err := p.Processor.ProcessFile(file)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error processing file %q ", file.Path)
+		}
+		if outputFile == nil {
```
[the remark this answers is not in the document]
martina-if: As we agreed on slack, I will change it to a value and we can always change it back to a pointer if we need to. :+1:

pkg/gitops/profile.go
martina-if: Done

go.mod
martina-if: wouldn't it then be more tricky to do this process? because we need to replace that hash in the Makefile. Unless it is done automatically... but then it's not so easy to know if a given image version is what you need when added a new dependency :/

errordeveloper: Yeah, I was thinking of automatic hashing. If we use git hashes, you can use 'git show <hash>' to tell what went in. I mean tree hashes, not actual commits.

errordeveloper: #1145

pkg/gitops/profile.go
2opremio: Did you?

2opremio: Ah, you meant you deleted it later, not that you added the cleanup function

PR #1134, opened 2019-08-12
Slim alternative to goformation

### LINKED ISSUE #1133: defork goformation
We've used a fork of goformation since #132. 

The initial attempt of contributing our changes back to goformation (https://github.com/awslabs/goformation/pull/103) didn't succeed, and goformation maintainers chose a different approach (https://github.com/awslabs/goformation/pull/114). However, their solution is not bi-directional, has a number of limitations/bugs (e.g. https://github.com/awslabs/goformation/issues/190 and https://github.com/awslabs/goformation/issues/194).

The main challenge is that mapping CloudFormation scheme to Go types is very complicated if you consider a general case and support of all types is required as well as Serverless Application Model (SAM), and that is what goformation attempts.
However, we only used a relative small number of types in a particular way.

It appears to be easy enough to create a slim library that offers a subset of goformation functionality that does things the way that suites our needs better without having to take all general cases into account. Aside from being able to round-trip templates from running stacks back to internal structures for mutation and updates, we could provide convenient abstractions for testing of template rendering.

errordeveloper:
### Description

A starting point for #1133.

### Checklist

### DIFF
--- a/pkg/cfn/template/api_test.go
+++ b/pkg/cfn/template/api_test.go
@@ -1,0 +1,72 @@
+package template
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("CloudFormation template", func() {
+	It("can create and render a minimal template", func() {
+		t := NewTemplate()
+		t.Description = "a template"
+
+		roleRef := t.NewResource("aRole", &IAMRole{
+			RoleName:          "foo",
+			ManagedPolicyArns: []string{"abc"},
+		})
+
+		jsRoleRef, err := roleRef.MarshalJSON()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(jsRoleRef).To(MatchJSON(`{"Ref": "aRole"}`))
+
+		policyRef := t.NewResource("aPolicy", &IAMPolicy{
+			PolicyName: NewString("foo"),
+			Roles:      MakeSlice(roleRef),
+		})
+
+		jsPolicyRef, err := policyRef.MarshalJSON()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(jsPolicyRef).To(MatchJSON(`{"Ref": "aPolicy"}`))
+
+		js, err := t.RenderJSON()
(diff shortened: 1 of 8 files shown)


errordeveloper:
To be clear, most of the code was copied from the fork.

pkg/cfn/template/iam.go
```diff
@@ -0,0 +1,39 @@
+package template
+
+// IAMPolicy represents a CloudFormation AWS::IAM::Policy resource
+type IAMPolicy struct {
+	PolicyName *Value `json:",omitempty"`
+
+	Roles          []*Value        `json:",omitempty"`
```
errordeveloper: The idea is that we only define the fields we care about and use types according to our usage, i.e. this one is always a reference, so it's of generic `Value` type.

pkg/cfn/template/iam.go
```diff
@@ -0,0 +1,39 @@
...
+	PolicyDocument MapOfInterfaces `json:",omitempty"`
```
errordeveloper: We don't use references for policy documents, so we declare it as a map of interfaces, as that's how we always use it.

pkg/cfn/template/api_test.go
```diff
@@ -0,0 +1,72 @@
+package template
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("CloudFormation template", func() {
```
errordeveloper: These are basic tests for now, but in the future it would be nice to provide a high-level testing capability through custom gomega matchers.

pkg/cfn/template/iam.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Type will return the full type name for the resource
+func (r *IAMPolicy) Type() string {
```
cPu1: Minor nit: I think `p` would be a better name for the receiver.

errordeveloper: We can refine these later on, for now it's `r` for 'Resource`...

PR #1136, opened 2019-08-12
Add the Helm Operator and Tiller to `eksctl install flux`

### LINKED ISSUE #1076: Move eksctl install flux's logic to pkg/addons/flux
I think we should adopt the pattern that is used in `eksctl utils update-*`, much of this is very similar except that manifests are obtained from a library call, and we have a way for testing things if we use `pkg/kubernetes` the same way we do it in `pkg/addons/default`. Interaction Git and Flux can be tested as part of integration suite.

I think we should move business logic into [...] `pkg/addons/flux` for getting Flux resources created or updated in Kubernetes (similar to `pkg/addons/default`)
[...]

_Originally posted by @errordeveloper in https://github.com/weaveworks/eksctl/pull/1064#issuecomment-515935316_

2opremio:
### Description

Adds the Helm Operator and Tiller to `eksctl install flux`. Draft for now, since there still seems to be a TLS problem I need to fix.

Also, fixes #1076 

### Checklist
- [x] Manually tested
- [ ] Push new version of the dependencies image (0.10)


pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
+package install
```
martina-if: This could go into its own package outside of `pkg/ctl`

pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
...
+		if err = pki.loadFrom(pkiPaths); err != nil {
+			return nil, nil, err
+		}
+		return pki, pkiPaths, nil
+	}
+
+	logger.Info("Generating public key infrastructure for the Helm Operator and Tiller")
+	logger.Info("  this may take up to a minute, please be patient")
+	fi.opts.tillerHost = tillerServiceName + "." + fi.opts.tillerNamespace
+	pki, err := newPKI(fi.opts.tillerHost, 5*365*24*time.Hour, 4096)
```
martina-if: If we create this here instead of using dependency injection with an interface it will be hard to mock, or how will flux installer be tested?

martina-if:
I think site/content/usage/experimental/01-gitops.md should also be updated mentioning helm and tiller

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+type publicKeyInfrastructure struct {
+	caCertificate     []byte
+	caKey             []byte
+	serverCertificate []byte
+	serverKey         []byte
+	clientCertificate []byte
+	clientKey         []byte
+}
+
+type devnull struct{}
```
errordeveloper:
```suggestion
type nullLogger struct{}
```

pkg/ctl/install/fluxinstaller_test.go
```diff
@@ -0,0 +1,100 @@
+package install
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
```
errordeveloper: All other packages use ginko/gomega, did you have a particular reason for not using it here?

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+// Command will create the `install` commands
+func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
```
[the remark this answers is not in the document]
errordeveloper: Yes.

pkg/ctl/install/fluxinstaller.go
errordeveloper: Per earlier discussion I've opened #1076.

pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
...
+)
+
+const (
+	fluxNamespaceFileName    = "flux-namespace.yaml"
+	tillerManifestPrefix     = "tiller-"
+	tillerNamespaceFileName  = tillerManifestPrefix + "namespace.yaml"
+	tillerServiceName        = "tiller-deploy" // do not change at will, hardcoded in Tiller's manifest generation
+	tillerServiceAccountName = "tiller"
+	tillerRBACTemplate       = `apiVersion: v1
+kind: ServiceAccount
```
errordeveloper: At very least please put into a separate file and use bindata. The convention is to have such files in `assets` subdirectory of a package.

pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to read Flux manifest file %s: %s", manifestFile.Name(), err)
+		}
+		manifests[manifestFile.Name()] = manifest
+	}
+	return manifests, nil
+}
+
+func runShell(workDir string) error {
+	shell, err := loginshell.Shell()
```
errordeveloper: Why not use an ordinary exec?

pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
...
+		}
+	}()
+
+	logger.Info("Writing Flux manifests")
+	fluxManifestDir := filepath.Join(cloneDir, fi.opts.gitFluxPath)
+	if err := writeFluxManifests(fluxManifestDir, manifests); err != nil {
+		return err
+	}
+
+	if fi.opts.amend {
```
errordeveloper: What is this for?

pkg/ctl/install/fluxinstaller.go
```diff
@@ -0,0 +1,558 @@
...
+			return err
+		}
+		// Re-read the manifests, as they may have changed:
+		manifests, err = readFluxManifests(fluxManifestDir)
+		if err != nil {
+			return err
+		}
+	}
+
+	logger.Info("Applying manifests")
```
errordeveloper: The convention is for log message begin with lower case letter, unless it's a name.

pkg/ctl/install/fluxinstaller.go
errordeveloper: Please use a constant for the certificate TTL.

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+type publicKeyInfrastructurePaths struct {
+	caCertificate     string
+	caKey             string
+	serverCertificate string
+	serverKey         string
+	clientCertificate string
+	clientKey         string
+}
+
+type publicKeyInfrastructure struct {
```
errordeveloper: These structs seem a little too verbose. Why not have just a map or a single struct?

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (devnull) Debug(string)   {}
+func (devnull) Info(string)    {}
+func (devnull) Warning(string) {}
+func (devnull) Err(string)     {}
+func (devnull) Crit(string)    {}
+func (devnull) Emerg(string)   {}
+func init() {
+	// Disable logging of cfssl since we handle errors explicitly
+	log.SetLogger(devnull{})
```
errordeveloper: Does it have to be an init function?

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+	hostName string, commonName string, validFor time.Duration) ([]byte, []byte, error) {
+	policy := &config.Signing{
+		Default: &config.SigningProfile{
+			Expiry: validFor,
+			Usage:  []string{"signing", "key encipherment", "server auth", "client auth"},
+		},
+	}
+	parsedCa, err := helpers.ParseCertificatePEM(caCert)
+	if err != nil {
+		return nil, nil, fmt.Errorf("malformed generated private CA certificate: %s", err)
```
errordeveloper: Please use `errors.Wrapf`.

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+	certSigner, err := local.NewSigner(priv, parsedCa, signer.DefaultSigAlgo(priv), policy)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot create signer: %s", err)
+	}
+	serverReq := &csr.CertificateRequest{
+		KeyRequest: keyReq,
+		CN:         commonName,
+		Hosts:      []string{hostName},
+	}
+	g := &csr.Generator{Validator: genkey.Validator}
```
errordeveloper: From the docs, my first guess would be  that `csr.Generate` is what you wan to call?

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
...
+	csrBytes, key, err := g.ProcessRequest(serverReq)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot process server-certificate signing request: %s", err)
```
errordeveloper:
```suggestion
		return nil, nil, fmt.Errorf("cannot process server certificate signing request: %s", err)
```

errordeveloper: Also, are you sure these errors need to be wrapped? Maybe it'd be okay to just return whatever the library returns, and wrap the result to say that it's to do with Tiller certificate generation?

pkg/ctl/install/pki.go
errordeveloper: I does seem all a little verbose, are you sure there's in a higher-level option?

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Git branch to be used by Flux")
+		fs.StringSliceVar(&opts.gitPaths, "git-paths", []string{},
+			"Relative paths within the Git repo for Flux to locate Kubernetes manifests")
+		fs.StringVar(&opts.gitLabel, "git-label", "flux",
+			"Git label to keep track of Flux's sync progress; overrides both --git-sync-tag and --git-notes-ref")
+		fs.StringVar(&opts.gitUser, "git-user", "Flux",
+			"Username to use as Git committer")
+		fs.StringVar(&opts.gitEmail, "git-email", "",
+			"Email to use as Git committer")
+		fs.StringVar(&opts.gitFluxPath, "git-flux-subdir", "flux/",
```
errordeveloper: Doesn't seem like an essential flag to expose.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Directory within the Git repository where to commit the Flux manifests")
+		fs.StringVar(&opts.namespace, "namespace", "flux",
+			"Cluster namespace where to install Flux and the Helm Operator")
+		fs.StringVar(&opts.tillerNamespace, "tiller-namespace", "kube-system",
```
errordeveloper: Is there a requirement for this? I'd put it into same namespace where flux is, avoid having a flag, and restrict it to use just that namespace.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Cluster namespace where to install Tiller")
+		fs.BoolVar(&opts.noHelmOp, "no-helmop", false,
```
errordeveloper: We don't use negative flags anywhere else. If you must add one here, call it `--without-flux`, but I'd much rather have these as `--flux` and `--helm-operator`.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Do not install the Flux Helm Operator (implies --no-tiller)")
+		fs.BoolVar(&opts.noTiller, "no-tiller", false,
```
errordeveloper: Let's avoid separate flag, it's either helm and the operator or neither.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Do not install Tiller. If the Helm Operator is installed, make sure to appropriately set --tiller-host and --helmop-tls-* options")
+		fs.StringVar(&opts.tillerHost, "tiller-host", "",
```
errordeveloper: Also, don't see a good reason for exposing this flag.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+			"Only applicable when --no-tiller is set. Hostname to use in order to connect to Tiller")
+		fs.StringVar(&opts.helmOpTLSKeyFile, "helmop-tls-key-path", "",
+			"Only applicable when --no-tiller is set. Path to TLS client key to connect to Tiller. Empty means no TLS.")
```
errordeveloper: Let's not make such long helm messages, it's quite obvious that `--helm=false` will imply that any other helm-related flags mean nothing.

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+		fs.StringVar(&opts.helmOpTLSCACertFile, "helmop-tls-ca-cert-path", "",
+			"Only applicable when --no-tiller is set. Path to CA certificate to verify Tiller's certificate. Empty means no TLS.")
+		fs.BoolVar(&opts.amend, "amend", false,
+			"Stop to manually tweak the Flux manifests before pushing them to the Git repository")
+	})
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		cmdutils.AddNameFlag(fs, cmd.ClusterConfig.Metadata)
+		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
+		cmdutils.AddTimeoutFlagWithValue(fs, &opts.timeout, 20*time.Second)
```
errordeveloper: Please use `cmd.ProviderConfig.WaitTimeout` everywhere, there is not need for an extra variable.

errordeveloper:
By the way, I do recall that we've agreed to use cert-manager, didn't we?

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
2opremio: @errordeveloper why? also, that train passed in #1064

pkg/ctl/install/fluxinstaller_test.go
2opremio: No reason. I will look into ginko gomega

pkg/ctl/install/fluxinstaller.go
2opremio: It's overkill for a single file which nicely in the code.

pkg/ctl/install/fluxinstaller.go
2opremio: To stop for amending. Please read the parameter description.

pkg/ctl/install/fluxinstaller.go
2opremio: Unrelated to this PR, this was done in #1064

pkg/ctl/install/fluxinstaller.go
2opremio: sure

pkg/ctl/install/pki.go
2opremio: Because of their use. They can be used separately. Please suggest a specific change if you think it can be done differently.

pkg/ctl/install/pki.go
2opremio: Where would you put it?

pkg/ctl/install/pki.go
2opremio: I am basing this code on what the command line tool does. If you think there is a better way, please suggest it.

pkg/ctl/install/cmd.go
2opremio: Unrelated to this PR (this was done in #1064 )

pkg/ctl/install/cmd.go
2opremio: There is no requirement, but tiller is usually installed in `kube-system`.

If we put it in the same namespace then we don't need to make tiller optional, which would probably make things simpler though.

@kalbir thoughts?

pkg/ctl/install/cmd.go
2opremio: sure

pkg/ctl/install/cmd.go
2opremio: I invoke @kalbir for this one as well

pkg/ctl/install/cmd.go
2opremio: It's necessary if you don't install tiller

pkg/ctl/install/cmd.go
2opremio: I don't think it's obvious for everybody

pkg/ctl/install/cmd.go
2opremio: This is unrelated to this PR, it was done before

2opremio:
> By the way, I do recall that we've agreed to use cert-manager, didn't we?

No, we didn't

pkg/ctl/install/pki.go
2opremio: I will try with `csr.Generate`

pkg/ctl/install/pki.go
2opremio: sure

pkg/ctl/install/fluxinstaller.go
2opremio: Then we can solve it when #1076 is tackled, or should I address it now?

pkg/ctl/install/pki.go
```diff
@@ -0,0 +1,167 @@
+package install
```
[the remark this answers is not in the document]
2opremio: should we solve it as part of #1076?

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+		fs.StringVar(&opts.tillerHost, "tiller-host", "",
+			"Only applicable when --no-tiller is set. Hostname to use in order to connect to Tiller")
+		fs.StringVar(&opts.helmOpTLSKeyFile, "helmop-tls-key-path", "",
+			"Only applicable when --no-tiller is set. Path to TLS client key to connect to Tiller. Empty means no TLS.")
+		fs.StringVar(&opts.helmOpTLSCertFile, "helmop-tls-cert-path", "",
+			"Only applicable when --no-tiller is set. Path to TLS client certificate to connect to Tiller. Empty means no TLS.")
+		fs.StringVar(&opts.helmOpTLSCACertFile, "helmop-tls-ca-cert-path", "",
+			"Only applicable when --no-tiller is set. Path to CA certificate to verify Tiller's certificate. Empty means no TLS.")
+		fs.BoolVar(&opts.amend, "amend", false,
+			"Stop to manually tweak the Flux manifests before pushing them to the Git repository")
```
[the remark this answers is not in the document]
2opremio: Uhm, this as part of PR #1064 but, I don't think it means what you think. It will literally stop the process and shell-out to let you tweak the manifests. It clearly didn't convey that properly (otherwise you would had understood it). How would you phrase it?

pkg/ctl/install/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+	cmd.SetRunFuncWithNameArg(func() error {
+		installer, err := newFluxInstaller(context.Background(), cmd, &opts)
+		if err != nil {
+			return err
+		}
+		return installer.run(context.Background())
+	})
+	cmd.FlagSetGroup.InFlagSet("Flux installation", func(fs *pflag.FlagSet) {
+		fs.StringVar(&opts.gitURL, "git-url", "",
+			"URL of the Git repository to be used by Flux, e.g. git@github.com:<your username>/flux-get-started")
```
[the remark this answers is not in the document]
2opremio: sure

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
2opremio: > Yes, and it must have just one function.
Why?

pkg/ctl/install/cmd.go
kalbir: I can see reasons for doing it either way. I like the idea of putting everything we install into our "own" namespaces rather than polluting `kube-system` but I have no strong opinion on what users would expect or prefer.

pkg/ctl/install/cmd.go
kalbir: I don't understand the pro/con of the options, can someone outline what happens in each case?

pkg/ctl/install/pki.go
2opremio: I would rather add context

pkg/ctl/install/cmd.go
2opremio: My idea was to make it more flexible and less opinionated by reusing an existing Tiller installation if the user wanted to. But then again, this is probably about novice users. So, I am on the fence about this one.

pkg/ctl/install/cmd.go
2opremio: It's related to https://github.com/weaveworks/eksctl/pull/1136#discussion_r313822189 . This flag makes the installation of Tiller optional (in case Tiller was already installed in the cluster).

kalbir: I guess if we put everything into a `flux` namespace we can assume that the user doesn't have a tiller in that namespace (since we are creating it). So that should get us around the case where a user already has tiller installed in say `kube-system`?

So it seems to me sensible to put tiller in `flux` and then we have less difficulty dealing with pre-existing tillers. Also as we are encouraging the user to do everything through gitops so the I think we have less worry about them having to supply a `tiller-namespace=flux` (though we should put that in the documentation somewhere.

Does that make sense?

2opremio: > So that should get us around the case where a user already has tiller installed in say `kube-system`?

As a user I may not be satisfied to have two Tillers since I may already have machinery in place to deal with one Tiller, where I installed it. However, we could probably consider that a corner case?

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
errordeveloper: That is the current convention, do you disagree?

pkg/ctl/install/fluxinstaller.go
errordeveloper: That is understood, I am questioning the purpose. Is it being assumed that user may wish to have that options?

pkg/ctl/install/fluxinstaller.go
errordeveloper: Why is the diff look like its a new file?

pkg/ctl/install/pki.go
errordeveloper: We only use it one place, just having it in the entry point function would be fine.

pkg/ctl/install/pki.go
errordeveloper: Do you know what the errors say?

Having errors that say the same things twice or more is a extremely annoying.

pkg/ctl/install/cmd.go
errordeveloper: Sure, I am just saying that there are many non-essential flags. We should expose flags for theoretical cases where user may need an option. If user needs many options, they should just use a different way of installing these components, there are several to chose from.

pkg/ctl/install/cmd.go
errordeveloper: Ok, I think we should reverse that. There's is no point, many options of installing these components are available to the user, we only provide a convenient shortcut for most part. We must enable higher level use-cases, and ungrnore low-level knobs whenever possible until multiple users request a particular kind of a knob.

pkg/ctl/install/cmd.go
errordeveloper: I think we are talking about the thing multiple threads - how opinionated the flux-tiller installation should be. My point being that it needs to only expose essential options to begin with, any customisation can be added later (when users request it).
Also, we have a strategy to keep less flags, and anything extra only gets exposed via config file.
I do think we need to join threads and agree on something, and that doesn't have to happen before this PR gets merged. I am mostly commenting on what I am seeing, this is an experimental feature and will have to go through an overall review before we graduate it.

pkg/ctl/install/cmd.go
errordeveloper: That is how docker and kubectl flags work, thereby that is what most of our users are used to.

pkg/ctl/install/cmd.go
errordeveloper: Again, looks like this file is all new for some reason.

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
errordeveloper: More specifically, each handler lives in separate file. All business logic is separated, top-level files only handles registration.

pkg/ctl/install/fluxinstaller.go
errordeveloper: Perhaps, just looks like it is a new file and it is not that much effort to structure code from the start. Clearly something is wrong with the diff...

pkg/ctl/install/fluxinstaller.go
2opremio: unrelated to this PR, this was introduced in #1064

pkg/ctl/install/fluxinstaller.go
2opremio: because I moved code around

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
2opremio: I don't see a reason to do this in this case, since there is only a subcommand.

pkg/ctl/install/fluxinstaller.go
2opremio: What do you mean? I don't think there is anything wrong with the diff.

pkg/ctl/install/fluxinstaller.go
2opremio: Uhm, do we need to mock it? That function doesn't really need any explicit I/O

pkg/ctl/install/cmd.go
errordeveloper: Again, don't think there is a need for all the options. We either install flux alone or flux with tiller and the operator. If user has tiller already, they will need to remove it. Also, I do think we should have our own namespaces tiler instance, but that clearly needs more discussion outside of this PR.

pkg/ctl/install/pki.go
2opremio: I don't think it says the same thing twice. Find a counter-example and I will fix it.

pkg/ctl/install/cmd.go
2opremio: Right, please raise a ticket. Let's please stick to reviewing the code added in the PR.

pkg/ctl/install/cmd.go
2opremio: Alright. I will always install tiller.

pkg/ctl/install/cmd.go
2opremio: We are not optionally installing tiller anymore

pkg/ctl/install/cmd.go
2opremio: Please check the history. I moved the code to a new file. That's why it shows as new.

pkg/ctl/install/cmd.go
2opremio: we are not installing tiller optionally anymore

pkg/ctl/install/pki.go
2opremio: Resolving for now. Please unresolve it if you find a nicer design.

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
martina-if: Aaah, it could be something like "stop process to...." or "pause"?

pkg/ctl/install/fluxinstaller.go
martina-if: don't we call something about reading or writing the certs to the filesystem? if there is no IO in the PKI package then I don't see a strong need for it to be mockable

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
martina-if: I think it will help new joiners and external contributors that we follow the same convention in this case. What do you think?

Regarding the business logic, moving it to it's own package makes sense from a design and testability perspective. If we do that in another PR that is fine with me.

pkg/ctl/install/pki.go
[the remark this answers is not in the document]
martina-if: @2opremio Oh, I found another reason to move all the business logic outside of this package, and it's that we need to be able to reuse this as part of `gitops apply`. Would it be possible to separate it in this PR?

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
2opremio: Alright, I will move it to `pkg/addons/flux`as per #1076

pkg/ctl/install/cmd.go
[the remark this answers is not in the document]
2opremio: I can't think of a better way to put it :S

pkg/ctl/install/pki.go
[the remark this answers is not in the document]
2opremio: Yep, will do

pkg/ctl/install/fluxinstaller.go
2opremio: Ah, that's done when `saveTo()`, yes. Uhm, let me think about it.

pkg/ctl/install/pki.go
2opremio: I simplified it

pkg/ctl/install/cmd.go
2opremio: I went with `--with-helmop`

pkg/ctl/install/fluxinstaller.go
2opremio: I tried to use `afero` but  I couldn't, because `tillerinstall.Secret()` expects a real file.

go.mod
```diff
@@ -10,24 +10,29 @@ require (
...
 	github.com/coreos/etcd v3.3.13+incompatible // indirect
 	github.com/coreos/go-semver v0.3.0 // indirect
 	github.com/coreos/go-systemd v0.0.0-20190620071333-e64a0ec8b42a // indirect
 	github.com/dave/jennifer v1.3.0
 	github.com/dlespiau/kube-test-harness v0.0.0-20190110151726-c51c87635b61
+	github.com/docker/distribution v2.7.1+incompatible // indirect
 	github.com/docker/docker v1.13.1 // indirect
-	github.com/evanphx/json-patch v4.1.0+incompatible
+	github.com/evanphx/json-patch v4.2.0+incompatible
+	github.com/fluxcd/helm-operator v0.0.0-20190813115430-0c769f5cbb3e
```
stefanprodan: This can be `github.com/fluxcd/helm-operator v1.0.0-rc1`

pkg/gitops/flux/pki.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Generate Server Certificate
+	serverCert, serverKey, err := generateCertificate(caCert, caKey, keyReq, hostName, "server", validFor)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot generate server certificate")
+	}
+
+	// Generate Client certificate
+	generateCertificate(caCert, caKey, keyReq, hostName, "client", validFor)
```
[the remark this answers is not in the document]
2opremio: well spotted!! Thanks

pkg/ctl/install/pki.go
2opremio: My answer got lost, sorry. I mentioned that there isn't an entry point function as such.

pkg/ctl/install/fluxinstaller.go
martina-if: @2opremio perhaps we should have an interface for PKI that we can mock later then?

PR #1146, opened 2019-08-15
Readded the ability to delete a cluster before it's finished creating

### LINKED ISSUE #1069: cannot delete cluster that is still being created
```
$ ./eksctl create cluster 
[ℹ]  using region us-west-2
[ℹ]  setting availability zones to [us-west-2d us-west-2c us-west-2a]
[ℹ]  subnets for us-west-2d - public:192.168.0.0/19 private:192.168.96.0/19
[ℹ]  subnets for us-west-2c - public:192.168.32.0/19 private:192.168.128.0/19
[ℹ]  subnets for us-west-2a - public:192.168.64.0/19 private:192.168.160.0/19
[ℹ]  nodegroup "ng-535bb482" will use "ami-03a55127c613349a7" [AmazonLinux2/1.13]
[ℹ]  using Kubernetes version 1.13
[ℹ]  creating EKS cluster "attractive-unicorn-1564408795" in "us-west-2" region
[ℹ]  will create 2 separate CloudFormation stacks for cluster itself and the initial nodegroup
[ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=us-west-2 --name=attractive-unicorn-1564408795'
[ℹ]  3 sequential tasks: { create cluster control plane "attractive-unicorn-1564408795", create nodegroup "ng-535bb482", 1 task: { (skip) update CloudWatch logging configurtaion } }
[ℹ]  building cluster stack "eksctl-attractive-unicorn-1564408795-cluster"
[ℹ]  deploying stack "eksctl-attractive-unicorn-1564408795-cluster"
^C
$ ./eksctl delete cluster attractive-unicorn-1564408795
[ℹ]  using region us-west-2
[ℹ]  deleting EKS cluster "attractive-unicorn-1564408795"
[✖]  status of cluster "attractive-unicorn-1564408795" is "CREATING", has to be "ACTIVE"
$
```

This was introduced in #1010. 

Main use-case is that one must be able to interrupt `eksctl create cluster` and delete it right away, which was possible before #1010.

The error is due to that we can only construct a Kubernetes client once cluster status is `ACTIVE`.

https://github.com/weaveworks/eksctl/blob/92cbf446b6168f834098e9c8a9f50eaf0bfe06ab/pkg/ctl/delete/cluster.go#L111-L113

https://github.com/weaveworks/eksctl/blob/9fde2b994c59d523a1a5d4ec8bc9d047af9f81a9/pkg/eks/eks.go#L50-L56

D3nn:
### Description

A previous PR (#1010) made it a prerequisite that an EKS cluster must
be in an ACTIVE state before `eksctl delete cluster` would be able to
delete the cluster.  This was done to find all the ELBs that were
created due to services in the cluster.  However, if a cluster isn't
created, there can be no ELBs created for the cluster, so this
PR removes that restriction and adds some integration tests for
this use case.

Resolves #1069

## Checklist

- [x] Manually tested

### DIFF
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -42,9 +42,7 @@ func eksctl(args ...string) *gexec.Session {
 	command.Env = append(command.Env, "EKSCTL_EXPERIMENTAL=true")
 	fmt.Fprintf(GinkgoWriter, "calling %q with %s and %v\n", eksctlPath, "EKSCTL_EXPERIMENTAL=true", args)
 	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
-	if err != nil {
-		Fail(fmt.Sprintf("error starting process: %v\n", err), 1)
-	}
+	Expect(err).To(BeNil())
 
 	t := time.Minute
 	switch args[0] {
@@ -65,25 +63,23 @@ func eksctl(args ...string) *gexec.Session {
 
 func eksctlSuccess(args ...string) *gexec.Session {
 	session := eksctl(args...)
-	Expect(session.ExitCode()).To(Equal(0))
+	Expect(session.ExitCode()).To(BeZero())
 	return session
 }
 
 func eksctlFail(args ...string) *gexec.Session {
 	session := eksctl(args...)
-	Expect(session.ExitCode()).To(Not(Equal(0)))
+	Expect(session.ExitCode()).ToNot(BeZero())
 	return session
 }
 
-//eksctlStart starts running an eksctl command, waits 45 seconds, but doesn't wait for it to finish the command
-//This is primarily so that we can run eksctl create ... and then subsequently call eksctl delete on the same cluster.
-func eksctlStart(args ...string) error {
-	cmd := exec.Command(eksctlPath, args...)
+//eksctlStart starts running an eksctl command but doesn't wait for it to finish the command
+//This is primarily so that we can run eksctl create and then subsequently call eksctl delete
+//on the same cluster, but might be useful for other test scenarios as well.
+func eksctlStart(args ...string) (err error) {
 	fmt.Fprintf(GinkgoWriter, "calling %q with %v\n", eksctlPath, args)
-	err := cmd.Start()
-	if err != nil {
-		return err
-	}
-	time.Sleep(45 * time.Second)
-	return nil
+	cmd := exec.Command(eksctlPath, args...)
+	err = cmd.Start()
+	Expect(err).To(BeNil())
+	return
 }
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -76,10 +76,9 @@ func eksctlFail(args ...string) *gexec.Session {
 //eksctlStart starts running an eksctl command but doesn't wait for it to finish the command
 //This is primarily so that we can run eksctl create and then subsequently call eksctl delete
 //on the same cluster, but might be useful for other test scenarios as well.
-func eksctlStart(args ...string) (err error) {
+func eksctlStart(args ...string) {
 	fmt.Fprintf(GinkgoWriter, "calling %q with %v\n", eksctlPath, args)
 	cmd := exec.Command(eksctlPath, args...)
-	err = cmd.Start()
+	err := cmd.Start()
 	Expect(err).To(BeNil())
-	return
 }
--- a/integration/common_test.go
+++ b/integration/common_test.go
@@ -74,3 +74,16 @@ func eksctlFail(args ...string) *gexec.Session {
 	Expect(session.ExitCode()).To(Not(Equal(0)))
 	return session
 }
+
+//eksctlStart starts running an eksctl command, waits 45 seconds, but doesn't wait for it to finish the command
+//This is primarily so that we can run eksctl create ... and then subsequently call eksctl delete on the same cluster.
+func eksctlStart(args ...string) error {
+	cmd := exec.Command(eksctlPath, args...)
+	fmt.Fprintf(GinkgoWriter, "calling %q with %v\n", eksctlPath, args)
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+	time.Sleep(45 * time.Second)
+	return nil
+}
--- a/integration/createdeletebeforeactive_test.go
+++ b/integration/createdeletebeforeactive_test.go
@@ -1,0 +1,83 @@
+// +build integration
+
+package integration_test
+
+import (
+	"fmt"
+
+	awseks "github.com/aws/aws-sdk-go/service/eks"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/testutils/aws"
+	. "github.com/weaveworks/eksctl/pkg/testutils/matchers"
+)
+
+var _ = Describe("(Integration) Create & Delete before Active", func() {
+	const (
+		initNG = "ng-0"
+		testNG = "ng-1"
+	)
+
+	Describe("when creating a cluster with 1 node", func() {
+		var clName string
+		Context("for deleting a creating cluster", func() {
+			It("should run eksctl and not wait for it to finish", func() {
+
+				fmt.Fprintf(GinkgoWriter, "Using kubeconfig: %s\n", kubeconfigPath)
+
+				if clName == "" {
+					clName = cmdutils.ClusterName("", "") + "-delb4active"
+				}
+
+				eksctlStart("create", "cluster",
+					"--verbose", "4",
+					"--name", clName,
+					"--tags", "alpha.eksctl.io/description=eksctl delete before active test",
+					"--nodegroup-name", initNG,
+					"--node-labels", "ng-name="+initNG,
+					"--node-type", "t2.medium",
+					"--nodes", "1",
+					"--region", region,
+					"--version", version,
+				)
+			})
+		})
+
+		Context("when deleting the (creating) cluster", func() {
+
+			It("should not return an error", func() {
+
+				eksctlSuccess("delete", "cluster",
+					"--verbose", "4",
+					"--name", clName,
+					"--region", region,
+					"--wait",
+				)
+			})
+
+			It("and should have deleted the EKS cluster and both CloudFormation stacks", func() {
+
+				awsSession := aws.NewSession(region)
+
+				Expect(awsSession).ToNot(HaveExistingCluster(clName, awseks.ClusterStatusActive, version))
+
+				Expect(awsSession).ToNot(HaveExistingStack(fmt.Sprintf("eksctl-%s-cluster", clName)))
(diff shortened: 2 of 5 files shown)


integration/common_test.go
```diff
@@ -71,3 +71,16 @@ func eksctlFail(args ...string) *gexec.Session {
...
+//eksctlStart starts running an eksctl command, waits 45 seconds, but doesn't wait for it to finish the command
+//This is primarily so that we can run eksctl create ... and then subsequently call eksctl delete on the same cluster.
+func eksctlStart(args ...string) error {
+	cmd := exec.Command(eksctlPath, args...)
+	fmt.Fprintf(GinkgoWriter, "calling %q with %v\n", eksctlPath, args)
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+	time.Sleep(45 * time.Second)
```
rndstr: for this function to be usable by other tests I would move that Sleep to the caller

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+var _ = Describe("(Integration) Create & Delete before Active", func() {
+	const (
+		initNG = "ng-0"
+		testNG = "ng-1"
+	)
+
+	Describe("when creating a cluster with 1 node", func() {
+		var clName string
+		Context("for deleting a creating cluster", func() {
+			It("should run eksctl and not wait for it to finish", func() {
```
rndstr: i don't see any assertions, should this be in an initialization block instead? (I'm not familiar with ginkgo enough to know where to put that)

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/testutils/aws"
+	. "github.com/weaveworks/eksctl/pkg/testutils/matchers"
+)
+
+var _ = Describe("(Integration) Create & Delete before Active", func() {
+	const (
+		initNG = "ng-0"
+		testNG = "ng-1"
```
rndstr: unused?

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+	)
+
+	Describe("when creating a cluster with 1 node", func() {
+		var clName string
```
rndstr: you probably want to initalize this with `clName := clusterName` (rhs being the global config option). maybe a better variable name is `deleteClusterName := clusterName`

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+				Expect(awsSession).ToNot(HaveExistingStack(fmt.Sprintf("eksctl-%s-cluster", clName)))
+				Expect(awsSession).ToNot(HaveExistingStack(fmt.Sprintf("eksctl-%s-nodegroup-ng-%d", clName, 0)))
+			})
+		})
+
+		Context("when trying to delete the cluster again", func() {
+
+			It("should return an a non-zero exit code", func() {
+
```
rndstr: nit: lots of blank lines above

pkg/eks/eks.go
```diff
@@ -33,6 +33,7 @@ func (c *ClusterProvider) DescribeControlPlane(cl *api.ClusterMeta) (*awseks.Clu
 }
 
 // DescribeControlPlaneMustBeActive describes the cluster control plane and checks if status is active
+// If status isn't active and error will be returned unless strict is set to false.
```
rndstr:
```suggestion
// If status isn't active an error will be returned unless strict is set to false.
```

pkg/eks/eks.go
```diff
@@ -56,6 +57,7 @@ func (c *ClusterProvider) RefreshClusterConfig(spec *api.ClusterConfig) error {
 	if err != nil {
 		return err
 	}
+
```
rndstr: nit: extra blank line

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+					"--region", region,
+					"--wait",
+				)
+			})
+
+			It("and should have deleted the EKS cluster and both CloudFormation stacks", func() {
+
+				awsSession := aws.NewSession(region)
+
+				Expect(awsSession).ToNot(HaveExistingCluster(clName, awseks.ClusterStatusActive, version))
```
rndstr: this doesn't fail if I comment out the deletion above

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+				Expect(awsSession).ToNot(HaveExistingStack(fmt.Sprintf("eksctl-%s-cluster", clName)))
+				Expect(awsSession).ToNot(HaveExistingStack(fmt.Sprintf("eksctl-%s-nodegroup-ng-%d", clName, 0)))
```
rndstr: shouldn't this technically be `fmt.Sprintf("eksctl-%s-nodegroup-%s", clName, initNG)`? are there methods we can re-use to build that string that eksctl uses itself?

pkg/ctl/delete/cluster.go
```diff
@@ -108,19 +108,18 @@ func doDeleteCluster(cmd *cmdutils.Cmd) error {
 	{
 
 		logger.Info("cleaning up LoadBalancer services")
```
rndstr: i'd move that log message inside the block

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+		initNG = "ng-0"
+		testNG = "ng-1"
+	)
+
+	Describe("when creating a cluster with 1 node", func() {
+		var clName string
+		Context("for deleting a creating cluster", func() {
+			It("should run eksctl and not wait for it to finish", func() {
+
+				fmt.Fprintf(GinkgoWriter, "Using kubeconfig: %s\n", kubeconfigPath)
```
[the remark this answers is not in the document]
D3nn: I'm assuming you're referring to wrapping the eksclStart in a test case with no assertions.  If so, the main reason _I_ did it was so the test output is more explicit about what the test suite is doing.

integration/createdeletebeforeactive_test.go
D3nn: My primary purpose of updating the name is that it makes it much easier to distinguish in output (and in the AWS console) the clusters from this test vs. the clusters from creategetdelete_test.go.  It was time-consuming to navigate in the web console and figure out which clusters were being managed by which test case.

integration/createdeletebeforeactive_test.go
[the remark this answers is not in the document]
rndstr: the comment is about printing the kubeconfig

integration/createdeletebeforeactive_test.go
rndstr: i don't mean use the same name but re-using it as a prefix.

so instead of
```
cmdutils.ClusterName("", "") + "-delb4active"
```
use
```
clusterName + "-delb4active"
```
(that needs changes on various lines considering `clusterName` could be empty)

integration/createdeletebeforeactive_test.go
D3nn: There is a method in pkg/cfn/manager/nodegroup.go:
```
func (c *StackCollection) makeNodeGroupStackName(name string) string {
	return fmt.Sprintf("eksctl-%s-nodegroup-%s", c.spec.Metadata.Name, name)
}
```
As you can see, it's not an exported outside of its package.  Essentially though it is 'eksctl-<clusterName>-nodegroup-<nodegroup name>, so in this case, the issue with the test is that we're generating "eksctl-<clustername>-nodegroup-ng-ng-0".  It was never going to match an existing nodegroup!

rndstr: I don't think so, the test builds the right string

D3nn:
> Looks good :) I think roli has some good points. Once those are fixed it should be good to go. Have you seen if there are tests for the ELB case? Or have you manually tested that that part still works ok?

Thanks, @martina-if.  I opted to test the case where we needed to clean up (delete) the ELBs manually.

integration/createdeletebeforeactive_test.go
D3nn: changed the name `clName` to `delb4activeName`.

pkg/ctl/delete/cluster.go
D3nn: Done.

integration/createdeletebeforeactive_test.go
D3nn: I re-worked this.  The check for an error happens within the helper function `eksctlStart`.

integration/common_test.go
D3nn: Removed sleep altogether in favor of Eventually assertions.

integration/createdeletebeforeactive_test.go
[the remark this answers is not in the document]
D3nn: Removed.

integration/createdeletebeforeactive_test.go
D3nn: It should now that we properly wait until the cluster shows up as "creating" before subsequently deleting it.

integration/createdeletebeforeactive_test.go
```diff
@@ -0,0 +1,87 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/testutils/aws"
+	. "github.com/weaveworks/eksctl/pkg/testutils/matchers"
+)
+
+const (
+	POLLINTERVAL = 15   //seconds
```
rndstr: convention for Go constants is camel case

integration/common_test.go
```diff
@@ -65,12 +63,23 @@ func eksctl(args ...string) *gexec.Session {
...
 	session := eksctl(args...)
-	Expect(session.ExitCode()).To(Not(Equal(0)))
+	Expect(session.ExitCode()).ToNot(BeZero())
 	return session
 }
+
+//eksctlStart starts running an eksctl command but doesn't wait for it to finish the command
+//This is primarily so that we can run eksctl create and then subsequently call eksctl delete
+//on the same cluster, but might be useful for other test scenarios as well.
+func eksctlStart(args ...string) (err error) {
```
rndstr: can remove return value since we assert it to be `nil`

integration/createdeletebeforeactive_test.go
D3nn: I have verified this by commenting out the deletion code and validating that the test fails.

integration/common_test.go
D3nn: done.

integration/createdeletebeforeactive_test.go
D3nn: switched to camelCase

PR #1148, opened 2019-08-15
Add integration tests for `eksctl install flux`

### LINKED ISSUE #1078: Add integration test for eksctl install flux
**What should be changed?**

There are currently no integration tests for the experimental command `eksctl install flux`.
We should add some to get all the usual benefits of testing (ensure this feature always works, to catch regressions early, to be able to refactor with confidence, etc.)

marccarre:
### Description

Fixes #1078.

### Checklist

- [x] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~
- [x] Manually tested.


marccarre:
Passing when running any of the below:
1) Full integration tests, containerised, on a new cluster:
```console
make integration-test-container TEST_V=1 AWS_PROFILE="default-mfa"
```
2) Full integration tests, containerised, on an existing cluster:
```console
make integration-test-container TEST_V=1 AWS_PROFILE="default-mfa" INTEGRATION_TEST_ARGS="-eksctl.region=ap-northeast-1 -eksctl.cluster=autoscaler -eksctl.create=false -eksctl.delete=false"
```
3) Full integration tests, non-containerised:
```
make integration-test-dev TEST_CLUSTER=autoscaler AWS_PROFILE="default-mfa"
```
---- 
```
Cloning into '/tmp/eksctl-install-flux-test-clone-406850672'...
Switched to a new branch 'oy6vn78r'
remote: 
remote: Create a pull request for 'oy6vn78r' on GitHub by visiting:        
remote:      https://github.com/eksctl-bot/my-gitops-repo/pull/new/oy6vn78r        
remote: 
To github.com:eksctl-bot/my-gitops-repo.git
 * [new branch]      oy6vn78r -> oy6vn78r
Cloning into '/tmp/eksctl-install-flux-test-branch-073885455'...
Cloning into '/tmp/eksctl-install-flux-test-branch-481368354'...
To github.com:eksctl-bot/my-gitops-repo.git
 - [deleted]         oy6vn78r
• [SLOW TEST:40.597 seconds]
(Integration) Create, Get, Scale & Delete
/src/integration/creategetdelete_test.go:22
  when creating a cluster with 1 node
  /src/integration/creategetdelete_test.go:47
    and configuring Flux for GitOps
    /src/integration/creategetdelete_test.go:111
      should not return an error
      /src/integration/creategetdelete_test.go:112
```
Note that a new SSH key is required to run these tests.
As per the `CONTRIBUTING.md`'s instructions.

integration/common_test.go
```diff
@@ -36,16 +36,21 @@ func newKubeTest() (*harness.Test, error) {
 	return test, nil
 }
 
-func eksctl(args ...string) *gexec.Session {
-	command := exec.Command(eksctlPath, args...)
-	command.Env = os.Environ()
-	command.Env = append(command.Env, "EKSCTL_EXPERIMENTAL=true")
-	fmt.Fprintf(GinkgoWriter, "calling %q with %s and %v\n", eksctlPath, "EKSCTL_EXPERIMENTAL=true", args)
+type params struct {
```
errordeveloper: Why is this an unexported struct with exported fields?

integration/common_test.go
```diff
@@ -63,12 +68,17 @@ func eksctl(args ...string) *gexec.Session {
 
 func eksctlSuccess(args ...string) *gexec.Session {
 	session := eksctl(args...)
+	return eksctlSuccessWith(params{Args: args})
+}
+
+func eksctlSuccessWith(params params) *gexec.Session {
```
errordeveloper: I know this is what we are doing right now, but there is fairly major issue with functions like this in ginkgo....

When the assertions fail, you get a trace, but the trace leads to function definition and not the call site. We should start using matchers, and I've started working on that. I about to open a PR that changes `eksctlSuccess` to a matcher (the syntax will be something like `Expect(cmd).To(RunSuccessfully())`). Just FYI.

I'll take what you did here into account make sure it's possible to set custom environment variables.

pkg/utils/random/string.go
```diff
@@ -0,0 +1,24 @@
+package random
```
errordeveloper: We have a couple of random name generator in `pkg/ctl/cmdutils/namers.go` already.

integration/creategetdelete_test.go
```diff
@@ -109,6 +121,34 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 			})
 		})
 
+		Context("and configuring Flux for GitOps", func() {
+			It("should not return an error", func() {
+				// Use a random branch to ensure test runs don't step on each others.
+				branch := random.String(8)
```
errordeveloper: I'd just import `github.com/kubicorn/kubicorn/pkg/namer` and call `namer.RandomName()`, we use it in `pkg/ctl/cmdutils/namers.go` already, but here it'd be fine to just import and call it directly.

pkg/git/git_test.go
```diff
@@ -80,6 +86,14 @@ var _ = Describe("GitClient", func() {
 
 		Expect(fakeExecutor.Calls[1].Arguments[0]).To(Equal("git"))
 		Expect(fakeExecutor.Calls[1].Arguments[2]).To(
+			Equal([]string{"config", "user.email", "test-user@example.com"}))
+
+		Expect(fakeExecutor.Calls[2].Arguments[0]).To(Equal("git"))
+		Expect(fakeExecutor.Calls[2].Arguments[2]).To(
+			Equal([]string{"config", "user.name", "test-user"}))
+
+		Expect(fakeExecutor.Calls[3].Arguments[0]).To(Equal("git"))
```
errordeveloper: I was looking at gomega docs a little more lately, and spotted this: https://godoc.org/github.com/onsi/gomega#ConsistOf

```Go
Expect(fakeExecutor.Calls[2].Arguments).To(ConsistOf("git", "<what's here?>", "commit", "-m", "test commit", "--author=test-user <test-user@example.com>"))
```

Perhaps also would be nice to make this table-driven...

pkg/git/git_test.go
```diff
@@ -80,6 +86,14 @@ var _ = Describe("GitClient", func() {
...
+		Expect(fakeExecutor.Calls[3].Arguments[2]).To(
```
errordeveloper: I was looking at gomega docs a little more lately, and spotted this: https://godoc.org/github.com/onsi/gomega#ConsistOf

```Go
Expect(fakeExecutor.Calls[3].Arguments[2]).To(ConsistOf("commit", "-m", "test commit", "--author=test-user <test-user@example.com>"))
```

Perhaps also would be nice to make this table-driven...

pkg/git/executor/executor.go
```diff
@@ -14,15 +15,17 @@ type Executor interface {
 
 // ShellExecutor an executor that shells out to run commands
 type ShellExecutor struct {
-	parentCtx context.Context
-	timeout   time.Duration
+	parentCtx         context.Context
+	timeout           time.Duration
+	privateSSHKeyPath string
```
martina-if: I think this executor, even though today it is only used for git, it could be reused for command (like for kubectl or anything else we need to shell out to in the future). So perhaps this field instead of being specific for the ssh key it could be something like environment variables. What do you think?

pkg/utils/random/string.go
```diff
@@ -0,0 +1,24 @@
...
+
+const charBytes = "abcdefghijklmnopqrstuvwxyz0123456789"
+
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
+}
+
+// String returns a random alphanumerical string of length n.
+func String(n int) string {
+	if n < 0 {
```
martina-if: `if n <= 0` ?

integration/common_test.go
marccarre: No specific reason, what would you suggest instead?

integration/common_test.go
marccarre: Good to know, and yes, this (among other things) makes it pretty difficult to debug integration tests at the moment.

integration/creategetdelete_test.go
marccarre: Thanks. Done.

pkg/utils/random/string.go
marccarre: Thanks. Resolved this in latest version of this PR.

integration/flux_utils_test.go
```diff
@@ -0,0 +1,539 @@
...
+	return cloneDir, nil
+}
+
+func deleteBranch(branch, cloneDir string) error {
+	defer os.RemoveAll(cloneDir)
+	return gitWith(gitParams{Args: []string{"push", "origin", "--delete", branch}, Dir: cloneDir, Env: gitSSHCommand()})
+}
+
+func getBranch(branch string) (string, error) {
+	cloneDir, err := ioutil.TempDir(os.TempDir(), "eksctl-install-flux-test-branch-")
```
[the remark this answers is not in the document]
marccarre: I haven't noticed any slowness at this level, no.

pkg/utils/random/string.go
marccarre: Sure, why not.
I was test-driving this with, among others:
```go
	assert.Equal(t, "", String(-1))
	assert.Equal(t, "", String(0))
```
and allocation was failing only for negative values iirc, hence I didn't go for `n <= 0`.
But this whole utility function is gone anyway, now, so we don't even have that problem anymore 🙂

pkg/git/executor/executor.go
marccarre: I'd rather stay specific for now, but good to keep this in mind if it grows, yep 👍

pkg/git/git_test.go
martina-if: I think `ConsistsOf()` is nice when you need to check two arrays contain the same elements in any order. In this case I would say we also care about the order so I would keep the `Equal()`.

pkg/git/executor/executor.go
martina-if: Let's fix it in another PR. There are also some git parameters grouping to be done

marccarre: Sure, I'll open a follow-up PR _soonishly_ then 👍

pkg/git/git_test.go
marccarre: +1

marccarre: +1

pkg/git/executor/executor.go
marccarre: Follow-up PR: https://github.com/weaveworks/eksctl/pull/1168

marccarre: Follow-up PR: https://github.com/weaveworks/eksctl/pull/1168

Issue #892, opened 2019-06-17
shared VPC


errordeveloper:
It would be useful to provide a way to create eksctl-managed VPC that can be used for one or more clusters. This would help users who want to have multiple clusters in one VPC, but also cases where different level of access is required for managing VPC resources.

Something like this was also proposed by @whereisaaron in https://github.com/weaveworks/eksctl/issues/560#issuecomment-465807918.

errordeveloper:
It's plausible to consider having VPC in a separate stack by default, per https://github.com/weaveworks/eksctl/issues/1123, but also because user doesn't know when they create cluster whether they may need to share the VPC with another cluster.

errordeveloper:
It's plausible to consider having VPC in a separate stack by default.

Per @sstkwayne's comments in https://github.com/weaveworks/eksctl/issues/1123:

> When using eksctl to create a cluster with a new VPC, it would be great to have the VPC and all it's resources created as a separate Cloudformation stack from the EKS cluster. I recently had to tear down a cluster and because it is in the same stack with the VPC i had to tear that down as well. This is inconvenient because there wasn't anything wrong with the VPC configuration,
>
> **Why do you want this feature?**
> Having a separate VPC stack would allow users to create and destroy EKS clusters without having to teardown the VPC which might have other resources in it such as a DB, Peering connections, etc.
>
> **What feature/behavior/change do you want?**
> I would like to see eksctl create cluster command result in the creation of a VPC stack, a cluster stack and a nodegroup stack if defined in the yaml. 

Also because user simply doesn't know when they create cluster whether they may need to share the VPC with other clusters or not.

errordeveloper:
Another important uses-case is RDS, one may wish to keep VPC as it has RDS beyond lifecycle of a cluster, e.g. they wish to re-create the cluster for any reason.

oemdaro:
I have an issue when using [VPC Sharing](https://aws.amazon.com/blogs/networking-and-content-delivery/vpc-sharing-a-new-approach-to-multiple-accounts-and-vpc-management/). When I try to create cluster on the sharing VPC and subnets, I got the following error.

```
Error: couldn't import subnet subnet-2222222222: subnet ID "subnet-1111111111" is not the same as "subnet-2222222222"
```

How to solve this problem?

michaelbeaumont:
Hi, thanks for bringing this up! Can you please include the exact config or the command and arguments you're using as well as the log output? Please feel free to open a new issue as well.

[...]

Issue #764, opened 2019-04-30
Create S3 endpoints for VPC


jahpola-futurice:
Hi,

It would a nice addition that network creates at least VPC S3 endpoints optionally.

errordeveloper:
hi @jahpola-futurice! I was wondering if we might need to create endpoints for APIs the that are being used by eksctl. S3 is actually not being used by eksctl itself yet, so I'm not sure if it'd be fair to add support for it before other APIs that one might use (DynamoDB, RDS, Lambda ...etc).
Wouldn't you want to add some S3 IAM role to your nodegroup(s) as well?

jahpola-futurice:
Yes, we have either an instance profile or kube2iam setup that allows access to S3 buckets based on naming policy. The s3 bucket contains items which the container stores&reads items .

errordeveloper:
I suppose one option to consider would be something like:

```yaml
vpc:
   withServiceEndpoints: [s3, dynamodb]
```

But I am not sure what to do about `PolicyDocument` here vs a nodegroup-level.

Another option could be this:

```yaml
vpc:
   endpoints:
     dynamodb:
     - "Sid": "Statement"
        "Action": "dynamodb:*"
        "Effect": "Allow"
        "Principal": "*"
```

But that means we will also need to allow inline policy documents in other places. A few things to think about here!

hryamzik:
Looks like creating an s3 endpoint with eksctl is still not possible.

michaelbeaumont:
With the fully-private cluster feature we now have the code and the config to enable VPC endpoints, it's just a matter of not depending on `fullyPrivate.enabled == true`

PR #1149, opened 2019-08-15
Enable management of access to public/private API endpoints

### LINKED ISSUE #649: Support Private/Public Endpoints

D3nn:
### Description

Enable the ability to allow/deny access to the API server endpoint publicly (from the Internet) or privately (from Worker nodes, Direct Connect networks, VPN networks, etc).

Resolve #649 

### Checklist

- [x] Manually tested


D3nn:
It's a WIP, so I just wanted to get the PR started and add commits as I
go.  I'll let you know when it's ready to review.

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,29 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+the VPC networks (but not Amazon's network) and then re-enter to reach the API server. Allowing
+direct access from
+within the VPC subnets is possible but not without some caveats.
+See [Amazon EKS Cluster
+Endpoint Access Control](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html) for
+further information on additional steps needed when public=false, private=true.
+
+The Kubernetes API server endpoint access for a cluster can be cofigured for public and private
+access by specifiying
+the `--public-endpoint-access` and/or `--private-endpoint-access` CLI flags to be `true` (access
```
rndstr: my first thought (without looking up how we name other flags/clusterconfigs) would be to do this the same as the cluster config path which would be `--vpc-endpoint-access-{public,private}`

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,29 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+enabled) or
+`false` (access disabled).  Alternatively, one can configure access in the cluster config file like
+the example below:
+
+```yaml
+vpc:
+  endpointAccess:
```
rndstr: do we need the word API in there somewhere or does `endpoint` make it unambiguous enough?

D3nn: Good question.  The AWS SDK option names for this are:

EndpointPrivateAccess *bool `locationName:"endpointPrivateAccess" type:"boolean"`
and 
EndpointPublicAccess *bool `locationName:"endpointPublicAccess" type:"boolean"`

So thought it might be less confusing for those already familiar with this terminology not to add API in the keys.

site/content/usage/06-vpc-networking.md
D3nn: > documentation is easy to understand 👍 I would want to know what the default is, though.
> 
> Edit: or is the default going to stay as is (Public=true, Private=false)?

That's a good question.  The submitter of #649 argues for making the default in eksctl to public = true, private = true, which seems like a rational choice.  I wonder though if those using eksctl today might be caught unaware by the default behavior (where eksctl uses the default EKS settings for this) changing to this new default.

D3nn: > my first thought (without looking up how we name other flags/clusterconfigs) would be to do this the same as the cluster config path which would be `--vpc-endpoint-access-{public,private}`

I've made this change as it appears to be consistent with other CLI flags.

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,32 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+must first exit
+the VPC networks (but not Amazon's network) and then re-enter to reach the API server. Allowing direct access from
+within the VPC subnets is possible but not without some caveats. See [Amazon EKS Cluster
+Endpoint Access Control](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html) for
+further information
+on additional steps needed when public=false, private=true.
+
+The Kubernetes API server endpoint access for a cluster can be cofigured for public and private
+access by
+specifiying the `--vpc-endpoint-access-public` and/or`--vpc-endpoint-access-private` CLI flags to be
```
martina-if: Have you checked with @errordeveloper about adding these new flags? I think we are trying to minimize the amount of CLI flags we have so eksctl doesn't become too complicated.

D3nn: Didn't realize, I'll check with him.

errordeveloper: Yes, we are no adding new flags, especially in `eksctl create cluster`. We should only expose config fields for this, and sever flag-only users with whatever we pick as default.

site/content/usage/06-vpc-networking.md
errordeveloper: I think we actually need to make it more specific. We may, in the future add support for endpoints in general, e.g. #764, so I'd reserve any generic terms referring to endpoints for now and make this one more specific. I think best name for it is `clusterEndpoints`, so `vpc.clusterEndpoints` will match what we have for logging - `cloudWatch.clusterLogging`.

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,32 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+or
+`false` (access disabled).
+
+Alternatively, one can configure access in the cluster config file like
+the example below:
+
+```yaml
+vpc:
+  endpointAccess:
+    public:  <true|false>
```
errordeveloper: One alternative I can think of is to make it a string field, e.g. `vpc.clusterEndpoints: PrivateOnly`, `vpc.clusterEndpoints: Default`, `vpc.clusterEndpoints: PublicOnly`... but let's not rush this decision just yet.

errordeveloper:
I think we will need document considerations around using private-only.

The main challenge is that aside from AWS API calls we do need to make calls to Kubernetes API, e.g. to update `kube-system:ConfigMap/aws-auth`. It's possible to design it in a way that lets you create a cluster from any machine and disable public right at the very end. The reason we'd have to do it that way is that we create the VPC, we could explore shared VPC more (#892), but in any case there are a few things to think about.

errordeveloper:
I opened #1230.

D3nn:
> I think we will need document considerations around using private-only. E.g. it should be doable if you are managing the VPC externally.
> 
> The main challenge is that aside from AWS API calls we do need to make calls to Kubernetes API, e.g. to update `kube-system:ConfigMap/aws-auth`. It's possible to design it in a way that lets you create a cluster from any machine and disable public right at the very end. The reason we'd have to do it that way is that we create the VPC, we could explore shared VPC more (#892), but in any case there are a few things to think about.

I'll remove that as an option until we can address the kubernetes API calls.

site/content/usage/06-vpc-networking.md
D3nn: I think, in this scheme, we'd also need to have vpc.clusterEndpoints: All (or PublicPrivate).  And we'd defer PrivateOnly until we resolve how to deal with the times when we need kubernetes API access.

site/content/usage/06-vpc-networking.md
D3nn: I'll remove the flags.

site/content/usage/06-vpc-networking.md
errordeveloper: Let's stick to boolean, but change the parent key to `clusterEndpoints`.

D3nn:
> I think we will need document considerations around using private-only. E.g. it should be doable if you are managing the VPC externally.
> 
> The main challenge is that aside from AWS API calls we do need to make calls to Kubernetes API, e.g. to update `kube-system:ConfigMap/aws-auth`. It's possible to design it in a way that lets you create a cluster from any machine and disable public right at the very end. The reason we'd have to do it that way is that we create the VPC, we could explore shared VPC more (#892), but in any case there are a few things to think about.

I'll remove that as an option until we can address the kubernetes API calls.

> I think we will need document considerations around using private-only. E.g. it should be doable if you are managing the VPC externally.
> 
> The main challenge is that aside from AWS API calls we do need to make calls to Kubernetes API, e.g. to update `kube-system:ConfigMap/aws-auth`. It's possible to design it in a way that lets you create a cluster from any machine and disable public right at the very end. The reason we'd have to do it that way is that we create the VPC, we could explore shared VPC more (#892), but in any case there are a few things to think about.

As it stands currently, the vpc endpoint updates happen as a separate task after the other tasks have completed since it is an AWS API call and not a CloudFormation configuration item.  That means that a cluster creation wouldn't be affected by this issue but it would then subsequently block access to the "writekubeconfig" functionality (as well as any other calls that rely on K8s API access to function properly).

site/content/usage/06-vpc-networking.md
D3nn: Maybe just a slight tweak:
```
clusterEndpointsAccess:
    Public: true
    Private: false

Comment (D3nn) [reply site/content/usage/06-vpc-networking.md]: I'm going to start with the first option (clusterEndpointsAccess), but it should be an easy change to make if we want to go with something else.

Comment (D3nn) [reply site/content/usage/06-vpc-networking.md]: Actually, I'm now thinking of a slightly different naming scheme, to correspond with the EKS API names:
```
clusterEndpoints:
     PublicAccess: <bool>
     PrivateAccess: <bool>
```
Where the API names are:
1. EndpointPublicAccess
1. EndpointPrivateAccess

Again, not too difficult to rename if necessary.

**Comment by InAnimaTe** [comment]
Just wanted to stop by and say awesome work @D3nn and others involved in making this happen! Been wanting this feature for a while and glad to see some awesome people are pushing it forward.

File: pkg/apis/eksctl.io/v1alpha5/vpc.go [hunk 5]
```diff
@@ -58,6 +66,14 @@ const (
 	SubnetTopologyPrivate SubnetTopology = "Private"
 	// SubnetTopologyPublic represents publicly-routed subnets
 	SubnetTopologyPublic SubnetTopology = "Public"
+
+)
+
+var (
+	// True holds true value and can have it's address taken
```
rndstr:
```suggestion
	// True holds true value and can have its address taken
```

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -58,6 +66,14 @@ const (
...
+	True = true
+	// False holds false value and can have it's address taken
```
rndstr:
```suggestion
	// False holds false value and can have its address taken
```

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -186,3 +202,18 @@ func (c *ClusterConfig) HasSufficientSubnets() error {
...
+
+//HasClusterEndpointAccess determines if endpoint access was configured in config file or not
+func (c *ClusterConfig) HasClusterEndpointAccess() bool {
+	hasAccess := false
+	if !(c.VPC == nil || c.VPC.ClusterEndpoints == nil) {
+		hasPublicAccess := &c.VPC.ClusterEndpoints.PublicAccess != nil && *c.VPC.ClusterEndpoints.PublicAccess
+		hasPrivateAccess := &c.VPC.ClusterEndpoints.PrivateAccess != nil && *c.VPC.ClusterEndpoints.PrivateAccess
+
+		hasAccess = hasPublicAccess || hasPrivateAccess
+	} else {
```
rndstr: nit: could remove this else, put a return at end of `then` block and `return true` at end of function

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
```diff
@@ -0,0 +1,84 @@
+package v1alpha5
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+type EndpointAccessCases struct {
```
rndstr: nit: no need to export

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+	describeAccessToUpdate := []string{"no updates to make"}
+
+	needsUpdate := false
+	if newPrivate != *curPrivate || newPublic != *curPublic {
+		needsUpdate = true
+	}
+
+	if needsUpdate {
+		cfg.VPC.ClusterEndpoints.PrivateAccess = &newPrivate
+		describeAccessToUpdate[0] = fmt.Sprintf("private access:%v", newPrivate)
```
rndstr: what about `privateAccess=%v` which reflects the config format

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if newPrivate == false && newPublic == false {
+		logger.Critical(api.NoAccessMsg(
+			&api.ClusterEndpoints{PublicAccess: &newPublic, PrivateAccess: &newPrivate}))
+		os.Exit(2)
+	} else if newPrivate == true && newPublic == false {
+		logger.Warning(api.PrivateOnlyAwsChangesNeededMsg())
+	}
+
+	describeAccessToUpdate := []string{"no updates to make"}
```
rndstr: i don't think this is even printed. either it's overwritten or then ignored.

pkg/ctl/utils/update_cluster_endpoint_access.go
rndstr: this doesn't need to be a slice that is joined below. can just do
```
describeAccessUpdate := fmt.Sprintf("privateAccess=%v, publicAccess=%v", newPrivate, newPublic)
```
?

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+		cmdutils.LogIntendedAction(
+			cmd.Plan, "update Cluster API Endpoint Access for cluster %q in %q to: (%s)",
+			meta.Name, meta.Region, describeAccessUpdate)
+		if !cmd.Plan {
+			if err := ctl.UpdateClusterConfigForEndpoints(cfg); err != nil {
+				return err
+			}
+			cmdutils.LogCompletedAction(
+				false,
+				"Cluster API Endpoint Access for cluster %q in %q has been upddated to: {%s}",
```
rndstr: inconsistency: `(%s)` above and `{%s}` here

pkg/eks/tasks.go
```diff
@@ -15,6 +18,37 @@ type clusterConfigTask struct {
...
+	if err != nil {
+		return false
+	}
+	return string(ajson) == string(bjson)
+}
+
+//DefaultEndpointsMsg returns a message that the EndpointAccess is the same as the default
+func DefaultEndpointsMsg(cfg *api.ClusterConfig) string {
+	return fmt.Sprintf(
+		"Cluster API server endpoint access will use default of {Pulic: true, Private false} for cluster %q in %q", cfg.Metadata.Name, cfg.Metadata.Region)
```
rndstr:
```suggestion
		"Cluster API server endpoint access will use default of {Pulic: true, Private: false} for cluster %q in %q", cfg.Metadata.Name, cfg.Metadata.Region)
```

pkg/eks/tasks.go
```diff
@@ -15,6 +18,37 @@ type clusterConfigTask struct {
...
+}
+
+//CustomEndpointsMsg returns a message indicating the EndpointAccess given by the user
+func CustomEndpointsMsg(cfg *api.ClusterConfig) string {
+	return fmt.Sprintf(
+		"Cluster API server endpoint access will use provided values {Public: %v, Private %v} for cluster %q in %q", *cfg.VPC.ClusterEndpoints.PublicAccess, *cfg.VPC.ClusterEndpoints.PrivateAccess, cfg.Metadata.Name, cfg.Metadata.Region)
```
rndstr:
```suggestion
		"Cluster API server endpoint access will use provided values {Public: %v, Private: %v} for cluster %q in %q", *cfg.VPC.ClusterEndpoints.PublicAccess, *cfg.VPC.ClusterEndpoints.PrivateAccess, cfg.Metadata.Name, cfg.Metadata.Region)
```

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+func updateEndpointAccessCmd(cmd *cmdutils.Cmd) {
+	cfg := api.NewClusterConfig()
+	cmd.ClusterConfig = cfg
+
+	cmd.SetDescription("update-cluster-api-access", "Update cluster API endpoint configuration", "")
+
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		cmdutils.AddNameFlag(fs, cfg.Metadata)
```
rndstr: this should be `--cluster` here (which will be fixed globally in #1329)

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+1. Updating a a cluster to have private only cluster api endpoint access means that kubernetes commands
+   (e.g. `kubectl`) as well as `eksctl delete cluster`, `eksctl utils write-kubeconfig`, and possibly the command
+   `eksctl utils update-kube-proxy` must be run within the cluster VPC.  This requires some changes to various AWS
+   resources.  See:
+   https://docs.aws.amazon.com/en_pv/eks/latest/userguide/cluster-endpoint#private-access
+   
+The following is an example of how one could configure the cluster api endpoint access using the `utils` subcommand:
+
+```
+eksctl utils update-cluster-api-access--private-access=true --public-access=false --approve
```
rndstr: missing whitespace and missing `--name` (or `--cluster`)

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
 
 **Note**: Specifying the NAT Gateway is only supported during cluster creation and it is not touched during a cluster
 upgrade. There are plans to support changing between different modes on cluster update in the future.
+
+### Managing Access to the Kubernetes API Server Endpoints
+
+The default creation of an EKS cluster exposes the Kubernetes API server publicly but not directly from within the
+VPC subnets (Public=true, Private=false). Traffic destined for the API server from within the VPC must first exit the
```
rndstr: there is an inconsistency with capitalized and lowercase public/Public. unless they refer to an actual setting that is capitalized (and then it should probably be in `backticks`) i think lowercase makes more sense

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+the cluster using the cluster config file. Example below:
+
+```yaml
+vpc:
+  clusterEndpoints:
+    publicAccess:  <true|false>
+    privateAccess: <true|false>
+```
+
+There are some additional caveats when configuring cluster api endpoint access:
```
rndstr: should this be 'Cluster API' (capitalization)? (same for other mentions below)

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+
+1. EKS doesn't allow one to create or update a cluster without at least one of Private or Public access being
```
rndstr: lowercase private/public?

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+   enabled.
+1. EKS does allow creating a configuration which allows only Private access to be enabled, but eksctl doesn't
+   support it during cluster creation as it prevents eksctl from being able to join the worker nodes to the cluster.
+1. To create private-only cluster api endpoint access, one must first create the cluster *with* public cluster api
+   endpoint access, and then use `/eksctl utils update-cluster-api-access` to change it after the cluster is finished
+   creating.
+1. Updating a a cluster to have private only cluster api endpoint access means that kubernetes commands
```
rndstr: capitalized 'Kubernetes'

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -58,6 +66,14 @@ const (
 	SubnetTopologyPrivate SubnetTopology = "Private"
 	// SubnetTopologyPublic represents publicly-routed subnets
 	SubnetTopologyPublic SubnetTopology = "Public"
+
+)
+
+var (
+	// True holds true value and can have it's address taken
+	True = true
```
rndstr: usages of this should be replaced by `Enabled()`

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -58,6 +66,14 @@ const (
...
+	// False holds false value and can have it's address taken
+	False = false
```
rndstr: usages of this should be replaced by `Disabled()`

integration/cluster_api_endpoints_test.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+const (
+	createCluster    = `CreateCluster`
+	updateCluster    = `UpdateCluster`
+	endpointPubTmpl  = `EndpointPublicAccess: %v`
+	endpointPrivTmpl = `EndpointPrivateAccess: %v`
+)
+
+var False = false
```
rndstr: usages of this should be replaced by `Disabled()`

integration/cluster_api_endpoints_test.go
```diff
@@ -0,0 +1,220 @@
...
+		Name    string
+		Private bool
+		Public  bool
+		Type    string // createCluster or updateCluster
+		Output  string
+		Error   error
+		Delete  bool
+	}
+
+	FDescribeTable("Can create/update Cluster Endpoint Access",
```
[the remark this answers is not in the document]
D3nn: No, it is not supposed to (still) be there.  I'll remove it.

integration/cluster_api_endpoints_test.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	return fmt.Sprintf("%v-%v", prefix, clusterName)
+}
+
+func setMetadata(cfg *api.ClusterConfig, name, region string) {
+	cfg.Metadata.Name = name
+	cfg.Metadata.Region = region
+}
+
+func printSeparator(op, clname string) {
```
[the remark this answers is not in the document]
D3nn: It doesn't work with the standard BeforeEach or JustBeforeEach functions, because they can only reside within a Describe, Context, or It block.  DescribeTable only allows the fun(x entryStruct) and the entry's themselves.  Putting it before DescribeTables (under the main Describe block opening) doesn't work because the Entries are passed in only to the DescribeTable function.  So putting it in that function is about as close as you can get to using the Ginkgo provided tools.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -196,8 +196,11 @@ func IsEnabled(v *bool) bool { return v != nil && *v }
 // IsDisabled will only return true if v is not nil and false
 func IsDisabled(v *bool) bool { return v != nil && !*v }
 
+// IsSet will only return true if v is not nil
+func IsSet(v *string) bool { return v != nil }
```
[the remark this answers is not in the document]
D3nn: Thanks for that.  It reminded me that I had other uses for it.  I'll be pushing those changes shortly.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -63,9 +64,61 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
...
+func PrivateOnlyUseUtilsMsg() string {
+	return "eksctl cannot join worker nodes to the EKS cluster when public access isn't allowed. " +
+		"use 'eksctl utils update-cluster-api-access ...' after creating cluster with default access"
+}
+
+// PrivateOnlyAwsChangesNeededMsg returns a message that warns that not having pulbic access
+// requires changes to AWS resource configuration in order to effectively use clients in the VPC
+func PrivateOnlyAwsChangesNeededMsg() string {
+	return "warning, having public access disallowed will subsequently interfere with some " +
+		"features of eksctl. This will require running subsequent eksctl (and kubernetes) " +
```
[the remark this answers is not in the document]
D3nn: I was trying to be a bit more generic here as I'm sure there are other tools outside of 'kubectl' that access the Kubernetes API.  I'll try to reword that better.

integration/cluster_api_endpoints_test.go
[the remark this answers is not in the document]
D3nn: I did end up making 'deleteCluster' another endpointAccessCase.Type value though and did a little refactoring to the test as a result.   Pushing that shortly.

integration/integration_test.go
```diff
@@ -55,7 +55,7 @@ func init() {
 	flag.StringVar(&privateSSHKeyPath, "eksctl.git.sshkeypath", defaultPrivateSSHKeyPath, fmt.Sprintf("Path to the SSH key to use for Git operations (default: %s)", defaultPrivateSSHKeyPath))
 
 	eksctlCmd = runner.NewCmd(eksctlPath).
-		WithArgs("--region", region).
+		// WithArgs("--region", region).
```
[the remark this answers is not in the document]
D3nn: Unfortunately, the --region argument is incompatible with -f \<configfile\> that I use in my integration test for this feature.

I'll need to find a way around that.

integration/cluster_api_endpoints_test.go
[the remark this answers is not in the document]
D3nn: I have updated this so that the printSeparator call happens at the beginning of each test case and passes the 'op' from the use case definition.

integration/integration_test.go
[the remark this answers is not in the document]
D3nn: I have fixed this by adding a WithoutArg function in the runner.  WithoutArg removes a specified argument if it exists.

pkg/apis/eksctl.io/v1alpha5/types.go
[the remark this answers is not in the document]
D3nn: Re-worked IsSet to take an interface and expanded it's use.

pkg/apis/eksctl.io/v1alpha5/validation.go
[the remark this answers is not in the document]
D3nn: I updated the wording on this.

site/content/usage/06-vpc-networking.md
D3nn: They are now lowercased.

pkg/apis/eksctl.io/v1alpha5/vpc.go
D3nn: Replaced all uses of False/True with Disabled()/Enabled()

pkg/apis/eksctl.io/v1alpha5/vpc.go
D3nn: Done.

integration/cluster_api_endpoints_test.go
D3nn: Done.

site/content/usage/06-vpc-networking.md
D3nn: Done.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -196,8 +197,11 @@ func IsEnabled(v *bool) bool { return v != nil && *v }
 // IsDisabled will only return true if v is not nil and false
 func IsDisabled(v *bool) bool { return v != nil && !*v }
 
+// IsSet will only return true if v is not nil
+func IsSet(v interface{}) bool { return !reflect.ValueOf(v).IsNil() }
```
rndstr: i'm not sure this function has any benefit, what's wrong with `foo != nil`?

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -63,9 +64,61 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 		}
 	}
 
+	if !cfg.HasClusterEndpointAccess() {
+		return fmt.Errorf("Cluster endpoint access invalid, at least one of Public, Private " +
```
rndstr:
```suggestion
		return fmt.Errorf("cluster endpoint access invalid, either public or private " +
```

integration/cluster_api_endpoints_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+var _ = Describe("(Integration) Create and Update Cluster with Endpoint Configs", func() {
+
+	type EndpointAccessCase struct {
+		Name    string
+		Private bool
+		Public  bool
+		Type    string
+		Output  string
```
rndstr: `Output`, `Delete`, and `fun` seem unused

integration/cluster_api_endpoints_test.go
```diff
@@ -0,0 +1,223 @@
...
+				"--name", clName,
+				"-o", "yaml",
+			)
+			Expect(getCmd).To(RunSuccessfullyWithOutputStringLines(
+				ContainElement(ContainSubstring(endpointPubTmpl, e.Public)),
+				ContainElement(ContainSubstring(endpointPrivTmpl, e.Private)),
+			))
+			if e.Type == deleteCluster {
+				// nned to update public access to allow access to delete when it isn't allowed
+				if e.Public == false {
```
rndstr:
```suggestion
				if !e.Public {
```

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
```diff
@@ -0,0 +1,84 @@
...
+				cc.VPC.ClusterEndpoints = e.endpoints
+			}
+			if e.Public != nil && cc.VPC.ClusterEndpoints != nil {
+				cc.VPC.ClusterEndpoints.PublicAccess = e.Public
+			}
+			if e.Private != nil && cc.VPC.ClusterEndpoints != nil {
+				cc.VPC.ClusterEndpoints.PrivateAccess = e.Private
+			}
+			if e.Public != nil && e.Private != nil {
+				if *e.Public == true || *e.Private == true {
```
rndstr: `if *e.Public || e.Private` (same below, use operators over comparison)

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
```diff
@@ -0,0 +1,84 @@
...
+					Expect(cc.HasClusterEndpointAccess()).Should(BeTrue())
```
rndstr: could add an `Expected bool` to the struct and pass it in through entries instead of figuring out the result here

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if ok, err := ctl.CanUpdate(cfg); !ok {
+		return err
+	}
+
+	curPrivate, curPublic, err := ctl.GetCurrentClusterConfigForEndpoints(cfg)
+	if err!= nil {
+		return err
+	}
+	logger.Info("Current endpoint access: private access:%v, public access: %v",
```
rndstr: afaik our log messages start with a lowercase character

pkg/eks/update.go
```diff
@@ -91,6 +91,37 @@ func (c *ClusterProvider) UpdateClusterConfigForLogging(cfg *api.ClusterConfig)
 	return nil
 }
 
+// GetCurrentClusterConfigForEndpoints fetches current cluster endpoint configuration for public and private access types
+func (c *ClusterProvider) GetCurrentClusterConfigForEndpoints(spec *api.ClusterConfig) (*bool, *bool, error) {
```
rndstr: i don't think this should return pointers unless they can be `nil` but then one actually needs to verify they aren't `nil` before dereferencing (which currently isn't done)

pkg/vpc/vpc.go
```diff
@@ -232,3 +239,21 @@ func ImportAllSubnets(provider api.ClusterProvider, spec *api.ClusterConfig) err
 
 	return nil
 }
+
+//UseEndpointAccessFromCluster retrieves the Cluster's endpoint access configuration via the SDK
+// as the CloudFormation Stack doesn't support that configuration currently
+func UseEndpointAccessFromCluster(provider api.ClusterProvider, spec *api.ClusterConfig) error {
+	input := &awseks.DescribeClusterInput{
+		Name: aws.String(spec.Metadata.Name),
```
rndstr: nit: can just use `Name: &spec.Metadata.Name`. I think main use case for `aws.String()` is if you have a string literal whose address you need.

pkg/eks/tasks.go
```diff
@@ -15,6 +18,37 @@ type clusterConfigTask struct {
...
+}
+
+//UpdateEndpointsMsg gives message indicating that they need to use eksctl utils to make this config
+func UpdateEndpointsMsg(cfg *api.ClusterConfig) string {
+	return fmt.Sprintf(
+		"you can update endpoint access with `eksctl utils update-cluster-api-access --region=%s --name=%s --private-access bool --public-access bool", cfg.Metadata.Region, cfg.Metadata.Name)
```
rndstr: nit: consistency with `--private-access=bool --public-access=bool`. also there seems to be a missing closing backtick

pkg/eks/tasks.go
```diff
@@ -87,3 +119,23 @@ func (c *ClusterProvider) appendCreateTasksForIAMServiceAccounts(cfg *api.Cluste
...
+
+func (c *ClusterProvider) maybeAppendTasksForEndpointAccessUpdates(cfg *api.ClusterConfig, newTasks, tasks *manager.TaskTree) {
+	// if a cluster config doesn't have the default api endpoint access, append a new task
+	// so that we update the cluster with the new access configuration.  This is a
+	// non-CloudFormation context, so we create a task to send it through the EKS API.
+	// A caveat is that sending the default endpoint parameters for a cluster as an update will
+	// return an error from the EKS API, so we must check for this before sending the request.
+	if cfg.HasClusterEndpointAccess() && endpointsEqual(*cfg.VPC.ClusterEndpoints, *api.ClusterEndpointAccessDefaults()) {
+		logger.Info(DefaultEndpointsMsg(cfg))
+		logger.Info(UpdateEndpointsMsg(cfg))
```
rndstr: this skips the `newTasks` which include OIDC and cloudwatch logging. is that the intention here?

pkg/eks/tasks.go
```diff
@@ -87,3 +119,23 @@ func (c *ClusterProvider) appendCreateTasksForIAMServiceAccounts(cfg *api.Cluste
...
+	} else {
+		logger.Info(CustomEndpointsMsg(cfg))
+		newTasks.Append(&clusterConfigTask{
+			info: "update cluster VPC endpoint access configuration",
+			spec: cfg,
+			call: c.UpdateClusterConfigForEndpoints,
+		})
+		tasks.Append(newTasks)
```
rndstr: i think this should be moved back to where it was

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: I did it like this as it's describing what's changing, not what the endstate is going to be after the update.

rndstr: from what i see my suggestion gives the exact same output without the bells and whistles :)

site/content/usage/06-vpc-networking.md
D3nn: I'm going to use --name for now and let the other PR make the changes for now so they can all be consistent at once.

integration/cluster_api_endpoints_test.go
D3nn: Yes.  Experiments that I abandoned.

pkg/apis/eksctl.io/v1alpha5/types.go
D3nn: It was the function name below that got me thinking about this.  IsSet just reads more naturally to me for cases (further down) where it is:
`IsSet(c.VPC.ClusterEndpoints.PublicAccess) && *c.CPC.ClusterEndpoints.PublicAccess`
vs.
`c.VPC.ClusterEndpoints.PublicAccess != nil && *c.VPC.ClusterEndpoints.PublicAccess)`

I could be convinced though if you feel strongly about it.

rndstr: i do because i think it adds unnecessary overhead (for reading, writing, and running code). I wouldn't know what `IsSet()` does without looking at it.

if the long variable name bothers you you could assign them to temp variables

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Keeping this the same (using --name) for now so the change can happen all at once rather than piecemeal.

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+			fs.BoolVar(&public, "public-access", true, "access for public clients")
+		})
+	cmd.SetRunFuncWithNameArg(func() error {
+		return doConfigureEndpointAccess(cmd, private, public)
+	})
+
+	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, cmd.ProviderConfig, false)
+}
+
+func doConfigureEndpointAccess(cmd *cmdutils.Cmd, newPrivate bool, newPublic bool) error {
```
errordeveloper: based on current convention this should be called `do<Action>`

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+	"strings"
+
+	"github.com/kris-nova/logger"
+	"github.com/spf13/pflag"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+func updateEndpointAccessCmd(cmd *cmdutils.Cmd) {
```
errordeveloper: based on current convention this should be called `<action>Cmd`, where `<action>` is a camel case version of the name of the command

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+	cfg := api.NewClusterConfig()
+	cmd.ClusterConfig = cfg
+
+	cmd.SetDescription("update-cluster-api-access", "Update cluster API endpoint configuration", "")
```
errordeveloper: Let's keep this consistent with config file, so I think it should be `update-cluster-endpoints`

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
+		cmdutils.AddApproveFlag(fs, cmd)
+		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
+	})
+
+	var private, public bool
+	cmd.FlagSetGroup.InFlagSet("Allow/Disallow Cluster API access flags",
+		func(fs *pflag.FlagSet) {
+			fs.BoolVar(&private, "private-access", false, "access for private (VPC) clients")
+			fs.BoolVar(&public, "public-access", true, "access for public clients")
```
errordeveloper: both flag names are consistent with config file fields 👍

pkg/ctl/utils/update_cluster_endpoint_access.go
errordeveloper: Also, 'Cluster API' means something else, let's avoid using it here;  let's change the description too

```suggestion
	cmd.SetDescription("update-cluster-endpoints", "Update Kubernetes API endpoint configuration", "")
```

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,135 @@
...
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		cmdutils.AddNameFlag(fs, cfg.Metadata)
+		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
+		cmdutils.AddApproveFlag(fs, cmd)
+		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
+	})
+
+	var private, public bool
+	cmd.FlagSetGroup.InFlagSet("Allow/Disallow Cluster API access flags",
```
errordeveloper: - word "flags" will be appended by the library
- 'Cluster API' means something else, let's avoid using it here

```suggestion
	cmd.FlagSetGroup.InFlagSet("Update private/public Kubernetes API endpoint configuration",
```

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -123,3 +123,11 @@ func DefaultClusterNAT() *ClusterNAT {
 		Gateway: &single,
 	}
 }
+
+// ClusterEndpointAccessDefaults returns a ClusterEndpoints pointer with default values set.
+func ClusterEndpointAccessDefaults() *ClusterEndpoints {
+	return &ClusterEndpoints{
+		PrivateAccess: Disabled(),
```
errordeveloper: I think it's acceptable to change the default for new clusters and enable private access, as currently private nodegroups have to go via NAT gateway to access the API...

D3nn: Should I then give the message that they'll need to update securitygroups/etc in order for clients within the VPC to communicate with the Kubernetes API whenever we create a cluster with these defaults?

pkg/apis/eksctl.io/v1alpha5/types.go
D3nn: I've got that change teed up for my next batch of updates.

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -58,6 +66,7 @@ const (
 	SubnetTopologyPrivate SubnetTopology = "Private"
 	// SubnetTopologyPublic represents publicly-routed subnets
 	SubnetTopologyPublic SubnetTopology = "Public"
+
```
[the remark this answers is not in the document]
D3nn: That brings up a question.  Should I gofmt any file I touch?  It seems like I've seen some cases where a file hasn't been gofmt'd previously.

pkg/apis/eksctl.io/v1alpha5/vpc.go
D3nn: Queued up for next updates.

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Queued up for next round of updates.

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Queued up for next round of updates.

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Also queued up for next round of updates.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: I like the name change, I'm thinking though that I should change the description slightly then to "Update Kubernetes API endpoint access configuration" to disambiguate it from changing the URL for the API endpoint.  Thoughts?

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Seems like another case were "endpoint configuration" might be "endpoint access configuration"?

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Isn't it that way, where ConfigureEndpointAccess is the action?  Or should it be doConfiguredClusterEndpointAccess?

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: It'll be gone in the next round of updates.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Shoot, I had it the way I described but changed it.  I like the idea of showing the actual changes being made and then showing the endstate after the change.  If I (re)implement it that way, would it be acceptable?

pkg/eks/tasks.go
D3nn: No, fixing that.

site/content/usage/06-vpc-networking.md
D3nn: Per @errordeveloper's comment, maybe we shouldn't use Cluster API as it might be confused with the Kubernetes Cluster API work.

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Done.

pkg/apis/eksctl.io/v1alpha5/types.go
D3nn: Done

integration/cluster_api_endpoints_test.go
D3nn: Done.

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Done

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
D3nn: Done.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Done.

pkg/eks/update.go
D3nn: Done.

pkg/vpc/vpc.go
D3nn: Changed.

pkg/eks/tasks.go
D3nn: fixed.

pkg/eks/tasks.go
D3nn: moved.

pkg/apis/eksctl.io/v1alpha5/vpc.go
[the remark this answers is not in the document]
D3nn: cleaned up.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: changed.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: changed.

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: changed the name and the description.

pkg/apis/eksctl.io/v1alpha5/defaults.go
D3nn: Need to discuss the UX of this.  Seems like we need to say something about what users will need to do to make private access actually work.  I can put the AWS URL that speaks to this in the output, but seems odd that the default configuration will flag immediately put this issue in front of them.

pkg/ctl/utils/update_cluster_endpoint_access.go
rndstr: what would the output be?

pkg/apis/eksctl.io/v1alpha5/vpc_test.go
```diff
@@ -0,0 +1,91 @@
...
+				}
+				if e.Expected {
+					Expect(cc.HasClusterEndpointAccess()).Should(BeTrue())
+				}
+				if e.Expected {
+					Expect(cc.HasClusterEndpointAccess()).Should(BeTrue())
+				}
+				if !e.Expected {
+					Expect(cc.HasClusterEndpointAccess()).Should(BeFalse())
+				}
```
rndstr: the whole block can be replaced with `Expect(cc.HasClusterEndpointAccess()).To(Equal(e.Expected))`

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,140 @@
...
+		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
+	})
+
+	var private, public bool
+	cmd.FlagSetGroup.InFlagSet("Update private/public Kubernetes API endpoint access configuration",
+		func(fs *pflag.FlagSet) {
+			fs.BoolVar(&private, "private-access", false, "access for private (VPC) clients")
+			fs.BoolVar(&public, "public-access", true, "access for public clients")
+		})
+	cmd.SetRunFuncWithNameArg(func() error {
```
rndstr: i think this should be `cmd.SetRunFunc(func() error {` since we are not actually looking at the `Cmd#NameArg` here

pkg/apis/eksctl.io/v1alpha5/vpc.go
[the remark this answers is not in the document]
errordeveloper: Yes, it's recommended to use either gofmt, goimports or goreturns. In VSCode I have one of them enabled on save. We have seen some discrepancies between versions, as it turns the formatting rules are not set in stone yet...

pkg/ctl/utils/update_cluster_endpoint_access.go
errordeveloper: Up to you, as long as it's consistent :)

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+There are some additional caveats when configuring Kubernetes API endpoint access:
+
+1. EKS doesn't allow one to create or update a cluster without at least one of private or public access being
+   enabled.
+1. EKS does allow creating a configuration which allows only private access to be enabled, but eksctl doesn't
+   support it during cluster creation as it prevents eksctl from being able to join the worker nodes to the cluster.
+1. To create private-only Kubernetes API endpoint access, one must first create the cluster *with* public Kubernetes API
+   endpoint access, and then use `/eksctl utils update-cluster-endpoints` to change it after the cluster is finished
+   creating.
+1. Updating a a cluster to have private only Kubernetes API endpoint access means that Kubernetes commands
```
errordeveloper: _  Updating a cluster_

site/content/usage/06-vpc-networking.md
```diff
@@ -160,3 +160,40 @@ See the complete example [here](https://github.com/weaveworks/eksctl/blob/master
...
+   (e.g. `kubectl`) as well as `eksctl delete cluster`, `eksctl utils write-kubeconfig`, and possibly the command
+   `eksctl utils update-kube-proxy` must be run within the cluster VPC.  This requires some changes to various AWS
+   resources.  See:
+   https://docs.aws.amazon.com/en_pv/eks/latest/userguide/cluster-endpoint#private-access
```
errordeveloper: I'd change this to `(see [EKS user guide](ttps://docs.aws.amazon.com/en_pv/eks/latest/userguide/cluster-endpoint#private-access))` at the end of the previous sentence.

pkg/ctl/utils/update_cluster_endpoint_access.go
errordeveloper: Yes, let's use `--name` here, same as in `eksctl utils update-cluster-logging`. I agree that consistency can be improved going forward, but I'd rather rename once we settled on what the consistent solution should be.

site/content/usage/06-vpc-networking.md
D3nn: Done

pkg/ctl/utils/update_cluster_endpoint_access.go
rndstr: https://github.com/weaveworks/eksctl/pull/1363

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -0,0 +1,168 @@
...
+			fs.BoolVar(&private, "private-access", false, "access for private (VPC) clients")
+			fs.BoolVar(&public, "public-access", false, "access for public clients")
+		})
+	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, cmd.ProviderConfig, false)
+}
+
+func accessFlagsSet(cmd *cmdutils.Cmd) (privateSet, publicSet bool) {
+	cmd.FlagSetGroup.InFlagSet("Update private/public Kubernetes API endpoint access configuration",
+		func(fs *pflag.FlagSet) {
+			fs.VisitAll(func(f *pflag.Flag) {
```
rndstr: can use `fs.Lookup(name)` instead

D3nn: Possibly, though there are only two flags in this flagset anyway, and the difference amounts to one line of code once you account for the lookup possibly returning nil.

Before:
```
		func(fs *pflag.FlagSet) {
			fs.VisitAll(func(f *pflag.Flag) {
				switch f.Name {
				case "private-access":
					privateSet = f.Changed
				case "public-access":
					publicSet = f.Changed
				default:
					// do nothing
				}
			})
```
After:
```
		func(fs *pflag.FlagSet) {
			priv := fs.Lookup("private-access")
			if priv != nil {
				privateSet = priv.Changed
			}
			pub := fs.Lookup("public-access")
			if pub != nil {
				publicSet = pub.Changed
			}
		}
```

pkg/ctl/utils/update_cluster_endpoint_access.go
D3nn: Sorry I missed this comment previously.  @rndstr, are you suggesting that we make the change in this PR or should we continue with the plan of fixing this all in #1363.  Or is this just in reply to @errordeveloper?

rndstr: > should we continue with the plan of fixing this all in #1363

this, or whichever gets merged last. if we merge #1363 first then we need to incorporate this here, otherwise there.

D3nn: I think that this is the last bit we need to merge this particular PR, with the VPC-peering and customer network <-> VPC configuration coming in the next.

pkg/ctl/utils/update_cluster_endpoint_access.go
rndstr: what about
```
                func(fs *pflag.FlagSet) {
			if priv := fs.Lookup("private-access"); priv != nil {
				privateSet = priv.Changed
			}
			if pub := fs.Lookup("public-access"); pub != nil {
				publicSet = pub.Changed
			}
		}
```
and technically, there is a bug if the lookup returns `nil`. in that case it will just ignore the value passed in.

pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -111,11 +108,10 @@ func doUpdateClusterEndpoints(cmd *cmdutils.Cmd, newPrivate bool, newPublic bool
...
 		meta.Name, meta.Region, describeAccessToUpdate)
 
-	fatal, err := validateClusterEndpointConfig(newPrivate, newPublic)
-	if err != nil {
-		if fatal {
-			logger.Critical(err.Error())
-			os.Exit(2)
+	if err := cfg.ValidateClusterEndpointConfig(); err != nil {
+		// Accept private-only since it leaves the cluster accessible
+		if err != api.ErrClusterEndpointPrivateOnly {
```
D3nn: Shouldn't this be:
```
         if err := api.ErrClusterEndpointNoAccess; err != nil {
...
```
Since api.ErrClusterEnpointPrivateOnly *does* leave the cluster accessible?

rndstr: the code that i wrote exits for all `errs != nil and errs != private-only` which is what we want i think? i think the comment is confusing, though, let me edit/move that

PR #1165, opened 2019-08-20
Add command eksctl gitops apply

martina-if:
`eksctl gitops apply` combines the functionality of `eksctl install flux` and
`eksctl profile generate` in one command. 

- [x] Manually tested
- [x] Add integration test
- [x] Check acceptance criteria: add flag --quickstart-profile (replace name argument?)

### DIFF
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"github.com/weaveworks/eksctl/pkg/ctl/generate"
+	"github.com/weaveworks/eksctl/pkg/ctl/gitops"
+
 	"os"
 
 	"github.com/kris-nova/logger"
@@ -30,6 +32,7 @@ func addCommands(rootCmd *cobra.Command, flagGrouping *cmdutils.FlagGrouping) {
 	if os.Getenv("EKSCTL_EXPERIMENTAL") == "true" {
 		rootCmd.AddCommand(install.Command(flagGrouping))
 		rootCmd.AddCommand(generate.Command(flagGrouping))
+		rootCmd.AddCommand(gitops.Command(flagGrouping))
 	}
 	rootCmd.AddCommand(utils.Command(flagGrouping))
 	rootCmd.AddCommand(completion.Command(rootCmd))
--- a/cmd/eksctl/main.go
+++ b/cmd/eksctl/main.go
@@ -2,9 +2,6 @@ package main
 
 import (
 	"fmt"
-	"github.com/weaveworks/eksctl/pkg/ctl/generate"
-	"github.com/weaveworks/eksctl/pkg/ctl/gitops"
-
 	"os"
 
 	"github.com/kris-nova/logger"
@@ -15,7 +12,9 @@ import (
 	"github.com/weaveworks/eksctl/pkg/ctl/create"
 	"github.com/weaveworks/eksctl/pkg/ctl/delete"
 	"github.com/weaveworks/eksctl/pkg/ctl/drain"
+	"github.com/weaveworks/eksctl/pkg/ctl/generate"
 	"github.com/weaveworks/eksctl/pkg/ctl/get"
+	"github.com/weaveworks/eksctl/pkg/ctl/gitops"
 	"github.com/weaveworks/eksctl/pkg/ctl/install"
 	"github.com/weaveworks/eksctl/pkg/ctl/scale"
 	"github.com/weaveworks/eksctl/pkg/ctl/update"
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	github.com/weaveworks/flux v0.0.0-20190729133003-c78ccd3706b5
 	github.com/weaveworks/github-release v0.6.2
 	github.com/weaveworks/launcher v0.0.0-20180711153254-f1b2830d4f2d
+	github.com/whilp/git-urls v0.0.0-20160530060445-31bac0d230fa
 	github.com/zmap/zlint v0.0.0-20190806182416-88c3f6b6f2f5 // indirect
 	go.etcd.io/bbolt v1.3.3 // indirect
 	go.uber.org/atomic v1.4.0 // indirect
--- a/integration/creategetdelete_test.go
+++ b/integration/creategetdelete_test.go
@@ -8,8 +8,7 @@ import (
 	"os"
 	"time"
 
-	awseks "github.com/aws/aws-sdk-go/service/eks"
-	harness "github.com/dlespiau/kube-test-harness"
+	"github.com/kubicorn/kubicorn/pkg/namer"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
@@ -17,7 +16,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/yaml"
 
-	"github.com/kubicorn/kubicorn/pkg/namer"
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 	"github.com/weaveworks/eksctl/pkg/authconfigmap"
 	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
@@ -168,8 +166,7 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 						"--git-url", Repository,
 						"--git-email", Email,
 						"--git-branch", branch,
-						// FIXME
-						// "--git-private-ssh-key-path", privateSSHKeyPath,
+						"--git-private-ssh-key-path", privateSSHKeyPath,
 						"--output-path", tempOutputDir,
 						"--quickstart-profile", "app-dev",
 						"--cluster", clusterName,
--- a/integration/creategetdelete_test.go
+++ b/integration/creategetdelete_test.go
@@ -149,6 +149,41 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 			})
 		})
 
+		Context("gitops apply", func() {
+			It("should add quickstart to the repo and the cluster", func() {
+				// Use a random branch to ensure test runs don't step on each others.
+				branch := namer.RandomName()
+				cloneDir, err := createBranch(branch)
+				Expect(err).ShouldNot(HaveOccurred())
+				defer deleteBranch(branch, cloneDir)
+
+				tempOutputDir, err := ioutil.TempDir(os.TempDir(), "gitops-repo-")
+				assertFluxManifestsAbsentInGit(branch)
+
+				deleteFluxInstallation(kubeconfigPath)
+				assertFluxPodsAbsentInKubernetes(kubeconfigPath)
+
+				eksctlSuccessWith(params{
+					Args: []string{"gitops", "apply",
+						"--git-url", Repository,
+						"--git-email", Email,
+						"--git-branch", branch,
+						// FIXME
+						// "--git-private-ssh-key-path", privateSSHKeyPath,
+						"--output-path", tempOutputDir,
+						"--quickstart-profile", "app-dev",
+						"--cluster", clusterName,
+						"--region", region,
+					},
+					Env: []string{"EKSCTL_EXPERIMENTAL=true"},
+				})
+
+				assertQuickStartComponentsPresentInGit(branch)
+				assertFluxManifestsPresentInGit(branch)
+				assertFluxPodsPresentInKubernetes(kubeconfigPath)
+			})
+		})
+
 		Context("and scale the initial nodegroup", func() {
 			It("should not return an error", func() {
 				eksctlSuccess("scale", "nodegroup",
--- a/integration/creategetdelete_test.go
+++ b/integration/creategetdelete_test.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"time"
 
+	awseks "github.com/aws/aws-sdk-go/service/eks"
+	harness "github.com/dlespiau/kube-test-harness"
 	"github.com/kubicorn/kubicorn/pkg/namer"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
--- a/pkg/ctl/generate/generate.go
+++ b/pkg/ctl/generate/generate.go
@@ -1,0 +1,14 @@
+package generate
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+// Command creates `generate` commands
+func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
+	verbCmd := cmdutils.NewVerbCmd("generate", "Generate GitOps manifests", "")
+	cmdutils.AddResourceCmd(flagGrouping, verbCmd, generateProfileCmd)
+	return verbCmd
+}
--- a/pkg/ctl/generate/profile.go
+++ b/pkg/ctl/generate/profile.go
@@ -2,7 +2,6 @@ package generate
 
 import (
 	"context"
-	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
@@ -14,10 +13,6 @@ import (
 	"github.com/weaveworks/eksctl/pkg/git"
 	"github.com/weaveworks/eksctl/pkg/gitops"
 	"github.com/weaveworks/eksctl/pkg/gitops/fileprocessor"
-)
-
-const (
-	defaultGitTimeout = 20 * time.Second
 )
 
 type options struct {
@@ -43,8 +38,6 @@ func generateProfileCmd(cmd *cmdutils.Cmd) {
 		fs.StringVarP(&o.GitOptions.Branch, "git-branch", "", "master", "Git branch")
 		fs.StringVarP(&o.ProfilePath, "profile-path", "", "./", "Path to generate the profile in")
 		_ = cobra.MarkFlagRequired(fs, "git-url")
-		fs.StringVar(&o.PrivateSSHKeyPath, "git-private-ssh-key-path", "",
-			"Optional path to the private SSH key to use with Git, e.g.: ~/.ssh/id_rsa")
 
 		cmdutils.AddNameFlag(fs, cfg.Metadata)
 		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)
@@ -71,7 +64,7 @@ func doGenerateProfile(cmd *cmdutils.Cmd, o options) error {
 		Path:      o.ProfilePath,
 		GitOpts:   o.GitOptions,
 		GitCloner: git.NewGitClient(context.Background(), git.ClientParams{
-			Timeout:           defaultGitTimeout,
+			Timeout:           git.DefaultGitTimeout,
 			PrivateSSHKeyPath: o.PrivateSSHKeyPath,
 		}),
 		FS: afero.NewOsFs(),
@@ -79,7 +72,6 @@ func doGenerateProfile(cmd *cmdutils.Cmd, o options) error {
 	}
 
 	err := profile.Generate(context.Background())
-
 	if err != nil {
 		return errors.Wrap(err, "error generating profile")
 	}
--- a/pkg/ctl/generate/profile.go
+++ b/pkg/ctl/generate/profile.go
@@ -21,7 +21,7 @@ const (
 )
 
 type options struct {
-	gitops.GitOptions
+	GitOptions        git.Options
 	ProfilePath       string
 	PrivateSSHKeyPath string
 }
@@ -39,8 +39,8 @@ func generateProfileCmd(cmd *cmdutils.Cmd) {
 	})
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
-		fs.StringVarP(&o.URL, "git-url", "", "", "URL for the quickstart base repository")
-		fs.StringVarP(&o.Branch, "git-branch", "", "master", "Git branch")
+		fs.StringVarP(&o.GitOptions.URL, "git-url", "", "", "URL for the quickstart base repository")
+		fs.StringVarP(&o.GitOptions.Branch, "git-branch", "", "master", "Git branch")
 		fs.StringVarP(&o.ProfilePath, "profile-path", "", "./", "Path to generate the profile in")
 		_ = cobra.MarkFlagRequired(fs, "git-url")
 		fs.StringVar(&o.PrivateSSHKeyPath, "git-private-ssh-key-path", "",
@@ -84,5 +84,6 @@ func doGenerateProfile(cmd *cmdutils.Cmd, o options) error {
 		return errors.Wrap(err, "error generating profile")
 	}
 
+	profile.DeleteClonedDirectory()
 	return nil
 }
--- a/pkg/ctl/generate/profile.go
+++ b/pkg/ctl/generate/profile.go
@@ -20,13 +20,6 @@ const (
 	defaultGitTimeout = 20 * time.Second
 )
 
-// Command creates `generate` commands
-func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
-	verbCmd := cmdutils.NewVerbCmd("generate", "Generate GitOps manifests", "")
-	cmdutils.AddResourceCmd(flagGrouping, verbCmd, generateProfileCmd)
-	return verbCmd
-}
-
 type options struct {
 	gitops.GitOptions
 	ProfilePath       string
--- a/pkg/ctl/gitops/apply.go
+++ b/pkg/ctl/gitops/apply.go
@@ -154,7 +154,7 @@ func doApplyGitops(cmd *cmdutils.Cmd, opts options) error {
 		Dir:               usersRepoDir,
 	})
 
-	gitOps := gitops.GitOps{
+	gitOps := gitops.Applier{
 		UserRepoPath:     usersRepoDir,
 		UsersRepoOpts:    opts.gitOptions,
 		GitClient:        gitClient,
@@ -164,7 +164,7 @@ func doApplyGitops(cmd *cmdutils.Cmd, opts options) error {
 		QuickstartName:   opts.quickstartNameArg,
 	}
 
-	if err = gitOps.Apply(context.Background()); err != nil {
+	if err = gitOps.Run(context.Background()); err != nil {
 		return err
 	}
 	return nil
--- a/pkg/ctl/gitops/apply.go
+++ b/pkg/ctl/gitops/apply.go
@@ -18,12 +18,14 @@ import (
 	"github.com/weaveworks/eksctl/pkg/gitops"
 	"github.com/weaveworks/eksctl/pkg/gitops/fileprocessor"
 	"github.com/weaveworks/eksctl/pkg/gitops/flux"
+	"github.com/weaveworks/eksctl/pkg/utils/file"
 )
 
 type options struct {
-	gitOptions        git.Options
-	quickstartNameArg string
-	outputPath        string
+	gitOptions           git.Options
+	quickstartNameArg    string
+	outputPath           string
+	gitPrivateSSHKeyPath string
 }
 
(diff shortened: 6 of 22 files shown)


pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,183 @@
...
+		return doApplyGitops(cmd, opts)
+	})
+
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&opts.git.URL, "git-url", "", "", "URL for the git repository that will contain the cluster components")
+		fs.StringVarP(&opts.git.Branch, "git-branch", "", "master", "Git branch")
+		fs.StringVarP(&opts.outputPath, "output-path", "", "./", "Path to directory where the GitOps repo will be cloned")
+		fs.StringVar(&opts.git.User, "git-user", "Flux", "Username to use as Git committer")
+		fs.StringVar(&opts.git.Email, "git-email", "", "Email to use as Git committer")
+		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "name of the EKS cluster to add the nodegroup to")
```
dholbach: Is this what `--name` is when using `create cluster`?

martina-if: Yes, but in this case the name could be the name of the quickstart profile so I went for `--cluster` like we do in `eksctl create nodegroup`. What do you think? I don't have a strong opinion here.

dholbach: Mh, ok. I wasn't aware of that.

I think it'd be nice if we had consistent options and flags across everything, but I guess cluster/name is not too confusing.

martina-if: I think @errordeveloper is the one who knows what it should be to be consistent.

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/git"
+	"github.com/weaveworks/eksctl/pkg/gitops"
+	"github.com/weaveworks/eksctl/pkg/gitops/fileprocessor"
+	"github.com/weaveworks/eksctl/pkg/gitops/flux"
+)
+
+const (
+	defaultGitTimeout = 20 * time.Second
```
marccarre: Nit: given this constant also exists in `pkg/ctl/generate/profile.go`, should we move it elsewhere, e.g. within the `git` package and only use one?

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+
+	var opts options
+
+	cmd.SetRunFuncWithNameArg(func() error {
+		return doApplyGitops(cmd, opts)
+	})
+
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&opts.quickstartNameArg, "quickstart-profile", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.git.URL, "git-url", "", "", "URL for the git repository that will contain the cluster components")
```
marccarre: Nit: I think we're starting to have several commands which accept Git-related arguments.
It may be worth factorising these -- to help reduce boilerplate and enforce consistency across commands -- in, for example, a `AddCommonFlagsForGit` function in `pkg/ctl/cmdutils/cmdutils.go`?

cmd/eksctl/main.go
```diff
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"github.com/weaveworks/eksctl/pkg/ctl/generate"
+	"github.com/weaveworks/eksctl/pkg/ctl/gitops"
```
2opremio: This should go at the end (together with the generate import)

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+
+func repoURLForQuickstart(quickstartArgument string) (string, error) {
+	// FIXME: urls like git@github.com:weaveworks/eksctl produce the error "first path segment in URL cannot contain colon"
+	parsedURL, err := url.Parse(strings.ReplaceAll(quickstartArgument, ":", "/"))
+	if err == nil && parsedURL.IsAbs() {
+		return quickstartArgument, nil
+	}
+	if quickstartArgument == "app-dev" {
+		// FIXME rename to eks-quickstart-app-dev once the repo is renamed
+		return "git@github.com:weaveworks/eks-gitops-example.git", nil
```
marccarre: Given the above `strings.ReplaceAll(quickstartArgument, ":", "/")`, will this URL not cause issues? Or is it okay because of the `.git` at the end? 🤔

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if err = gitOps.Apply(context.Background()); err != nil {
+		return err
+	}
+	return nil
+}
+
+func repoURLForQuickstart(quickstartArgument string) (string, error) {
+	// FIXME: urls like git@github.com:weaveworks/eksctl produce the error "first path segment in URL cannot contain colon"
```
marccarre: What's the plan regarding this `FIXME`? Is it to fix it in this PR or in a follow-up issue? If the latter, it may be worth adding a link to the issue in the comment.

pkg/git/git.go
```diff
@@ -97,6 +120,21 @@ func (git Client) DeleteLocalRepo() error {
 	return fmt.Errorf("no cloned directory to delete")
 }
 
+// RepoName returns the name of the repository given its URL
+func RepoName(repoURL string) (string, error) {
+	// FIXME: urls like git@github.com:weaveworks/eksctl produce the error "first path segment in URL cannot contain colon"
+	u, err := url.Parse(strings.ReplaceAll(repoURL, ":", "/"))
```
marccarre: Duplicate: `strings.ReplaceAll(repoURL, ":", "/")`
It may be worth having a `SanitizeGitURL` function factorising this. And ideally some unit tests to cover it and add more context on what we're trying to achieve there, and why.

pkg/ctl/generate/generate.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+)
+
+// Command creates `generate` commands
+func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
+	verbCmd := cmdutils.NewVerbCmd("generate", "Generate GitOps manifests", "")
+	cmdutils.AddResourceCmd(flagGrouping, verbCmd, generateProfileCmd)
```
2opremio: Unrelated to the PR but I think `AddResourceCmd` should simply be called `AddCmd` since it's also used for non-resources.

pkg/ctl/generate/profile.go
```diff
@@ -20,15 +20,8 @@ const (
...
-// Command creates `generate` commands
-func Command(flagGrouping *cmdutils.FlagGrouping) *cobra.Command {
-	verbCmd := cmdutils.NewVerbCmd("generate", "Generate GitOps manifests", "")
-	cmdutils.AddResourceCmd(flagGrouping, verbCmd, generateProfileCmd)
-	return verbCmd
-}
-
 type options struct {
-	gitops.GitOptions
+	GitOptions  git.Options
```
2opremio: 👍

pkg/git/git.go
```diff
@@ -97,6 +120,21 @@ func (git Client) DeleteLocalRepo() error {
 	return fmt.Errorf("no cloned directory to delete")
 }
 
+// RepoName returns the name of the repository given its URL
+func RepoName(repoURL string) (string, error) {
```
marccarre: It looks like a function which is complex enough to have unit tests?

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/client-go/tools/clientcmd"
+
+	kubeclient "k8s.io/client-go/kubernetes"
```
2opremio: This should be part of the imports above, leading to a single group of external imports (we voted on this on Slack).

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/weaveworks/eksctl/pkg/gitops/fileprocessor"
+	"github.com/weaveworks/eksctl/pkg/gitops/flux"
+)
+
+const (
+	defaultGitTimeout = 20 * time.Second
+)
+
+type options struct {
+	git               git.Options
```
2opremio:
```suggestion
	gitOptions               git.Options
```

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+		return errors.New("please supply a valid gitops Quick Start URL or name in --quickstart-profile")
+	}
+
+	if opts.git.URL == "" {
+		return errors.New("please supply a valid --git-url argument")
+	}
+
+	quickstartRepoURL, err := repoURLForQuickstart(opts.quickstartNameArg)
+	if err != nil {
+		return errors.Wrapf(err, "please supply a valid Quick Start name or url")
```
marccarre: Über nit: `url` -> `URL`, as other messages to the end-user have the uppercased version, e.g.: `URL for the quickstart base repository`.

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&opts.quickstartNameArg, "quickstart-profile", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.git.URL, "git-url", "", "", "URL for the git repository that will contain the cluster components")
+		fs.StringVarP(&opts.git.Branch, "git-branch", "", "master", "Git branch")
+		fs.StringVarP(&opts.outputPath, "output-path", "", "./", "Path to directory where the GitOps repo will be cloned")
+		fs.StringVar(&opts.git.User, "git-user", "Flux", "Username to use as Git committer")
+		fs.StringVar(&opts.git.Email, "git-email", "", "Email to use as Git committer")
+		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "name of the EKS cluster to add the nodegroup to")
+		_ = cobra.MarkFlagRequired(fs, "quickstart-profile")
+		_ = cobra.MarkFlagRequired(fs, "git-url")
```
2opremio: Hah, I didn't know cobra allowed for this, we should use it in `eksctl install flux`

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	// FIXME: urls like git@github.com:weaveworks/eksctl produce the error "first path segment in URL cannot contain colon"
+	parsedURL, err := url.Parse(strings.ReplaceAll(quickstartArgument, ":", "/"))
+	if err == nil && parsedURL.IsAbs() {
+		return quickstartArgument, nil
+	}
+	if quickstartArgument == "app-dev" {
+		// FIXME rename to eks-quickstart-app-dev once the repo is renamed
+		return "git@github.com:weaveworks/eks-gitops-example.git", nil
+	}
+	return "", fmt.Errorf("invalid url or unknown Quick Start %s ", quickstartArgument)
```
marccarre: Über nit: `url` -> `URL`, as other messages to the end-user have the uppercased version, e.g.: `URL for the quickstart base repository`.

pkg/git/git.go
```diff
@@ -97,6 +120,21 @@ func (git Client) DeleteLocalRepo() error {
...
+	// FIXME: urls like git@github.com:weaveworks/eksctl produce the error "first path segment in URL cannot contain colon"
+	u, err := url.Parse(strings.ReplaceAll(repoURL, ":", "/"))
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to parse git url '%s'", repoURL)
```
marccarre: Über nit: `url` -> `URL`, as other messages to the end-user have the uppercased version, e.g.: `URL for the quickstart base repository`.

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+
+		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
+	})
+
+	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, cmd.ProviderConfig, false)
+}
+
+func doApplyGitops(cmd *cmdutils.Cmd, opts options) error {
+	if opts.quickstartNameArg == "" {
```
2opremio: Is this required even when using `MarkFlagRequired`? If so, what's the purpose of `MarkFlagRequired`?

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+		return errors.New("please supply a valid gitops Quick Start URL or name in --quickstart-profile")
+	}
+
+	if opts.git.URL == "" {
```
2opremio: Same here

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+	fluxOpts := flux.InstallOpts{
+		GitURL:      opts.git.URL,
+		GitBranch:   opts.git.Branch,
+		OutputPath:  opts.outputPath,
+		GitEmail:    opts.git.Email,
+		GitUser:     opts.git.User,
+		Namespace:   "flux",
+		GitFluxPath: "flux/",
+		Timeout:     defaultGitTimeout,
+	}
```
2opremio: `AdditionalFluxArgs: []string{"--sync-garbage-collection", "--manifest-generation"}` is missing

We should probably add them implicitly inside the flux package and not expose `AdditionalFluxArgs` externally

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,178 @@
...
+
+	quickstartRepoURL, err := repoURLForQuickstart(opts.quickstartNameArg)
+	if err != nil {
+		return errors.Wrapf(err, "please supply a valid Quick Start name or url")
+	}
+
+	if err := cmdutils.NewGitopsMetadataLoader(cmd).Load(); err != nil {
+		return err
+	}
+	cfg := cmd.ClusterConfig
```
2opremio: Most of this setup preamble is shared with the `install flux` command, maybe we can factor it out? (maybe it's not worth it though)

pkg/ctl/gitops/apply.go
2opremio: Flux uses https://github.com/whilp/git-urls , maybe we can use that

pkg/git/git.go
```diff
@@ -23,13 +27,29 @@ type Client struct {
...
+
 // NewGitClient returns a client that can perform git operations
 func NewGitClient(ctx context.Context, timeout time.Duration) *Client {
 	return &Client{
 		executor: executor.NewShellExecutor(ctx, timeout),
 	}
 }
 
+// NewGitClientDir returns a client that can perform git operations in an already cloned directory
+func NewGitClientDir(ctx context.Context, timeout time.Duration, path string) *Client {
```
2opremio: Unrelated to this PR (sorry to bring this up again), but contexts shouldn't be stored in objects which may have a long life.

pkg/git/git.go
```diff
@@ -50,6 +70,9 @@ func (git *Client) CloneRepo(cloneDirPrefix string, branch string, gitURL string
 // CloneRepoInPath clones a repo to the specified directory
 func (git *Client) CloneRepoInPath(clonePath string, branch string, gitURL string) error {
 	git.dir = clonePath
+	if err := os.MkdirAll(git.dir, 0755); err != nil {
```
2opremio: I would go for `0700`

pkg/gitops/flux/installer.go
```diff
@@ -203,17 +207,28 @@ func (fi *Installer) Run(ctx context.Context) error {
...
-	cleanCloneDir = true
 
 	logger.Info("Flux will only operate properly once it has write-access to the Git repository")
 	logger.Info("please configure %s so that the following Flux SSH public key has write access to it\n%s",
 		fi.opts.GitURL, fluxSSHKey.Key)
 	return nil
 }
 
+// DeleteCloneDir deletes the directory used for cloning the user's repository
+func (fi *Installer) DeleteCloneDir() {
```
2opremio: I think that the since `gitops apply` also clones a repository, we should decouple the git repository actions from either of them. I don't think it's a good design to have `gitops apply` piggybacking on `install flux`'s repository

pkg/gitops/apply.go
```diff
@@ -0,0 +1,56 @@
...
+	FluxInstaller    *flux.Installer
+	ProfileGenerator *Profile
+	GitClient        *git.Client
+}
+
+// Apply setups gitops in a repository and a cluster and installs flux, helm, tiller and a quickstart into the cluster
+func (g *GitOps) Apply(ctx context.Context) error {
+
+	// Install Flux, Helm and Tiller. Clones the user's repo
+	err := g.FluxInstaller.Run(context.Background())
```
2opremio: I think we should move the cloning/pushing out of the flux installer, so that it doesn't own it (it's a shared resource by  both of them).

Among other things, it will allow for having a single add, commit and push.

pkg/gitops/profile.go
```diff
@@ -70,10 +65,15 @@ func (p *Profile) Generate(ctx context.Context) error {
 	return nil
 }
 
-func (p *Profile) deleteClonedDirectory(path string) {
-	logger.Debug("deleting cloned directory %q", path)
-	if err := p.IO.RemoveAll(path); err != nil {
-		logger.Warning("unable to delete cloned directory %q", path)
+// DeleteClonedDirectory deletes the directory where the repository was cloned
+func (p *Profile) DeleteClonedDirectory() {
```
2opremio: I think that after the refactoring I suggested this is not needed. If not, I think its name should be consistent with the method added to the flux installer.

pkg/ctl/gitops/apply.go
errordeveloper: Yes, I think the current situation could be improved, but it's logical.

`--name` is used to refer to name of the things you are looking at, most of the time it's the name of the cluster, but in the case of nodegroups it's the name of the nodegroup.

It's also always the case that name is equivalent to argument:
- `eksctl create cluster foo` is equivalent to `eksctl create cluster --name=foo` 
- `eksctl create ng --cluster=foo bar` is equivalent to `eksctl create ng --cluster=foo --name=bar`

I'd like to depart from this, but renaming existing flags is hard and I'd rather see this exercise being done  on something simpler at first.

If you are going to have multiple profiles per cluster, and/or you want to allow user to specify profile name as argument - use `--name` for it.

Also, this command is still experiment, so we can change it later.

pkg/ctl/gitops/apply.go
martina-if: I fixed it by using a small library that can parse git urls

pkg/git/git.go
martina-if: fixed in new helper functions in the git package that call a library to parse the git urls

pkg/git/git.go
martina-if: added unit tests

pkg/ctl/gitops/apply.go
martina-if: Ok, then, for now I will keep the profile in the `--qiuckstart-profile` argument like Kalbir proposed in the issue, and the cluster name under `--cluster`. Then when he's back we can decide if we use the name argument and `--name` flag

pkg/ctl/gitops/apply.go
martina-if: git urls have that colon in between, so I have to keep it like that or we won't be able to clone it. But I fixed the replace hack.

pkg/ctl/gitops/apply.go
martina-if: I think the problem is that a user can do `--cluster ""` so we still need to check that the value is valid

pkg/ctl/gitops/apply.go
martina-if: I found that this setup is slightly different. Maybe we can see later a pattern and what we can extract. But that was why I insisted in the other flux pr that we extract all non-ui logic into it's own package.

pkg/git/git.go
martina-if: yeahp, we need to fix this at some point.

pkg/gitops/flux/installer.go
martina-if: I think you are right. I'll make gitops apply get it's own clone of the repo so it doesn't rely on an implementation detali.

martina-if: Agree. I left flux installer as it was before, making gitops apply create its own clone. We can decide later if we extract all git operations outside the flux installer if we want.

pkg/ctl/gitops/apply.go
martina-if: I think I would need to sit for a bit and reorganize that. We can address that in another Pr.

pkg/gitops/apply.go
martina-if: Good idea. To be done in a different PR

pkg/ctl/gitops/apply.go
martina-if: Not needed since this is in the Template

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,173 @@
...
+
+	cmd.SetRunFuncWithNameArg(func() error {
+		return doApplyGitops(cmd, opts)
+	})
+
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&opts.quickstartNameArg, "quickstart-profile", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.gitOptions.URL, "git-url", "", "", "URL for the git repository that will contain the cluster components")
+		fs.StringVarP(&opts.gitOptions.Branch, "git-branch", "", "master", "Git branch")
+		fs.StringVarP(&opts.outputPath, "output-path", "", "./", "Path to directory where the GitOps repo will be cloned")
```
dholbach: Also: is this the same as `profile-path` when using `generate profile`?

martina-if: Not exactly :/ I am happy to get name suggestions by the way. This is the directory where the user's repo will be cloned, (if path is "/home/m/myrepos/", it will be cloned in "/home/m/myrepos/repo1/"). Then, in the repo, we will create two folders: `flux` and `base`.

integration/creategetdelete_test.go
```diff
@@ -149,6 +149,41 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
...
+
+				deleteFluxInstallation(kubeconfigPath)
+				assertFluxPodsAbsentInKubernetes(kubeconfigPath)
+
+				eksctlSuccessWith(params{
+					Args: []string{"gitops", "apply",
+						"--git-url", Repository,
+						"--git-email", Email,
+						"--git-branch", branch,
+						// FIXME
```
martina-if: @marccarre The private ssh key in 1password doesn't work for me.

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// A git client that operates in the user's repo
+	gitClient := git.NewGitClient(context.Background(), git.ClientParams{
+		PrivateSSHKeyPath: opts.gitPrivateSSHKeyPath,
+		Timeout:           git.DefaultGitTimeout,
+		Dir:               usersRepoDir,
+	})
+
+	gitOps := gitops.GitOps{
```
martina-if: Suggestions for new names for the struct are welcome. GitopsApplier?

integration/creategetdelete_test.go
marccarre: After our discussion offline, are things working for you now?
(As also discussed with Martina, I've hopefully improved & clarified the instructions to setup the private SSH key in 1Password, in the notes for the test account.)

pkg/ctl/gitops/apply.go
marccarre: Given the package is `gitops` (and the file `apply.go`) then maybe just `Applier`?
(That way, it would read as `gitops.Applier` when called elsewhere.)

go.mod
```diff
@@ -73,6 +73,7 @@ require (
 	github.com/weaveworks/flux v0.0.0-20190729133003-c78ccd3706b5
 	github.com/weaveworks/github-release v0.6.2
 	github.com/weaveworks/launcher v0.0.0-20180711153254-f1b2830d4f2d
+	github.com/whilp/git-urls v0.0.0-20160530060445-31bac0d230fa
```
2opremio: You need to bump the dependency cache image

integration/creategetdelete_test.go
2opremio: Why not simply generate it on the fly and use https://github.com/stefanprodan/gitsrv ?

pkg/ctl/gitops/apply.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&opts.quickstartNameArg, "quickstart-profile", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.gitOptions.URL, "git-url", "", "", "URL for the git repository that will contain the cluster components")
+		fs.StringVarP(&opts.gitOptions.Branch, "git-branch", "", "master", "Git branch")
+		fs.StringVarP(&opts.outputPath, "output-path", "", "./", "Path to directory where the GitOps repo will be cloned")
+		fs.StringVar(&opts.gitOptions.User, "git-user", "Flux", "Username to use as Git committer")
+		fs.StringVar(&opts.gitOptions.Email, "git-email", "", "Email to use as Git committer")
+		fs.StringVar(&opts.gitPrivateSSHKeyPath, "git-private-ssh-key-path", "",
```
2opremio: Sorry I missed this. I think that if we add this flag here, it should also be part of the `eksctl install flux`

2opremio: And, the inconsistency between `eksctl gitops apply` and `eksctl install flux` is going to be confusing for the end users.

pkg/git/git.go
```diff
@@ -6,12 +6,22 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/kris-nova/logger"
+	"github.com/pkg/errors"
+
+	giturls "github.com/whilp/git-urls"
```
2opremio: Please use a single block of external imports

integration/creategetdelete_test.go
martina-if: It is working now. I think I would live the tests like this and try to set up gitsrv in a new PR that improves the IT

pkg/ctl/gitops/apply.go
martina-if: `install flux' already has this. It was added by marc

integration/creategetdelete_test.go
marccarre: Sounds good to me! (And thanks a lot for the pointer @2opremio! I wish I had known about it at the time 👍 )

go.mod
martina-if: Thanks. I'm pushing the image now. It just takes long with my internet connection...

pkg/ctl/gitops/apply.go
martina-if: :+1: yep, I did that

pkg/ctl/gitops/apply.go
2opremio: Ah ... sorry, I missed that

PR #1168, opened 2019-08-21
Make git shell executor more generic by accepting arbitrary env vars

marccarre:
### Description

After discussion with @martina-if, we decided it would be more flexible to have the git `ShellExecutor` accept an array of environment variables.

### Checklist

- [x] ~Added tests that cover your change (if possible)~
- [x] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~
- [x] ~Manually tested~

Follows up on #1148.

### DIFF
--- a/pkg/git/executor/executor.go
+++ b/pkg/git/executor/executor.go
@@ -2,7 +2,6 @@ package executor
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"os/exec"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/git/executor/executor.go
```diff
@@ -34,8 +33,8 @@ func (e ShellExecutor) Exec(command string, dir string, args ...string) error {
 	ctx, ctxCancel := context.WithTimeout(e.parentCtx, e.timeout)
 	defer ctxCancel()
 	cmd := exec.CommandContext(ctx, command, args...)
-	if e.privateSSHKeyPath != "" {
-		cmd.Env = []string{fmt.Sprintf("GIT_SSH_COMMAND=ssh -i %s", e.privateSSHKeyPath)}
+	if len(e.envVars) > 0 {
+		cmd.Env = e.envVars
```
martina-if: doesn't this override any other env vars we have? Should we first get os.Env()?

marccarre: It does, but we were also overriding them before, and it was actually what I wanted to avoid inheriting a "polluted" shell/Git config.

PR #1171, opened 2019-08-21
Improve integration test helpers

errordeveloper:
### Description

We had been using helper function `eksctlSuccessWith`, `eksctlSuccessWith`, `eksctlStart` and `eksctlFail` to call `eksctl` binary from in the integration test suite.

These function are handy, but there is one major issue - any assertions in these functions end up being reference by the the line inside the function, not where the function had been called.

I've rewritten these as gomega matchers, which also read a little nicer:

```Go
c1 := NewCmd("true")
Expect(c1).To(RunSuccessfully())

c2 := NewCmd("false")
Expect(c1).ToNot(RunSuccessfully())

c4 := NewCmd("env").WithEnv("FOO=bar")

Expect(c4).To(RunSuccessfullyWithOutputString(ContainSubstring("FOO=bar\n")))
Expect(c4).To(RunSuccessfullyWithOutputString(ContainSubstring("PATH=")))

Expect(c4.WithCleanEnv()).To(RunSuccessfullyWithOutputString(Equal("FOO=bar\n")))
Expect(c4.WithCleanEnv()).To(RunSuccessfullyWithOutputString(Not(ContainSubstring("PATH="))))

c5 := NewCmd("echo").WithArgs("[]")

Expect(c5).To(RunSuccessfullyWithOutputString(MatchJSON("[]")))
```

### Checklist


integration/matchers/matchers.go
```diff
@@ -10,9 +10,46 @@ import (
...
+	m.stackNotFound = !found
+	return found, nil
+}
+
+func (m *existingStack) FailureMessage(actual interface{}) (message string) {
+	return fmt.Sprintf("Expected to find a Cloudformation stack named %s but it wasn't found", m.expectedStackName)
+}
+
+func (m *existingStack) NegatedFailureMessage(_ interface{}) (message string) {
+	return fmt.Sprintf("Expected NOT to find a Cloudformation stack named %s but it found", m.expectedStackName)
```
rndstr:
```suggestion
	return fmt.Sprintf("Expected NOT to find a Cloudformation stack named %s but it was found", m.expectedStackName)
```

integration/runner/runner.go
```diff
@@ -0,0 +1,184 @@
...
+func (m *runCmdMatcher) Match(actual interface{}) (bool, error) {
+	switch actual.(type) {
+	case Cmd:
+		return m.run(actual.(Cmd)), nil
+	case []Cmd:
+		for _, cmd := range actual.([]Cmd) {
+			if !m.run(cmd) {
+				return false, nil
+			}
+			continue
```
rndstr: remove

integration/runner/runner.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if !m.runCmdMatcher.run(cmd) {
+		m.failureMessage = m.runCmdMatcher.FailureMessage(nil)
+		return false, nil
+	}
+
+	output := string(m.runCmdMatcher.session.Buffer().Contents())
+
+	if ok, err := m.outputMatcher.Match(output); !ok {
+		m.failureMessage = m.outputMatcher.FailureMessage(output)
```
rndstr: shouldn't we capture the `negatedFailureMessag` here and return it in `runCmdOutputMatcher#NegatedFailureMessage`?

integration/runner/runner.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (m *runCmdOutputMatcher) Match(actual interface{}) (bool, error) {
+	cmd, ok := actual.(Cmd)
+	if !ok {
+		return false, fmt.Errorf("not a Cmd")
+	}
+
+	if !m.runCmdMatcher.run(cmd) {
+		m.failureMessage = m.runCmdMatcher.FailureMessage(nil)
```
rndstr: (not capturing the `negatedFailureMessage` here is fine because we always expect this to succeed in order to match the output)

integration/gitopscreateprofile_test.go
```diff
@@ -8,21 +8,25 @@ import (
...
+	. "github.com/weaveworks/eksctl/integration/runner"
 )
 
 var _ = Describe("(Integration) generate profile", func() {
 
 	Describe("when generating a profile", func() {
 		It("should write the processed repo files in the supplied directory", func() {
 
-			eksctlSuccess("generate", "profile",
+			cmd := eksctlExperimentalCmd.WithArgs(
```
errordeveloper: Need to add this above:
```
			if clusterName = "" {
				clusterName = "amazing-testing-gopher"
			}
			if region = "" {
				region = "us-west-2"
			}
```

rndstr: could we initialize the `clusterName` at the beginning of the integration test for all the suites at once?

rndstr: could we initialize the clusterName at the beginning of the integration test for all the suites at once?

errordeveloper: Yes, good idea. I just thought it maybe better to do that in a separate PR.

marccarre: Yes, please, I had troubles with hardcoded-values mutating the state of the tests as they run, like the above ones, when running tests on an existing cluster, see also: https://github.com/weaveworks/eksctl/pull/1148/commits/df6773f4045218ccb4bf32321d8fc87b6bc61048.

integration/runner/runner.go
errordeveloper: Good idea, thanks!

integration/runner/runner.go
errordeveloper: Hm.. not exactly, we could be expecting a failure with particular output too.

errordeveloper:
Merging on the basis of @rndstr approval.

PR #1186, opened 2019-08-22
Avoid `nodeGroups[0].name must be set` validation error

### LINKED ISSUE #1164: get nodegroups command no longer display all nodegroups
**What happened?**
You can no longer get all nodegroups via a single command. You can only get individual nodegroup settings. 

Expect to display all nodegroups without entering a nodegroup name, which was capable in v0.3.1. At least another flag/option to display all groups. 

**How to reproduce it?**
eksctl get nodegroup --cluster amg-non-prod-use1
[✖]  nodeGroups[0].name must be set

**Versions**
```
eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.4.0"}
```

errordeveloper:
### Description

This fixes #1164 and related cases of the same error. This is a regression that was introduced by  #1116 and #1124.

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/configfile_test.go
+++ b/pkg/ctl/cmdutils/configfile_test.go
@@ -123,11 +123,95 @@ var _ = Describe("cmdutils configfile", func() {
 					ProviderConfig:    &api.ProviderConfig{},
 				}
 
-				Expect(NewCreateClusterLoader(cmd, nil).Load()).To(Succeed())
+				Expect(NewCreateClusterLoader(cmd, NewNodeGroupFilter(), nil, true).Load()).To(Succeed())
 				cfg := cmd.ClusterConfig
 				Expect(cfg.VPC.NAT.Gateway).To(Not(BeNil()))
 				Expect(*cfg.VPC.NAT.Gateway).To(Equal(natTest.expectedGateway))
 			}
+		})
+
+		It("loader should handle named and unnamed nodegroups without config file", func() {
+			unnamedNG := api.NewNodeGroup()
+
+			namedNG := api.NewNodeGroup()
+			namedNG.Name = "ng-1"
+
+			loaderParams := []struct {
+				ng               *api.NodeGroup
+				withoutNodeGroup bool
+			}{
+				{unnamedNG, false},
+				{unnamedNG, true},
+				{namedNG, false},
+				{namedNG, false},
+			}
+
+			for _, loaderTest := range loaderParams {
+				cmd := &Cmd{
+					CobraCommand:   newCmd(),
+					ClusterConfig:  api.NewClusterConfig(),
+					ProviderConfig: &api.ProviderConfig{},
+				}
+
+				ngFilter := NewNodeGroupFilter()
+
+				Expect(cmd.ClusterConfig.NodeGroups).To(HaveLen(0))
+
+				Expect(NewCreateClusterLoader(cmd, ngFilter, loaderTest.ng, loaderTest.withoutNodeGroup).Load()).To(Succeed())
+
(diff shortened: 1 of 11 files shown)


errordeveloper:
I'm, going to have a go at adding some unit test to cover this, possibly also integration tests.

pkg/ctl/get/nodegroup.go
```diff
@@ -62,6 +54,20 @@ func doGetNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, params *getCmdParams)
 		ng.Name = cmd.NameArg
 	}
 
+	if ng.Name == "" {
+		// avoid validation errors when listing all nodegroups
+		cfg.NodeGroups = nil
```
cPu1: While this skips the validation for node groups, it'd be ideal if we could avoid calling `ValidateClusterConfig` for this usage of the command as there's no node group input to validate here.

errordeveloper: At the moment the idea is that `ValidateClusterConfig` is mean to be call in all instances, whether `ClusterConfig` instance was constructed by the command, or parsed from a config file. I would rather review how it's constructed. I'll see if there is an easy fix to this that is a more coherent.

errordeveloper:
@cPu1 what do you make of most recent commit?

cPu1:
>@cPu1 what do you make of the most recent commit?

LGTM

The unit test is failing because the change to `NewCreateClusterLoader` isn't updated in `configfile_test.go:126`

errordeveloper:
Yes, I'll fix the unit tests and try to add some new ones also....

errordeveloper:
I'd like to add `eksctl get ng` to integration suite as well.

pkg/ctl/cmdutils/configfile_test.go
```diff
@@ -123,11 +123,95 @@ var _ = Describe("cmdutils configfile", func() {
...
+					Expect(cmd.ClusterConfig.NodeGroups[0]).To(Equal(loaderTest.ng))
+				}
+
+				_, err := cmd.NewCtl()
+				Expect(err).ToNot(HaveOccurred())
+			}
+		})
+
+		It("loader should handle nodegroup exclusion with config file", func() {
+			loaderParams := []struct {
```
cPu1: Alternative way:

```golang
		loaderParams := []struct {
			configFile     string
			nodeGroupCount int
		}{
			{"01-simple-cluster.yaml", 1, true},
			{"02-custom-vpc-cidr-no-nodes.yaml", 0, true},
			{"03-two-nodegroups.yaml", 2, true},
			{"05-advanced-nodegroups.yaml", 3, true},
			{"07-ssh-keys.yaml", 5, true},
		}
		
		for _, loaderTest := range loaderParams {
			cmd := &Cmd{
				CobraCommand:      newCmd(),
				ClusterConfigFile: filepath.Join(examplesDir, loaderTest.configFile),
				ClusterConfig:     api.NewClusterConfig(),
				ProviderConfig:    &api.ProviderConfig{},
			}
		
			ngFilter := NewNodeGroupFilter()
		
			for _, withoutNodeGroup := range []bool{true, false} {
				Expect(NewCreateClusterLoader(cmd, ngFilter, nil, withoutNodeGroup).Load()).To(Succeed())
		
				Expect(ngFilter.ExcludeAll).To(Equal(withoutNodeGroup))
		
				Expect(cmd.ClusterConfig.NodeGroups).To(HaveLen(loaderTest.nodeGroupCount))
		
				_, err := cmd.NewCtl()
		
			}
		}
```

errordeveloper: yeah, I thought about it, but didn't seem to make that much difference

PR #1200, opened 2019-08-27
Build improvements

### LINKED ISSUE #1145: automate tagging of build image
Git hashes would be something to investigate. 

There maybe other solutions.

### LINKED ISSUE #1202: build cache misses
We have seen a few instances where build cache was missed on PR builds for external contributors:

- #1158 (https://circleci.com/gh/weaveworks/eksctl/4152)
- #1201 (https://circleci.com/gh/weaveworks/eksctl/4288)

errordeveloper:
### Description

Basic summary:
- streamlined CircleCI configuration
- store dependency cache in the build image
- 2 minute speed-up of CircleCI builds
- Fix #1145
- Fix #1202

See proposal document within for more details.

### Checklist

- [x] Manually tested


Dockerfile
```diff
@@ -1,3 +1,48 @@
-FROM scratch
-CMD eksctl
-COPY --from=weaveworks/eksctl-builder:latest /out /
+# Make sure to bump the version of EKSCTL_DEPENDENCIES_IMAGE if you make any changes
+# to this file
```
errordeveloper: Will need to update this and other files. Now you need to run `make update-build-image`, we might want to add add `make publish-build-image` as well.

Makefile
```diff
@@ -1,14 +1,42 @@
...
+git_commit = $(shell git describe --dirty --always)
+
 git_toplevel := $(shell git rev-parse --show-toplevel)
 version_pkg := github.com/weaveworks/eksctl/pkg/version
 
-# The dependencies version should be bumped every time the build dependencies are updated
-EKSCTL_DEPENDENCIES_IMAGE ?= weaveworks/eksctl-build:deps-0.13
-EKSCTL_BUILDER_IMAGE ?= weaveworks/eksctl-builder:latest
-EKSCTL_IMAGE ?= weaveworks/eksctl:latest
+build_image_input := Dockerfile install-build-deps.sh go.mod go.sum
```
errordeveloper: It maybe a good idea to add `Makefile` here in the future.

Makefile
```diff
@@ -174,34 +202,64 @@ site/content/usage/20-schema.md: $(call godeps,cmd/schema/generate.go)
...
+
+.PHONY: update-build-image-manifest
+update-build-image-manifest: ## Update build image manifest and commits the changes
+	$(update_build_image_manifest)
+	git commit --quiet .build_image_manifest --message 'Update build image manifest'
+	$(info build image name is $(build_image_name))
+
+.PHONY: build-image
+build-image: check-build-image-manifest-up-to-date ## Build the build image that has all of external dependencies
+	-docker pull $(build_image_name)
```
D3nn: I know I'm not on the reviewer's list but looks like an extra `-` here...

Dockerfile
```diff
@@ -1,3 +1,48 @@
...
+RUN ./install-build-deps.sh && \
+    go install github.com/goreleaser/goreleaser && \
+    go build github.com/kubernetes-sigs/aws-iam-authenticator/cmd/aws-iam-authenticator && \
+    rm -rf /root/.cache/go-build /go/pkg/mod
+
+
+
+# Runtime dependencies. Build the root filesystem of the eksctl image at /out
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb --root /out \
```
D3nn: I realize it's a bit easier to read this way, but installing all the apks in one RUN command would make the image even smaller.

Makefile
errordeveloper: The list of reviewers is just a suggestion :)

You mean `-docker`? That's make syntax for ignoring errors.

Dockerfile
errordeveloper: Yes, however we don't aim to make this small, we actually optimise for layer cache here, but the resulting image gets squashed into a small number of layers.

Makefile
```diff
@@ -174,34 +202,64 @@ site/content/usage/20-schema.md: $(call godeps,cmd/schema/generate.go)
...
 
 .PHONY: eksctl-image
-eksctl-image: eksctl-deps-image## Create the eksctl image
-	time docker run -t --name eksctl-builder -e TEST_TARGET=$(TEST_TARGET) \
-	  -v "$(git_toplevel)":/src -v "$$(go env GOCACHE):/root/.cache/go-build" -v "$$(go env GOPATH)/pkg/mod:/go/pkg/mod" \
-          $(EKSCTL_DEPENDENCIES_IMAGE) /src/eksctl-image-builder.sh || ( docker rm eksctl-builder; exit 1 )
-	time docker commit eksctl-builder $(EKSCTL_BUILDER_IMAGE) && docker rm eksctl-builder
-	docker build --tag $(EKSCTL_IMAGE) .
+eksctl-image: intermediate-image ## Build the eksctl image that has release artefacts and no build dependencies
+	printf 'FROM scratch\nCMD eksctl\nCOPY --from=%s /out /' $(intermediate_image_name) \
```
errordeveloper: @D3nn this is the inline `Dockerfile` for the end image.

errordeveloper:
Looks like remote Docker feature in Circle actually uses Docker on a different machine, seems like regular worker is in AWS and the remote Docker host is on GCP. I'll try caching dependencies inside the image, otherwise will have to revert back to using those Ubuntu 16.04 machines with outdated everything...

errordeveloper:
This works nicely in GitHub Actions now, we will need to see if we can get CircleCI setup upgraded to have more RAM. I'd rather not suggest we switch to GitHub Action in one go and have it in parallel for a little while.

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+
+At the start of the project we've picked CircleCI as it appears to be a default choice at Weaveworks.
+
+The original solution involved:
+- running `docker build` to create a build image
+- use build image to run tests produce final image
+
+Publishing build and final images hasn't been automated, but some users requested it.
+
+When we switched [to modules](https://github.com/weaveworks/eksctl/pull/917), we have increased the
```
gemagomez: s/to modules/to go modules ?

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+compexity of how build image worked. It's hard for a new person to understand it, e.g. for a new
```
gemagomez: why is it hard, what is required now that was not there before (brief explanation, doesn't need to go into deep details)

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+contributor.
+
+We currently build images in CI and run tests as part of that, we should only run tests as a primary
+task.
```
gemagomez: Do you mean that build and running tests should be two separate build steps? Rather than one combined one?

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+
+Downside of current setup:
+
+- CircleCI requires delegation of access to GitHub
+  - for releases we use a token
+  - we don't have a way to automate workflows that require pushing to the repo
+- CircleCI cache 
+  - it is not very fast, it turns out pulling image with cached dependencies is as fast
+  - it is very specifig to CircleCI, not portable
+  - we do not have a way to controll it
```
gemagomez: we do not have a way to control what? and how does the new solution address this?

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+task.
+
+Downside of current setup:
+
+- CircleCI requires delegation of access to GitHub
+  - for releases we use a token
+  - we don't have a way to automate workflows that require pushing to the repo
+- CircleCI cache 
+  - it is not very fast, it turns out pulling image with cached dependencies is as fast
+  - it is very specifig to CircleCI, not portable
```
gemagomez: what parts are specific to circle ci?

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+We currently build images in CI and run tests as part of that, we should only run tests as a primary
+task.
+
+Downside of current setup:
+
+- CircleCI requires delegation of access to GitHub
+  - for releases we use a token
+  - we don't have a way to automate workflows that require pushing to the repo
+- CircleCI cache 
+  - it is not very fast, it turns out pulling image with cached dependencies is as fast
```
gemagomez: what does not being very fast mean? compared to what? I am not sure i understand this sentence as a whole.

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+  - it is very specifig to CircleCI, not portable
+  - we do not have a way to controll it
+- the task of building images should not be mixed with the task of running test
```
gemagomez: agreed.

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+- images are not being automatically pushed to a registry
+- update of build image is manual
```
gemagomez: how does this change in the new proposed solution?

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposed Improvements
```
gemagomez: what alternatives have been considered? Have you looked into travis/circleci/github actions? Are there any other automation services that are worth considering?

I'd like to see how they compare to each other in terms of what we are trying to achieve. I.e. advantages / disadvantages.

docs/proposal-006-build-improvements.md
```diff
@@ -0,0 +1,55 @@
...
+
+This proposal [#1200](https://github.com/weaveworks/eksctl/pull/1200) aims to streamline the following
```
gemagomez: I don't think we can review this without the analysis requested before, unless you want to use it as an experiment only.

docs/proposal-006-build-improvements.md
errordeveloper: Yes, especially as there is overhead associated with building the image and requirement of being able to run any `docker` commands, vs just running a Docker image natively.

docs/proposal-006-build-improvements.md
errordeveloper: this section is about cache, we don't have way to control cache in CircleCI

docs/proposal-006-build-improvements.md
errordeveloper: as above, this is under 'CircleCI cache'

docs/proposal-006-build-improvements.md
errordeveloper: Right now we do this:

- restore cache from previous build
- mount cache as docker volume
- run tests and build new binary

In my experiment it's as fast (or faster) to do the following:

- pull docker image
- run tests and build new binary

So, `docker pull` for a large image that contains all cached objects takes as long (or less time) then using CircleCI cache feature. There are other benefits to this, as we can tell exactly what's being cached and can reproduce it anywhere else.

docs/proposal-006-build-improvements.md
errordeveloper: The version is determined based on git object hashes, but publishing of the image is not yet automated in the PR, that can be done easily later on.

docs/proposal-006-build-improvements.md
errordeveloper: At the moment I'm trying to evolve what we have in CircleCI, and enable GitHub Actions in parallel.

docs/proposal-006-build-improvements.md
errordeveloper: Well, there two main parts:

- streamline what we have in CircleCI (which has never been review previously)
- enable GitHub Actions as an experiment (as it turned out to be easy enough)

docs/proposal-006-build-improvements.md
errordeveloper: It's just complicated. Try reading what we have.

https://github.com/weaveworks/eksctl/blob/950f9bc695234107725234b9e1a9c9d2ee54e51f/.circleci/config.yml#L3-L38

https://github.com/weaveworks/eksctl/blob/950f9bc695234107725234b9e1a9c9d2ee54e51f/Makefile#L188-L204


And what becomes of that section of `.circleci/config.yml` on this branch:

https://github.com/weaveworks/eksctl/blob/2bae42e5bcd8991ed734851144b12e2bd5af68a8/.circleci/config.yml#L3-L16

Which now points you at much easier to understand make target:

https://github.com/weaveworks/eksctl/blob/2bae42e5bcd8991ed734851144b12e2bd5af68a8/Makefile#L68-L73

There is still some complexity in how the build image is put together, but that's in a separate [`Makefile.docker`](https://github.com/weaveworks/eksctl/blob/2bae42e5bcd8991ed734851144b12e2bd5af68a8/Makefile.docker) and you don't have to dig into it just to figure out why your tests may have failed in CI and passed locally.

docs/proposal-006-build-improvements.md
gemagomez: Ok, so what happens to the cache in the new proposed solution?

docs/proposal-006-build-improvements.md
gemagomez: Please, add this explanation to the doc, sounds good.

docs/proposal-006-build-improvements.md
errordeveloper: Also, I have used TravisCI many times in the past, it is largely the same as CricleCI, except that the UI is a little cleaner. There is no significant proposition for a swap. GitHub Action is being proposed as it offered many additional features that go beyond just running builds.

docs/proposal-006-build-improvements.md
errordeveloper: It will be stored inside of the build image.

docs/proposal-006-build-improvements.md
marccarre: Bonus points for:

- mentioning build time before this change,
- mentioning time after this change,
- mentioning this in `docs/proposal-006-build-improvements.md`.

errordeveloper:
I have managed to bypass the need for upgrading CircleCI setup (at least for the time being).

New timings are:

- image pull - 01:17
- make test - 06:11
- total - 07:44

I still need to add `make build` to this, which will add another minute, perhaps.

Current timings on master are between 10 & 11 minutes in total, e.g.:

- retore cache - 02:00
- make image - 04:18
- save cache - 03:20

I think image build on master is faster then `make test`, as some of the dependencies are pre-compiled and test are cached also. Personally, I am not very keen on test being cached, as it can prevent flakes from surfacing.

errordeveloper:
Ok, perhaps, but we haven't made that decision here. We are only adding GHA
as an experiment, in parallel to CircleCI. We can get back to this document
once we have more info to make a decision.

docs/proposal-006-build-improvements.md
errordeveloper: Ok, perhaps, but we haven't made that decision here. We are only adding GHA
as an experiment, in parallel to CircleCI. We can get back to this document
once we have more info to make such decision.

docs/proposal-006-build-improvements.md
gemagomez: Please mention that and resolve this comment.

docs/proposal-006-build-improvements.md
gemagomez: Could you reflect this in the document and resolve this comments? Sounds very good.

docs/proposal-006-build-improvements.md
gemagomez: Can you add it to a "Next steps" section in this document?

docs/proposal-006-build-improvements.md
gemagomez: I think it is worth comparing them so we can (when the time comes) justify the decision we are making of moving to GHA, if it turns out to be the consensus then.

docs/proposal-006-build-improvements.md
gemagomez: So maybe we should move to GHAs for the things that are not automated yet, and then consider, based on that whether it is worth moving the rest? Please, add your previous explanation to the doc, it is useful.

gemagomez:
> Personally, I am not very keen on test being cached, as it can prevent flakes from surfacing.

Agreed. Let's not do that.

gemagomez:
I have a more generic question for the analysis doc. When you separate build from test, how are you going to know that you are testing the right image when it comes to downloading it and testing?

errordeveloper:
> When you separate build from test, how are you going to know that you are testing the right image when it comes to downloading it and testing?

It's configured explicitly in the CircleCI config at the moment, and for any given revision it can be determined based on git tree hashes. You can read `Makefile.docker` to find out how it works, if comments are not clear there - please let me know, I'm not very keen on duplicating the details in another file. In short, we have ` .build_image_manifest` that tracks git tree hashes of each of the files that determine what goes into an image, that file is checked in and we use `git ls-tree --full-tree @ -- .build_image_manifest` to get the git tree hash of that file and use that as image tag.
So, given image tag you can use `git show` to find out what was in `.build_image_manifest`, and use `git show` again for each of the hashes that file returns.

E.g. given `weaveworks/eksctl-build:b0aa27c30e3f533cdc861a471a0f01f624c7f3d5`:
```
$ git show b0aa27c30e3f533cdc861a471a0f01f624c7f3d5
100644 blob 47cb6bf90db7b43772d0a77f4082d8552d0724fc    Dockerfile
100644 blob 5a627d2f0babe888166ab54168f5d11cdaa11f8e    go.mod
100644 blob 68e10db9f70d4e1d0e0159dba3cdb57669be64c0    go.sum
100755 blob 180a7df24779458c0bb748261c6dc4ed78185c86    install-build-deps.sh
$ git show 47cb6bf90db7b43772d0a77f4082d8552d0724fc | head -5
# Make sure to run the following commands after changes to this file are made:
# `make -f Makefile.docker update-build-image-manifest && make -f Makefile.docker push-build-image`

# This digest corresponds to golang:1.12.9-alpine3.10
FROM golang@sha256:e0660b4f1e68e0d408420acb874b396fc6dd25e7c1d03ad36e7d6d1155a4dff6 AS base
$
```

errordeveloper:
cc @stefanprodan

errordeveloper:
@marccarre I've updated the proposal doc, and rebased this - could you please take another look?

docs/proposal-006-build-improvements.md
errordeveloper: Perhaps, let's see how well it works for us :)

docs/proposal-006-build-improvements.md
errordeveloper: There is such section at the bottom already.

docs/proposal-006-build-improvements.md
errordeveloper: I'd rather avoid this for now, the document was meant to be lightweight.

.github/workflows/commit.yaml
```diff
@@ -0,0 +1,34 @@
+name: Commit Workflow
+on: [push]
+
+jobs:
+  build:
+    name: Test & Build
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - uses: docker://weaveworks/eksctl-build:19ff80e010d2fdb7be3c44fc6491e13395a682ad
```
marccarre: Nit: do you know if there is a way to factorise `docker://weaveworks/eksctl-build:19ff80e010d2fdb7be3c44fc6491e13395a682ad`?

.circleci/config.yml
```diff
@@ -1,122 +1,77 @@
...
-  make-eksctl-image:
-    # use machine build, as otherwise we cannot use make and have to duplicate makefile
+  test-and-build:
     machine:
       image: ubuntu-1604:201903-01
     environment:
-        GO_VERSION: 1.12.6
-        GOCACHE: /home/circleci/.cache/go-build/
-        GOPATH: /home/circleci/go
+      BUILD_IMAGE: weaveworks/eksctl-build:19ff80e010d2fdb7be3c44fc6491e13395a682ad
```
marccarre: Nit: do you know if there is a way to factorise `BUILD_IMAGE: weaveworks/eksctl-build:19ff80e010d2fdb7be3c44fc6491e13395a682ad` in order to have it defined at a higher-level in the YAML file?

marccarre:
This looks reasonable to me. The only way is forward. And we can always refine later.

.github/workflows/commit.yaml
errordeveloper: Not sure. For now `make -f Makefile.docker update-build-image-manifest` handles updating this.

errordeveloper: Some of the fields in github configs accept shell syntax (like `foo-$(echo 123)`), but that didn't work here.

PR #1215, opened 2019-08-28
Fix error messages in nodegroup IAM config validation, add tests

### LINKED ISSUE #1214: IAM add-on validation errors are incorrect for some of the add-ons
We print correct path here:

```
		if IsEnabled(ng.IAM.WithAddonPolicies.ExternalDNS) {
			return fmt.Errorf("%s.withAddonPolicies.externalDNS cannot be set at the same time", p)
		}
```

But not here:
```
		if IsEnabled(ng.IAM.WithAddonPolicies.ImageBuilder) {
			return fmt.Errorf("%s.imageBuilder cannot be set at the same time", p)
		}
```

errordeveloper:
### Description

Fix #1214

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation_test.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation_test.go
@@ -6,7 +6,7 @@ import (
 )
 
 var _ = Describe("ClusterConfig validation", func() {
-	Describe("nodeGroups", func() {
+	Describe("nodeGroups[*].name", func() {
 		var (
 			cfg *ClusterConfig
 			err error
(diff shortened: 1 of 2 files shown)


pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 var _ = Describe("ClusterConfig validation", func() {
-	Describe("nodeGroups", func() {
+	Describe("nodeGroups[%d].name", func() {
```
cPu1: Minor nit: since this doesn't actually format it, should we use `[*]` to refer to the index?

```suggestion
	Describe("nodeGroups[*].name", func() {
```

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -46,6 +46,105 @@ var _ = Describe("ClusterConfig validation", func() {
 		})
 	})
 
+	Describe("nodeGroups[%d].iam", func() {
```
cPu1: Same as above.

PR #1227, opened 2019-08-29
Enforce --git-email flag for `eksctl gitops apply`

### LINKED ISSUE #1209: `gitops apply`should enforce a --git-email argument
**What happened?**
`eksctl apply gitops` allows for an empty `--git-email`. This leads to commit messages from Flux without an author.

`eksctl apply gitops` fails in no `--git-email` is provided

cPu1:
### Description

Closes #1209 

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/gitops/apply.go
+++ b/pkg/ctl/gitops/apply.go
@@ -50,9 +50,13 @@ func applyGitops(cmd *cmdutils.Cmd) {
 		fs.StringVar(&opts.gitPrivateSSHKeyPath, "git-private-ssh-key-path", "",
 			"Optional path to the private SSH key to use with Git, e.g.: ~/.ssh/id_rsa")
 		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "name of the EKS cluster to add the nodegroup to")
-		_ = cobra.MarkFlagRequired(fs, "quickstart-profile")
-		_ = cobra.MarkFlagRequired(fs, "git-url")
-		_ = cobra.MarkFlagRequired(fs, "cluster")
+
+		requiredFlags := []string{"quickstart-profile", "git-url", "cluster", "git-email"}
+		for _, f := range requiredFlags {
+			if err := cobra.MarkFlagRequired(fs, f); err != nil {
+				panic(fmt.Sprintf("unexpected error: %v", err))
+			}
+		}
 
 		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)
 		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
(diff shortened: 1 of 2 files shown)


pkg/ctl/gitops/apply.go
```diff
@@ -50,9 +50,13 @@ func applyGitops(cmd *cmdutils.Cmd) {
...
 			"Optional path to the private SSH key to use with Git, e.g.: ~/.ssh/id_rsa")
 		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "name of the EKS cluster to add the nodegroup to")
-		_ = cobra.MarkFlagRequired(fs, "quickstart-profile")
-		_ = cobra.MarkFlagRequired(fs, "git-url")
-		_ = cobra.MarkFlagRequired(fs, "cluster")
+
+		requiredFlags := []string{"quickstart-profile", "git-url", "cluster", "git-email"}
+		for _, f := range requiredFlags {
+			if err := cobra.MarkFlagRequired(fs, f); err != nil {
+				panic(fmt.Sprintf("unexpected error: %v", err))
```
errordeveloper: Let's make this function return the error. I also wanted to point out that we don't normally use `cobra.MarkFlagRequired`, it's on config loader and validation to determine what's required; could you please add a comment to suggest that this should get moved into a loader going forward?

cPu1: >Let's make this function return the error

The rationale for using a panic is that `MarkFlagRequired` throws an error only if the supplied flag isn't known/registered, which can only happen if there's a bug in our code.

cPu1: >could you please add a comment to suggest that this should get moved into a loader going forward

I can add that, but it does mean that we'll have to write the validation ourselves.

errordeveloper: I see, so perhaps let's just double-check flag exists and we can safely ignore the error.

```Go
if cmd.Flag(f) != nil {
   _ = cobra.MarkFlagRequired(fs, f)
}
```

errordeveloper: > I can add that, but it does mean that we'll have to write the validation ourselves.

Let's leave it until we get to implementing Flux configuration fields in `ClusterConfig` :)

cPu1: >I see, so perhaps let's just double-check flag exists and we can safely ignore the error.

That'd assume that the implementation of `MarkFlagRequired` throws an error only on this particular condition (it does now but, although unlikely, it may not later). It does the same check and the error check guards us against possible misuse of that function in the future. It also creates a confusion for someone new to this part of the code as to why the error is being ignored.

errordeveloper: I just think we shouldn't panic, print the error and `os.Exit(1)` if you must.

cPu1: Okay, I'll make the change.

Issue #1212, opened 2019-08-28
`gitops apply` git urls should be ssh


gemagomez:
**What happened?**

Tried to run `gitops apply` with http urls as parameters and it fails. 

**What you expected to happen?**

`eksctl` should enforce the right format of the git parameters and return a meaningful error if the format is not met.

**How to reproduce it?**

Run `eksctl gitops apply` with git urls in http format, rather than ssh format.

errordeveloper:
Generally speaking, I think any valid git URL should work. However, I am aware that due to a number of reasons we've been forcing Flux user to requite SSH URLs.

errordeveloper:
So are we saying that we will have to allow only SSH URLs for now? There is a number of implications as in some organisations use of SSH ports is not permitted.

errordeveloper:
@gemagomez do you have more context of what exactly didn't work? Just to be clear...

2opremio:
This applies to `install flux` as well

cPu1:
`eksctl gitops apply` delegates to the git CLI for cloning the repo, which does support HTTPS URLs by prompting you to enter your credentials. If you have 2FA on GitHub enabled (we all do since joining the weaveworks org requires having it enabled), your regular credentials won't work and you'd need to create a [personal access token](https://github.com/settings/tokens) and use it as your username (leaving the password blank).

Alternatively, you could add this directive to your gitconfig to rewrite HTTPS URLs to SSH:
```
url "git@github.com:"]
        insteadOf = https://github.com/
```

cPu1:
The UX for using HTTPS URLs in `eksctl gitops apply` though is cumbersome as you'd be prompted for credentials for every remote git operation (clone, push etc)

I tried `eksctl gitops apply` with an HTTPS URL and it successfully ran to completion.

2opremio:
> I tried `eksctl gitops apply` with an HTTPS URL and it successfully ran to completion.

Yeah, but Flux won't work because the credentials are not currently propagated to it.

It is theoretically possible to make everything work with `https` URLS but, as I mentioned privately, right now:

1. I private https repos, Git will ask you for credentials to push to the repo, which will cause the operation to time out.
2. The http credentials won't be passed to Flux and it won’t work at all if the repo is private. Autoupdates definitely won't work even if the repo is public and I am not sure if it will be able to run at all. I will work with a public repo _if_ Flux is run in readonly mode (which still don't currently have support for).

[...]

PR #1231, opened 2019-08-30
Restructure Git-related tests

marccarre:
### Description

As a preparation for the work on #1212, I'd like to "clean" / restructure these tests a bit.
Given the `diff` isn't easily reviewable, I'd rather have this in a separate PR.
You can find the description of the various transformations done in the commit messages.
Nothing else was changed so I'd recommend the reviewers to not make your eyes bleed unnecessarily, and not to look too much at the `diff`.

### DIFF
--- a/pkg/git/git_suite_test.go
+++ b/pkg/git/git_suite_test.go
@@ -1,4 +1,4 @@
-package git
+package git_test
 
 import (
 	"testing"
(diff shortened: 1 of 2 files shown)


marccarre:
Merged without any review, as trivial, and needed to move fast.

PR #1238, opened 2019-08-30
Update generated code in `pkg/nodegroup/` and improve the build

### LINKED ISSUE #1229: Missing m5.16xlarge from max_pods.map due to which kubelet fails to start
**What happened?**
A description of actual behavior (with error messages).
Missing m5.16xlarge from max_pods.map due to which kubelet fails

Missing m5.16xlarge instance type from the max_pods.map is causing kubelet daemon not able to start issue. 

**How to reproduce it?**
Include the steps to reproduce the bug
Launch m5.16xlarge instance type which would fail to join the cluster. If you check kubelet logs, you would see below error:

```
Aug 29 18:58:50 ip-192-168-5-5.us-west-2.compute.internal kubelet[25640]: F0829 18:58:50.172283   25640 server.go:147] invalid argument "" for "--max-pods" flag: strconv.ParseInt: p
```

errordeveloper:
### Description

- Fix #1229 
- Improve makefile
- Stop relying on asset permissions

### Checklist

- [x] Manually tested


errordeveloper:
@cPu1 thanks!

pkg/nodebootstrap/userdata.go
```diff
@@ -51,20 +46,19 @@ func addFilesAndScripts(config *cloudconfig.CloudConfig, files configFiles, scri
 				Path: dir + fileName,
 			}
 			if file.isAsset {
-				data, info, err := getAsset(fileName)
+				data, err := getAsset(fileName)
 				if err != nil {
 					return err
 				}
 				f.Content = data
-				f.Permissions = fmt.Sprintf("%04o", uint(info.Mode()))
```
errordeveloper: We didn't actually use it for anything, as it posed a challenge on the build, I figured there is no point of keeping this (untested) feature.

pkg/nodebootstrap/assets.go
```diff
@@ -86,7 +86,7 @@ func _10EkscltAl2Conf() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "10-eksclt.al2.conf", size: 999, mode: os.FileMode(420), modTime: time.Unix(1, 0)}
+	info := bindataFileInfo{name: "10-eksclt.al2.conf", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
```
cPu1: Where do we use `asset.info`? I could only find `asset.bytes` being used but not `info`

errordeveloper: It was used It's no longer used for setting remote permissions, but we didn't make use of it in a long time, hence it's being killed in this PR as it made the build messy - as local permissions cannot be relied on... (well git track's owner's bit, but not group's or others').

cPu1: I see, thanks for the info. We should plan to remove having this field completely.

cPu1: Looks like go-bindata doesn't support making these fields optional.

Issue #1240, opened 2019-08-30
Clean up cmdutils package


2opremio:
In particular the `cmdutils/cmdutils.go` contains flag parsing helpers which:

1. Should probably belong to the ctl package where they are used
2. Uses constant strings for flags (e.g. "--name" "--cluster") , which can get out of sync with the actual flags being declared, see

errordeveloper:
The idea is that there are _many_ flags in common. If you don't put them in one place, they again get out of sync different ways.

2opremio:
That's not the case for some commands, and doesn't conflict with using constants  instead of string literals.

PR #1241, opened 2019-08-30
 Fix --cluster parameter validation for gitops and install commands

### LINKED ISSUE #1223: `gitops apply` should allow providing a config file instead of --cluster
```
$ EKSCTL_EXPERIMENTAL=true ./eksctl gitops apply -f ~/fons-cluster-1.yaml --quickstart-profile git@github.com:weaveworks/eks-gitops-example --git-url url  --git-email email
Error: required flag(s) "cluster" not set
Usage: eksctl gitops apply [flags]

General flags:
      --quickstart-profile string         name or URL of the Quick Start profile. For example, app-dev.
      --git-url string                    URL for the git repository that will contain the cluster components
      --git-branch string                 Git branch (default "master")
      --output-path string                Path to directory where the GitOps repo will be cloned (default "./")
      --git-user string                   Username to use as Git committer (default "Flux")
      --git-email string                  Email to use as Git committer
      --git-private-ssh-key-path string   Optional path to the private SSH key to use with Git, e.g.: ~/.ssh/id_rsa
      --cluster string                    name of the EKS cluster to add the nodegroup to
  -r, --region string                     AWS region
  -f, --config-file string                load configuration from a file (or stdin if set to '-')

AWS client flags:
  -p, --profile string   AWS credentials profile to use (overrides the AWS_PROFILE environment variable)

Common flags:
  -C, --color string   toggle colorized logs (valid options: true, false, fabulous) (default "true")
  -h, --help           help for this command
  -v, --verbose int    set log level, use 0 to silence, 4 for debugging and 5 for debugging with AWS debug logging (default 3)

Use 'eksctl gitops apply [command] --help' for more information about a command.


required flag(s) "cluster" not set
```

2opremio:
We were not validating the flags properly in the `gitops apply` and `install flux` commands:

1. `gitops apply` enforced `--cluster` (which shouldn't since the user can provide a config file)
2. `install flux` declared a `--cluster` parameter, but validated whether `--name` was present.

This PR doesn't take any measures towards preventing this from happening in the future. I have created https://github.com/weaveworks/eksctl/issues/1240 for that purpose.

Fixes #1223

### DIFF
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -126,8 +126,13 @@ func NewGitopsMetadataLoader(cmd *Cmd) ClusterConfigLoader {
 	l := newCommonClusterConfigLoader(cmd)
 
 	l.validateWithoutConfigFile = func() error {
+		meta := l.ClusterConfig.Metadata
+		if meta.Name == "" {
+			return ErrMustBeSet("--cluster")
+		}
 		return nil
 	}
+
 	return l
 }
 
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -121,8 +121,8 @@ func NewMetadataLoader(cmd *Cmd) ClusterConfigLoader {
 	return l
 }
 
-// NewGitopsMetadataLoader handles loading of clusterConfigFile vs using flags for gitops commands
-func NewGitopsMetadataLoader(cmd *Cmd) ClusterConfigLoader {
+// NewGitopsApplyLoader handles loading of clusterConfigFile vs using flags for gitops commands
+func NewGitopsApplyLoader(cmd *Cmd) ClusterConfigLoader {
 	l := newCommonClusterConfigLoader(cmd)
 
 	l.validateWithoutConfigFile = func() error {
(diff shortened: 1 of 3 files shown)


pkg/ctl/cmdutils/configfile.go
```diff
@@ -126,8 +126,28 @@ func NewGitopsMetadataLoader(cmd *Cmd) ClusterConfigLoader {
...
+			return ErrMustBeSet("--cluster")
+		}
+		return nil
+	}
+
+	return l
+}
+
+// NewInstallMetadataLoader handles loading of clusterConfigFile vs using flags for install commands
+func NewInstallMetadataLoader(cmd *Cmd) ClusterConfigLoader {
```
errordeveloper: Let's call it `NewInstallFluxLoader`, just like for each of other commands that don't use supposed-to-be-generic metadata loader... This package does need some thought going forward, it's a half-baked idea of a shared abstraction that clearly has a few issues.

2opremio: Sure, I will rename the GitOps one too for consistency

2opremio:
Fixed the nit. Merging without an explicit approval to move thing forward fast.

PR #1243, opened 2019-08-30
Create branch if none existed when cloning gitops repo

### LINKED ISSUE #1137: `eksctl install flux` fails when you have a brand new github repo
**What happened?**
I created a brand new github repo called `my-eks-cluster` and then ran `eksctl install flux --name=kalbir-test-1 --git-url=git@github.com:kalbir/my-eks-cluster.git --git-email=kalbir@gmail.com`. 

I got the following error:

Screenshot 2019-08-09 at 10 20 59

It looks like when you create a new repo on github *without any files in it* (ie. no readme, no license) it does not set up any branches for that repo and this means that the initial clone in `eksctl install flux` fails.

`eksctl install flux` to install flux and add the config to my git repo. I think I would be expecting at the end of the command there is a master branch in the repo and the relevant files are committed to the master branch. So probably eksctl should create the master branch somewhere in this process. 

**How to reproduce it?**

1. Create a new repo, "foo", on github, *without a readme or license*. It should look like this after creation:

Screenshot 2019-08-12 at 17 25 49

2. Run `eksctl install flux --name=<clustername> --git-url=git@github.com:<user>/foo.git --git-email=<email>`

Mac OS 10.14

**Versions**

```
└>> eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.3.0"}
```

```
└>> kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:38:32Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.8-eks-a977ba", GitCommit:"a977bab148535ec195f12edc8720913c7b943f9c", GitTreeState:"clean", BuildDate:"2019-07-29T20:47:04Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
```
└>> eksctl install flux --name=kalbir-test-01 --git-url=git@github.com:kalbir/my-eks-cluster.git --git-email=kalbir@gmail.com --region=eu-west-1
[ℹ]  Cloning git@github.com:kalbir/my-eks-cluster.git
Cloning into '/var/folders/7b/3kr7dlnn6231zcwcqjy_6pdh0000gn/T/eksctl-install-flux-clone018782554'...
fatal: Remote branch master not found in upstream origin
[✖]  cannot clone repository git@github.com:kalbir/my-eks-cluster.git: exit status 128
```

2opremio:
If (and only if) the gitops repo provided by `--git-url` is empty (no branches), we create the branch provided by the user.

Fixes #1137

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -128,12 +128,11 @@ func (git *Client) cloneRepoInPath(clonePath string, options CloneOptions) error
 		// Switch to target branch
 		args := []string{"checkout", options.Branch}
 		if options.Bootstrap {
-			// create the branch if the repository is empty (no pre-existing branches)
-			files, err := ioutil.ReadDir(filepath.Join(git.dir, ".git", "refs", "heads"))
+			empty, err := git.isRepoEmpty()
 			if err != nil {
 				return err
 			}
-			if len(files) == 0 {
+			if empty {
 				args = []string{"checkout", "-b", options.Branch}
 			}
 		}
@@ -143,6 +142,15 @@ func (git *Client) cloneRepoInPath(clonePath string, options CloneOptions) error
 	}
 
 	return nil
+}
+
+func (git *Client) isRepoEmpty() (bool, error) {
+	// A repository is empty if it doesn't have branches
+	files, err := ioutil.ReadDir(filepath.Join(git.dir, ".git", "refs", "heads"))
+	if err != nil {
+		return false, err
+	}
+	return len(files) == 0, nil
 }
 
 // Add performs can perform a `git add` operation on the given file paths
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -21,9 +22,9 @@ const (
 	DefaultGitTimeout = 20 * time.Second
 )
 
-// Cloner can clone git repositories
-type Cloner interface {
-	CloneRepo(cloneDirPrefix string, branch string, gitURL string) (string, error)
+// TmpCloner can clone git repositories in temporary directories
+type TmpCloner interface {
+	CloneRepoInTmpDir(cloneDirPrefix string, options CloneOptions) (string, error)
 }
 
 // Client can perform git operations on the given directory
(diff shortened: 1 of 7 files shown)


pkg/git/git.go
```diff
@@ -88,28 +89,59 @@ func NewGitClientFromExecutor(executor executor.Executor) *Client {
...
 }
 
-// CloneRepo clones a repo specified in the gitURL and checks out the specified branch
-func (git *Client) CloneRepo(cloneDirPrefix string, branch string, gitURL string) (string, error) {
-	cloneDir, err := ioutil.TempDir(os.TempDir(), cloneDirPrefix)
+// CloneOptions are the options for cloning a Git repository
+type CloneOptions struct {
+	URL       string
+	Branch    string
+	Bootstrap bool // create the branch if the repository is empty
```
marccarre: Nit: `CreateBranchIfMissing`? (In order to avoid having a comment, and having something hopefully more explicit? It is admittedly longer/more verbose, but hopefully reduces mental load a bit.)

pkg/git/git.go
```diff
@@ -88,28 +89,59 @@ func NewGitClientFromExecutor(executor executor.Executor) *Client {
...
+	// only do it after the clone so that it doesn't create an
+	// undesirable nested directory
 	git.dir = clonePath
+
+	if options.Branch != "" {
+		// Switch to target branch
+		args := []string{"checkout", options.Branch}
+		if options.Bootstrap {
+			// create the branch if the repository is empty (no pre-existing branches)
+			files, err := ioutil.ReadDir(filepath.Join(git.dir, ".git", "refs", "heads"))
```
marccarre: Nit: extract to utility function:
```go
func doBranchesExist(dir string) (bool, error) {
	files, err := ioutil.ReadDir(filepath.Join(dir, ".git", "refs", "heads"))
	if err != nil {
		return false, err
	}
	return len(files) != 0, nil
}
```
to separate the "what" from the "how", and hopefully increase readability.

pkg/git/git.go
marccarre: Also, given it looks like we always pass `Bootstrap: true` in production code, isn't it worth to _always_ "bootstrap"? 🤔

2opremio: You don't always create the branch, only if the repo is empty. So the name would need to be something like `CreateBranchIfRepoEmpty` which maybe is too long?

We do set it to false when cloning the profile repo

pkg/git/git.go
2opremio: Good point. Will do

pkg/git/git.go
marccarre: Fair enough 👍

2opremio:
rebasing ...

PR #1250, opened 2019-09-02
Fix node group deletion

cPu1:
### Checklist

### DIFF
--- a/pkg/cfn/manager/delete_tasks.go
+++ b/pkg/cfn/manager/delete_tasks.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/service/cloudformation"
-
-	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 )
 
 // NewTasksToDeleteClusterWithNodeGroups defines tasks required to delete all the nodegroup
@@ -13,7 +11,9 @@ import (
 func (c *StackCollection) NewTasksToDeleteClusterWithNodeGroups(wait bool, cleanup func(chan error, string) error) (*TaskTree, error) {
 	tasks := &TaskTree{Parallel: false}
 
-	nodeGroupTasks, err := c.NewTasksToDeleteNodeGroups(nil, true, cleanup)
+	deleteAll := func(_ string) bool { return true }
+	nodeGroupTasks, err := c.NewTasksToDeleteNodeGroups(deleteAll, true, cleanup)
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/cfn/manager/delete_tasks.go
```diff
@@ -13,7 +13,7 @@ import (
 func (c *StackCollection) NewTasksToDeleteClusterWithNodeGroups(wait bool, cleanup func(chan error, string) error) (*TaskTree, error) {
 	tasks := &TaskTree{Parallel: false}
 
-	nodeGroupTasks, err := c.NewTasksToDeleteNodeGroups(nil, true, cleanup)
+	nodeGroupTasks, err := c.NewTasksToDeleteNodeGroups(c.spec.NodeGroups, true, cleanup)
```
errordeveloper: We do want to delete all nodegroups that are present remotely, they may or may not be listed in `c.spec.NodeGroups`, hence currently we pass `nil` as the means of forcing this behaviour (but that part doesn't work since #1198, which got missed in review).

Perhaps we should introduce `all bool` argument to make it explicit?

errordeveloper: To be clear, we won't be able to delete the cluster stack before all nodegroups get deleted. And current behaviour is very weird, as imported outputs prevent cluster stack deletion you will see the following:

```
[ℹ]  will delete stack "eksctl-irp-test-4-cluster"
[ℹ]  waiting for stack "eksctl-irp-test-4-cluster" to get deleted
[✖]  unexpected status "CREATE_COMPLETE" while waiting for CloudFormation stack "eksctl-irp-test-4-cluster"
[ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
[ℹ]  1 error(s) occurred while deleting cluster with nodegroup(s)
[✖]  waiting for CloudFormation stack "eksctl-irp-test-4-cluster" to reach "DELETE_COMPLETE" status: ResourceNotReady: failed waiting for successful resource state
[✖]  failed to delete cluster with nodegroup(s)
```

That implies deletion never happens, so stack never goes into `DELETE_IN_PROGRESS` state...

cPu1: I thought the nil argument was used to ignore the filter list. 

I think passing a predicate function to `NewTasksToDeleteNodeGroups` to determine node group deletion would avoid the need of both adding a specialised case and taking more parameters.

errordeveloper: I'd say booleans argument would be easier... We could provide a separate functions otherwise, or perhaps optimise `NewTasksToDeleteClusterWithNodeGroups` to make a single list of all stacks that need to be deleted?

I'd rather see simplest fix for now.

errordeveloper: I guess we could have the predicate function passed to a private function and wrap it to keep footprint of 'NewTasksToDeleteNodeGroups' as is. Then we could have different predicate in 'NewTasksToDeleteClusterWithNodeGroups'. I mean, just keep the predicate an implementation detail of a private function... What do you think?

cPu1:
```go
func (c *StackCollection) NewTasksToDeleteNodeGroups(predicateFn func(string) bool, wait bool, cleanup func(chan error, string) error) (*TaskTree, error) {
	nodeGroupStacks, err := c.DescribeNodeGroupStacks()
	if err != nil {
		return nil, err
	}

	tasks := &TaskTree{Parallel: true}

	for _, s := range nodeGroupStacks {
		name := c.GetNodeGroupName(s)

		if !predicateFn(name) {
			continue
		}
```

And then one of the calls will be:
```go
		hasNodeGroup := func(name string) bool {
			for _, ng := range filteredNodeGroups {
				if ng.Name == name {
					return true
				}
			}
			return false
		}

		tasks, err := stackManager.NewTasksToDeleteNodeGroups(hasNodeGroup, cmd.Wait, nil)
```

PR #1252, opened 2019-09-02
Simplify predicate function implementation

errordeveloper:
### Description

A small refinement on top of #1250.

### Checklist

### DIFF
--- a/pkg/ctl/delete/nodegroup.go
+++ b/pkg/ctl/delete/nodegroup.go
@@ -115,16 +115,8 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 	cmdutils.LogIntendedAction(cmd.Plan, "delete %d nodegroups from cluster %q", len(filteredNodeGroups), cfg.Metadata.Name)
 
 	{
-		hasNodeGroup := func(name string) bool {
-			for _, ng := range filteredNodeGroups {
-				if ng.Name == name {
-					return true
-				}
-			}
-			return false
-		}
-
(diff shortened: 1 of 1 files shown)


pkg/ctl/delete/nodegroup.go
```diff
@@ -115,16 +115,7 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
...
-			for _, ng := range filteredNodeGroups {
-				if ng.Name == name {
-					return true
-				}
-			}
-			return false
-		}
-
-		tasks, err := stackManager.NewTasksToDeleteNodeGroups(hasNodeGroup, cmd.Wait, nil)
+		tasks, err := stackManager.NewTasksToDeleteNodeGroups(ngFilter.Match, cmd.Wait, nil)
```
cPu1: While this is terser, the issue with using `Match` here is the entire list of rules and glob matching need to be performed for every call. Since we already have the filtered node groups, the `hasNodeGroup` call would be more efficient.

errordeveloper: Yes, I've changed this just now to pass `ngSubset.Has` instead. I guess I'm not too concerned with performance at this stage tend to go for reduced verbosity and readaability.

I think when we will work on a controller, we will need to care more about performance, and I actually don't see a good reason for this whole nodegroup filter ideal to be of use in controller code. That's also why the filter lives in `pkg/ctl/cmdutils` and not `pkg/apis`. At the same time, I'm very happy we've reduced the number of repeated calls to `Match` and started cleared clearing abstraction boundaries.

PR #1138, opened 2019-08-13
Gitops quickstart doc

dholbach:
Document GitOps workflow.

PR #1256, opened 2019-09-03
Rename eks-gitops-example to eks-quickstart-app-dev

martina-if:
The first quickstart repo has been renamed to eks-quickstart-app-dev


martina-if:
partially done in https://github.com/weaveworks/eksctl/pull/1138/files so I will do it there

PR #1259, opened 2019-09-03
Enhance cluster status checks

### LINKED ISSUE #1254: status check in ctl.RefreshClusterConfig should not be relied on implicitly
We currently happen to rely on the fact that `ctl.RefreshClusterConfig` check if cluster status is `ACTIVE`, or return an error otherwise. This makes deletions complicated, e.g. we attempted to address #1069 by checking the error, but it now prevents one from running `eksctl delete cluster --wait` on a cluster where deletion was already triggered in order to wait until it actually gets deleted (which worked previously).

Also in other cases we don't have to have an `ACTIVE` cluster, e.g. we should be able to operate it when it is in `UPDATING` state.

errordeveloper:
## Description

This fixes #1254.

- rename `ctl.RefreshClusterConfig` to `ctl.RefreshClusterStatus`, so
  it's clear what it does and doesn't do
- enable simple short-term cache of cluster info
- imporove how `ctl.RefreshClusterStatus` works
- defined concrete methods - `ctl.CanOperate`, `ctl.CanUpdate`, `ctl.CanDelete`,
  for perfomaing high-level status checks and avoid usign
  `ctl.RefreshClusterStatus` directly
- remove `ctl.DescribeControlPlaneMustBeActive`
- ensure Kuberntes API deletion operations are only perfoermed on operable clusters
- improve early deletion test case

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/delete/nodegroup.go
+++ b/pkg/ctl/delete/nodegroup.go
@@ -2,7 +2,6 @@ package delete
 
 import (
 	"github.com/kris-nova/logger"
-	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
@@ -62,8 +61,8 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 		return err
 	}
 
-	if err := ctl.RefreshClusterConfig(cfg); err != nil {
-		return errors.Wrapf(err, "getting credentials for cluster %q", cfg.Metadata.Name)
+	if ok, err := ctl.CanOperate(cfg); !ok {
+		return err
 	}
 
 	clientSet, err := ctl.NewStdClientSet(cfg)
--- a/pkg/eks/cache.go
+++ b/pkg/eks/cache.go
@@ -1,0 +1,31 @@
+package eks
+
+import (
+	"time"
+
+	awseks "github.com/aws/aws-sdk-go/service/eks"
+)
+
+const clusterInfoCacheTTL = 15 * time.Second
+
+type clusterInfo struct {
+	timestamp time.Time
+	cluster   *awseks.Cluster
+}
+
+func (c *ClusterProvider) clusterInfoNeedsUpdate() bool {
+	if c.Status.clusterInfo == nil {
+		return true
+	}
+	if time.Since(c.Status.clusterInfo.timestamp) > clusterInfoCacheTTL {
+		return true
+	}
+	return false
+}
+
+func (c *ClusterProvider) setClusterInfo(cluster *awseks.Cluster) {
+	c.Status.clusterInfo = &clusterInfo{
+		timestamp: time.Now(),
+		cluster:   cluster,
+	}
+}
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -21,9 +21,9 @@ import (
 )
 
 // DescribeControlPlane describes the cluster control plane
-func (c *ClusterProvider) DescribeControlPlane(cl *api.ClusterMeta) (*awseks.Cluster, error) {
+func (c *ClusterProvider) DescribeControlPlane(meta *api.ClusterMeta) (*awseks.Cluster, error) {
 	input := &awseks.DescribeClusterInput{
-		Name: &cl.Name,
+		Name: &meta.Name,
 	}
 	output, err := c.Provider.EKS().DescribeCluster(input)
 	if err != nil {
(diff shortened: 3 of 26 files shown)


pkg/utils/kubeconfig/kubeconfig.go
```diff
@@ -208,11 +208,11 @@ func isValidConfig(p, name string) error {
 }
 
 // MaybeDeleteConfig will delete the auto-generated kubeconfig, if it exists
-func MaybeDeleteConfig(cl *api.ClusterMeta) {
-	p := AutoPath(cl.Name)
+func MaybeDeleteConfig(meta *api.ClusterMeta) {
```
errordeveloper: I'll split this into another commit, this is just an ancient inconsistency that jumped at me...

pkg/ctl/delete/nodegroup.go
```diff
@@ -62,8 +61,8 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 		return err
 	}
 
-	if err := ctl.RefreshClusterConfig(cfg); err != nil {
-		return errors.Wrapf(err, "getting credentials for cluster %q", cfg.Metadata.Name)
+	if ok, err := ctl.CanOperate(cfg); !ok {
+		return err
```
martina-if: Don't we need the wrap message we had before?

pkg/eks/cache.go
```diff
@@ -0,0 +1,34 @@
...
+type clusterInfo struct {
+	timestamp time.Time
+	cluster   *awseks.Cluster
+}
+
+func (c *ClusterProvider) clusterInfoNeedsUpdate() bool {
+	if c.Status.clusterInfo == nil {
+		return true
+	}
+	if c.Status.clusterInfo == nil {
```
martina-if: This check is duplicated

pkg/eks/cache.go
```diff
@@ -0,0 +1,34 @@
...
+		return true
+	}
+	if time.Since(c.Status.clusterInfo.timestamp) > clusterInfoCacheTTL {
+		return true
+	}
+	return false
+}
+
+func (c *ClusterProvider) newClusterInfo(cluster *awseks.Cluster) {
```
martina-if: To follow conventions, I think this should be renamed to `setClusterInfo(cluster *awseks.Cluster)`

pkg/eks/eks.go
```diff
@@ -32,56 +32,94 @@ func (c *ClusterProvider) DescribeControlPlane(cl *api.ClusterMeta) (*awseks.Clu
...
+		return c.RefreshClusterStatus(spec)
+	}
 	return nil
 }
 
+// CanDelete return true when a cluster can be deleted, otherwise it returns false along with an error explaing the reason
+func (c *ClusterProvider) CanDelete(spec *api.ClusterConfig) (bool, error) {
+	err := c.maybeRefreshClusterStatus(spec)
+	if err != nil {
+		return false, errors.Wrapf(err, "fetching cluster status to determin if it can be deleted")
```
martina-if:
```suggestion
		return false, errors.Wrapf(err, "fetching cluster status to determine if it can be deleted")
```

pkg/eks/eks.go
```diff
@@ -32,56 +32,94 @@ func (c *ClusterProvider) DescribeControlPlane(cl *api.ClusterMeta) (*awseks.Clu
...
+	}
+	// it must be possible to delete cluster in any state
+	return true, nil
+}
+
+// CanOperate return true when a cluster can be operated, otherwise it returns false along with an error explaing the reason
```
martina-if:
```suggestion
// CanOperate return true when a cluster can be operated, otherwise it returns false along with an error explaining the reason
```

pkg/ctl/delete/nodegroup.go
errordeveloper: No, because new one is more meaningful...

errordeveloper:
@martina-if I see your point about better overall interface, I'd definitely like to to rework this package a bit more, as it's a historic one and carries around a lot of things... But I'm a little cautious making more radical changes a the moment, and I do see many things that can be improved.

pkg/eks/eks.go
```diff
@@ -32,56 +32,94 @@ func (c *ClusterProvider) DescribeControlPlane(cl *api.ClusterMeta) (*awseks.Clu
...
+// CanOperate return true when a cluster can be operated, otherwise it returns false along with an error explaining the reason
+func (c *ClusterProvider) CanOperate(spec *api.ClusterConfig) (bool, error) {
+	err := c.maybeRefreshClusterStatus(spec)
+	if err != nil {
+		return false, errors.Wrapf(err, "fetching cluster status to determine operability")
+	}
+
+	switch status := *c.Status.clusterInfo.cluster.Status; status {
+	case awseks.ClusterStatusCreating, awseks.ClusterStatusDeleting, awseks.ClusterStatusFailed:
+		return false, fmt.Errorf("cannot operate cluster %q in %q region due to status %q", spec.Metadata.Name, spec.Metadata.Region, status)
```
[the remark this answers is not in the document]
errordeveloper: You are right, I'll change it!

errordeveloper:
I think CI should be happy now, there were a few typos and annoying bugs in tests that I fix now...

PR #1291, opened 2019-09-05
Set 1.14 as the latest version

cPu1:
### Checklist


pkg/ctl/update/cluster.go
```diff
@@ -81,7 +81,7 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 	case api.Version1_12:
 		cfg.Metadata.Version = api.Version1_13
 	case api.Version1_13:
-		cfg.Metadata.Version = api.Version1_13
+		cfg.Metadata.Version = api.Version1_14
```
discosultan: Shouldn't there be a new case for `api.Version1_14`?

PR #1299, opened 2019-09-06
Fix launch template errors

cPu1:
Not all instance types support EBS-optimized volumes, this fix makes the configuration optional

- Fixes #1293 
- Fixes #1257 

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] All unit tests passing (i.e. `make test`)
- [x] Manually tested

PR #1303, opened 2019-09-06
Test all values of `EBSOptimized`

errordeveloper:
### Description

As discussed in https://github.com/weaveworks/eksctl/pull/1299#discussion_r321642012.

### Checklist


errordeveloper:
> We should consider using table-driven tests for new changes and even port some of the existing ones.

I did consider, but wasn't sure how to do it in this case... I'm also curious if the there is a pattern that works nicely with gomega.

PR #1306, opened 2019-09-06
Return gracefully when updating 1.14 clusters

### LINKED ISSUE #1304: 1.14 is not known in this version of eksctl
**What happened?**
Upgraded control plane to 1.14 using `eksctl update cluster --name ats-auto --approve` but then tried to use a config file to update the node groups and got an error saying 1.14 is not known in this version of eksctl.

```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.5.2"}

ejsmith at ejsmith in ~/infrastructure (master●)
$ eksctl update cluster -f eks-cluster.yaml
[ℹ]  using region us-east-1
[!]  NOTE: config file is used for finding cluster name and region
[!]  NOTE: cluster VPC (subnets, routing & NAT Gateway) configuration changes are not yet implemented
[✖]  control plane version "1.14" is not known to this version of eksctl, try to upgrade eksctl first
```

cPu1:
### Description

Closes #1304 

### Checklist

### DIFF
--- a/pkg/ctl/update/cluster.go
+++ b/pkg/ctl/update/cluster.go
@@ -82,6 +82,8 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 		cfg.Metadata.Version = api.Version1_13
 	case api.Version1_13:
 		cfg.Metadata.Version = api.Version1_14
+	case api.Version1_14:
+		cfg.Metadata.Version = api.Version1_14
 	default:
 		// version of control plane is not known to us, maybe we are just too old...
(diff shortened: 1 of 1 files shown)


pkg/ctl/update/cluster.go
```diff
@@ -82,6 +82,8 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 		cfg.Metadata.Version = api.Version1_13
 	case api.Version1_13:
```
cPu1: This can also be:

```golang
case api.Version1_13, api.Version1_14:
```

but the cases aren't combined because when releasing a newer version, it'll help us remind of changing the `api.Version1_14` case.

PR #1318, opened 2019-09-10
Fix creating IAM roles for Service Accounts (IRSA) when using attachPolicy

### LINKED ISSUE #1311: IAM Service Accounts error when using attachPolicy
**What happened?**
Recently, eksctl was updated to support IAM for service accounts. I followed the guide https://eksctl.io/usage/iamserviceaccounts/ and am using the cluster.yaml approach. 

I receive the following error in CloudFormation during the LogicalId = Policy1 step: 
Property PolicyName cannot be empty.

Here's part of the cluser.yaml:
iam:
  withOIDC: true
  serviceAccounts:
  - metadata:
      name: cluster-autoscaler
      namespace: kube-system
      labels: {aws-usage: "cluster-ops"}
    attachPolicy: # inline policy can be defined along with `attachPolicyARNs`
      Version: "2012-10-17"
      Statement:
      - Effect: Allow
        Action:
        - "autoscaling:DescribeAutoScalingGroups"
        - "autoscaling:DescribeAutoScalingInstances"
        - "autoscaling:DescribeLaunchConfigurations"
        - "autoscaling:DescribeTags"
        - "autoscaling:SetDesiredCapacity"
        - "autoscaling:TerminateInstanceInAutoScalingGroup"
        Resource: '*'

I expected no error and to have the cluster and iamserviceaccount to create successfully. 

**How to reproduce it?**
I have attempted to deploy the cluster and iamserviceaccount resources by running the following commands:
eksctl create cluster -f clusters/cluster.yaml
eksctl create iamserviceaccount -f cluster.yaml --profile=xxxxxxxx --approve

**Versions**
```
$ eksctl version
[i]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.5.2"}
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
```

cPu1:
### Description

Fixes #1311

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/iam.go
+++ b/pkg/cfn/builder/iam.go
@@ -158,7 +158,7 @@ func (n *NodeGroupResourceSet) addResourcesForIAM() {
 	}
 
 	role := gfn.AWSIAMRole{
-		Path:                     gfn.NewString("/"),
+		Path: gfn.NewString("/"),
 		AssumeRolePolicyDocument: cft.MakeAssumeRolePolicyDocumentForServices("ec2.amazonaws.com"),
 		ManagedPolicyArns:        makeStringSlice(n.spec.IAM.AttachPolicyARNs...),
 	}
@@ -431,6 +431,7 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
 
 	if len(rs.spec.AttachPolicy) != 0 {
 		rs.template.NewResource("Policy1", &cft.IAMPolicy{
+			PolicyName:     cft.NewString(fmt.Sprintf("%s-%s", rs.spec.Namespace, rs.spec.Name)),
 			Roles:          cft.MakeSlice(roleRef),
 			PolicyDocument: rs.spec.AttachPolicy,
 		})
--- a/pkg/cfn/builder/iam.go
+++ b/pkg/cfn/builder/iam.go
@@ -430,22 +430,10 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
 	})
 
 	if len(rs.spec.AttachPolicy) != 0 {
-		rs.template.NewResource("Policy1", &cft.IAMPolicy{
-			PolicyName:     cft.NewString(truncateResourceName(fmt.Sprintf("%s-%s", rs.spec.Namespace, rs.spec.Name))),
-			Roles:          cft.MakeSlice(roleRef),
-			PolicyDocument: rs.spec.AttachPolicy,
-		})
+		rs.template.AttachPolicy("Policy1", roleRef, cft.MapOfInterfaces(rs.spec.AttachPolicy))
 	}
 
 	return nil
-}
-
-func truncateResourceName(name string) string {
-	const maxLength = 128
-	if len(name) > maxLength {
-		return name[:maxLength]
-	}
-	return name
 }
 
 // RenderJSON will render iamserviceaccount stack as JSON
--- a/pkg/cfn/builder/iam.go
+++ b/pkg/cfn/builder/iam.go
@@ -431,13 +431,21 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
 
 	if len(rs.spec.AttachPolicy) != 0 {
 		rs.template.NewResource("Policy1", &cft.IAMPolicy{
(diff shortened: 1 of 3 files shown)


errordeveloper:
@cPu1 could you please explain more specifically what went wrong exactly?

pkg/cfn/builder/iam.go
```diff
@@ -431,6 +431,7 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
 
 	if len(rs.spec.AttachPolicy) != 0 {
 		rs.template.NewResource("Policy1", &cft.IAMPolicy{
+			PolicyName:     cft.NewString(truncateResourceName(fmt.Sprintf("%s-%s", rs.spec.Namespace, rs.spec.Name))),
```
errordeveloper: My understanding was that this will get named automatically by CloudFormation, and setting a name will require `CAPABILITY_NAMED_IAM`. Could please clarify? Will CloudFormation pre-pend a unique suffix, but user is still required to provide the name?

pkg/cfn/builder/iam.go
```diff
@@ -439,6 +440,14 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
 	return nil
 }
 
+func truncateResourceName(name string) string {
```
errordeveloper: If CloudFormation will prepend a unique suffix, it will truncate it automatically also. We shouldn't need this function.

errordeveloper:
Please make sure to add a test.

pkg/cfn/builder/iam.go
errordeveloper: I can see that my understanding was wrong, we do set `PolicyName` in all other instances and it doesn't require `CAPABILITY_NAMED_IAM`.

Please see `pkg/cfn/template/iam_helpers.go`, we should just add `AttachPolicy` function there and use that. Be sure to use `MakeName` also, there is no need to have custom naming scheme here.

cPu1:
>@cPu1 could you please explain more specifically what went wrong exactly?

PolicyName is a required field but it wasn't being set.

pkg/cfn/builder/iam.go
cPu1: It does not prepend a unique suffix. The field needs to be set as mentioned below.

pkg/cfn/builder/iam_test.go
```diff
@@ -166,6 +167,7 @@ var _ = Describe("template builder for IAM", func() {
 		]`))
 
 		Expect(t).To(HaveResourceWithPropertyValue("Role1", "AssumeRolePolicyDocument", expectedAssumeRolePolicyDocument))
+		Expect(t).To(HaveResourceWithProperties("Policy1", "PolicyName"))
```
errordeveloper: Could you also check the value also, I believe it should be deterministic.

pkg/cfn/template/iam_helpers.go
```diff
@@ -2,14 +2,19 @@ package template
...
 // AttachAllowPolicy constructs a role with allow policy for given resources and actions
 func (t *Template) AttachAllowPolicy(name string, refRole *Value, resources interface{}, actions []string) {
+	t.AttachPolicy(name, refRole, MakePolicyDocument(MapOfInterfaces{
+		"Effect":   "Allow",
+		"Resource": resources,
+		"Action":   actions,
+	}))
+}
+
+// AttachPolicy attaches the specified policy document
```
errordeveloper: nit: move it up?

pkg/cfn/builder/iam_test.go
cPu1: It is deterministic. I thought about using it first but it leaks the CloudFormation abstraction. Anyway, I just saw another assertion using an intrinsic function, so I'll add it for this one as well.

pkg/cfn/template/iam_helpers.go
cPu1: There's no standard convention for ordering dependent functions (like you said, it's a nit).

PR #1320, opened 2019-09-10
Fix invalid go module version

bboreham:
The Google go proxy is flagging the current version as an error:
`invalid version: unknown revision 000000000000`

It shouldn't really matter as we have a `replace` line that overrides it, but ¯\_(ツ)_/¯

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/MakeNowJust/heredoc v0.0.0-20171113091838-e9091a26100e // indirect
 	github.com/alecthomas/jsonschema v0.0.0-20190530235721-fd8d96416671
 	github.com/aws/aws-sdk-go v1.23.15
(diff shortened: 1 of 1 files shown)


errordeveloper:
Thanks, Bryan! We do need to update this:

https://github.com/weaveworks/eksctl/blob/20c52b905a02f5c8129ae71e7eceb240803ea685/Makefile#L7

Then run `make eksctl-image && docker push weaveworks/eksctl-build:deps-0.16`.

bboreham:
I don't think this is really a change to the dependencies, since you have a `replace` line overriding the version.

errordeveloper:
With the way build image works we need to update it, as otherwise the CI
will slow down dramatically. We are aiming to automate this in #1200, but
we aren't quite there yet.

PR #1335, opened 2019-09-16
Document current design of node bootstrap process

errordeveloper:
The purpose of this document is to summaries how node bootstrap process is currently implemented in `eksctl`,  so it can be referred to in any discussions.

### DIFF
--- a/docs/current-001-node_bootstrap.md
+++ b/docs/current-001-node_bootstrap.md
@@ -1,0 +1,255 @@
+# Current Design #001: Node Bootstrap
+
+> NOTE: the purpose of this document is to summaries how node bootstrap process is currently implemented in `eksctl`,
+> so it can be referred to in any discussions.
+
+## Guiding Principles
+
+To enable reliable node management at large-scale, the bootstrap process of a node should bare the following attributes:
+
+- stateless
+- immutable
+- deterministic
+- simple
+
+It **should avoid**:
+- non essential runtime parameters
+- mutation of configuration during execution
+- generating complex configuration files
+- download of configuration files
+- installation of software
+- making unnecessary look-up calls to remote APIs
+- only call out to commands when absolutely essential
+
+Current implementation is build for official AL2 EKS AMIs, that includes dependencies such as AWS CLI and jq, however
+it's ought to be possible for one to create a stripped-down AMI and re-use the bootstrap script.
+
+Aside from all of the above concerns, it's crucial to bare in mind that the dependencies of bootstrap process should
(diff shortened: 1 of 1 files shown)


whereisaaron:
This is super awesome @errordeveloper, thank you for writing it. Reads well. This definitely belongs on the website please!

The only content thing I noted it that it doesn't cover or reference how you can customise the kubelet config with the `eksctl` config (e.g. to reserve kubelet resources). It is not clear to me how that works to modify the drop-in files?

I think it is great that the actual script/files are included in the discussion (with links to the particular commit), but wonder if that sets it up to fall our of date later? Kinda not DRY 😄 

Could we integrate this document into the website and employ the static site generator to embed the matching version of the relevant files as the documentation website is generated?

errordeveloper:
> The only content thing I noted it that it doesn't cover or reference how you can customise the kubelet config with the `eksctl` config (e.g. to reserve kubelet resources). It is not clear to me how that works to modify the drop-in files?

Resource reservation is documented here: https://eksctl.io/usage/customizing-the-kubelet/

If you really must have a drop-in for some special case, you will need to add a shell script via `preBootstrapCommands`.

> I think it is great that the actual script/files are included in the discussion (with links to the particular commit), but wonder if that sets it up to fall our of date later?

Well, that's a general issue with documentation, hence this document didn't exist for a while...

> Could we integrate this document into the website and employ the static site generator to embed the matching version of the relevant files as the documentation website is generated?

I suppose we could, I just didn't see this as use-facing kind of documentation, but perhaps I'm wrong on that one. To me this was an implementation detail that need documenting, so new contributors are informed of motivation and goals, as until now was mostly in my head.

whereisaaron:
> Resource reservation is documented here: https://eksctl.io/usage/customizing-the-kubelet/

Yep that is cool, but that document and this one, don’t say how those kubelet configuration values are incorporated for the bootstrap process (that I spotted?).

> I suppose we could, I just didn't see this as use-facing kind of documentation, but perhaps I'm wrong on that one. To me this was an implementation detail that need documenting, so new contributors are informed of motivation and goals, as until now this was mostly in my head.

Well I’d read it 😀 This is a key difference between `eksctl` clusters and ‘standard’ EKS clusters, so it worth mentioning. And if you are ever investigating a node problem, it is helpful to know what goes were. It could be buried in the back or under ‘Advanced Topics’ to be sure.

errordeveloper:
@whereisaaron fair enough, I've opened https://github.com/weaveworks/eksctl/issues/1357.

Issue #1315, opened 2019-09-09
[CoreDNS] use forward plugin instead of proxy plugin  in kubernetes 1.14


s-tokutake:
**Why do you want this feature?**

I want to use CoreDNS forward plugin in EKS.

- Kubernetes 1.14 use CoreDNS forward plugin instead of proxy plugin by default. 
  - cf https://github.com/kubernetes/kubernetes/pull/73267
- CoreDNS proxy plugin has problems and it was deprecated.
  - cf https://github.com/coredns/coredns/issues/1443

**What feature/behavior/change do you want?**

- run ` eksctl utils update-coredns ` , and update coredns configmap to use  forward plugin.

sayboras:
Related issue in EKS as reference https://github.com/aws/containers-roadmap/issues/489

sayboras:
@s-tokutake @martina-if seems like we can close this as per https://github.com/weaveworks/eksctl/pull/1729

PR #1343, opened 2019-09-18
[coredns] use forward plugin instead of proxy plugin

s-tokutake:
- coredns in kubernetes 1.14 use forward plugin instead of proxy plugin, so eks 1.14 should use forward plugin.

### Description

- please see https://github.com/weaveworks/eksctl/issues/1315

### Checklist

- [x] Manually tested
  - run ` eksctl utils update-coredns `, and ` kubectl -n kube-system get configmap coredns -o yaml ` , output is as expected ↓ .

```yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          upstream
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
kind: ConfigMap
metadata:
  creationTimestamp: "2019-09-18T05:30:34Z"
  labels:
    eks.amazonaws.com/component: coredns
    k8s-app: kube-dns
  name: coredns
  namespace: kube-system
  resourceVersion: "4193"
  selfLink: /api/v1/namespaces/kube-system/configmaps/coredns
  uid: 70479f91-d9d5-11e9-9353-0af93430349e
```


errordeveloper:
The manifest is based on what's shipped in EKS cluster when one is created. We have exported this manifest when Kubernetes 1.14 was released in EKS.

So the way to update this manifest is to run [`./hack/dump-coredns-manifests.sh`](https://github.com/weaveworks/eksctl/blob/master/hack/dump-coredns-manifests.sh) against a newly-created cluster, is that what you've done here you just updated the file?

I would rather avoid a divergence between this manifest and what EKS ships.

s-tokutake:
@errordeveloper 

Thank you for your reply.

I created EKS cluster Kubernetes 1.14 using aws console and ` aws eks ` command  , and confirmed that coredns in that cluster use proxy plugin. 
So I'm going to talk with EKS team.

errordeveloper:
Thanks, perhaps you close this PR as it stands?

PR #1337, opened 2019-09-17
gitops apply: use temp dir instead of --output-path

marccarre:
### WARNING: UX decision

This PR should only be merged if we decide we no longer want to generate a local clone of the user's configuration repository as an output.

### Description

Given end-users provide their configuration Git repository as an input, it makes little sense to return it, cloned locally, as an output. Instead, `eksctl gitops apply` now:

- no longer expose `--output-path`,
- uses a temporary directory instead, for cloning and file operations,
- expects that the end-user already has, or soon will clone this
  repository, and will "pull" the latest changes themselves, once
  they have run `eksctl gitops apply`.

Fixes #1274.

### Checklist

- [x] Code compiles correctly (i.e `make build`)
- [x] ~Added tests that cover your change (if possible)~
- [x] All unit tests passing (i.e. `make test`)
- [x] ~Added/modified documentation as required (such as the `README.md`, and `examples` directory)~
- [x] Manually tested

### Manual tests

```console
$ cd $GOPATH/src/github.com/weaveworks/eksctl
$ make build
```

Using the default (current directory) fails, as it is a non-empty directory:
```console
$ EKSCTL_EXPERIMENTAL=true ./eksctl \
    -f examples/eks-quickstart-app-dev.yaml \
    gitops apply \
    --quickstart-profile git@github.com:marccarre/eks-quickstart-app-dev.git \
    --git-url git@github.com:marccarre/my-gitops-repo.git \
    --git-email carre.marc+eksctl@gmail.com

[ℹ]  Generating public key infrastructure for the Helm Operator and Tiller
[ℹ]    this may take up to a minute, please be patient
[!]  Public key infrastructure files were written into directory "/var/folders/24/d3mml6bn20nftpt91cfldq1h0000gn/T/eksctl-helm-pki208351179"
[!]  please move the files into a safe place or delete them
[ℹ]  Generating manifests
[ℹ]  Cloning git@github.com:marccarre/my-gitops-repo.git
Cloning into '/var/folders/24/d3mml6bn20nftpt91cfldq1h0000gn/T/eksctl-install-flux-clone-244413614'...
remote: Enumerating objects: 61, done.
remote: Counting objects: 100% (61/61), done.
remote: Compressing objects: 100% (47/47), done.
remote: Total 367 (delta 23), reused 45 (delta 10), pack-reused 306
Receiving objects: 100% (367/367), 147.66 KiB | 440.00 KiB/s, done.
Resolving deltas: 100% (130/130), done.
Already on 'master'
[...]

PR #1349, opened 2019-09-20
Add release notes for 0.6

martina-if:


docs/release_notes/0.6.0.md
```diff
@@ -0,0 +1,7 @@
+# Release 0.6.0
+
+## Features
+
+- Rename experimental command `eksctl gitops apply` to `eksctl enable profile` (#1347) and use a temporary directory when cloning the repository (#1337)
+
```
cPu1: We need to add this as well:
```suggestion
- fix `eksctl update cluster` for 1.14 clusters (#1304)
```

docs/release_notes/0.6.0.md
```diff
@@ -0,0 +1,7 @@
+# Release 0.6.0
+
+## Features
+
+- Rename experimental command `eksctl gitops apply` to `eksctl enable profile` (#1347) and use a temporary directory when cloning the repository (#1337)
```
cPu1: We've been using the convention of not capitalising the first word:

```suggestion
- rename experimental command `eksctl gitops apply` to `eksctl enable profile` (#1347) and use a temporary directory when cloning the repository (#1337)
```

Also, any reason why we can't have #1337  as its own item, perhaps under "Code Improvements"?

PR #1350, opened 2019-09-20
NodeGroup: Define Subnets per node group

dnascimento:
### Description

Define multiple subnets per node group. This allows deploying node groups in subnets other than the EKS control plane. It also allows having multiple subnets in the same availability zone, for instance: internet-facing, application and database subnets

1. We need to verify that these subnets exist. `n.clusterSpec.VPC.Subnets.Private` contains a list of subnets that are verified but if we define two subnets in the same AZ, it fails. The code assumes one subnet per AZ [here](https://github.com/weaveworks/eksctl/blob/e9fc98eef14a8bbe86f417ba10b56b9bb438964c/pkg/apis/eksctl.io/v1alpha5/vpc.go#L133). How can we validate these subnets without breaking the code structure?
2. Should I create a test in `api_test.go` or what is the best place to test it?

@errordeveloper could you review this PR?

### Checklist

- [x] Manually tested

### DIFF
--- a/humans.txt
+++ b/humans.txt
@@ -63,6 +63,7 @@ Miguel Pereira          @onemorepereira
 Pedro Tôrres            @t0rr3sp3dr0
 Michael Treacher        @treacher
 Ajay Kemparaj           @ajayk
+Dario Nascimento        @dnascimento
 
 /* Thanks */
 
--- a/pkg/apis/eksctl.io/v1alpha5/types.go
+++ b/pkg/apis/eksctl.io/v1alpha5/types.go
@@ -552,6 +552,8 @@ type NodeGroup struct {
 	// +optional
 	AvailabilityZones []string `json:"availabilityZones,omitempty"`
 	// +optional
+	Subnets []string `json:"subnets,omitempty"`
+	// +optional
 	Tags map[string]string `json:"tags,omitempty"`
 	// +optional
 	PrivateNetworking bool `json:"privateNetworking"`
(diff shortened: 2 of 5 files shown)


errordeveloper:
hi @dnascimento! sorry for taking so long to read this, just didn't have the time...

Could you please illustrate what sort of configuration file have you tested this with?

Ideally, we should create a design proposal for any config schema changes. It would be great to gather all of the use-cases for this, I believe there are quite a few distinct ones. If you are happy to start on a design document, I can gather a few more pointers and help you drafting the document.

martina-if:
@dnascimento I see there hasn't been much activity since October, do you still have time to work on this?

dnascimento:
Hi @martina-if  @errordeveloper  Could you help to create a design proposal to add a field `Subnet` to the NodeGroup? 
The current code base assumes a subnet per AZ, which may not be 
This allows to "hard-code" the value of the subnets in which a node group will be deployed instead of using an AZ-based search.

dnascimento:
@martina-if @errordeveloper What should I do to have this merged?

dnascimento:
@kalbir may you review this PR and assist on what should I do to merge it?

martina-if:
Hi @dnascimento , Thanks for taking the time again to keep this up to date. Given the current limitations of eksctl we realize we need to redesign how eksctl works with subnets. I think we need to align this with the roadmap of the team so we can put some thought into how this would look like in terms of schema, use cases, etc.

Unfortunately we don't have this in our near term plans but we would like to tackle it at some point so it would be great if you could create a short design document where you described the use cases  and schema changes. The more information we have about a problem the easier it is to give it the right priority.

michaelbeaumont:
@dnascimento Hi, I'm going to close this PR for now. This seems like a useful feature though. Can you open a feature request and describe your use cases and the implications of config changes there?

PR #1353, opened 2019-09-20
test: Fix unit-tests when aws is the only install authenticator

dlespiau:
We were testing against either the heptio or AWS authenticator but I have none
of the above installed. Just aws.


pkg/eks/client_test.go
```diff
@@ -69,11 +69,21 @@ var _ = Describe("eks auth helpers", func() {
 					Expect(k.AuthInfos[ctx].Token).To(BeEmpty())
 					Expect(k.AuthInfos[ctx].Exec).To(Not(BeNil()))
 
-					Expect(k.AuthInfos[ctx].Exec.Command).To(MatchRegexp("(heptio-authenticator-aws|aws-iam-authenticator)"))
-
-					expectedArgs := "token -i auth-test-cluster"
+					Expect(k.AuthInfos[ctx].Exec.Command).To(MatchRegexp("(heptio-authenticator-aws|aws-iam-authenticator|aws)"))
+
+					var expectedArgs, roleARNArg string
```
errordeveloper: Could you add a TODO to say that ideally this should be tested with a mocked-out system somehow, so we exercise all of the 3 paths, and not just an accidental one?

dlespiau:
Thank you for the review!

PR #1360, opened 2019-09-24
[site] change the apiVersion to  eksctl.io/v1alpha5

ajayk:
[site] change the apiVersion to  eksctl.io/v1alpha5

### Description
[site] change the apiVersion to  eksctl.io/v1alpha5


errordeveloper:
Thank you!

PR #1363, opened 2019-09-24
Use --cluster for non-cluster commands

### LINKED ISSUE #1329: Use --cluster instead of --name for all non-cluster resource commands
In #1187 we fixed a command that works on non-cluster resource to use the proper flag name `--cluster` (instead of `--name`).

This also needs to be fixed in other areas, such as
- `eksctl <verb> iamidentitymapping --name=cluster`.
- `eksctl utils describe-stack --name=cluster`
- `eksctl generate profile --name=cluster`
- etc.

Consistency.

`cmdutils.AddNameFlag()` should be replaced for the commands where the main resource is not a cluster with a function that adds `--cluster` (and hides `--name`)

rndstr:
Commands act on a principal resource which may be a cluster or not. Most
commands act in the context of a cluster and as such need to receive its
name.

In case of the prinicpal resource being a cluster (such as `eksctl get cluster`)
the flag name is `--name`. For all other commands, the cluster name is
being passed in as `--cluster`.

Closes #1329

- [x] Once this is merged, create an issue to update AWS EKS documentation to use the renamed flags instead #1394 

### Checklist

- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -124,9 +124,28 @@ func AddTimeoutFlag(fs *pflag.FlagSet, p *time.Duration) {
 	AddTimeoutFlagWithValue(fs, p, api.DefaultWaitTimeout)
 }
 
-// AddNameFlag adds common --name flag for cluster
-func AddNameFlag(fs *pflag.FlagSet, meta *api.ClusterMeta) {
+// AddClusterFlag adds a common --cluster flag for cluster name.
+// Use this for commands whose principal resource is *not* a cluster.
+func AddClusterFlag(fs *pflag.FlagSet, meta *api.ClusterMeta) {
+	fs.StringVarP(&meta.Name, "cluster", "c", "", "EKS cluster name")
+}
+
+// AddClusterFlagWithDeprecated adds a common --cluster flag for
+// cluster name as well as a deprecated --name flag.
+// Use AddClusterFlag() for new commands.
+func AddClusterFlagWithDeprecated(fs *pflag.FlagSet, meta *api.ClusterMeta) {
+	AddClusterFlag(fs, meta)
 	fs.StringVarP(&meta.Name, "name", "n", "", "EKS cluster name")
+	_ = fs.MarkDeprecated("name", "use --cluster")
+}
+
+// ClusterNameFlag returns the flag to use for the cluster name
+// taking the principal resource into account.
+func ClusterNameFlag(cmd *Cmd) string {
+	if cmd.CobraCommand.Use == "cluster" {
+		return "--name"
+	}
+	return "--cluster"
 }
 
 // AddRegionFlag adds common --region flag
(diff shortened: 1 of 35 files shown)


rndstr:
'Manual test' output
```bash
$ ./eksctl get iamidentitymapping
[✖]  --cluster must be set
$ ./eksctl get iamidentitymapping --cluster=roli-dev
ARN                                                                                             USERNAME                                GROUPS
arn:aws:iam::840210520287:role/eksctl-roli-dev-nodegroup-ng-1-NodeInstanceRole-1AWFATM7MYTQ4    system:node:{{EC2PrivateDNSName}}       system:bootstrappers,system:nodes
$ ./eksctl get iamidentitymapping --name=roli-dev                                                                                                      
Flag --name has been deprecated, use --cluster
ARN                                                                                             USERNAME                                GROUPS
arn:aws:iam::840210520287:role/eksctl-roli-dev-nodegroup-ng-1-NodeInstanceRole-1AWFATM7MYTQ4    system:node:{{EC2PrivateDNSName}}       system:bootstrappers,system:nodes
```

martina-if:
LGTM I can approve when it's rebased

martina-if:
I ran the integration tests and they passed :+1: Let's merge it after this week's release.

pkg/ctl/cmdutils/cmdutils.go
```diff
@@ -124,9 +124,28 @@ func AddTimeoutFlag(fs *pflag.FlagSet, p *time.Duration) {
...
+	fs.StringVarP(&meta.Name, "cluster", "c", "", "EKS cluster name")
+}
+
+// AddClusterFlagWithDeprecated adds a common --cluster flag for
+// cluster name as well as a deprecated --name flag.
+// Use AddClusterFlag() for new commands.
+func AddClusterFlagWithDeprecated(fs *pflag.FlagSet, meta *api.ClusterMeta) {
+	AddClusterFlag(fs, meta)
 	fs.StringVarP(&meta.Name, "name", "n", "", "EKS cluster name")
+	_ = fs.MarkDeprecated("name", "use --cluster")
```
errordeveloper: How about we use `MakeHidden` and produce a warning when it is used to begin with. After 3 release we can switch to `MarkDeprecated`?

rndstr: > How about we use MakeHidden and produce a warning when it is used to begin with

my understanding is this is exactly what `MarkDeprecated` does

rndstr: it doesn't exit, see https://github.com/weaveworks/eksctl/pull/1363#issuecomment-537191382

our logger doesn't support writing to `stderr` (unless we enable something called `TestMode` that turns off colored output) which `MarkDeprecated` does and I would consider an important part to make sure not to break things for clients that parse the output for now.

also if we re-implement this ourselves it needs a fair amount of extra code and new people joining the project need to discover this. `MarkDeprecated` is known by everyone that knows cobra.

> it would be good make the warning more informative, to say that flag will be removed in a specific release.

can do this in a new PR, we have a couple deprecated flags. also not pressing to remove those IMHO.

martina-if: Mark deprecated just prints a warning but it allows the operation to continue. I just tested this. Here is an example:

```
$ eksctl create iamidentitymapping --name martina-test-4 --role arn:aws:iam::123:user/eksctl-testing-XYZ --group "system:something"
Flag --name has been deprecated, use --cluster
Flag --role has been deprecated, use --arn
[ℹ]  eksctl version 
[ℹ]  using region eu-north-1
[ℹ]  adding identity "arn:aws:iam::123:user/eksctl-testing-XYZ" to auth ConfigMap

$ eksctl create iamidentitymapping --cluster  martina-test-4 --role arn:aws:iam::123:user/eksctl-testing-XYZ --group "system:something"
Flag --role has been deprecated, use --arn
[ℹ]  eksctl version 
[ℹ]  using region eu-north-1
[!]  found existing mappings with same arn "arn:aws:iam::123:user/eksctl-testing-XYZ" (which will be shadowed by your new mapping)
[ℹ]  adding identity "arn:aws:iam::123:user/eksctl-testing-XYZ" to auth ConfigMap

```

I think with that we can proceed with the PR so I will merge it.

PR #1372, opened 2019-09-30
build: Update to Kubernetes 1.14.7 packages

dlespiau:
Ahead of starting to implement a Kubernetes controller with kubebuilder, we
need to update the Kubernetes packages to be compatible with
k8s.io/controller-runtime.

As always, updating from 1.12.x to 1.14.7 has quite a few repercussions:

- Updated most k8s.io/* packages to the kubernetes-1.14.7 tag

  go get k8s.io/apiextensions-apiserver@kubernetes-1.14.7
  go get k8s.io/api@kubernetes-1.14.7
  go get k8s.io/apimachinery@kubernetes-1.14.7
  go get k8s.io/apiserver@kubernetes-1.14.7
  go get k8s.io/client-go@kubernetes-1.14.7
  go get k8s.io/cli-runtime@kubernetes-1.14.7
  go get k8s.io/cloud-provider@kubernetes-1.14.7
  go get k8s.io/csi-translation-lib@kubernetes-1.14.7
  go get k8s.io/csi-api@kubernetes-1.14.7
  go get k8s.io/cloud-provider@kubernetes-1.14.7

- Had to bump a few other dependencies to compile:

  go get k8s.io/helm@v2.14.3
  github.com/Azure/go-autorest -> v11.1.0

- Kept (but updated) the replacement rules that were in place to pin various
  Kubernetes packages.

- A few things has to change in the code:

  * Use clienset.CoreV1() instead of clientset.Core() which doesn't exist
    anymore.
  * A couple of packages in cli-runtime had moved (actually Ilya did that!)
  * The cloud-provider package has moved out of Kubernetes


site/content/usage/20-schema.md
```diff
@@ -216,6 +227,21 @@ ListMeta:
     selfLink:
       type: string
   type: object
+ManagedFieldsEntry:
```
martina-if: Where did this changes come from?

dlespiau: This comes from the updated `metav1.ObjectMeta` struct that has changed between Kubernetes 1.12.x and 1.14.x.

The whole chain looks like:

- We use `ObjectMeta` in `ClusterIAMServiceAccount`:
    ```go
    type ClusterIAMServiceAccount struct {
            metav1.ObjectMeta `json:"metadata,omitempty"`
            // +optional
            AttachPolicyARNs []string `json:"attachPolicyARNs,omitempty"`
            // +optional
            AttachPolicy InlineDocument `json:"attachPolicy,omitempty"`
            // +optional
            Status *ClusterIAMServiceAccountStatus `json:"status,omitempty"`
    }
    ```
- `ObjectMeta` has a new field `managedFields`: https://github.com/kubernetes/apimachinery/blob/a9f1d8a9c10182d101acf19b5145c3d4e9299adb/pkg/apis/meta/v1/types.go#L267). This is reflected in the schema of `ObjectMeta`

- This new field is of type `ManagedFieldsEntry`, which needs the `Fields` type. Both of those are now in the schema as definitions.

martina-if: Aha, ok. Thanks for the detailed explanation :) I guess we shouldn't be reusing third-party objects in our API... :/ 

Please rebase and fix the conflicts and I'll approve :+1:

errordeveloper:
@dlespiau you will need to rebase this since #1200 was merged and you will need to follow instruction on rebuilding the build image per comments in `Dockerfile`.

Also, I am not sure if we can jump to 1.14, we do need to support 1.11 (until #1375)... what do you think?

dlespiau:
We hopefully should still be able to use create 1.11 clusters with 1.14 libraries, we were using 1.12.x packages anyway. Probably a good idea to test it though!

stefanprodan:
@dlespiau when you get back to this, please bump `github.com/fluxcd/helm-operator` to ` v1.0.0-rc3`

Thanks

martina-if:
@stefanprodan I remember you were interested in this. Do you think you have time to take this work over?

stefanprodan:
We've moved Flux and Helm Operator to Kubernetes 1.16 and I've tried to do the same with eksctl but no luck, the kops dependency has some pinned packages that I don't seem to match. I think we got into a point where eksctl is stuck doe to so many conflicting deps, maybe someone else could give it a try - I gave up ofter spending half a day on this.

martina-if:
@stefanprodan :( That's unfortunate. Thanks a lot for taking time to look into this! We will pick it up at some point.

PR #1389, opened 2019-10-03
Update Flux and Helm operator dependencies to `v1.15.0` and `v1.0.0-rc2`

### LINKED ISSUE #1385: Update Flux to 1.15 and enable Kustomize support
Today we've released Flux [v1.15.0](https://github.com/fluxcd/flux/blob/master/CHANGELOG.md#1150-2019-10-02) and the Kustomize support has been graduated from experimental to stable. 

We should update the go packages:
* `github.com/weaveworks/flux v0.0.0-20190729133003-c78ccd3706b5` > `github.com/fluxcd/flux v1.15.0`
* `github.com/fluxcd/helm-operator v1.0.0-rc1` > `github.com/fluxcd/helm-operator v1.0.0-rc2`

We should also enable Kustomize support by setting `--manifest-generation=true` in Flux command args.

hiddeco:
### Description

This PR updates the Flux and Flux Helm operator dependencies to their latest available releases (`v1.15.0` and `v1.0.0-rc2`), and closes #1385.

The Flux update includes a change in import paths, as all components have been
moved to `pkg/`, and Flux has moved away from the `weaveworks` GitHub organisation.


martina-if:
Hi @hiddeco ! Thanks for the PR! have you run some manual tests to check everything still works like before?

hiddeco:
@martina-if I have trouble with running `make build` and `make test` on my machine as `make` has a hard time finding `time`, which is literally used everywhere in the `Makefile`, so if you can run them for me that would be great.

(The reason the `make` commands fail is probably because I used `asdf` (which does some magic with `$PATH` to be able to switch versions) to downgrade my version of Go to `1.12.x`. Given this is a rather unusual setup, I would not spend much time on debugging it).

martina-if:
I manually tested `eksctl install flux....` and it works fine :+1:

PR #1412, opened 2019-10-07
WIP: Added --cache as authorizer option in kubeconf

gekart:
### Description

I think the credential cache option should be added to the aws-iam-authorizer call as a default, since awscli also uses caching by default. https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/193

This is how the resulting config would look after the PR:

```
       users:
        - name: kubelet@ferocious-mushroom-1532594698.us-west-2.eksctl.io
          user:
            exec:
              apiVersion: client.authentication.k8s.io/v1alpha1
              args:
              - token
              - -i
              - ferocious-mushroom-1532594698
              - --cache
              command: heptio-authenticator-aws
              env: null
```

Comments requested: Should I use this as the default or make it configurable? Then I'll complete the PR.

BTW, why doesn't everyone use roles & MFA for doing anything on AWS? Or am I missing something?


gekart:
Closed due to lack of comment.

PR #1420, opened 2019-10-08
Bump eksctl-dep image version to 0.17

martina-if:
Bump deps image to 0.17 after updating different dependency versions


rndstr:
I presume we need to bump it to have circleci rebuild the image and most importantly update `go-bindata` to generate its new format? because builds are failing right now

PR #1421, opened 2019-10-08
Rename `install flux` to `enable repo`

### LINKED ISSUE #1417: Rename `install flux` to `enable repo`
In order to unify the UI of the gitops commands `eksctl install flux` should be renamed to `eksctl enable repo`. That also makes the command's interface more agnostic to the underlying technology used for gitops.

martina-if:
closes #1417

This branch contains a commit from another PR that needs to be merged first: #1420

- [x] Manually tested


pkg/nodebootstrap/assets.go
```diff
@@ -1,16 +1,16 @@
-// Code generated by go-bindata. DO NOT EDIT.
```
rndstr: this reverts it to the old bindata version, no longer required after #1420

integration/creategetdelete_test.go
```diff
@@ -124,7 +124,7 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 			})
 		})
 
-		Context("and configuring Flux for gitops", func() {
+		Context("and enabling a gitops repo", func() {
```
marccarre: Nit: what does "_and enabling a gitops repo_" mean? 🤔 
Do we want to say "_enabling a repo for gitops_"? "_enabling gitops on the provided repo_"? Or so?
```suggestion
		Context("and enabling gitops on the configured repo", func() {
```

pkg/nodebootstrap/assets.go
martina-if: Thanks! I think I still had the old go-bindata locally. Fixed

integration/creategetdelete_test.go
martina-if: yeah... Now I dislike the word "enable" for this :smile: I went with your suggestion. Thanks!

marccarre: Yes, I don't like it too much either 😅 ... but we said we wouldn't change it so... 🤷‍♂

Issue #1282, opened 2019-09-05
eksctl version displays 0.2.0-xxx when built from the master branch


dlespiau:
**What happened?**

- Built `eksctl` from the `master` branch
- `./eksctl version` reports:
```
[ℹ]  version.Info{BuiltAt:"1567603534", GitCommit:"0.2.0-296-g94dfbe2", GitTag:""}
```

**What you expected to happen?**

I was expecting to see `0.5.0-2` in the GitCommit version

**How to reproduce it?**

- Checkout master
- make build
- ./eksctl version

errordeveloper:
Does it still do that when if you fetch all tags?

dlespiau:
No it doesn't change things.

dlespiau:
Also tried `--tags` at first in case you were using lightweight tags before realising the tags weren't on the master branch.

If you do a `git log --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit`, searching for `tag: ` shows `0.2.0` is the latest tag on `master`.

errordeveloper:
We should review release scripts and see what needs to be done to get tags onto master. I always assumed that tags weren't associated with branches and were just independent pointers to commits - TIL!

rndstr:
> assumed that tags weren't associated with branches and were just independent pointers to commits

Your assumption was correct. The issue at hand is that the commit that was tagged is not reachable from master. Just by merging the `release-X.Y` branch–which contains the tagged commit–back to master, the tag would be reachable and show up with `git describe`.

Having tags on master would be nice but not sure feasible without risking races.

PR #1425, opened 2019-10-08
Remove outdated and generated version/release.go

rndstr:
The file `pkg/version/release.go` is generated when doing a release
through `tag-release.sh` and also committed into its release branch but
never merged back to master.

The one in master is from before we changed this process and can be
confusing for third parties vendoring and seeing the code with an
outdated tag.


martina-if:
Should we just commit it when we release instead? It might be useful because then when we run `eksctl version` locally we would get the actual version instead of `[ℹ]  version.Info{BuiltAt:"1570548488", GitCommit:"0.2.0-565-gef183767", GitTag:""}`. https://github.com/weaveworks/eksctl/issues/1282

What do you think?

rndstr:
we already do commit it when we release but never merge it back into master i think. also the tag only lives in the `release-X.Y` branch that's why the `GitCommit` only finds 0.2.0 in parent commits (which is what `git describe` does)

martina-if:
can we commit it before branching out?

rndstr:
@martina-if that won't update it for bugfix releases

martina-if:
Ok, if @errordeveloper agrees I'm fine with merging this

Issue #1251, opened 2019-09-02
eksctl create times out when adding nodes to the cluster


bonelli:
**What happened?**
The creation of a new cluster with eksctl times out when adding nodes to the cluster. The EC2 instances are created, but nodes do not join the nodegroup.

**What you expected to happen?**
I expect the nodes to join the cluster and the creation to finish successfully.

**How to reproduce it?**
In the [bug-report.zip](https://github.com/weaveworks/eksctl/files/3566676/bug-report.zip) i have included the eksctl config file, the eksctl log and the cloudformation stack description taken after the moment in which eksctl failed with the timeout. 

**Anything else we need to know?**
Compiled binary of eksctl version 0.4.3
Linux Ubuntu 18.04.2 LTS

**Versions**

Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.4.3"}
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-21T15:34:43Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
In the [bug-report.zip](https://github.com/weaveworks/eksctl/files/3566676/bug-report.zip) i have included the eksctl config file, the eksctl log and the cloudformation stack description taken after the moment in which eksctl failed with the timeout.

achandak123:
👍 i am hitting the same issue. Here is the output of the describe nodes

```
Name:               ip-192-168-107-166.us-west-2.compute.internal
Roles:              <none>
Labels:             alpha.eksctl.io/cluster-name=spy
                    alpha.eksctl.io/instance-id=i-0d21257e1b4c1d3e8
                    alpha.eksctl.io/nodegroup-name=ng-2dd3b513
                    beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=m5.8xlarge
                    beta.kubernetes.io/os=linux
                    failure-domain.beta.kubernetes.io/region=us-west-2
                    failure-domain.beta.kubernetes.io/zone=us-west-2b
                    kubernetes.io/hostname=ip-192-168-107-166.us-west-2.compute.internal
Annotations:        node.alpha.kubernetes.io/ttl=0
[...]

PR #1426, opened 2019-10-08
Add correct Ubuntu bootstrap script

codyshepherd:
### Description

eksctl is injecting the wrong bootstrap script into ubuntu images, causing incorrect behavior of worker nodes running ubuntu EKS images. This PR adds the correct bootstrap script to the nodegroup assets. The following text describes testing with a k8s 1.12 configuration, but the bootstrap script will work for all Ubuntu EKS images built for k8s 1.11 and higher.

Please note that this PR represents a workaround only. The eksctl tool should not need to inject custom bootstrapping behavior into an official Ubuntu EKS image, as it comes with a bootstrap script baked in (the same one added by this PR). Further discussion on this topic is forthcoming.

I've tested these changes manually (description follows) with success. This is great, considering these tests fail when using the current stable version of eksctl, following the instructions on the AWS EKS getting started guide. However, `make build` produces a `pkg/nodebootstrap/assets.go` file different from the one here (in ways other than simply the addition of code artifacts as a result of the new bootstrap script), and I'm not quite sure why. I cannot be sure I am reproducing a production build of eksctl, so my test successes come with that caveat.

Testing:

1. Run the three `make` commands listed on the contributor readme for this project
2. Launch an EKS cluster with the following command:
```
./eksctl create cluster \
  --name <cluster name> \
  --version 1.12 \
  --nodegroup-name test-workers \
  --node-type m5.large \
  --nodes 10 \
  --nodes-min 10 \
  --nodes-max 10 \
  --node-ami ami-0551b6dc0b4079e1d \ # this is the official Ubuntu 18.04 ami for k8s 1.12 in us-west-2: https://cloud-images.ubuntu.com/aws-eks/
  --node-ami-family Ubuntu1804 \
  --region us-west-2 \
  --ssh-access \
  --ssh-public-key <AWS .pem key>
```
3. Wait for cluster launch to complete
4. Install kubectl
`sudo snap install kubectl --classic --channel 1.12/stable`
5. Verify connectivity
`kubectl get all --all-namespaces`
6. Install/Config/Run the tests (should take < 2 hours):
```
sudo snap install kubernetes-test-eks --classic --channel 1.12.9/stable
export SERVER=$(cat $HOME/.kube/config | grep server | sed 's/    server: //') ## (Note that this only works if you have only one cluster listed in your config)
export FOCUS="\\[Conformance\\]"
export SKIP="\\[Flaky\\]|\\[Serial\\]"
export ACTION_JUNIT=$HOME/eks/e2e
export ACTION_LOG=$HOME/eks/e2e.log
export KUBECONFIG=$HOME/.kube/config
export PARALLELISM=10 ## ← how many e2e threads to run
mkdir -p $HOME/eks/e2e
GINKGO_ARGS="-nodes=$PARALLELISM" kubernetes-test-eks.e2e \
  -kubeconfig $HOME/.kube/config \
  -host $SERVER \
  -ginkgo.focus "$FOCUS" \
  -ginkgo.skip "$SKIP" \
  -report-dir $ACTION_JUNIT 2>&1 | tee -a $ACTION_LOG
```
Success: 171 Passed, 7 Failed, 0 Pending, 1834 Skipped
Failure: 10 or more tests fail

Testing using the current stable eksctl package results in failure due to, perhaps among other problems, an apparent misconfiguration of CA configs, which results in worker node communication failures.

### Checklist
- [x] Manually tested


codyshepherd:
Possibly a fix for #1251

errordeveloper:
@codyshepherd thanks for opening this up, I'll look into the details soon. In the meantime, could you please confirmed whether you have reviewed [the current design of node bootstrap process](https://github.com/weaveworks/eksctl/blob/master/docs/current-001-node_bootstrap.md)?

codyshepherd:
@errordeveloper Ah I had not seen that doc. I'll be sure to review, thank you.

errordeveloper:
@codyshepherd I cannot see how those test failures would relate to node configuration. Generally speaking, I believe that any of Kubernetes-level test that assumes a certain node runtime configuration would be flawed, and I doubt there are any such tests in the end-to-end suite.

errordeveloper:
From reading the output you share, it looks like `kubectl logs` and `kubectl exec` are failing due to kubelet presenting a bad server certificate, that functionality is part of kubelet and isn't depended on the bootstrap script as such.

Could you please run `eksctl create cluster --node-ami-family=Ubuntu1804 --version=1.12` and simply test `kubectl logs` and  `kubectl exec` on any pod? If that is broken, please open an issue as a bug, and we will deal with that.
If you also manage to reproduce failures in e2e tests, please open an issue as well, especially if those results contradict manual `kubectl logs` and `kubectl exec` tests.

codyshepherd:
@errordeveloper I've rolled back through several eksctl versions and re-run tests, and discovered that these errors were occurring with version 0.3.1. I'm not prepared to say which change since that version has seemed to have alleviated the test failures mentioned above.

I will work on a better test plan in support of this PR, as the fact remains that the kubelet is being misconfigured (compared to the baked-in script) by the eksctl bootstrap script in at least the following ways:

* CA certificate path
* kubeconfig path
* cluster DNS IP, which is dependent upon the VPC subnet cidr block

And possibly:

* api server endpoint
* max-pods per instance type
* enabling docker bridge network

errordeveloper:
I don't agree that kubelet is misconfigured, it's configured correctly, just that the configuration process is designed differently (for reason outlined in [the design document](https://github.com/weaveworks/eksctl/blob/master/docs/current-001-node_bootstrap.md)).

> - CA certificate path
> - kubeconfig path

Both of these are different from what Ubuntu script uses, however kubelet flags are set accordingly to accommodate for these. Per design document, the aim is to contain all of the files in a single directory ( `/etc/eksctl`) which is fully owned by eksctl.

> - cluster DNS IP, which is dependent upon the VPC subnet cidr block

This is take care of externally from the node, again per design document, the goal is to prepare all of configuration outside of the node itself.

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/userdata.go#L108-L110

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/userdata.go#L83-L93

> - api server endpoint

It is set as part of kubeconfig.

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/userdata.go#L69-L81

> - max-pods per instance type

This is done as part of bootstrap process, in order to support ASGs with mix instance types.

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/assets/bootstrap.ubuntu.sh#L7-L14

The lookup table is generated in a format that is easier to parse and is determined by version of eksctl (not the AMI).

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/userdata.go#L161-L167

https://github.com/weaveworks/eksctl/blob/36ca6e0ac4998ea299222104b9ef93f8732f5a0e/pkg/nodebootstrap/userdata_ubuntu.go#L40




> - enabling docker bridge network

We don't support customisation to Docker daemon configuration at present, but adding logic inside the bootstrap script is not going to serve support for this. The path forward would be to expose Docker daemon config file in eksctl config file (similarly to how we do it for [kubelet](https://eksctl.io/usage/customizing-the-kubelet/)).

codyshepherd:
Thank you for that breakdown, that was very helpful. I should have clarified "misconfigured;" as you said, the kubelet configuration is different than that of the "original" boot script, and points at eksctl-controlled artifacts rather than those shipped on the image and produced by Canonical. Perhaps my PR is inappropriate, given the design philosophy you have pointed to above. But then I think there are some remaining concerns about how far a departure this represents, or could represent as the Ubuntu image evolves, from a) the Ubuntu image when used with an EKS cluster created manually and b) the expected behavior of the Ubuntu image as shipped.

This PR can probably be closed, and I'll open an issue if I can articulate some tests that distinguish the concerns described above.

errordeveloper:
@codyshepherd yes, let's discuss those concerns separately!

Also, something I forgot to mention that you may not be aware of, is that config file has `overrideBootstrapCommand` and `preBootstrapCommands`.
For the purpose of your test, you can create a cluster config that specifically calls the script that is shipped in the AMI.

```
--- 
kind: ClusterConfig

metadata:
  name: test-cluster-1
  region: eu-west-1
  version: 1.13

nodeGroups:

  - name: ng1
    instanceType: m5.large
    desiredCapacity: 2
    volumeSize: 100
    ami: ami-073dafa91555a49ca
    amiFamily: Ubuntu1804

  - name: ng2
    instanceType: m5.large
    desiredCapacity: 2
    volumeSize: 100
    ami: ami-073dafa91555a49ca
    amiFamily: Ubuntu1804
    overrideBootstrapCommand:  /etc/eks/bootstrap.sh  test-cluster-1
```

I haven't tested this configuration, but I think it should work. If it doesn't, see if some IAM permissions need to be enabled, you can set ` iam: { attachPolicyARNs: [ ] }` to add any extra policy ARNs to a nodegroup. In any case, such configuration should result in 2 nodegroups (one using bootstrap script provided by `eksctl`, and the other using the script provided by the AMI).

PR #1427, opened 2019-10-08
Generate nodebootstrap assets with new go-bindata

rndstr:
After upgrading go-bindata (#1395) and fixing the circleci job to rebuild the
eksctl-dep image (#1420) we also need to update those assets to use the new
go-bindata format to make the checks pass everywhere.


rndstr:
turns out we already have the latest assets, this PR would be reverting them.

Issue #1358, opened 2019-09-23
Make VPC private endpoint access work within VPC


D3nn:
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
A clear and concise description of why you want this feature. Include use-cases and any specific constraints you may have.

**What feature/behavior/change do you want?**
A clear and concise description of the change you would like to see in eksctl.

Do not hesitate, when appropriate, to share the exact commands or API you would like, and/or to share a diagram (e.g.: asciiflow.com): "a picture is worth a thousand words".

D3nn:
When enabling access to the Kubernetes API endpoint from within the cluster, make the changes necessary for allowing such access within the VPC (security groups, DNS updates, etc).  Currently, we are pointing customers at the documentation for them to enable this access, but would be better if we could manage this for them, at least for nodes inside the VPC.

morinap:
@D3nn I was discussing this use case on the recently resolved https://github.com/weaveworks/eksctl/issues/649 and have just opened a PR that takes one simple approach to address this. I'm definitely willing to work on a more robust approach as well if desired (such as dynamically adding the ingress rules to a eksctl-created security group), but this approach worked perfectly for my use case.

D3nn:
Scenarios we would like to handle for the user rather than forcing them to make changes:
1.  Worker nodes should be able to make Kubernetes API calls
1. Private access through VPC peering needs to forward DNS to the EKS Control Plane Private Route53 DNS domain.  Control Plane security group updated for API access from VPC peer network(s)
1. VPN access from customer sites should be able to resolve Endpoint from the EKS Control Plane Private Route53 DNS domain.  Control Plane security updated for API access from corp networks
1. DirectConnect access from customer sites should be able to resolve Endpoint from EKS Control Plane Private Route 53 DNS domain.  Control Plane security updated for API access from corp networks.

PR #1434, opened 2019-10-10
Support Create of Cluster with Private VPC Endpoint

morinap:
### Description
This enables the cluster to be created without public access while allowing the creator to specify a control plane security group that can allow bastion or other hosts on the VPC to access the control plane for further eksctl calls.

This works well for our use case; I created a security group that allows HTTPS access inbound from our bastion host before creating the cluster, and then I specified that security group under `VPC` in my cluster create YML.

This addresses https://github.com/weaveworks/eksctl/issues/1358

### Checklist
- [x] Manually tested

### DIFF
--- a/humans.txt
+++ b/humans.txt
@@ -59,6 +59,7 @@ Hidde Beydals           @hiddeco
 Cedric Kring            @cedrickring
 Tam Mach                @sayboras
 Leigh Capili            @stealthybox
+Andrew Morin            @morinap
 
 /* Thanks */
 
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -20,7 +20,9 @@ var (
 		"commands/API calls from within the VPC.  Running these in the VPC requires making " +
 		"updates to some AWS resources.  See: " +
 		"https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html#private-access " +
-		"for more details")
+		"for more details. Specifying a SecurityGroup under the VPC configuration will suppress " +
+		"this error; the specified SecurityGroup can be used to allow eksctl traffic from a " +
+		"bastion host, for example.")
 )
 
 // NOTE: we don't use k8s.io/apimachinery/pkg/util/sets here to keep API package free of dependencies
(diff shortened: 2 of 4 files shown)


D3nn:
Someone creating a new EKS cluster using the PR while NOT using an existing VPC/Bastion Host 
to run eksctl will currently fail since eksctl will not have access to the Kubernetes API to join the nodes to the security group as it now stands, at least not in the scenario where public access is disabled.

There are a couple of solutions where I think a single command approach might suffice:
1. This PR, but with up-front verification of prerequisites mentioned above being met (and possibly the newly created security group being used (and then discarded) in the creation of the ControlPlaneSecurityGroup).
1. We delay setting Public access to false until the worker nodes have been joined to the cluster. (Investigating this one currently)

morinap:
@D3nn Thank you for the feedback; I can start down the path of meeting the 3 requisites you mentioned. Very good point about creating a cluster without an existing VPC; that's not my use case, and obviously it isn't something I'd considered because of that.

Just a couple of notes/questions:
* While it's true that the `ControlPlaneSecurityGroup` is created as part of cluster creation, I'm simply leveraging an existing code path that instead of creating that group modifies the specified group when one is specified in the VPC spec.
* Instead of allowing specification of a security group, the other (more complex) path we'd considered was to allow specification of a set of ingress rules to be automatically added to the `ControlPlaneSecurityGroup` when it's created. Does that seem more in line with a more robust solution you're looking for? And if so, any suggestions on the structure of such a definition?

D3nn:
There are a few things that need to be fixed so we can merge:
1. There are generated files that no longer matching the previously existing generated files.  For these, you'll need to check in the new generated files to fix that circleci build error.
1. There are a few conflicts (humans.txt and site/content/usage/06-vpc-networking.md) which need to be resolved.

Please make these updatess so we can validate that circleci isn't failing (at least for these issues)

morinap:
@D3nn I'm in the midst of some other things and may not have further time to work on this in the near future. Would you prefer for managing the project for me to leave this open and come back to it when I'm ready, or close it and open a new PR when I can address your feedback?

D3nn:
Thanks for the update @morinap.  I have no issues leaving this open for the time being.

martina-if:
Hi, I am doing a little bit of cleaning so I will be closing this but please feel free to reopen it whenever you have time to work on this again :)

Issue #1436, opened 2019-10-10
Add revision option for quick start profiles 


stefanprodan:
Add the `--revision` arg to the `eksctl enable profile` command, the revision would allow users to use sem ver releases or stable branches of profile repositories e.g. 
```
eksctl enable profile appmesh --revision=v1.0.0
```

PR #1437, opened 2019-10-10
Add revision arg to enable profile command

stefanprodan:
This allows pulling a specific branch or release of a profile, e.g.

```sh
eksctl enable profile appmesh --revision 1.0.0
```

Fix: #1436 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/enable/profile.go
+++ b/pkg/ctl/enable/profile.go
@@ -27,13 +27,14 @@ import (
 
 type options struct {
 	gitOptions           git.Options
-	quickstartNameArg    string
+	profileNameArg       string
+	profileRevision      string
 	gitPrivateSSHKeyPath string
 }
 
 func (opts options) validate() error {
-	if opts.quickstartNameArg == "" {
-		return errors.New("please supply a valid gitops Quick Start URL or name")
+	if opts.profileNameArg == "" {
+		return errors.New("please supply a valid Quick Start profile URL or name")
 	}
 	if err := opts.gitOptions.ValidateURL(); err != nil {
 		return errors.Wrap(err, "please supply a valid --git-url argument")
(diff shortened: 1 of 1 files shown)


pkg/ctl/enable/profile.go
```diff
@@ -27,13 +27,14 @@ import (
 
 type options struct {
 	gitOptions           git.Options
-	quickstartNameArg    string
+	profileName          string
```
martina-if: This is not really the name of the profile. It can be its URL. How about `profileNameArg`?

pkg/ctl/enable/profile.go
```diff
@@ -156,8 +158,8 @@ func doEnableProfile(cmd *cmdutils.Cmd, opts options) error {
 		Processor: processor,
 		Path:      profileOutputPath,
 		GitOpts: git.Options{
-			URL:    quickstartRepoURL,
-			Branch: "master",
```
martina-if: Oh I see now this was broken. Thanks!

pkg/ctl/enable/profile.go
```diff
@@ -57,14 +58,15 @@ func enableProfileCmd(cmd *cmdutils.Cmd) {
 	})
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
-		fs.StringVarP(&opts.quickstartNameArg, "name", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.profileName, "name", "", "", "name or URL of the Quick Start profile. For example, app-dev.")
+		fs.StringVarP(&opts.profileRevision, "revision", "", "master", "revision of the Quick Start profile.")
```
martina-if: Please remove this new flag and use `--git-branch` from line 64.

martina-if:
Thanks for all the improvements in the logging!

pkg/ctl/enable/profile.go
stefanprodan: The revision is for the profile repo not the destination one. Please see #1436 on why I've added this.

Issue #1416, opened 2019-10-08
Simplify `eksctl enable profile`


martina-if:
The functionality of this command is to install flux to set up gitops and to install the components of a quickstart profile on the cluster.

This two actions will be performed by `eksctl create cluster -f cluster-01.yaml` when #1415 is finished. After that is done, `eksctl enable profile` should just perform the second action: install quickstart components on the cluster by adding them to a gitops repo.

This way the command will be doing one thing instead of two different things. Another advantage is that it can allow applying multiple profiles to the same cluster in the future.

stefanprodan:
I think `enable profile` should check if Flux is running and if it isn't, only then install it. 

How would users upgrade Flux and Helm Operator after installing them with eksctl? For minor releases `kubectl set image` should work but it's not gitopsy.

marccarre:
@stefanprodan, new issue created to address this: https://github.com/weaveworks/eksctl/issues/1534

Issue #1415, opened 2019-10-08
Add support for quickstart profiles to ClusterConfig manifests for a more declarative way to create cluster


martina-if:
Today eksctl has the experimental feature `eksctl enable profile` that sets up a gitops repo with flux and then installs the components from a specified quickstart profile.

We would like that functionality to be moved to the `create cluster -f` command so that:

```cluster-1.yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
 name: cluster-1
 region: eu-north-1

nodeGroups:
 - name: ng-1
   instanceType: m5.large
   desiredCapacity: 1

gitops:
  repo:
    url: git@github.com:myorg/cluster-prod
    branch: master
    paths: [ "kube-system", "base" ]
    fluxSubdir: "flux/"
    user: john
    email: john@myorg.com
    privateSshKeyPath: ~/.ssh/.id_rsa_flux
 operator:
    label: flux
    namespace: flux
     withHelm: true
 bootstrapProfiles:
   - source: app-dev
     revision: master
```
```
eksctl create cluster -f cluster-1.yaml
```

Would create a cluster, set up gitops on `git@github.com:myorg/cluster-prod` and install the components from the `app-dev` quickstart profile.

(The next step will be to simplify the command `eksctl enable profile` so that it only installs profile components on the cluster and it doesn't install flux anymore: #1416)

martina-if:
- `enable profile` -> if flux is not installed it should do something reasonable: either install flux or fail to install the profile. See https://github.com/weaveworks/eksctl/issues/1534

PR #1440, opened 2019-10-11
Add GitOps-related types to ClusterConfig

marccarre:
### Description

Pre-requisite for #1415 and #1416.
Add GitOps-related types to `ClusterConfig`, e.g.:

```yaml
kind: ClusterConfig
git:
  repo:
    url: git@github.com:org/cluster-prod
    branch: master
  operator:
    label: flux
    namespace: flux
    withHelm: true
  profiles:
   - source: app-dev
     revision: master
   - source: https://github.com/aws/eks-templates/templates/app-dev
     revision: 2.0.0-preview.1
```

### Checklist
- [x] ~Added tests that cover your change (if possible)~
- [x] ~Manually tested~ Ran `make test`


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,56 @@ type (
...
+// GitOps.
+type Repo struct {
+	// +optional
+	URL string `json:"url,omitempty"` // the Git SSH URL to the repository which will contain the cluster configuration, e.g. git@github.com:org/repo
+	// +optional
+	Branch string `json:"branch,omitempty"` // the branch under which cluster configuration files will be committed & pushed, e.g. master
+	// +optional
+	Paths []string `json:"paths,omitempty"` // the paths to the directories under which cluster configuration files will be written, e.g. ["kube-system", "base"]
+	// +optional
+	FluxSubDir string `json:"fluxSubDir,omitempty"` // the directory under which Flux configuration files will be written, e.g. flux/
```
stefanprodan:
```suggestion
	FluxDir string `json:"fluxDir,omitempty"` // the directory under which Flux configuration files will be written, e.g. flux/
```

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,52 @@ type (
 	}
 )
 
+// Git groups all configuration options related to enabling GitOps on a
+// cluster and linking it to a Git repository.
+type Git struct {
+	// +optional
```
stefanprodan: Repo should be required I think.

site/content/usage/20-schema.md
```diff
@@ -504,6 +532,34 @@ OwnerReference:
...
+      type: string
+  type: object
+Repo:
+  additionalProperties: false
+  properties:
+    branch:
+      type: string
+    email:
+      type: string
+    fluxSubDir:
```
stefanprodan: Rename this one as well to `fluxDir`, I though this is was generated...

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
+}
+
+// Repo groups all configuration options related to a Git repository used for
+// GitOps.
+type Repo struct {
+	URL string `json:"url,omitempty"` // the Git SSH URL to the repository which will contain the cluster configuration, e.g. git@github.com:org/repo
+	// +optional
+	Branch string `json:"branch,omitempty"` // the branch under which cluster configuration files will be committed & pushed, e.g. master
+	// +optional
+	Paths []string `json:"paths,omitempty"` // the paths to the directories under which cluster configuration files will be written, e.g. ["kube-system", "base"]
```
martina-if: I would explain this as "the paths to the directories that will be scanned to look for manifests, e.g...."

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
+	// +optional
+	FluxDir string `json:"fluxDir,omitempty"` // the directory under which Flux configuration files will be written, e.g. flux/
+	User    string `json:"user,omitempty"`    // Git user to authenticate as to pull/commit/push changes
+	Email   string `json:"email,omitempty"`   // email of the Git user to authenticate as to pull/commit/push changes
```
martina-if: IIRC this user and email are used only for commiting

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
+	// +optional
+	PrivateSSHKeyPath string `json:"privateSshKeyPath,omitempty"` // path to the private SSH key to use to authenticate
```
errordeveloper: It'd have to be `privateSSHKeyPath`, but let's make it a object that allow user to specify body or path. Body would be desirable for users who want to ensure the configuration file is fully deterministic.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
 )
 
+// Git groups all configuration options related to enabling GitOps on a
+// cluster and linking it to a Git repository.
+type Git struct {
+	Repo *Repo `json:"repo,omitempty"`
+	// +optional
+	Operator *Operator `json:"operator,omitempty"`
+	// +optional
+	Profiles []*Profile `json:"profiles,omitempty"` // one or many profiles to enable on this cluster
```
errordeveloper: Have we already implemented support for multiple profiles? I was hoping we could avoid that and keep it simple by only allowing a single profile per cluster. If we haven't implemented that yet, let's drop it from here.

errordeveloper:
I think it's worth considering to take controller-first perspective.

I previously noted (in various places) the following idea:
```
controllers:
  - kind: FluxController
    metadata: {name: flux-cluster-admin}
    spec:
      gitURL: git@github.com:JaneJane/gitops-examples
 - kind: ClusterController
    metadata: {name: eksctld}
    spec:
      mode: SelfManagedCluster
```

This accommodates Cluster API controller, and Flux.

I am not entirely sure of profiles should be exposed directly via the configuration file, as fundamentally everything in config file should be actionable on by `eksctl apply`.

There are things that won't be actionable on due to limitation in AWS APIs, e.g. adding new subnets to the cluster is currently not possible as EKS simply doesn't support that (yet). However, we should introduce new special cases ourselves if we can avoid them.

Consider the following scenario.

A simple config file was given on cluster creation:

https://github.com/weaveworks/eksctl/blob/30941eccc756b08581a978387e355d6770937a14/examples/01-simple-cluster.yaml#L3-L13

After cluster was created user adds configuration that specifies that flux should be installed and pointed to a git repository, user runs `eksctl apply` and Flux gets configured.

Next user pushes somme changes to a repository, perhaps adds some Kubernetes objects they'd like to deploy. Flux picks those up.

After that user finds that there exists a profile that suites their needs, they try add the profile to the config file, and wish to run `eksctl apply`...

(Perhaps it's not that hard to support this case, let's keep going.)

User makes changes to files that were created from the profile, Flux picks those up.

Days later user realises profile has evolved, they set new revision on the cluster config file and run `eksctl apply`. What is supposed to happen now? How are merge conflicts going to be resolved? If we are to allow multiple profiles, how overlaying is going to work and how do we prevent user mistakes as we have offered so much complexity?

I thin this brings too many new questions, let's try to avoid this for now and introduce Flux controller configuration first.

I would strongly recommend exploring alternatives to how profiles work. Perhaps profile definition should be defined by a file in the actual repo, and generated by a process other then `eksctl` itself.

I believe `eksctl enable profile` is meant to be simple to use, user should have to write complex config files before they can try a profile that way. But once they are to write a config file, we can step up and consider a more appropriate design, e.g. you would need to have a repo and have some files in it, perhaps `profile.yaml` or something like that. If user makes changes to the `profile.yaml` in the repo, Flux picks them up, and perhaps it runs the profile generation tool do all of the operations needed to generate a profile. We should also keep an open mind to what profile generation tool would like, we've discussed using jk and possible a remote service.

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: Remember that this is a private key, not public

errordeveloper: Yes, but user should still be able to specify the body, it's up to them how they secure the config file itself.

martina-if: I disagree. On one hand we try to promote gitops as the best way of working with kubernetes. By allowing this, on the other hand,  it looks like we are saying it's ok to do that which is quite insecure. I think we should be consistent and promote good practices.

site/content/usage/20-schema.md
marccarre: It is supposed to be generated, yes, but this doesn't seem to consistently re-generate things on my machine, not sure why/what I'm doing wrong! 🤔

marccarre: OK, it seems to generate it under `github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go` instead of `pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go` for some reason. I'll try to debug this and fix it in another PR.

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: Fixed.

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: Fixed.

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: `privateSshKeyPath` -> `privateSSHKeyPath` fixed 👍 
For the rest, I agree with @martina-if, so don't plan to change it for now.

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: I believe this is now a requirement (cc @kalbir).

Besides, it seems legit to me, in order to mix and match simple profiles instead of having a combinatorial explosion of heavyweight profiles each with small differences (e.g. "_oh, but I prefer to use X than Y, let's fork and change just this bit, and keep the 50 other tools as is_").

It is kind of similar to "_composition_" v.s. "_inheritance_": either you compose what you need in your `ClusterConfig` or you have to come up with the perfect profile for everyone.

marccarre:
**A.**:

> I think it's worth considering to take controller-first perspective.
> ```yaml
> controllers:
>   - kind: FluxController
>     metadata: {name: flux-cluster-admin}
>     spec:
>       gitURL: git@github.com:JaneJane/gitops-examples
>  - kind: ClusterController
>     metadata: {name: eksctld}
>     spec:
>       mode: SelfManagedCluster
> ```

Why? Would you mind linking the relevant issues and/or documents @errordeveloper?
Is there any other place where we're already using these or plan to very soon? (cc @dlespiau)

----
**B.**:

> I am not entirely sure of profiles should be exposed directly via the configuration file, as fundamentally everything in config file should be actionable on by `eksctl apply`.

Why would the two be incompatible?
We currently don't do anything to prevent errors if an user runs `eksctl generate profile` twice in a row. How is it different? (I'm not saying these issues don't exist, but just that we cannot solve all of them here, and that they are out of scope of this PR imho.)

----
**C.**:

> There are things that won't be actionable on due to limitation in AWS APIs, e.g. adding new subnets

This is unrelated to adding the new GitOps-related fields that this PR introduces, though, isn't it?

----
**D.**: 

> I believe eksctl enable profile is meant to be simple to use, user should have to write complex config files before they can try a profile that way.

They don't have to. They can still use the CLI arguments instead. This PR aims just to have the file-equivalent of the CLI arguments, something you yourself previously said would be valuable if I recall and understood correctly.

---- 
**E.**: Many of your points are completely valid, but we cannot possibly tackle all possible issues raised by the introduction of an `eksctl apply` command-to-be here. Indeed, there are many issues caused by limitations of the AWS API, of non-idempotent behaviour of `eksctl`, or simply unsolvable situations (e.g. conflicts in profiles). 

Here, I'm trying to unblock us for #1415 and #1416, and to at least have the changes required to enable:

- a YAML equivalent of the CLI arguments for the GitOps commands
- a more user-friendly `eksctl enable repo` (callable using either a YAML file or CLI arguments), just installing & configuring a GitOps operator on the current cluster.
- a more user-friendly `eksctl enable profile` (callable using either a YAML file or CLI arguments), just installing a profile to an already GitOps-enabled cluster.
- a `eksctl create cluster -f ...` which would do the equivalent of the two aforementioned commands, and should be semantically better than what we currently have.

We can always (and will have to!) refine this iteratively, and especially when `eksctl apply` is introduced. Likewise, there are many corner cases to consider in the implementation of #1415, #1416, et al. I just would rather not tackle nor discuss all of this here, as it may prevent any progress at all.

pkg/apis/eksctl.io/v1alpha5/types.go
kalbir: Yes the intention was to have some simple predictable behaviour here. Not any smart composition but just something that we can communicate to people e.g. if you add a second profile it will overwrite files in the first.

martina-if: Note that even if the API has a list it doesn't mean we need to support it. It could be done just for the shake of "extendability" and forward compatibility. If we are unsure of the effects of adding more than one profile or we don't want to implement it just yet, we can always return an error when there is more than one item.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
+// Repo groups all configuration options related to a Git repository used for
+// GitOps.
+type Repo struct {
+	URL string `json:"url,omitempty"` // the Git SSH URL to the repository which will contain the cluster configuration, e.g. git@github.com:org/repo
+	// +optional
+	Branch string `json:"branch,omitempty"` // the branch under which cluster configuration files will be committed & pushed, e.g. master
+	// +optional
+	Paths []string `json:"paths,omitempty"` // relative paths within the Git repository which the GitOps operator will monitor to find Kubernetes manifests to apply, e.g. ["kube-system", "base"]
+	// +optional
+	FluxDir string `json:"fluxDir,omitempty"` // the directory under which Flux configuration files will be written, e.g. flux/
```
martina-if: I think roli suggested to rename this to "FluxPath" so it's more consistent with "paths" and other flags that relate to directories and files. What do you think?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -628,6 +631,51 @@ type (
...
+	User    string `json:"user,omitempty"`    // Git user which will be used to commit changes
```
martina-if: I think the user, technically speaking, was optional

errordeveloper:
> **A.**:

> > I think it's worth considering to take controller-first perspective.
> > ```yaml
> > controllers:
> >   - kind: FluxController
> >     metadata: {name: flux-cluster-admin}
> >     spec:
> >       gitURL: git@github.com:JaneJane/gitops-examples
> >  - kind: ClusterController
> >     metadata: {name: eksctld}
> >     spec:
> >       mode: SelfManagedCluster
> > ```
> 

> Why? Would you mind linking the relevant issues and/or documents @errordeveloper?
> Is there any other place where we're already using these or plan to very soon? (cc @dlespiau)

I have suggested this on a few internal threads.

The reasoning behind this is as follows:

1. A way of configuring flux in a cluster is required
   - there are multiple modes of operation (single namespace, subset of namespaces, all of namespaces)
   - it should be possible to easily configure multiple instance in the same cluster, each in a different mode; e.g. a common use-case is to have an instance of Flux that manages infrastructure like ingress controller and other platform components and a separate instance that manages applications
2. A way of configuring Cluster API controller is required
   - there are multiple modes of operation (e.g. whether controller's aim is to manage only the cluster it is running inside of, or only other cluster)


So with 1 in mind, it's essential to provide a way for configuring multiple instances of Flux, and not just one.

----
**B.**:

> I am not entirely sure of profiles should be exposed directly via the configuration file, as fundamentally everything in config file should be actionable on by `eksctl apply`.

Why would the two be incompatible?
We currently don't do anything to prevent errors if an user runs `eksctl generate profile` twice in a row. How is it different? (I'm not saying these issues don't exist, but just that we cannot solve all of them here, and that they are out of scope of this PR imho.)

Yes, this PR only introduces new fields. However, by introducing these new fields we commit to implementing (not immediately, of course) all of the behaviours that these fields imply. We shouldn't introduce something where we don't see a clear path towards the implementation.

At present we manage to two types of resources through AWS and Kubernetes APIs.
The current implementation of AWS resource management has a few limitation, but there exists a clear path forward to implementing a fully-declarative solution. And same goes for Kubernetes APIs.
In general sense, we have most of CRUD operations implemented already and those can be generalised to provide what's needed for `eksctl apply`.

When it comes to profiles, we only have a way to render a profile to a repo, there is no update capability, and we don't even know what files belong to us and what files don't. I don't yet see a way forward for having all CRUD operations for files that are rendered based on a given profile. I also don't see a particularly good reason for going down this route, I'd be very happy if we only provide the capability of bootstrapping a repo, and I don't think that even has to be part of `eksctl` CLI as such, it could be decoupled, e.g. as a service or as jk program (or any other program) that gets executed inside of Flux pod (and I'm sure there are more options to explore).

----
**C.**:

>> There are things that won't be actionable on due to limitation in AWS APIs, e.g. adding new subnets

> This is unrelated to adding the new GitOps-related fields that this PR introduces, though, isn't it?

Yes, this is just to say that we already have some special cases to deal with, and we don't need to add more special cases ourselves.

----
**D.**: 

>> I believe eksctl enable profile is meant to be simple to use, user should have to write complex config files before they can try a profile that way.

> They don't have to. They can still use the CLI arguments instead. This PR aims just to have the file-equivalent of the CLI arguments, something you yourself previously said would be valuable if I recall and understood correctly.

Certainly, but now in the context of this I can see how managing files in a repo after the initial bootstrap is going to be a major challenge that we don't really have to take up. I am still advocating for Flux controller options to be exposed, but that is secondary to what's proposed in this PR.

---- 

> **E.**: Many of your points are completely valid, but we cannot possibly tackle all possible issues raised by the introduction of an `eksctl apply` command-to-be here. Indeed, there are many issues caused by limitations of the AWS API, of non-idempotent behaviour of `eksctl`, or simply unsolvable situations (e.g. conflicts in profiles). 
> 
> Here, I'm trying to unblock us for #1415 and #1416, and to at least have the changes required to enable:
> 
> - a YAML equivalent of the CLI arguments for the GitOps commands
> - a more user-friendly `eksctl enable repo` (callable using either a YAML file or CLI arguments), just installing & configuring a GitOps operator on the current cluster.
> - a more user-friendly `eksctl enable profile` (callable using either a YAML file or CLI arguments), just installing a profile to an already GitOps-enabled cluster.
> - a `eksctl create cluster -f ...` which would do the equivalent of the two aforementioned commands, and should be semantically better than what we currently have.
> 
> We can always (and will have to!) refine this iteratively, and especially when `eksctl apply` is introduced. Likewise, there are many corner cases to consider in the implementation of #1415, #1416, et al. I just would rather not tackle nor discuss all of this here, as it may prevent any progress at all.

I agree, we need a way forward, and I think that Flux-first approach will be sufficient.

Once we have the structure for Flux controller options, we can consider two routes:
- add a field in Flux controller config to set a _bootstrap profile_ spec, like `bootstrapFromProfile`, or anything that is sufficiently explicitly about _bootstrap_, and doesn't lead user to think that they could swapping the profile in the future or do anything else
- make profile rendering a completely separate concern and use in-repo config file to declare whatever configuration is (e.g. with `package.json` if we used jk, or some sort of `profile.yaml`)

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: Sure, why not 🙂

pkg/apis/eksctl.io/v1alpha5/types.go
marccarre: `Email` as well then, right?

marccarre:
**A. 1.**: Multiple Flux: I think this deserves its own issue. I'll be happy to have a look into it if this becomes a real requirement and will refine the API accordingly.

**A. 2.**: `FluxController`: sorry but I'm currently not comfortable changing the API all over again, given it was already discussed with several stakeholders, and we agreed on the current shape. I'm especially _not_ comfortable doing it given it mentions Flux, and we're trying to make `eksctl` more neutral. But again, happy to tackle this as part of another issue, in a different iteration and/or version of the API.

**B.**: I think it is actually clear how we will use these fields given the current implementation of the GitOps-related commands. As for future changes (incl. `eksctl apply`), I think this should be tackled separately, i.e. in other issues.

**D.**: Actually, if we go for what was suggested in [this comment](https://github.com/weaveworks/eksctl/pull/1440#discussion_r334796973), then there is also a clean diff between the commits from the previous profile and the ones from the current profile, so we should have conflicts. As for other operations on these files, I think it could be fine not to support them and have a warning for now. E.g. we only have `eksctl enable profile` at the moment, not `eksctl disable profile` or so, and it could be OK to let the end-user then delete things in their repository directly in my opinion. At least for now, we can discuss all this in the issue for `eksctl apply` or so.

**E.**: This isn't what was agreed with other stakeholders though 😞

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: No, I think the email was marked as required. But I had an idea on how to make it optional and read it so I would say it's required for now and I'll try to make it optional

marccarre: Done.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -629,6 +632,53 @@ type (
...
+	URL string `json:"url,omitempty"` // the Git SSH URL to the repository which will contain the cluster configuration, e.g. git@github.com:org/repo
+	// +optional
+	Branch string `json:"branch,omitempty"` // the branch under which cluster configuration files will be committed & pushed, e.g. master
+	// +optional
+	Paths []string `json:"paths,omitempty"` // relative paths within the Git repository which the GitOps operator will monitor to find Kubernetes manifests to apply, e.g. ["kube-system", "base"]
+	// +optional
+	FluxPath string `json:"fluxPath,omitempty"` // the directory under which Flux configuration files will be written, e.g. flux/
+	// +optional
+	User string `json:"user,omitempty"` // Git user which will be used to commit changes
+	// +optional
```
martina-if: Remember that the email was required

marccarre: Fixed.

marccarre:
Merging as it stands as we need to make progress.
I'm happy to keep discussing & designing a better target state in the meantime though! 👍 (In the appropriate Github issues and/or design docs, as it will get lost otherwise.)

PR #1442, opened 2019-10-11
A couple of nodegroup.ssh fixes

dlespiau:
- That field is actually optional, make sure to mark it as is
- A small fixup that makes sure the `validateNodeGroupSSH` error is propagated instead of being overriden


dlespiau:
Thanks for the review!

PR #1449, opened 2019-10-15
cfn: Fully internalize the regex to list stacks

dlespiau:
I've noticed a little buglet in `get` code: we were using a regex with a typo:
eksclt.

Instead of exposing the regex parameter to people wanting to use ListStacks
outside the package, make sure it stays an implementation detail that only the
cfn code will know about.

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -167,8 +167,7 @@ func (c *StackCollection) DescribeStack(i *Stack) (*Stack, error) {
 	return resp.Stacks[0], nil
 }
 
-// ListStacks gets all of CloudFormation stacks
-func (c *StackCollection) ListStacks(nameRegex string, statusFilters ...string) ([]*Stack, error) {
+func (c *StackCollection) listStacks(nameRegex string, statusFilters ...string) ([]*Stack, error) {
 	var (
 		subErr error
 		stack  *Stack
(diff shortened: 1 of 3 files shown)


pkg/cfn/manager/api.go
```diff
@@ -167,8 +167,7 @@ func (c *StackCollection) DescribeStack(i *Stack) (*Stack, error) {
 	return resp.Stacks[0], nil
 }
 
-// ListStacks gets all of CloudFormation stacks
-func (c *StackCollection) ListStacks(nameRegex string, statusFilters ...string) ([]*Stack, error) {
+func (c *StackCollection) listStacks(nameRegex string, statusFilters ...string) ([]*Stack, error) {
```
errordeveloper: Could you please export this as something like `ListStacksMatching`? It's handy when you want to discover stacks you have, started working on cluster deleter lambda the other day, and there I used `ListStacks("eksctl-.*")`.

dlespiau: Yup, makes sense. Will do.

errordeveloper: Ping!

dlespiau: Re-added a `ListStacksMatching` function.

errordeveloper:
Rebase should fix failing build.

dlespiau:
All green! merging!

PR #1458, opened 2019-10-18
WIP test fix for broken build

martina-if:


errordeveloper:
@martina-if could you please try `make -f Makefile.docker check-build-image-manifest-up-to-date` locally?

Makefile.docker
```diff
@@ -33,6 +33,8 @@ update-build-image-manifest: ## Update build image manifest and commits the chan
 	for i in .github/workflows/commit.yaml .circleci/config.yml ; do \
 	  sed "s|\(weaveworks/eksctl-build\):.*$$|\1:$(build_image_tag)|" "$${i}" > "$${i}.sedtmp" && mv "$${i}.sedtmp" "$${i}" ; \
 	done
+	git config --local user.email 'action@github.com'
+	git config --local user.name "Github Action"
```
errordeveloper: We don't actually push commit from GHA at the moment, but it's something we should probably do in the future. Right now this is expected to be called by a person that made a change to either `Dockerfile`, `go.mod` or `install-build-deps.sh`. All those files have appropriate comments in them.

.build_image_manifest
```diff
@@ -1,4 +1,4 @@
 100644 blob 47cb6bf90db7b43772d0a77f4082d8552d0724fc	Dockerfile
 100644 blob b46311e84b12ee620fa133b0e48e53ec80bff479	go.mod
-100644 blob d9cf89fd4955c2c5f618eeaf9073498ea8a0997f	go.sum
+100644 blob b83728b6ed5004fc1ea456b6771e9d997ef00086	go.sum
```
errordeveloper: What was the change in `go.sum`? Was it a clean-up?

.github/workflows/image-update.yaml
```diff
@@ -7,8 +7,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
-      - run: make -f Makefile.docker eksctl-image
```
errordeveloper: The idea here is to build the image, then update manifests and push new image. We are currently not able to actually push the image, it's something that we need to do.

.github/workflows/image-update.yaml
```diff
@@ -7,8 +7,8 @@ jobs:
...
       - run: make -f Makefile.docker check-build-image-manifest-up-to-date || (make -f Makefile.docker update-build-image-manifest && make -f Makefile.docker push-build-image)
```
errordeveloper: So this line is misleading right now, it should be just `make -f Makefile.docker check-build-image-manifest-up-to-date` for now, the rest is currently a wishlist.

.build_image_manifest
martina-if: :woman_shrugging: I don't know. I think it was done by the command I ran that updates the image(?)

Makefile.docker
martina-if: This is being ran by GHA, I saw it in the logs. That's why I knew that the `git commit` command on line 38 was failing. 

Look at .github/workflows/image-update.yaml:10 it is calling this part.

.github/workflows/image-update.yaml
martina-if: Not at the moment, I have other stuff to do. I just want the ci jobs to start working again :/

.build_image_manifest
errordeveloper: The command to update the image updates this file,  https://github.com/weaveworks/eksctl/commit/ad42b2044e3f4df8d374e9bd1f2d179830f5e212 was the culprit.

errordeveloper: So what should have happen is that after `go.sum` change the image should had been rebuilt and pushed. The GHA CI job is currently indicating exactly this - image is not up to date, albeit perhaps it's not spelling it out to clearly right now ;)

Issue #1461, opened 2019-10-21
go-bindata generates sporadic diffs


marccarre:
**What happened?**

Sporadic diffs have been observed, both when building locally, and in CI (e.g. [#5143](https://circleci.com/gh/weaveworks/eksctl/5143)), despite always using:
```console
$ go-bindata --version
go-bindata 3.11.0 (Go runtime go1.1x.y).
Copyright (c) 2010-2015, Jim Teeuwen.
Copyright (c) 2017-2018, Kevin Burke.
```

**What you expected to happen?**

No diff, since no file was changed, and the version of `go-bindata` didn't change either.

**How to reproduce it?**

Unclear how to reproduce it.
Rebasing on latest `master` seems to sometimes miraculously "fix" it.
Other people reported that re-applying each commit one-by-one on top of `master` also "fixed" it for them.

errordeveloper:
You should be able to run all the steps using `weaveworks/eksctl-build:19ff80e010d2fdb7be3c44fc6491e13395a682ad`, and results ought to be the same, unless something else is broke (e.g. go-bindata does something that is actually non-deterministic).

errordeveloper:
Right now we don't have an easy way to determine the image tag, not without reading `Makefile`. Perhaps it's something we could improve on that by providing `make print-docker-build-image-and-args` or something like that.

PR #1465, opened 2019-10-21
Fix build image, add target for running `make test` and `make build` inside build image

errordeveloper:
xref #1461


Makefile.docker
```diff
@@ -73,3 +78,11 @@ eksctl-image: intermediate-image ## Build the eksctl image that has release arte
 	      --iidfile=.eksctl_image.iid \
 	      --tag="$(eksctl_image_name)" -
 
+.PHONY: test build
+build test: ## Build and test using the build image
```
martina-if: test build or build test?

Makefile.docker
```diff
@@ -22,19 +23,23 @@ update_build_image_manifest = git ls-tree --full-tree @ -- $(build_image_input)
...
+	  || git commit --quiet .build_image_manifest --message 'Update build image manifest'
 	@# There is no portable way of using sed -i on BSD and GNU, so we use temporary files instead
 	for i in .github/workflows/commit.yaml .circleci/config.yml ; do \
-	  sed "s|\(weaveworks/eksctl-build\):.*$$|\1:$(build_image_tag)|" "$${i}" > "$${i}.sedtmp" && mv "$${i}.sedtmp" "$${i}" ; \
+	  sed "s|\(weaveworks/eksctl-build\):.*$$|\1:$$($(build_image_tag_cmd))|" "$${i}" > "$${i}.sedtmp" && mv "$${i}.sedtmp" "$${i}" ; \
 	done
-	git commit --quiet .build_image_manifest .github/workflows/commit.yaml .circleci/config.yml --message 'Update build image'
-	$(info build image name is $(build_image_name))
+	git diff --quiet -- .github/workflows/commit.yaml .circleci/config.yml \
+	  || git commit --quiet .github/workflows/commit.yaml .circleci/config.yml --message 'Update build image in CI'
```
martina-if: Perhaps say something like "Please update build image in CI with make...."

CONTRIBUTING.md
```diff
@@ -84,6 +84,20 @@ make integration-test-container TEST_V=1
...
+
+```
+make -f Makefile.docker test
+make -f Makefile.docker build
+```
+
+> NOTE: It is not the most convenient way of working on the project, as
+> binaries are built inside the container and cannot be tested manually,
+> also majority of end-users consume binaries and not Docker images.
+> It is recommended to use `make build` etc, unless there is issue in CI
```
martina-if:
```suggestion
> It is recommended to use `make build` etc, unless there is an issue in CI
```

Makefile.docker
martina-if: If this is correct go ahead and merge

errordeveloper: The order doesn't matter here, but I'll make it same the line above for cosmetic reasons.

PR #1475, opened 2019-10-23
make build: update go.mod to latest dependencies

marccarre:
### Description

Building latest `master` currently generates this diff in `go.mod`, which therefore should presumably be updated. This dependency was apparently introduced in https://github.com/weaveworks/eksctl/pull/1465, though I'm not sure why at the moment.


marccarre:
Note that GitHub's build failed, and the hint doesn't fix/change anything:
```
HINT: to fix this, run 'make -f Makefile.docker update-build-image-manifest'
```

errordeveloper:
@marccarre so when you run `make -f Makefile.docker update-build-image-manifest`, you won't see anything in `git diff` or `git status`, because it actually creates two commits.

PR #1476, opened 2019-10-23
Refactor package `drain`

### LINKED ISSUE #1474: Refactor package `drain`
Remove usage of deprecated Kubernetes APIs and ensure wait timeout in `drain.NodeGroup` is handled for retries.

cPu1:
### Description

Closes #1474

### DIFF
--- a/pkg/drain/cordon.go
+++ b/pkg/drain/cordon.go
@@ -31,32 +31,21 @@ import (
 
 // CordonHelper wraps functionality to cordon/uncordon nodes
 type CordonHelper struct {
-	node   *corev1.Node
-	status DesiredCordonStatus
-}
-
-type DesiredCordonStatus string
-
-const (
-	CordonNode   DesiredCordonStatus = "cordon"
-	UncordonNode DesiredCordonStatus = "uncordon"
-)
-
-func (n DesiredCordonStatus) String() string {
-	return string(n)
+	node    *corev1.Node
+	desired bool
 }
 
 // NewCordonHelper returns a new CordonHelper
-func NewCordonHelper(node *corev1.Node, desired DesiredCordonStatus) *CordonHelper {
+func NewCordonHelper(node *corev1.Node, desired bool) *CordonHelper {
 	return &CordonHelper{
-		node:   node,
-		status: desired,
+		node:    node,
+		desired: desired,
 	}
 }
 
 // NewCordonHelperFromRuntimeObject returns a new CordonHelper, or an error if given object is not a
 // node or cannot be encoded as JSON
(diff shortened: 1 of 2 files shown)


pkg/drain/nodegroup.go
```diff
@@ -142,12 +140,19 @@ func NodeGroup(clientSet kubernetes.Interface, ng *api.NodeGroup, waitTimeout ti
...
 			logger.Debug("will drain: %v", newPendingNodes.List())
 
+		evict_loop:
 			for _, node := range nodes.Items {
 				if newPendingNodes.Has(node.Name) {
 					pending, err := evictPods(drainer, &node)
 					if err != nil {
 						logger.Warning("pod eviction error (%q) on node %s – will retry after delay of %s", err, node.Name, retryDelay)
-						time.Sleep(retryDelay)
+						retryTimer := time.NewTimer(retryDelay)
```
martina-if: This also sleeps?

pkg/drain/cordon.go
```diff
@@ -31,32 +31,21 @@ import (
...
-
-const (
-	CordonNode   DesiredCordonStatus = "cordon"
-	UncordonNode DesiredCordonStatus = "uncordon"
-)
-
-func (n DesiredCordonStatus) String() string {
-	return string(n)
+	node    *corev1.Node
+	desired bool
```
martina-if: I think bools are usually terrible for communicating what they mean. I actually prefer the old constants. Perhaps rename `desired` to `cordon`?

pkg/drain/nodegroup.go
cPu1: It doesn't. The receive on the `retryTimer.C` channel below blocks until `retryDelay` elapses.

pkg/drain/cordon.go
cPu1: CordonHelper either cordons or uncordons nodes. The old constants were redundant to define that boolean state and also required more code. 
The field is named `desired` because it reflects the type's default behaviour (`CordonHelper`). I'm not opposed to renaming it to `cordon` though as I think it does make the intent more explicit.

cPu1: Some of the initial code was borrowed from `kubectl`, which also uses `CordonHelper.desired`, so I think we should keep that name.

martina-if: It was more code but it was more clear. less code is not always better :)

What do you think about `cordoned`?

martina-if: Hm, althought `cordoned` could be understood as  the actual state. Yeah, leave it as desired if you feel strongly against what we had before.

cPu1: >It was more code but it was more clear. less code is not always better :)

I don't think it was more clear :) It was a boolean state

Issue #1468, opened 2019-10-21
When creating a cluster, subnets are listed multiple times [0.7.0]


whereisaaron:
**What happened?**
When creating a cluster 0.7.0 lists the same subnets multiple times. I was testing with a config file with two node groups, but with command `eksctl create cluster --config-file=foo-config.yaml --without-nodegroup` so no node groups were created. The subnets are listed only once in the `vpc` section of the config file. I am guessing subnets get incorrectly duplicated as many times as there are node groups in the config file?

E.g. here the private subnets were each listed twice, though in different orders:
```
[ℹ]  eksctl version 0.7.0
[ℹ]  using region ap-southeast-2
[✔]  using existing VPC (vpc-012345678901234) and subnets (private:[subnet-0000000000000 
subnet-111111111111 subnet-22222222222 subnet-1111111111111 subnet-222222222222 
subnet-000000000000] public:[])
[!]  custom VPC/subnets will be used; if resulting cluster doesn't function as expected, 
make sure to review the configuration of VPC/subnets
[ℹ]  using Kubernetes version 1.14
```

**What you expected to happen?**
Each subnet to be listed once.

**How to reproduce it?**

1. Create a config file `foo-config.yaml` with definitions for two node groups
2. `eksctl create cluster --config-file=foo-config.yaml --without-nodegroup` 

**Anything else we need to know?**
Using downloaded 0.7.0 release.

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.7.0"}

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-eks-5047ed", GitCommit:"5047edce664593832e9b889e447ac75ab104f527", GitTreeState:"clean", BuildDate:"2019-08-21T22:32:40Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
Haven't got a `-v 4` log example sorry.

sayboras:
[...]

PR #1477, opened 2019-10-23
Bugfix/subnets duplicate

sayboras:
### Description

Fixed https://github.com/weaveworks/eksctl/issues/1468

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/vpc/vpc.go
+++ b/pkg/vpc/vpc.go
@@ -92,34 +92,6 @@ func describeVPC(provider api.ClusterProvider, vpcID string) (*ec2.Vpc, error) {
 	return output.Vpcs[0], nil
 }
 
-// importVPC will update spec with VPC ID/CIDR
-// NOTE: it does respect all fields set in spec.VPC, and will error if
-// there is a mismatch of local vs remote states
-func importVPC(provider api.ClusterProvider, spec *api.ClusterConfig, id string) error {
-	vpc, err := describeVPCFn(provider, id)
-	if err != nil {
-		return err
-	}
-	if spec.VPC.ID == "" {
-		spec.VPC.ID = *vpc.VpcId
-	} else if spec.VPC.ID != *vpc.VpcId {
-		return fmt.Errorf("VPC ID %q is not the same as %q", spec.VPC.ID, *vpc.VpcId)
-	}
-	if spec.VPC.CIDR == nil {
-		spec.VPC.CIDR, err = ipnet.ParseCIDR(*vpc.CidrBlock)
-		if err != nil {
-			return err
-		}
-	} else if cidr := spec.VPC.CIDR.String(); cidr != *vpc.CidrBlock {
-		return fmt.Errorf("VPC CIDR block %q is not the same as %q",
-			cidr,
-			*vpc.CidrBlock,
-		)
-	}
-
-	return nil
-}
-
 // UseFromCluster retrieves the VPC configuration from an existing cluster
 // based on stack outputs
 // NOTE: it doesn't expect any fields in spec.VPC to be set, the remote state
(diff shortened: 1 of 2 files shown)


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -2,6 +2,7 @@ package v1alpha5
 
 import (
 	"fmt"
+	"k8s.io/kops/util/pkg/slice"
```
sayboras: ops 😢

sayboras:
@errordeveloper @martina-if @cPu1 for your review and feedback.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -421,6 +423,21 @@ func (c *ClusterConfig) AppendAvailabilityZone(newAZ string) {
 	c.AvailabilityZones = append(c.AvailabilityZones, newAZ)
 }
 
+// CleanupSubnets clean up subnet entries having invalid AZ
+func (c *ClusterConfig) CleanupSubnets() {
```
martina-if: Since this is a very specific thing for subnets and it's an implementation detail I think I would move it to the vpc.go file, moving the receiver to an argument.

martina-if:
@sayboras since this was a regression, can you add some unit tests so that this doesn't happen again?

sayboras:
> @sayboras since this was a regression, can you add some unit tests so that this doesn't happen again?

Sure, I will address the lack of coverage soon. Please let me know if you have other comments as well.

sayboras:
@martina-if the test cases for `vpc.go` is very limited, and probably wrongly assert the result. I am slowly adding the test cases to improve coverage. This will take times to actually reach the point for the actual fixes.

pkg/vpc/vpc_test.go
```diff
@@ -91,4 +102,82 @@ var _ = Describe("VPC Endpoints", func() {
 			Error:       errors.New(eks.ErrCodeResourceNotFoundException),
 		}),
 	)
+
+	DescribeTable("can import VPC",
```
martina-if: Can we add a test that checks that given an input that would normally produce duplicate subnets checks that there are no dups?

martina-if:
One more nitpick: can you rewrite the commit message so it says something relevant to the change instead of "rebased from master"?

pkg/vpc/vpc_test.go
sayboras: Agreed that this test case must be added. However, all the methods in this file is public package methods, so it's hard to mock the entire method stack. 

So just want to check with you if I can refactor the code a little bit with dependency injection for private method like this commit https://github.com/weaveworks/eksctl/pull/1477/commits/1905d0b667c118b352943e629d0b253008b78cd3

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: @sayboras I think I prefer the way the tests were before. What I was looking for was something like this:
```
// For corner cases test closer to the function
testCleanup() {
  c = ClusterConfig{ .... }  // test data
  CleanupSubnets(c)
  Expect(c.subnets).To(//not have duplicates)
}

// One regression test for the public API that makes sure we don't have this issue again
testImport() {
// Mock EC2 calls
  p.MockEC2().On("DescribeVpcsInput", MatchedBy(func(input *ec2.DescribeVpcsInput) bool {
      return input != nil
  })).Return(...)

  ImportAllSubnets(p, spec) // Use the mocked provider
  
  Expect(spec).To(// not have duplicates...)
}
```
What do you think? At the top of the tests for "VPC - Cluster Endpoints" you can see an example on how we use the EC2 mocks.

sayboras: It's great idea, I should have thought about it earlier. Maybe I just want to understand and test the whole thing, which makes my mind so narrowed :cry:

pkg/vpc/vpc.go
```diff
@@ -137,11 +139,11 @@ func UseFromCluster(provider api.ClusterProvider, stack *cfn.Stack, spec *api.Cl
 	return outputs.Collect(*stack, requiredCollectors, optionalCollectors)
 }
 
-// Import will update spec with VPC ID/CIDR
+// importVPC will update spec with VPC ID/CIDR
```
sayboras: @martina-if this function is used in the same package only, so making private

pkg/vpc/vpc.go
```diff
@@ -74,7 +76,7 @@ func describeSubnets(provider api.ClusterProvider, subnetIDs ...string) ([]*ec2.
 	return output.Subnets, nil
 }
 
-func describe(provider api.ClusterProvider, vpcID string) (*ec2.Vpc, error) {
+func describeVPC(provider api.ClusterProvider, vpcID string) (*ec2.Vpc, error) {
```
sayboras: `describe` might be just a little bit too general, so I renamed to describe VPC.

martina-if:
Thanks @sayboras !

PR #1480, opened 2019-10-24
Fix a few issues with go-bindata

### LINKED ISSUE #1033: Bindata assets don't get regenerated when a file is removed
This should fail:
```
rm -f pkg/addons/default/assets/coredns-1.12.json 
make test
```
But currently it doesn't as we don't detect that this file was deleted, to fix it one needs to run `go generate ./pkg/addons/default/` manually, after which the test will fail.

### LINKED ISSUE #1461: go-bindata generates sporadic diffs

### LINKED ISSUE #1463: GitHub build fails on diff on .build_image_manifest
**What happened?**

GitHub build fails on diff on `.build_image_manifest`, see also https://github.com/weaveworks/eksctl/runs/265316014 for https://github.com/weaveworks/eksctl/pull/1457.
Running the command suggested by the build's hints doesn't change/update any file.

GitHub builds to pass, just like CircleCI builds.

errordeveloper:
- always generate go-bindata assets
- fixes #1461
- fixes #1033
- fixes #1463


errordeveloper:
Will re-run the build once new image is in Docker Hub...

Makefile
```diff
@@ -127,51 +131,38 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
...
 ##@ Code Generation
 
-.PHONY: print-go-bindata-version
-# In order to help debug issues & discrepancies when building:
-print-go-bindata-version:
-	go-bindata --version
+.PHONY: generate-always
+generate-always: ## Generate code (required for every build)
+	@# go-bindata targets must run every time, as dependencies are too complex to declare in make:
+	@# - deleting an asset is breaks the dependencies
```
martina-if:
```suggestion
	@# - deleting an asset breaks the dependencies
```

Makefile
```diff
@@ -127,51 +131,38 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
...
+	@# - different version of go-bindata generate different code
```
martina-if:
```suggestion
	@# - different versions of go-bindata generate different code
```

Makefile
```diff
@@ -127,51 +131,38 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
...
-# static_resolver_ami.go doesn't only depend on files (it should be refreshed whenever a release is made in AWS)
-# so we need to forcibly generate it
 .PHONY: generate-ami
 generate-ami: ## Generate the list of AMIs for use with static resolver. Queries AWS.
 	time go generate ./pkg/ami
 
 site/content/usage/20-schema.md: $(call godeps,cmd/schema/generate.go)
 	time go run ./cmd/schema/generate.go $@
 
+deep_copy_helper_input = $(shell $(call godeps_cmd,./pkg/apis/...) | sed 's|$(generated_code_deep_copy_helper)||' )
```
martina-if: omg what is this? I see that it's just moved but....

Makefile
```diff
@@ -7,21 +7,25 @@ gocache := $(shell go env GOCACHE)
...
 
 generated_code_deep_copy_helper := pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
 
-all_generated_code := \
-  pkg/addons/default/assets.go \
+generated_code_aws_sdk_mocks := $(wildcard pkg/eks/mocks/*API.go)
+
+conditionally_generated_files := \
+  site/content/usage/20-schema.md \
+  $(generated_code_deep_copy_helper) $(generated_code_aws_sdk_mocks)
```
martina-if: We need to always generate these files during `make build`  too since otherwise the build/tests will fail when something changes.

If anything, I see the "always_generate_in_packages" more optional than these. Why the separation now?

Makefile
errordeveloper: Yeah, that's how go dependencies have to be calculated... it results in something like `go list -deps -f <fmtString> ... | sed ... | sed ...`. I am not entirely clear if it carries much of a benefit at the end of the day.

Makefile
errordeveloper: Deep copy helpers and AWS SDK mocks take long time.

The markdown schema has relatively dependencies, albeit slightly wrong at present. I didn't want to touch it as part of this.

marccarre: > take long time.
How long? (I'd rather have something reliably long, than unreliably slightly less long. 😅 )

Makefile
marccarre: So given:
```console
$ git rev-parse --show-toplevel
$GOPATH/src/github.com/weaveworks/eksctl
```
then:
```console
$ go list -deps -f '{{if not .Standard}}{{ $dep := . }}{{range .GoFiles}}{{$dep.Dir}}/{{.}} {{end}}{{end}}' ./pkg/apis/... | sed "s|$(git rev-parse --show-toplevel)/||g"
```
lists a bunch (currently 125 lines worth of) of dependencies, and the `sed` part of the comment simply makes it relative instead of absolute.

Questions:
1. Why do we need this? (My understanding is that `go list -deps ./pkg/apis/...` will reliably pick all Go modules that `./pkg/api/...` relies upon, and `make` will treat these as dependencies when building `pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go`. Is that all there is to it?)
2. Why do we need the `sed` part to make these paths relative? If we don't could we remove it? If we do, could we get a comment explaining the reason why this is needed? (either in this PR or later)
3. Could we have a comment (either in this PR or later) explaining the `-f {{ ... }}` part?

Makefile
errordeveloper: I think mocks take on order of 30s-45s, and don't change most of the time.

Issue #1020, opened 2019-07-12
next API version: schema migration


errordeveloper:
Next time we change the API, we will need provide migration. There are know patterns of doing, e.g. see what [Skaffold does](https://github.com/GoogleContainerTools/skaffold/tree/9981c376ea21aff40ff81715735003645c93afa9/pkg/skaffold/schema), but this does need a design.

marccarre:
I think it may be time to bite the bullet?

> but this does need a design.

If you have already thoughts about this, capturing your insight either here or in a design doc. would be useful, so that it can reviewed, understood, iterated upon, etc. and so that we can implement it sooner rather than later.

marccarre:
Useful resources: 
- https://blog.openshift.com/kubernetes-deep-dive-code-generation-customresources/
- [Skaffold's folder structure and logic](https://github.com/GoogleContainerTools/skaffold/tree/master/pkg/skaffold/schema), which is rather clean, and therefore understandable.

marccarre:
@errordeveloper, would you know why we specifically pin this version of `code-generator`?
https://github.com/weaveworks/eksctl/blob/38daf7dec34127ef63c0da6bc2144fb091f5b20c/Makefile#L162

sayboras:
> would you know why we specifically pin this version of `code-generator`?
> https://github.com/weaveworks/eksctl/blob/38daf7dec34127ef63c0da6bc2144fb091f5b20c/Makefile#L162

Hope you don't mind if I chime in. I believed the above is to make sure same version of code-generator is used with go.mod. However, recent PRs have changed this behavior https://github.com/weaveworks/eksctl/pull/2007 https://github.com/weaveworks/eksctl/pull/1964 (i.e. use semantic verision retrieved from go.sum).

PR #1483, opened 2019-10-25
Revert "Add GitOps-related types to ClusterConfig"

marccarre:
### Description

We want to iterate & refine these changes before we actually commit to these API changes.
Reverts (the relevant bits of) #1440.


errordeveloper:
If we addressed #1020 and had an internal API version, we would have revert changes like this.

PR #1484, opened 2019-10-25
Push build image to Docker Hub and GitHub registries from master

errordeveloper:


errordeveloper:
This needs to land in master before I can test it properly. I'll be sure to revert/disable if something doesn't go well. We don't use GHA for relEase automation yet, so it shouldn't impact the RC.

PR #1486, opened 2019-10-25
Improve image update workflow

errordeveloper:
Improve on #1484, make it work correctly.


errordeveloper:
Merging without review as this fixes broken configuration in master.

PR #1491, opened 2019-10-25
Update build image manifest, create PR with new commits

errordeveloper:
Follow-up on https://github.com/weaveworks/eksctl/pull/1486 and https://github.com/weaveworks/eksctl/pull/1484.

### DIFF
--- a/.github/workflows/image-update.yaml
+++ b/.github/workflows/image-update.yaml
@@ -10,15 +10,50 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
-      - run: make -f Makefile.docker build-image
+      # Login to both registries
       - run: echo ${DOCKER_HUB_PASSWORD} | docker login --username weaveworkseksctlci --password-stdin
         env:
           DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
-      - run: make -f Makefile.docker push-build-image
+      - name: Update the build image, push it Docker Hub and update manifest in the repo
+        env:
+          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+        run: |
+          set -x
+
+          set -o errexit
+          set -o pipefail
+          set -o nounset
+
+          # Update image manifest and push new image to registry only when needed
+          if ! make -f Makefile.docker check-build-image-manifest-up-to-date ; then
+            # Commit changes to the manifest
+            git config --local user.email "weaveworksbot@users.noreply.github.com"
+            git config --local user.name "Weaveworks Bot"
+            make -f Makefile.docker update-build-image-manifest
+
+            # Build the image and push it to Docker Hub registry
+            make -f Makefile.docker build-image
+            make -f Makefile.docker push-build-image
+
+            # Configure private key
+            mkdir -p "${HOME}/.ssh"
+            ssh-keyscan -t rsa github.com > "${HOME}/.ssh/known_hosts"
+            echo "${DEPLOY_KEY}" > "${HOME}/.ssh/id_rsa"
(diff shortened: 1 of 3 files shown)


.github/workflows/image-update.yaml
```diff
@@ -10,15 +10,50 @@ jobs:
...
+            make -f Makefile.docker update-build-image-manifest
+
+            # Build the image and push it to Docker Hub registry
+            make -f Makefile.docker build-image
+            make -f Makefile.docker push-build-image
+
+            # Configure private key
+            mkdir -p "${HOME}/.ssh"
+            ssh-keyscan -t rsa github.com > "${HOME}/.ssh/known_hosts"
+            echo "${DEPLOY_KEY}" > "${HOME}/.ssh/id_rsa"
```
errordeveloper: So I was hoping GitHub Actions would allow you to easily push back to a branch.

It was actually fairly easy to do:

git push `https://${{ github.actor }}:${{ secrete.GITHUB_TOKEN }}@github.com/weaveworks/eksctl <ref>`

However, as we attempt to update file in `.github/workflows/`, the commits get rejected... so we have to use SSH, and for that a deploy key had to be configured.

.github/workflows/image-update.yaml
```diff
@@ -10,15 +10,50 @@ jobs:
...
+            chmod 400 "${HOME}/.ssh/id_rsa"
+            # Now push commits updated manifest and CI configs to a new branch and create a PR
+            new_branch="update-build-image-manifest-$(date +%s)"
+            git push "git@github.com:${GITHUB_REPOSITORY}.git" "HEAD:refs/heads/${new_branch}"
```
errordeveloper: There is an pre-defined action that enabled you to open a PR, but we do it conditionally in this script, so calling curl this way turned out to be the easiest option. We could also push to master, but it seemed potentially challenging (as merge conflicts happen).

.github/workflows/image-update.yaml
```diff
@@ -10,15 +10,50 @@ jobs:
...
+          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+        run: |
+          set -x
+
+          set -o errexit
+          set -o pipefail
+          set -o nounset
+
+          # Update image manifest and push new image to registry only when needed
+          if ! make -f Makefile.docker check-build-image-manifest-up-to-date ; then
```
martina-if: If we move this `if ! make -f Makefile.docker check-build-image-manifest-up-to-date...` to its own `run:` step maybe we achieve this if and we can use GHA push instead of the curl command below?

something like:

```
- name: Check image is up to date
  run: ! make -f Makefile.docker check-build-image-manifest-up-to-date
- name: Update manifest and push image to Docker Hub 
  run: |
    set -x
    ...
    # Commit changes to the manifest
    git config --local user.email "weaveworksbot@users.noreply.github.com"
    config --local user.name "Weaveworks Bot"
    .....
```
Then we can use the pre-defined action that enables you to open a PR?

That is assuming that when a `run` step fails it will not continue with the other steps.

errordeveloper: Ah, I see what you mean. Good idea, but one issue is that when `make -f Makefile.docker check-build-image-manifest-up-to-date` returns 0 (which is going to happen most often), the job will be marked as failing.

errordeveloper: > Then we can use the pre-defined action that enables you to open a PR?

Ideally, yes, but one that I found took would imply a few lines of config. I can have another look into this, but `curl` option didn't seem too problematic. The only issue to me is that the script is a bit verbose, and the error handling is subject to the usual concerns with bash...

martina-if:
@errordeveloper Can you please add a description in this PR of what it is doing. Also, is it ready to be merged?

martina-if:
We are not planning to use the Github actions for now so I am closing this.

PR #1499, opened 2019-10-28
Move release targets to `Makefile.docker`

errordeveloper:
Releasing [currently](https://circleci.com/gh/weaveworks/eksctl/5252) breaks as we call the version of Go local to CircleCI instead of the one from our build image.


errordeveloper:
@marccarre thanks for updating the description! :)

PR #1525, opened 2019-11-04
Schedule it circleci

martina-if:
WIP this is just a test for now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,41 @@ jobs:
           name: Create release binaries for all plaforms and upload to GitHub
           command: make -f Makefile.docker release
           no_output_timeout: 21m
+  integration-tests:
+    machine:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -77,3 +84,13 @@ workflows:
...
             tags:
               only: /[0-9]+\.[0-9]+\.[0-9]+/
+  scheduled-integration-tests:
+    triggers:
+      - schedule:
+          cron: "*/7 * * * *"
+          filters:
+            branches:
+              only:
+                - master
```
cPu1: Can you temporarily change this to the current branch `schedule-it-circleci` and also adjust the Cron schedule for testing before merging it?

martina-if: yeah, good idea, let me try that. I think I tried without the filter and it didn't work either but maybe that works.

PR #1530, opened 2019-11-06
Refactor Nodegroup, IAM and SSH configuration

cPu1:
### Checklist
- [x] Manually tested


pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -343,16 +345,13 @@ func validateInstancesDistribution(ng *NodeGroup) error {
 }
 
 func validateNodeGroupSSH(SSH *NodeGroupSSH) error {
-	if SSH == nil {
```
martina-if: I think this check was good to have. Safer against null pointers.

pkg/ssh/ssh.go
```diff
@@ -18,9 +20,56 @@ import (
...
+		logger.Info("using EC2 key pair %q", sshConfig.PublicKeyName)
+		return *sshConfig.PublicKeyPath, nil
+	}
+
+}
+
 // LoadKeyFromFile loads and imports a public SSH key from a file provided a path to that file.
 // returns the name of the key
-func LoadKeyFromFile(filePath, clusterName, ngName string, provider api.ClusterProvider) (string, error) {
+func LoadKeyFromFile(filePath, clusterName, ngName string, ec2API ec2iface.EC2API) (string, error) {
```
martina-if: Will we be able to inject a mock here if we use `ec2iface.EC2API`? I would think so but can you please double check?

pkg/ssh/ssh.go
```diff
@@ -18,9 +20,56 @@ import (
...
+
+	// Load Key by content
+	case sshConfig.PublicKey != nil:
+		keyName, err := LoadKeyByContent(sshConfig.PublicKey, clusterName, nodeGroupName, ec2API)
+		if err != nil {
+			return "", err
+		}
+		return keyName, nil
+
+		// Use key by name in EC2
```
martina-if: nit: Check the indentation of these comments

pkg/ssh/ssh.go
```diff
@@ -18,9 +20,56 @@ import (
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 )
 
+// LoadKey loads the SSH public key specified in NodeGroupSSH and returns it. The key should be specified
+// in only one way: by name (for a key existing in EC2), by path (for a key in a local file)
+// or by its contents (in the config-file). It also assumes that if ssh is enabled (SSH.Allow
+// == true) then one key was specified
+func LoadKey(sshConfig *api.NodeGroupSSH, clusterName, nodeGroupName string, ec2API ec2iface.EC2API) (string, error) {
```
martina-if: My initial  idea when I designed this logic was that the ssh package would only do the low level stuff, talking to aws apis.

This loading is closer to the UI, because it's very specific to how a user can specify ssh keys in the API schema. While the ssh package is UI agnostic, and just knows how to do certain things like loading a key by filename, but it doesn't know that a user can specify it as a key, name or content, that's a UI decision -> `ctl` layer.

I'm not completely against moving it, but let's decide on the strategy we want to follow since we will be doing more refactors for the versioning. @marccarre what do you think?

pkg/ssh/ssh.go
cPu1: Yes, we do generate mocks for the EC2API interface (`*mocks.EC2API`). I have also used that mock in this PR

pkg/apis/eksctl.io/v1alpha5/validation.go
marccarre: +1, however, is it expected to return `nil`?
Note that:

- with this check, then passing a `nil` `*NodeGroupSSH` **succeeds**,
- without this check, then passing a `nil` `*NodeGroupSSH` **fails** with a `nil` pointer error -- is it intended to no longer succeed?

Therefore, if we keep this check, should we keep `return nil` or should we return a proper error, e.g. `return errors.New("invalid NodeGroupSSH: nil")`?

pkg/ssh/ssh.go
marccarre: I don't have a strong opinion on this _at the moment_. I agree we generally shouldn't mix higher-level logic with lower-level logic. However, that being said, it looks like other methods of this file were also depending on `pkg/apis`' types, e.g. `api.ClusterProvider`, so it doesn't _feel_ too different.

pkg/apis/eksctl.io/v1alpha5/validation.go
martina-if: hmmm in the schema the NodeGroupSSH is optional so it's ok that it is nil

marccarre: @cPu1, what was the reason for removing this check? 🤔

cPu1: This check is now performed here: https://github.com/weaveworks/eksctl/pull/1530/files#diff-4572c83eda74aa47ee807019c6900162R159 before invoking this function.

Instead of having that function check if its argument is nil and then act on it, the caller performs the check and skips calling the function.

marccarre: Fair enough, either sounds good, given we only call `validateNodeGroupSSH` once anyway.

pkg/ssh/ssh.go
cPu1: After discussion with @martina-if , I have moved the low-level SSH functions to package `ssh/client` and kept LoadKey in the top-level `ssh` package.

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -368,11 +368,6 @@ var _ = Describe("ClusterConfig validation", func() {
 			checkItDetectsError(SSHConfig)
 		})
 
-		It("It doesn't panic when SSH is not enabled", func() {
```
martina-if: Uhm, if you remove this test from here then I think it should be moved somewhere else

cPu1: We have some other fields in `NodeGroup` that can be nil, but we don't have explicit tests for them. I think we should have a test that checks for default values after calling `SetNodeGroupDefaults`.

Issue #1508, opened 2019-10-30
Automate labeling of PRs with targeted version


marccarre:
**Why do you want this feature?**

`eksctl`'s team recently decided to use GitHub labels to indicate in which version of `eksctl` PRs would be released to the world. However, as with any manual task, it is easy to forget and not be consistent. We should therefore find a way to enforce & automate this.

**Possible solutions**

1. Have some GitHub "plugin" preventing us from merging until we add a version label?
2. Have the release scripts look at all PRs going out, and labeling them at the time of the release?

Feel free to suggest more below! 🙂 

**See also**:

- #1507 (Automate release notes.)

michaelbeaumont:
Closing this for now as the labels have little remaining relevance.

Issue #1507, opened 2019-10-30
Automate release notes


marccarre:
**Why do you want this feature?**

Currently, release notes are hand-crafted, which isn't ideal, as it is easy to miss something in these. It would be easier if these were (at least partially) automated.

Related to #1451. (Automatically mentioning external contributors in release notes.)

**Possible solutions**

1. Improve the current release script to scan all PRs going out and generate the release notes. (These release notes could then either be published as-is, or manually edited/improved.)

**Relevant resources**

- https://hypothesis.works/articles/continuous-releases/

martina-if:
Fixed via https://github.com/weaveworks/eksctl/pull/1799

PR #1544, opened 2019-11-12
Add draft.md to gather temporary release notes

marccarre:
### Description

... in order to:

- allow developers to gradually add the description of their work
  in the release notes to-be as they open PRs, and
- allow whoever is triggering the next release to simply move content
  from `draft.md` to `X.Y.Z.md` without having to scan all issues/PRs,
  understand them, and potentially rephrase them.

Related: #1507, #1508.

### DIFF
--- a/docs/release_notes/draft.md
+++ b/docs/release_notes/draft.md
@@ -1,0 +1,13 @@
+# Draft
+
+## Features
+
+- lowercased, user-focused description of the new feature (#0)
+
+## Improvements
+
+- lowercased, user-focused description of the improvement (#0)
+
+## Bug fixes
+
+- lowercased, user-focused description of the bugfix (#0)


docs/release_notes/draft.md
```diff
@@ -0,0 +1,13 @@
+# Draft
+
+## Features
+
+- lowercased, user-focused description of the new feature (#0)
+
+## Improvements
```
gemagomez: What is the difference between features and improvements?

marccarre: These terms predates me -- see other release notes -- so I cannot say for sure.
However, my understanding is that:
- a "feature" is a bigger (usually functional) change which typically adds capabilities to `eksctl`, e.g. support for GPU nodes, support for some new AWS feature, etc.
- an "improvement" would be a smaller (functional or non-functional, but usually non-functional) change which modifies `eksctl`'s behaviour, makes it operate more reliably, faster, with more transparency, etc.

martina-if: :+1: What marc said. Also improvements can be additions to docs.

PR #1545, opened 2019-11-13
No longer default ng.SSH.allow to true when SSH Public Keys are defined.

treacher:
### Description
As a consumer of EKSCTL I wanted to be able to define a KeyPair for my node groups without having to explicitly allow SSH. The reason for this is that I'd like to be able to define my own security group seperate to EKSCTL to attach to the node group which allows access to the node group from a particular jump host. In the organisation I'm working in we have all SSH actions go through a jumphost and then for each instance to have a security group which only allows access from that particular jumphost.

With the current implementation if we only define the SSH Key Name it also sets SSH.Allow to true which then opens up Port 22 to the entire VPC.

In this change I've removed the logic around setting SSH.Allow to true when SSH. PublicKeyName, SSH.PublicKeyPath or SSH.PublicKey are set.

I'm not 100% sure this solution fits with the original intention of this project, but it's a conversation starter. During the thought process of making this change, I got to thinking that it may be easier to completely decouple the KeyPair from SSH Access configuration as it empowers configurability and reduces the amount of logic required around different permutations.

### Testing done
Deployed a cluster with the following ssh configuration:
```
ssh:
  allow: false
  publicKeyName: <REDACTED>
```
With this configuration I observed that the Key Pair Name was set to <REDACTED> and there was 
no ingress rule opening port 22 to the entire VPC.

----

Deployed a cluster with the following ssh configuration:
```
ssh:
  publicKeyName: <REDACTED>
  allow: true
```
With this configuration I observed that the Key Pair Name was set to <REDACTED> and there was 
an ingress rule opening port 22 to the entire VPC.

----

Deployed a cluster with the following ssh configuration:
```
ssh:
  allow: true
```
With this configuration I observed that the Key Pair Name was set to the default key and there 
was an ingress rule opening port 22 to the entire VPC.

----

Deployed a cluster with the following ssh configuration:
```
ssh:
  publicKeyName: <REDACTED>
```
With this configuration I observed that the Key Pair Name was set to <REDACTED> and there was 
an ingress rule opening port 22 to the entire VPC.

----

Deployed a cluster with no ssh configuration:
With this configuration I observed that no Key Pair Name is set and that there was no ingress rule for port 22.

### Checklist
- [x] Manually tested

I was unable to get the integration tests working as I don't have a suitable account to run the tests against 😞 . I would highly appreciate it if somebody is able to do this on my behalf if the CI pipeline doesn't already do this.

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/defaults.go
+++ b/pkg/apis/eksctl.io/v1alpha5/defaults.go
@@ -67,7 +67,7 @@ func SetNodeGroupDefaults(ng *NodeGroup, meta *ClusterMeta) {
 		ng.SSH.PublicKeyPath,
 		ng.SSH.PublicKey)
 
-	if numSSHFlagsEnabled > 0 {
+	if numSSHFlagsEnabled > 0 && ng.SSH.Allow == nil {
 		ng.SSH.Allow = Enabled()
 	} else {
 		if IsEnabled(ng.SSH.Allow) {
--- a/pkg/apis/eksctl.io/v1alpha5/defaults.go
+++ b/pkg/apis/eksctl.io/v1alpha5/defaults.go
@@ -67,14 +67,15 @@ func SetNodeGroupDefaults(ng *NodeGroup, meta *ClusterMeta) {
 		ng.SSH.PublicKeyPath,
 		ng.SSH.PublicKey)
 
-	if numSSHFlagsEnabled > 0 && ng.SSH.Allow == nil {
-		ng.SSH.Allow = Enabled()
-	} else {
+	if numSSHFlagsEnabled == 0 {
 		if IsEnabled(ng.SSH.Allow) {
 			ng.SSH.PublicKeyPath = &DefaultNodeSSHPublicKeyPath
 		} else {
 			ng.SSH.Allow = Disabled()
 		}
+	} else if !IsDisabled(ng.SSH.Allow) {
+		// Enable SSH if not explicitly disabled when passing an SSH key
+		ng.SSH.Allow = Enabled()
 	}
 
 	if !IsSetAndNonEmptyString(ng.VolumeType) {
--- a/pkg/apis/eksctl.io/v1alpha5/defaults_test.go
+++ b/pkg/apis/eksctl.io/v1alpha5/defaults_test.go
@@ -97,6 +97,7 @@ var _ = Describe("ClusterConfig validation", func() {
 			SetNodeGroupDefaults(&testNodeGroup, &ClusterMeta{})
 
 			Expect(*testNodeGroup.SSH.Allow).To(BeFalse())
+			Expect(*testNodeGroup.SSH.PublicKeyPath).To(BeIdenticalTo(testKeyPath))
 		})
 	})
 
--- a/pkg/apis/eksctl.io/v1alpha5/defaults_test.go
+++ b/pkg/apis/eksctl.io/v1alpha5/defaults_test.go
@@ -55,13 +55,12 @@ var _ = Describe("ClusterConfig validation", func() {
 
 	Context("SSH settings", func() {
 
-		It("Providing an SSH key enables SSH", func() {
+		It("Providing an SSH key enables SSH when SSH.Allow not set", func() {
 			testKeyPath := "some/path/to/file.pub"
 
 			testNodeGroup := NodeGroup{
 				VolumeSize: &DefaultNodeVolumeSize,
 				SSH: &NodeGroupSSH{
-					Allow:         Disabled(),
 					PublicKeyPath: &testKeyPath,
 				},
 			}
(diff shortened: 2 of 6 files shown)


pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -62,14 +62,16 @@ func SetNodeGroupDefaults(ng *NodeGroup, meta *ClusterMeta) {
...
+	if ng.SSH.Allow == nil {
+		ng.SSH.Allow = Disabled()
+	}
+
 	numSSHFlagsEnabled := countEnabledFields(
 		ng.SSH.PublicKeyName,
 		ng.SSH.PublicKeyPath,
 		ng.SSH.PublicKey)
 
-	if numSSHFlagsEnabled > 0 {
```
martina-if:
```suggestion
	// Enable when an ssh key is passed and ssh access is not explicitly disabled
	if numSSHFlagsEnabled > 0 && !IsDisabled(ng.SSH.Allow) {
```

treacher: Hello @martina-if yeah I think that change makes sense and would work. I'll look at implementing that tomorrow at work. Thanks for the response :)

treacher:
@martina-if I've made the design changes you recommended. I've ran through all the test cases again above and added in a new test case of:
```

File: pkg/apis/eksctl.io/v1alpha5/defaults_test.go [hunk 2]
```diff
@@ -55,13 +55,12 @@ var _ = Describe("ClusterConfig validation", func() {
...
+		It("Providing an SSH key enables SSH when SSH.Allow not set", func() {
 			testKeyPath := "some/path/to/file.pub"
 
 			testNodeGroup := NodeGroup{
 				VolumeSize: &DefaultNodeVolumeSize,
 				SSH: &NodeGroupSSH{
-					Allow:         Disabled(),
 					PublicKeyPath: &testKeyPath,
 				},
 			}
```
martina-if: can you add in the end of this test please? So we cover for the case I explained above.

```
Expect(*testNodeGroup.SSH.PublicKeyPath).To(BeIdenticalTo(testKeyPath))
```

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -67,7 +67,7 @@ func SetNodeGroupDefaults(ng *NodeGroup, meta *ClusterMeta) {
...
+	if numSSHFlagsEnabled > 0 && ng.SSH.Allow == nil {
```
martina-if: What happens in this block if the config file has:

```
ssh:
  allow: true
  publicKeyPath: ./id_rsa_dev
```
or
```
ssh:
  allow: true
  keyName: ec2-dev-key
```
?

I think it would go to line 74 and override the key path.That's why I think it should only check for it being explicitly disabled.

If you want, you can simplify the logic a little with:

```
	if numSSHFlagsEnabled == 0 {
		if IsEnabled(ng.SSH.Allow) {
			ng.SSH.PublicKeyPath = &DefaultNodeSSHPublicKeyPath
		} else {
			ng.SSH.Allow = Disabled()
		}
	} else if !IsDisabled(ng.SSH.Allow) {
		// Enable SSH if not explicitly disabled when passing an SSH key
		ng.SSH.Allow = Enabled()
	}
```

martina-if:
Thanks for the changes @treacher ! and the manual tests :) I added a couple of comments.

pkg/apis/eksctl.io/v1alpha5/defaults_test.go
treacher: Good idea! :)

pkg/apis/eksctl.io/v1alpha5/defaults.go
treacher: Ah nice find! I've applied your changes and ran through the tests again, everything is working as expected.

Thanks for your patience, I haven't done a lot of Golang programming.

pkg/apis/eksctl.io/v1alpha5/defaults_test.go
martina-if: @treacher Sorry, what I meant was if we can add that `Expect` in this test under line 70 so that we make sure that the ssh key path is not overriden by the default value if we have a bug. 

Can you add that and make sure it's rebased, then I'll approve and merge! :)

pkg/cfn/builder/nodegroup.go
```diff
@@ -110,7 +110,7 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
 	launchTemplateName := gfn.MakeFnSubString(fmt.Sprintf("${%s}", gfn.StackName))
 	launchTemplateData := newLaunchTemplateData(n)
 
-	if api.IsEnabled(n.spec.SSH.Allow) && api.IsSetAndNonEmptyString(n.spec.SSH.PublicKeyName) {
```
martina-if: Even though we set proper values to this field it might be good to be cautious and check for `n.spec.SSH` being `nil` before checking the publicKeyName.

pkg/apis/eksctl.io/v1alpha5/defaults.go
martina-if: Thank you  for the contribution! We appreciate it!

pkg/apis/eksctl.io/v1alpha5/defaults_test.go
treacher: Ah, I've added it to that test case too :)

pkg/cfn/builder/nodegroup.go
treacher: Fair call, I've added that change now :)

PR #1554, opened 2019-11-14
Upgrade CNI plugin to version 1.5.5

### LINKED ISSUE #1549: Update AWS VPC CNI Plugin to 1.5.3 (latest default)
The command `eksctl utils update-aws-node` applies the asset https://github.com/weaveworks/eksctl/blob/master/pkg/addons/default/assets/aws-node.yaml which is pointing to `amazon-k8s-cni:v1.5.0` currently

The latest cni points to [v1.5.3](https://docs.aws.amazon.com/eks/latest/userguide/cni-upgrades.html) and the asset should reflect that.

There has been a previous attempt at upgrading to `v1.5.4` #1376 which must have been reverted.

### LINKED ISSUE #1553: Upgrade CNI plugin to version 1.5.5
When we [reverted the upgrade to 1.5.4] (https://github.com/weaveworks/eksctl/pull/1423) we went back to version 1.5.0. We should use 1.5.5 now that it is released

_Originally posted by @jicowan in https://github.com/weaveworks/eksctl/pull/945#issuecomment-552612797_

martina-if:
closes #1553
closes #1549 

 **TO DO**: 
- [x] 1.5.5 is released. Wait until https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.5/aws-k8s-cni.yaml is updated and use that instead. 
- [x] Manually tested


johnsushant:
https://github.com/aws/amazon-vpc-cni-k8s/releases
1.5.5 is released already. So I believe that should get used.

martina-if:
@johnsushant Right! Thanks. I'll wait until this https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.5/aws-k8s-cni.yaml gets updated then and go directly to 1.5.5

docs/release_notes/draft.md
```diff
@@ -8,6 +8,7 @@
 
 - `eksctl` now leverages `cobra` to print all errors in a more standardised way (#1664)
 - Improved Fargate documentation & example `ClusterConfig` manifest (#1666)
+- upgrade CNI plugin to version 1.5.3 (#0)
```
cPu1: Note before merging: 1.5.3 -> 1.5.5

martina-if: Thanks!

PR #1557, opened 2019-11-15
Fixed the wrong Expect() in existing test cases

sayboras:
### Description
To fix the wrong Expect() without any assert/comparision

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/vpc/vpc_test.go
+++ b/pkg/vpc/vpc_test.go
@@ -2,7 +2,6 @@ package vpc
 
 import (
 	"errors"
-
 	"github.com/aws/aws-sdk-go/service/eks"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -25,7 +25,6 @@ require (
 	github.com/fluxcd/flux v1.15.0
 	github.com/fluxcd/helm-operator v1.0.0-rc2
 	github.com/go-ini/ini v1.37.0 // indirect
-	github.com/go-stack/stack v1.8.0 // indirect
```
sayboras: un-used import, discovered by running `go mod tidy`

pkg/vpc/vpc_test.go
```diff
@@ -58,12 +57,11 @@ var _ = Describe("VPC Endpoints", func() {
 
 			err := UseEndpointAccessFromCluster(p, cfg)
 			if err != nil {
-				Expect(err.Error() == eks.ErrCodeResourceNotFoundException)
```
martina-if: Was this failing before?

sayboras: it's not failing before as it's doing nothing :). Expect(err.Error() == eks.ErrCodeResourceNotFoundException) is just equivalent to Expect(boolean), which is `Assertion` interface https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L150

sayboras: If you just commented this line, the test is still passing

```golang
			if err != nil {
				//Expect(err.Error() == eks.ErrCodeResourceNotFoundException)
			}
            // these 3 lines are doing nothing as well
			Expect(err == nil) 
			Expect(cfg.VPC.ClusterEndpoints.PrivateAccess == &e.Private)
			Expect(cfg.VPC.ClusterEndpoints.PublicAccess == &e.Public)
```

pkg/vpc/vpc_test.go
```diff
@@ -56,14 +55,12 @@ var _ = Describe("VPC Endpoints", func() {
 				return input != nil
 			})).Return(mockResultFn, e.Error)
 
-			err := UseEndpointAccessFromCluster(p, cfg)
-			if err != nil {
-				Expect(err.Error() == eks.ErrCodeResourceNotFoundException)
+			if err := UseEndpointAccessFromCluster(p, cfg); err != nil {
+				Expect(err.Error()).To(Equal(eks.ErrCodeResourceNotFoundException))
```
cPu1: FTR, in the real implementation, the equality test needs to be against `awserr.Error.Code()`
```golang
awsError, ok := err.(awserr.Error)
Expect(ok).To(BeTrue())
Expect(awsError.Code()).To(Equal(eks.ErrCodeResourceNotFoundException))
```

But it's fine here since the error is set in the test case.

cPu1:
@sayboras do you mind rebasing so we could merge it?

pkg/vpc/vpc_test.go
sayboras: Agreed :100:

sayboras:
@cPu1 @martina-if I think it's ready now, thanks for your review.

PR #1563, opened 2019-11-18
Add support for c5d.12x/c5d.24x/c5d.metal instances

marccarre:
... and to match upstream https://raw.github.com/awslabs/amazon-eks-ami/master/files/eni-max-pods.txt as it was updated in https://github.com/awslabs/amazon-eks-ami/commit/81ac166912ebbdb46c56549efe3d88331f524cad as part of https://github.com/awslabs/amazon-eks-ami/pull/371

N.B.: `make build` currently generates a `diff` because of this. Once this PR is merged, this will be resolved.


marccarre:
Thank you very much @cPu1! ✨

PR #1570, opened 2019-11-18
Update build image

errordeveloper:
### Description

This update build image, and provides a temporary fix for AWS SDK being vendored under`lib/`. Ideally we should use an another repo for the SDK fork, but if that doesn't work - this should suffice as a temporary fix.


errordeveloper:
@cPu1 if you get it to work with a fork, please close this PR :)

marccarre:
Superseded by #1582.

PR #1583, opened 2019-11-20
Fix `update cluster` for EKS clusters below platform version `eks.3`

### LINKED ISSUE #1574: Attribute: ClusterSecurityGroupId was not found for resource AWS::EKS::Cluster
**What happened?**
- I upgraded EKSCTL 0.9.0 -> 0.10.0
- Ran the following command: `eksctl update cluster -f <REDACTED> --approve`
- Then I observed the following error message:
```
 UPDATE_ROLLBACK_IN_PROGRESS – "Attribute: ClusterSecurityGroupId was not found for resource: <Redacted>" <- The resource here is AWS::EKS::Cluster
```

I had expected the cluster upgrade to complete successfully.

**How to reproduce it?**
- Upgrade to EKSCTL 0.9.0 -> 0.10.0
- Run the following command: `eksctl update cluster -f <cluster-name> --approve`

MacOS Majave
Downloaded the binary through Homebrew

**Versions**
```
$ eksctl version - version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.10.0"}
$ kubectl version - version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.10", GitCommit:"e3c134023df5dea457638b614ee17ef234dc34a6", GitTreeState:"clean", BuildDate:"2019-07-08T03:50:59Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"darwin/amd64"}
```

**Logs**
```
019-11-19 09:13:05 UTC+1100    eksctl-eks-sandpit-cluster    UPDATE_ROLLBACK_COMPLETE    -
2019-11-19 09:13:04 UTC+1100    eksctl-eks-sandpit-cluster    UPDATE_ROLLBACK_COMPLETE_CLEANUP_IN_PROGRESS    -
2019-11-19 09:12:51 UTC+1100    eksctl-eks-sandpit-cluster    UPDATE_ROLLBACK_IN_PROGRESS    Attribute: ClusterSecurityGroupId was not found for resource: eks-sandpit
2019-11-19 09:12:47 UTC+1100    eksctl-eks-sandpit-cluster    UPDATE_IN_PROGRESS    User Initiated
```

**Hypothesis**
This may be a bug as I don't think the attribute `ClusterSecurityGroupId` exists for https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-cluster.html.
I believe this issue may be coming from the following line of code: https://github.com/weaveworks/eksctl/blob/d80a1f60fd39ea5aa49b323f516cec5302774599/pkg/cfn/builder/cluster.go#L130

cPu1:
### Description

eksctl uses the ClusterSecurityGroupId to enable communication between both managed and unmanaged nodegroups. This field, however, was recently added to AWS EKS (in platform version eks.3) and the `update` command in eksctl wasn't amended to ignore it for older/existing clusters in the 0.10.0 release.

This PR exports that attribute only for new 1.14 clusters, and skips exporting it for existing clusters that are on platform version below `eks.3`.

Fixes #1574 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -68,34 +68,47 @@ func (c *ClusterProvider) RefreshClusterStatus(spec *api.ClusterConfig) error {
 	}
 }
 
-var (
-	platformVersionRegex = regexp.MustCompile(`^eks\.(\d+)`)
-)
-
-// SupportsManagedNodes reports whether the cluster supports Managed Nodes
-// The minimum required platform version for Managed Nodes is `eks.3`
+// VersionSupportsManagedNodes reports whether an existing cluster supports Managed Nodes
+// The minimum required control plane version and platform version are 1.14 and eks.3 respectively
 func (c *ClusterProvider) SupportsManagedNodes(clusterConfig *api.ClusterConfig) (bool, error) {
 	if err := c.maybeRefreshClusterStatus(clusterConfig); err != nil {
 		return false, err
 	}
 
-	if platformVersion := c.Status.clusterInfo.cluster.PlatformVersion; platformVersion != nil {
-		match := platformVersionRegex.FindStringSubmatch(*platformVersion)
-		if len(match) != 2 {
-			logger.Warning("failed to parse cluster's platform version: %q", *platformVersion)
-			return false, nil
-		}
-		versionStr := match[1]
-		minSupportedVersion := 3
-		version, err := strconv.Atoi(versionStr)
-		if err != nil {
-			return false, err
-		}
-		return version >= minSupportedVersion, nil
+	return ClusterSupportsManagedNodes(c.Status.clusterInfo.cluster)
+}
+
+var (
+	platformVersionRegex = regexp.MustCompile(`^eks\.(\d+)`)
+)
+
+// ClusterSupportsManagedNodes reports whether the EKS cluster supports managed nodes
+func ClusterSupportsManagedNodes(cluster *awseks.Cluster) (bool, error) {
+	versionSupportsManagedNodes, err := VersionSupportsManagedNodes(*cluster.Version)
+	if err != nil {
+		return false, err
 	}
 
-	logger.Warning("could not find cluster's platform version")
-	return false, nil
+	if !versionSupportsManagedNodes {
+		return false, nil
+	}
+
+	if cluster.PlatformVersion == nil {
+		logger.Warning("could not find cluster's platform version")
+		return false, nil
+	}
+
+	match := platformVersionRegex.FindStringSubmatch(*cluster.PlatformVersion)
+	if len(match) != 2 {
+		return false, fmt.Errorf("failed to parse cluster's platform version: %q", *cluster.PlatformVersion)
+	}
+	versionStr := match[1]
+	minSupportedVersion := 3
+	version, err := strconv.Atoi(versionStr)
+	if err != nil {
+		return false, err
+	}
+	return version >= minSupportedVersion, nil
 }
 
 func (c *ClusterProvider) maybeRefreshClusterStatus(spec *api.ClusterConfig) error {
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -68,7 +68,7 @@ func (c *ClusterProvider) RefreshClusterStatus(spec *api.ClusterConfig) error {
 	}
 }
 
-// VersionSupportsManagedNodes reports whether an existing cluster supports Managed Nodes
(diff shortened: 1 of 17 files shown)


pkg/eks/eks.go
```diff
@@ -66,6 +68,36 @@ func (c *ClusterProvider) RefreshClusterStatus(spec *api.ClusterConfig) error {
...
+
+// SupportsManagedNodes reports whether the cluster supports Managed Nodes
+// The minimum required platform version for Managed Nodes is `eks.3`
+func (c *ClusterProvider) SupportsManagedNodes(clusterConfig *api.ClusterConfig) (bool, error) {
+	if err := c.maybeRefreshClusterStatus(clusterConfig); err != nil {
+		return false, err
+	}
+
+	if platformVersion := c.Status.clusterInfo.cluster.PlatformVersion; platformVersion != nil {
+		match := platformVersionRegex.FindStringSubmatch(*platformVersion)
```
marccarre: Would it be possible to extract the version's parsing logic to an easily unit-testable utility function and to add unit tests for that? (It is easy for subtle bugs to appear and hide in such logic.)

cPu1: Good point, done.

cPu1:
>Which release are we targeting? (No label on the PR.)

Likely `0.10.2`, which we didn't have a label for. I created the label and added it to this PR.


>Mind sharing the commands used for manual testing, as well as their output (for reference for maintainers and users.)?


You need to have a 1.14 cluster that was created with a prior version of eksctl or, a 1.13 (or below) cluster, and run:
`eksctl update cluster`
This was failing prior to this PR (more details here https://github.com/weaveworks/eksctl/issues/1574)

I tried this PR on a 1.13 cluster. Here's the output:
```console
eksctl update cluster -f cluster.yaml -v4 --approve
2019-11-20T16:46:11+05:30 [ℹ]  eksctl version 
2019-11-20T16:46:11+05:30 [ℹ]  using region us-west-2
2019-11-20T16:46:19+05:30 [ℹ]  re-building cluster stack "eksctl-managed-13-20-cluster"
2019-11-20T16:46:19+05:30 [✔]  all resources in cluster stack "eksctl-managed-13-20-cluster" are up-to-date

2019-11-20T16:46:21+05:30 [ℹ]  checking security group configuration for all nodegroups
2019-11-20T16:46:21+05:30 [ℹ]  all nodegroups have up-to-date configuration
2019-11-20T16:46:21+05:30 [ℹ]  will upgrade cluster "managed-13-20" control plane from current version "1.13" to "1.14"
2019-11-20T17:06:20+05:30 [✔]  cluster "managed-13-20" control plane has been upgraded to version "1.14"
2019-11-20T17:06:20+05:30 [ℹ]  you will need to follow the upgrade procedure for all of nodegroups and add-ons

```

pkg/eks/eks.go
```diff
@@ -66,6 +68,49 @@ func (c *ClusterProvider) RefreshClusterStatus(spec *api.ClusterConfig) error {
 	}
 }
 
+// VersionSupportsManagedNodes reports whether an existing cluster supports Managed Nodes
```
marccarre: `VersionSupportsManagedNodes ` doesn't match `SupportsManagedNodes`, I suspect `lint` will fail once the build get fixed?

pkg/eks/eks.go
marccarre: Thanks for adding the tests! 👍 ✨ 
What I meant was also to extract such logic to a dedicated function so only `platformVersion string` gets passed, and unit tests are even simpler, but I'll leave that up to you:
```go
	match := platformVersionRegex.FindStringSubmatch(*cluster.PlatformVersion)
	if len(match) != 2 {
		return false, fmt.Errorf("failed to parse cluster's platform version: %q", *cluster.PlatformVersion)
	}
	versionStr := match[1]
	minSupportedVersion := 3
	version, err := strconv.Atoi(versionStr)
	if err != nil {
		return false, err
	}
	return version >= minSupportedVersion, nil
```

pkg/eks/eks.go
cPu1: Good catch (looks like the IDE failed to rename this when I refactored it)

pkg/eks/eks.go
cPu1: Both Kubernetes version and platform version are required to determine whether managed nodes are supported or not, so IMO, not much is gained by separating them into two functions. This way, the tests execute the same function as the actual code.

PR #1612, opened 2019-11-28
Add Makefile rule for generating builds

cPu1:
### Description
Adds a Makefile rule for generating builds for multiple OSes. 
We may want to rethink the need for separate files for `release` and normal builds later.

Additionally, the need for a separate goreleaser config could have been avoided if this feature of supplying an override config (via --extra-config) was supported: https://github.com/aidenkeating/goreleaser/commit/ffd8ea2c0444902e2facebab467296488b49ec6f
But it doesn't seem to be available in upstream goreleaser.

### Checklist
- [x] Manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,11 @@ godeps = $(shell $(call godeps_cmd,$(1)))
 build: generate-always ## Build main binary
 	CGO_ENABLED=0 time go build -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
 
+# Build binaries for Linux, Windows and Mac and place them in dist/
+.PHONY: build-all
+build-all: generate-always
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -44,6 +44,11 @@ godeps = $(shell $(call godeps_cmd,$(1)))
 build: generate-always ## Build main binary
 	CGO_ENABLED=0 time go build -ldflags "-X $(version_pkg).gitCommit=$(git_commit) -X $(version_pkg).builtAt=$(built_at)" ./cmd/eksctl
 
+# Build binaries for Linux, Windows and Mac and place them in dist/
+.PHONY: build-all
+build-all: generate-always
```
martina-if: build-release? Not a strong opinion..

cPu1: It doesn't do a release build though, in that the `release` tag isn't set. I used `all` to emphasise that it builds for all major OSes.

PR #1615, opened 2019-11-29
Enable loading of private Go modules

martina-if:
This enables loading private Go modules in Github from docker and CircleCi


Dockerfile
```diff
@@ -29,13 +29,19 @@ RUN apk add --no-cache --initdb --root /out \
...
 ENV KUBECTL_VERSION v1.11.5
 RUN curl --silent --location "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" --output /out/usr/local/bin/kubectl \
     && chmod +x /out/usr/local/bin/kubectl
 
 # Remaining dependencies are controlled by go.mod
 WORKDIR /src
-ENV CGO_ENABLED=0 GOPROXY=https://proxy.golang.org
+ENV CGO_ENABLED=0
+ENV GOPROXY=https://proxy.golang.org
+ENV GOPRIVATE=github.com/weaveworks-aws/*
```
martina-if: TO DO: Change the value of GOPRIVATE

Dockerfile
```diff
@@ -29,13 +29,19 @@ RUN apk add --no-cache --initdb --root /out \
...
+
 ENV KUBECTL_VERSION v1.11.5
 RUN curl --silent --location "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" --output /out/usr/local/bin/kubectl \
     && chmod +x /out/usr/local/bin/kubectl
 
 # Remaining dependencies are controlled by go.mod
 WORKDIR /src
-ENV CGO_ENABLED=0 GOPROXY=https://proxy.golang.org
+ENV CGO_ENABLED=0
+ENV GOPROXY=https://proxy.golang.org
```
cPu1: Now that we are on Go 1.13, we can omit this environment variable but I think we should be explicit. The default value for `GOPROXY` is `https://proxy.golang.org,direct`, so I think we should set it to that.

Issue #1552, opened 2019-11-14
Allow Cluster Endpoint API to be overriden


jitran:
**Why do you want this feature?**
We are using EKS in private mode and access to it is via a bastion host. Our bastion host is locked down so we're tunnelling the eks-cluster-api endpoint on port 443 to our local machine on 127.0.0.1:8001.

Subsequently, we've configured our kube config file to access the api endpoint on port 8001, mapped eks-cluster-api to 127.0.0.1 in the hosts file, and kubectl is working fine.

However, eksctl for operations such as `eksctl update cluster` or `eksctl delete nodegroup` does not read the api endpoint and port number from kube config. It appears to extract it from the cluster's cloudformation stack's Endpoint output value. The value itself is https://eks-cluster-api-endpoint and assumes the default port 443. This in turn causes eksctl operations to fail as the endpoint is not available locally on port 443.

**What feature/behavior/change do you want?**
Allow the user to specify `--endpoint <custom-eks-cluster-api-endpoint>` or some other means to override the default behaviour.

Thank you

martina-if:
closed via #1746

asthasr:
This needs to be re-opened. There's currently no way to handle a bastion host serving as a "tunnel" to a private endpoint, as PR #1746 was reverted. Ideally this should not impact _any_ state of the cluster, as it's entirely client side.

asthasr:
Okay, it looks like a SOCKS5 proxy can be used to overcome this issue. Go's HTTP client seemingly respects it inherently.

```
AWS_PROFILE=sts-qa \
HTTP_PROXY="socks5://127.0.0.1:4141" \
eksctl create cluster -f qa.yaml
```

After the cluster has been created, and a ingress rule added to the cluster's security group, `kubectl` can then be used similarly--although the name of the environment variable is different!

```
HTTPS_PROXY="socks5://127.0.0.1:4141" \
kubectl get nodes
```

michaelbeaumont:
@asthasr Nice solution, thanks!

PR #1618, opened 2019-12-02
Added support for overriding the EKS Endpoint

treacher:
### Description

Added support for overriding the EKS Endpoint for the following commands:

- `eksctl get iamidentitymapping`
- `eksctl create iamidentitymapping`
- `eksctl delete iamidentitymapping`
- `eksctl create iamserviceaccount`
- `eksctl delete iamserviceaccount`
- `eksctl create nodegroup`
- `eksctl delete nodegroup`
- `eksctl drain nodegroup`
- `eksctl delete cluster`

Updated the Kubeconfig class so that it accepts an override for the EKS Endpoint.

Refactored explicit arguments into a params object. I followed the pattern used in pkg/ctl/create/cluster.go.

Added an explicit `--cluster-endpoint-override` flag for defining the EKS Endpoint you'd like to use for the eksctl commands.

This change allows users with a private cluster to be able to use SSH tunnelling to access the cluster from their local machine via a specified port.

Issue #1552 

During the making of this change, I had considered that perhaps it makes more sense to allow a port override instead of the entire cluster URL. That being said changing the full URL gives you a bit more power if for example you decide to put your cluster endpoint behind a DNS record.

### Manual testing performed:
Ran the following commands with and without passing in the `--cluster-endpoint-override` flag:
- `eksctl get iamidentitymapping`
- `eksctl create iamidentitymapping`
- `eksctl delete iamidentitymapping`
- `eksctl create iamserviceaccount`
- `eksctl delete iamserviceaccount`
- `eksctl create nodegroup`
- `eksctl delete nodegroup`
- `eksctl drain nodegroup`
- `eksctl delete cluster`

1. I observed that the above commands were still using the EKSEndpoint output from the EKS Cluster stack when not passing in the `--cluster-endpoint-override` flag and that the command succeeded.
2. I observed that the above commands used my specified EKS Endpoint when passing the `--cluster-endpoint-override` flag and that the command succeeded.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -179,9 +179,9 @@ func AddUpdateAuthConfigMap(fs *pflag.FlagSet, updateAuthConfigMap *bool, descri
 	fs.BoolVar(updateAuthConfigMap, "update-auth-configmap", true, description)
 }
 
-// AddClusterEndpointOverrideFlag adds common cluster-endpoint-override flag
+// AddClusterEndpointOverrideFlag adds common cluster-endpoint flag
 func AddClusterEndpointOverrideFlag(fs *pflag.FlagSet, clusterEndpoint *string) {
-	fs.StringVar(clusterEndpoint, "cluster-endpoint-override", "", "cluster endpoint to use instead of the eks cluster endpoint provided by Cloudformation")
+	fs.StringVar(clusterEndpoint, "cluster-endpoint", "", "cluster endpoint to use instead of the eks cluster endpoint provided by Cloudformation")
 }
 
 // AddCommonFlagsForKubeconfig adds common flags for controlling how output kubeconfig is written
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -179,6 +179,11 @@ func AddUpdateAuthConfigMap(fs *pflag.FlagSet, updateAuthConfigMap *bool, descri
 	fs.BoolVar(updateAuthConfigMap, "update-auth-configmap", true, description)
 }
 
+// AddClusterEndpointOverrideFlag adds common cluster-endpoint-override flag
+func AddClusterEndpointOverrideFlag(fs *pflag.FlagSet, clusterEndpoint *string) {
+	fs.StringVar(clusterEndpoint, "cluster-endpoint-override", "", "cluster endpoint to use instead of the eks cluster endpoint provided by Cloudformation")
+}
+
 // AddCommonFlagsForKubeconfig adds common flags for controlling how output kubeconfig is written
 func AddCommonFlagsForKubeconfig(fs *pflag.FlagSet, outputPath, authenticatorRoleARN *string, setContext, autoPath *bool, exampleName string) {
 	fs.StringVar(outputPath, "kubeconfig", kubeconfig.DefaultPath, "path to write kubeconfig (incompatible with --auto-kubeconfig)")
--- a/integration/creategetdelete_test.go
+++ b/integration/creategetdelete_test.go
@@ -438,7 +438,7 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 					Expect(awsSession).To(HaveExistingStack(stackNamePrefix + "default-s3-read-only"))
 					Expect(awsSession).To(HaveExistingStack(stackNamePrefix + "app1-app-cache-access"))
 
-					clientSet, err := ctl.NewStdClientSet(cfg)
+					clientSet, err := ctl.NewStdClientSet(cfg, "")
 					Expect(err).ShouldNot(HaveOccurred())
 
 					{
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -341,7 +341,7 @@ func doCreateCluster(cmd *cmdutils.Cmd, ng *api.NodeGroup, params *cmdutils.Crea
 		}
 
 		// create Kubernetes client
-		clientSet, err := ctl.NewStdClientSet(cfg)
+		clientSet, err := ctl.NewStdClientSet(cfg, "")
 		if err != nil {
 			return err
 		}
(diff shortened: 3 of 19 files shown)


pkg/ctl/cmdutils/cmdutils.go
```diff
@@ -173,6 +173,11 @@ func AddUpdateAuthConfigMap(fs *pflag.FlagSet, updateAuthConfigMap *bool, descri
 	fs.BoolVar(updateAuthConfigMap, "update-auth-configmap", true, description)
 }
 
+// AddClusterEndpointOverrideFlag adds common cluster-endpoint-override flag
+func AddClusterEndpointOverrideFlag(fs *pflag.FlagSet, clusterEndpoint *string) {
+	fs.StringVar(clusterEndpoint, "cluster-endpoint-override", "", "cluster endpoint to use instead of the eks cluster endpoint provided by Cloudformation")
```
rndstr: it feels like we wouldn't need the `-override` suffix since cli flags generally take precedence anyway. i would just use `--cluster-endpoint` here.

jitran:
This change looks sound to me. Nice use of structs to improve the readability :)

Should we raise a new PR to deprecate the AWS_EKS_ENDPOINT environment variable?

pkg/ctl/cmdutils/cmdutils.go
treacher: Yeah, very good point.

treacher:
Thanks for your thorough review @rndstr, I've done the following:
- Fixed the grammar errors
- Rebased against master (this required one change as a NewClient call was made with the old interface [here](https://github.com/weaveworks/eksctl/blob/f483e1bace0e27220f42d9f5448939272728237f/pkg/ctl/create/fargate.go)
- Updated the CLI flag to `--cluster-endpoint`

I'll await a decision in regards to adding support in the ClusterConfig. :)

martina-if:
Hi @treacher , thank you for the PR! I am curious as I haven't used it myself yet: does setting the environment variable `AWS_EKS_ENDPOINT` to the specified endpoint url work for these cases? Was it only working for specific cases and not others?

jitran:
Hi @martina-if 

Here's what I tried:

Add the cluster's ca chain to the mac key chain so that the cli tools can trust the cluster endpoint
```
aws eks describe-cluster --name <cluster-name> --query cluster.certificateAuthority.data | xargs | base64 -D > ca.pem
sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ca.pem

# create the ssh tunnel and test the endpoint
ssh -L 8001:<cluster-endpoint>:443 <cluster-name-ssh-config>
curl -v https://<cluster-endpoint>:8001/clusters/<cluster-name>
```

Now attempting to use the environment export:
```
export AWS_EKS_ENDPOINT=https://<cluster-endpoint>:8001
eksctl delete nodegroup --name <node-group-name> --cluster <cluster-name>
```

It gets pass the untrusted ssl certificate trust issue, but fails with:
```
[✖]  fetching cluster status to determine operability: unable to describe cluster control plane: SerializationError: failed to unmarshal response error
    status code: 403, request id:
caused by: UnmarshalError: failed decoding error message
HEX DUMP (REDACTED) |{"kind":"Status"|
HEX DUMP (REDACTED) |,"apiVersion":"v|
HEX DUMP (REDACTED) |1","metadata":{}|
HEX DUMP (REDACTED) |,"status":"Failu|
HEX DUMP (REDACTED) |re","message":"f|
HEX DUMP (REDACTED) |orbidden: User \|
HEX DUMP (REDACTED) |"system:anonymou|
HEX DUMP (REDACTED) |s\" cannot get p|
HEX DUMP (REDACTED) |ath \"/clusters/|
HEX DUMP (REDACTED) |<cluster-name>\"|
HEX DUMP (REDACTED) |","reason":"Forb|
HEX DUMP (REDACTED) |idden","details"|
HEX DUMP (REDACTED) |:{},"code":403}.|

caused by: json: cannot unmarshal number into Go struct field jsonErrorResponse.code of type string
```

It seems as though the aws-iam-authenticator token is not passed correctly to the endpoint.

martina-if:
I guess that value of the variable is not passed to the kubeconfig properly (like you correctly do here for the flag https://github.com/weaveworks/eksctl/pull/1618/files#diff-811c261701d0690beed656bfd959b131R72). Do you think you can modify this PR to pass the endpoint all the way from the env variable instead of the flag?

jitran:
> I guess that value of the variable is not passed to the kubeconfig properly (like you correctly do here for the flag https://github.com/weaveworks/eksctl/pull/1618/files#diff-811c261701d0690beed656bfd959b131R72). Do you think you can modify this PR to pass the endpoint all the way from the env variable instead of the flag?

Hi @martina-if, I couldn't find any documentation for the AWS_EKS_ENDPOINT environment variable? Is that a widely used option?

I personally would prefer to explicitly define the cluster api via an eksctl argument or inside the config file. This follows the current pattern for the existing eksctl arguments.

treacher:
> I guess that value of the variable is not passed to the kubeconfig properly (like you correctly do here for the flag https://github.com/weaveworks/eksctl/pull/1618/files#diff-811c261701d0690beed656bfd959b131R72). Do you think you can modify this PR to pass the endpoint all the way from the env variable instead of the flag?

I believe the flag `--cluster-endpoint` and the environment variable `AWS_EKS_ENDPOINT` have very different use cases.

`AWS_EKS_ENDPOINT` seems to be used for overwriting the endpoint that's used for the EKS API calls, see: https://github.com/aws/aws-sdk-go/blob/master/service/eks/service.go#L59 that's why @jitran will be seeing errors.

`--cluster-endpoint` is used for overwriting the endpoint that's used in the generated kube config file.

If we pass `foobar.amazon.com:8001` as the AWS_EKS_ENDPOINT then the AWS EKS API calls will fail as the actual cluster url is `foobar.amazon.com` not `foobar.amazon.com:8001`.

We have a couple of options here:
1. We use the changes defined here.
2. We create a new environment variable that can be set like `KUBECONFIG_CLUSTER_ENDPOINT`

Option 1 will be a bit more complex as I'll need to rebase etc and there are a lot of code changes that'll be merged. It's also a bit annoying to have to explicitly define the cluster endpoint on each CLI call.

Option 2 is probably going to be the easiest from an implementation complexity point of view but there then needs to be some documentation added around this variable.

@martina-if let me know what you think :)

martina-if:
Hi @treacher , you are right, I confused both endpoints. 

What you say makes sense to me: having an environment variable means the user doesn't need to type it every time. I guess this value could be written into the kubeconfig and then picked up by the kube client for subsequent calls after the cluster has been created. I would say go for it!

treacher:
I'm going to create a new PR and link it to this PR just to reduce confusion as the suggested change is quite different from the original.

PR #1620, opened 2019-12-02
Make IAM OIDC provider configuration work with a proxy

### LINKED ISSUE #1619: IAM OIDC provider configuration should work with a proxy
Currently we use `crypto/tls` to make a connection to the OIDC endpoint and obtain fingerprint of the root CA certificate.

https://github.com/weaveworks/eksctl/blob/649935988fa4db895c16521b7887fadea3f4ad22/pkg/iam/oidc/api.go#L117-L133

This breaks when an HTTP proxy is used. We should switch to using `net/http`, which respects standard environment variables and will use a proxy when needed. We cannot do much in the way of verifying the certificate that we see, so proxy will have to be trusted in that regard.

errordeveloper:
## Description

This change enables use of `net/http` (over `crypto/tls`), as the
HTTP client respects standard environment variables and will use
a proxy when needed. We cannot do much in the way of verifying
the certificate that we see, so proxy will have to be trusted
in that regard.

This fixes #1619.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/iam/oidc/api.go
+++ b/pkg/iam/oidc/api.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha1"
 	"crypto/tls"
 	"fmt"
+	"net/http"
 	"net/url"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


errordeveloper:
@rndstr thanks, I also did a bunch of manual tests now. An integration test for this is not feasible, as it would require an HTTPS proxy...

PR #1626, opened 2019-12-03
Add docs for Fargate

cPu1:
### Description
Adds documentation for Fargate

### Checklist

### DIFF
--- a/site/content/usage/16-fargate-support.md
+++ b/site/content/usage/16-fargate-support.md
@@ -1,0 +1,301 @@
+---
+title: "EKS Fargate Support"
+weight: 160
+url: usage/fargate
+---
+
+## Fargate
+
+[AWS Fargate][fargate] is a managed compute engine for Amazon ECS that can run containers. In Fargate you don't need to
+manage servers or clusters.
+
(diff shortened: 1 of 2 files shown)


site/content/usage/16-fargate-support.md
```diff
@@ -0,0 +1,301 @@
...
+}
+```
+
+Note that the profile deletion is a process that can take up to a few minutes. When the `--wait` flag is not specified,
+`eksctl` optimistically expects the profile to be deleted and returns as soon as the aws request has been sent. To make
+`eksctl` wait until it has been successfully deleted use `--wait` like in the example above.
+
+### Further reading
+
+#<!-- TODO: Add link to AWS EKS Fargate docs -->
```
marccarre: Where is this documentation published? Can we update this `TODO`?

marccarre: Presumably https://docs.aws.amazon.com/eks/latest/userguide/fargate.html, will open a follow-up PR

marccarre: See #1627

PR #1632, opened 2019-12-04
Fix tar command if GOPATH is multiple paths

rothgar:
### Description

If a GOPATH has multiple paths (e.g. `/bin/go:$HOME/go`) then the tar command in install-build-deps.sh fails. This update will make it use the first directory in GOPATH instead.

### Checklist
- [x] Manually tested

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -33,4 +33,4 @@ go install \
 # a lot of work, so we install all of those from the release tarball
 METALINTER_VERSION="3.0.0"
 curl --silent --location "https://github.com/alecthomas/gometalinter/releases/download/v${METALINTER_VERSION}/gometalinter-${METALINTER_VERSION}-${OSARCH}.tar.gz" \
-  | tar -x -z -C "${GOBIN}" --strip-components 1
+  | tar -x -z -C "${GOBIN%%:*}" --strip-components 1
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -4,7 +4,7 @@
 # `make -f Makefile.docker update-build-image-manifest && make -f Makefile.docker push-build-image`
 
 if [ -z "${GOBIN+x}" ]; then
- GOBIN="$(go env GOPATH)/bin"
+ GOBIN="${GOPATH%:*}/bin"
 fi
 
 if [ "$(uname)" = "Darwin" ] ; then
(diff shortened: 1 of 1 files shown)


install-build-deps.sh
```diff
@@ -33,4 +33,4 @@ go install \
 # a lot of work, so we install all of those from the release tarball
 METALINTER_VERSION="3.0.0"
 curl --silent --location "https://github.com/alecthomas/gometalinter/releases/download/v${METALINTER_VERSION}/gometalinter-${METALINTER_VERSION}-${OSARCH}.tar.gz" \
-  | tar -x -z -C "${GOBIN}" --strip-components 1
+  | tar -x -z -C "${GOBIN%%:*}" --strip-components 1
```
rndstr: if `GOPATH` equals `/bin/go:$HOME/go` then [`GOBIN=/bin/go:$HOME/go/bin`](https://github.com/weaveworks/eksctl/pull/1632/files#diff-bad7ead3a3d2710cd37d93de0fa7ad24R7) and your parameter expansion `"${GOBIN%%:*}"` leads to `/bin/go` but it should be `/bin/go/bin` in my opinion.

Shouldn't the GOBIN assignment further up be changed to
```
GOBIN="${GOPATH%%:*}/bin"
```

rothgar: Sorry my example GOPATH was flipped. _Usually_ I see people add to their GOPATH with preferred (personal) directories at the beginning of the path to take precedence. This change would assume that they want to use the first directory in their GOPATH.

We can't change GOBIN earlier in the file because it's using `go env GOPATH` instead of `$GOPATH`. I might be the worst person to ask why we would use one vs the other here.

I'm fine making the change at that earlier declaration but we would still need to use the substitution in the tar command in case someone already had GOBIN set with multiple paths (see the if statement on line 6).

rndstr: > This change would assume that they want to use the first directory in their GOPATH.

My comment still stands, though. The code initially meant to append `/bin` to the `GOPATH` but your change does not do that anymore (in the case where `GOPATH` has multiple directories). It just picks the first in `GOPATH` without appending `/bin`.

> We can't change GOBIN earlier in the file because it's using go env GOPATH instead of $GOPATH. I might be the worst person to ask why we would use one vs the other here.

From what I can tell it's the same thing. You can just use `$GOPATH` instead if you want to keep bash's parameter expansion (or then assign `go env GOPATH` to a variable first)

> I'm fine making the change at that earlier declaration

Please do, thanks!

rothgar: Sorry about that. I see what you mean now. I think the latest push will work. It uses the *first* path in GOPATH and appends /bin

install-build-deps.sh
```diff
@@ -4,7 +4,7 @@
 # `make -f Makefile.docker update-build-image-manifest && make -f Makefile.docker push-build-image`
 
 if [ -z "${GOBIN+x}" ]; then
- GOBIN="$(go env GOPATH)/bin"
+ GOBIN="${GOPATH%:*}/bin"
```
rndstr: i think the greedy expansion you had before (with double `%%`) was better since that will support >2 paths

PR #1635, opened 2019-12-04
Fix CIDR typo in 02-custom-vpc-cidr-no-nodes.yaml

benbrandt:
Noticed a small typo in the example yaml file.


rndstr:
thank you @benbrandt!

PR #1647, opened 2019-12-06
Fix CircleCI config for IT

martina-if:
Changed two things in the IT configuration for CircleCI:

- changed the no_output_timeout to 35m
- Fixed a configuration error. For successful builds we were getting:

```

Ran 63 of 63 Specs in 7840.439 seconds
SUCCESS! -- 63 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSuite (7840.44s)
PASS
/bin/bash: line 8: -: command not found

Exited with code exit status 127
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,11 +78,10 @@ jobs:
             --volume=$(pwd):/src \
             --volume=$HOME/project/.ssh:/root/.ssh \
             $BUILD_IMAGE make integration-test TEST_V=1 SSH_KEY_PATH="/root/.ssh/id_rsa_82cbed40f32705e6169d6825407d8307"
-            - store_test_results:
-                path: ./test-results
-            - store_artifacts:
-                path: ./test-results
-          no_output_timeout: 140m
+      - store_test_results:
+        path: ./test-results
+      - store_artifacts:
+        path: ./test-results
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ jobs:
                 path: ./test-results
             - store_artifacts:
                 path: ./test-results
-          no_output_timeout: 120m
+          no_output_timeout: 140m
```
cPu1: Do we need to have this match the test timeout? Every time a test case runs, its description is output, and I don't think we have a test that can take longer than `120m`.

martina-if: yeah, I wasn't sure how this one works. I think I had to increase it at some point but maybe it wasn't this one that I needed to change. Let's see what happens tonight when the IT run and we can decide on Monday.

.circleci/config.yml
```diff
@@ -78,11 +78,11 @@ jobs:
             --volume=$(pwd):/src \
             --volume=$HOME/project/.ssh:/root/.ssh \
             $BUILD_IMAGE make integration-test TEST_V=1 SSH_KEY_PATH="/root/.ssh/id_rsa_82cbed40f32705e6169d6825407d8307"
-            - store_test_results:
-                path: ./test-results
-            - store_artifacts:
-                path: ./test-results
-          no_output_timeout: 120m
+          no_output_timeout: 35m
+      - store_test_results:
```
cPu1: This field was incorrectly placed under the `command` field before, right?

martina-if: Exactly! My bad

Issue #1655, opened 2019-12-10
Cluster stack update (AppendNewClusterStackResource) ignores creation of some new resources


cPu1:
`ClusterResourceSet` builds a CloudFormation template using the `ClusterConfig` as input and skips creation of certain resources if the user has supplied their own resource (e.g., `ClusterConfig.VPC.SecurityGroup`) This works fine during cluster creation. 

During an update, the VPC fields in `ClusterConfig` supplied by the user are ignored because user-supplied updates to VPC are not (yet) supported. However, before building a new template, the code loads the VPC config from the existing cluster's stack output, which results in `ClusterResourceSet` skipping execution of code that'd otherwise be executed when these fields aren't supplied by the user. As a consequence, any update logic that adds code that's conditionally executed, depending on whether a field was user-supplied or not, will be skipped and the updated stack will not have the new resources.

Issue #915, opened 2019-06-20
Update cluster not respecting version in config-file / automatically upgrading control plane


dhanvi:
**What happened?**
eksctl automatically upgraded the control plan version while doing `eksctl update cluster`

**What you expected to happen?**
Not upgrade the control plane version unless asked for.

**How to reproduce it?**
`eksctl create cluster -f eks_config.yaml` with config as 

```yaml
---
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: us-east-1-prod-eks
  region: us-east-1
  version: "1.12"

vpc:
  cidr: "10.0.0.0/16"
```
and then try to update cluster with `eksctl update cluster -f eks_config.yaml`

```
[ℹ]  using region us-east-1
[!]  NOTE: config file is only used for finding cluster name and region, deep cluster configuration changes are not yet implemented
[ℹ]  re-building cluster stack "eksctl-us-east-1-prod-eks-cluster"
[✔]  all resources in cluster stack "eksctl-us-east-1-prod-eks-cluster" are up-to-date
[ℹ]  (plan) would upgrade cluster "us-east-1-prod-eks" control plane from current version "1.12" to "1.13"
[!]  no changes were applied, run again with '--approve' to apply the changes
```

Related #909

martina-if:
Hi @dhanvi , thank you for your report. I see in the last line of the logs that it says 
```
[!]  no changes were applied, run again with '--approve' to apply the changes
```

Did you check if it actually upgraded the cluster version or not? You can see the version running `eksctl get cluster us-east-1-prod-eks`.

dhanvi:
It didn't upgrade the cluster version without the --approve flag. (It does upgrade with the approve flag)

Ideally I am expecting eksctl not to upgrade the cluster (even with the approve flag) as in the yaml we are defining the version to be as 1.12

martina-if:
Oh, I see. I think that makes sense. There are several commands where the config file is only used to get the region and the cluster name. We will discuss it in the team.

viktorradnai:
The behaviour of the upgrade ignoring the version flag is totally non-intuitive so I would consider this a bug.

We are trying to use eksctl as part of a larger automation piece to manage EKS clusters. Having to review the changes and approve them without having any control over what version of EKS we are upgrading to is something very difficult to script.

AT78:
[...]

PR #1656, opened 2019-12-10
Enable support for Managed Nodegroups when upgrading from EKS 1.13 to 1.14

### LINKED ISSUE #1622: `eksctl update cluster` updates the CloudFormation stack before updating the cluster version
`eksctl update cluster` updates the CloudFormation stack first and then the cluster version. This has the side effect that any new CloudFormation resource or exported attribute added to the stack template that's only available in the next EKS version (e.g,. an update from 1.13 to 1.14) will cause the stack creation to fail.

This can be fixed by updating the cluster version first and then the CloudFormation stack. A consequence of this approach is that the `--wait` flag would no longer be supported because the code would have to wait for the cluster update to complete before initiating the CloudFormation stack update. But this shouldn't affect most users as the default behaviour is also to wait for all updates to complete.

cPu1:
### Description
Enable support for Managed Nodegroups when upgrading from EKS 1.13 to 1.14 by allowing new CloudFormation resources to be created during cluster stack update, and reordering the update logic to perform control plane version update first and then CloudFormation stack update. 

Fixes #1622 
Partially addresses #1655 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/cluster.go
+++ b/pkg/cfn/builder/cluster.go
@@ -7,6 +7,7 @@ import (
 	cfn "github.com/aws/aws-sdk-go/service/cloudformation"
 	gfn "github.com/awslabs/goformation/cloudformation"
 	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 	"github.com/weaveworks/eksctl/pkg/cfn/outputs"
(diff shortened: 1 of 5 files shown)


pkg/cfn/builder/cluster.go
```diff
@@ -24,7 +25,10 @@ type ClusterResourceSet struct {
 }
 
 // NewClusterResourceSet returns a resource set for the new cluster
-func NewClusterResourceSet(provider api.ClusterProvider, spec *api.ClusterConfig, supportsManagedNodes bool) *ClusterResourceSet {
+func NewClusterResourceSet(provider api.ClusterProvider, spec *api.ClusterConfig, supportsManagedNodes bool, existingStack *gjson.Result) *ClusterResourceSet {
+	if existingStack != nil {
```
martina-if: Can this affect other commands? Like any of the `utils` commands for example? I'm worried that this mutates the ClusterConfig quite "late". Perhaps we should mutate a copy instead? Could it happen that another task that has to run later or in parallel needs the original SGs?

pkg/ctl/update/cluster.go
```diff
@@ -100,10 +100,27 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 
 	stackManager := ctl.NewStackManager(cfg)
 
+	if versionUpdateRequired {
+		msgNodeGroupsAndAddons := "you will need to follow the upgrade procedure for all of nodegroups and add-ons"
+		cmdutils.LogIntendedAction(cmd.Plan, "upgrade cluster %q control plane from current version %q to %q", cfg.Metadata.Name, currentVersion, cfg.Metadata.Version)
```
martina-if: Interesting: there are people reporting that eksctl upgrades to the next version regardless of what they specify in the config file. But here it mentions that it will upgrade to what they want. https://github.com/weaveworks/eksctl/issues/915

pkg/cfn/builder/cluster.go
cPu1: >Can this affect other commands? Like any of the utils commands for example?

Only `create cluster` and `update cluster` use this type. 

> I'm worried that this mutates the ClusterConfig quite "late". Perhaps we should mutate a copy instead

This type already mutates the ClusterConfig in AddAllResources, which happens later. If we pass a copy, we'd have to use the mutated copy later in code.

> Could it happen that another task that has to run later or in parallel needs the original SGs?
The original SGs will eventually be available when the stack creation/update completes.
This does not change the behaviour of `create cluster` as the `existingStack` is nil in that case.
Update has no concurrent tasks.

pkg/ctl/update/cluster.go
cPu1: That bug report is correct. It does upgrade to the next version without looking at the version in the config file.

martina-if: Thanks. Let's discuss that as a separate issue

PR #1663, opened 2019-12-12
kubectl.go: fix spelling error

otterley:
s/newever/newer/

### DIFF
--- a/pkg/utils/kubectl.go
+++ b/pkg/utils/kubectl.go
@@ -27,7 +27,7 @@ func CheckKubectlVersion(env []string) error {
 	ktl := &kubectl.LocalClient{Env: env}
 	kubectlPath, err := ktl.LookPath()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/utils/kubectl.go
```diff
@@ -27,7 +27,7 @@ func CheckKubectlVersion(env []string) error {
 	ktl := &kubectl.LocalClient{Env: env}
 	kubectlPath, err := ktl.LookPath()
 	if err != nil {
-		return fmt.Errorf("kubectl not found, v1.10.0 or newever is required")
+		return fmt.Errorf("kubectl not found, v1.10.0 or newer is required")
```
cPu1: FTR, this could also be:
```suggestion
		return errors.New("kubectl not found, v1.10.0 or newer is required")
```
as there is no formatting directive involved.

martina-if:
Hi @otterley Thanks for the PR! I think this typo has been fixed already.

PR #1681, opened 2019-12-18
enable all logging by default

zengelan:
Industry CIS best practice is to enable logging for Kubernetes and EKS clusters, so enabling all logging options during creation.


martina-if:
Hi @zengelan , thank you for this PR.

I see your point that logging is very important. As far as I know, enabling this by default will have an extra cost for users so we want that to be a conscious decision. We could consider enabling them by default if we could find a way for them to not cost extra, for example: define an appropriate retention time and delete the logs when the cluster is deleted.

martina-if:
Hi @zengelan I am closing this PR for the reasons I explained above. If you figure out a way to enable them without the extra cost  (or limit it) please feel free to open a new PR!

PR #1685, opened 2019-12-19
Fix build: Remove traces of weaveworks-experiments

martina-if:
This should have been discovered in #1668


martina-if:
@cPu1 I'm merging this since the build pipeline is broken

PR #1687, opened 2019-12-20
Release notes for 0.12.0

martina-if:


docs/release_notes/0.12.0.md
```diff
@@ -0,0 +1,30 @@
+# Draft
+
+## Features
+
+- `eksctl {get,delete} fargateprofile` can now read the cluster name & region from a `ClusterConfig` file (#1667)
+- support Managed Nodegroups and Fargate for auto-upgraded clusters (#1665)
+- upgrade CNI plugin to version 1.5.5 (#1554)
+- support using a proxy in the IAM OIDC provider configuration (#1620)
+- enable support for Managed Nodegroups when upgrading from EKS 1.13 to 1.14 (#1656)
+  - `--wait` flag is no longer supported in `eksctl update cluster`
```
cPu1: Minor nit:
```suggestion
- `--wait` flag is no longer supported in `eksctl update cluster`
```

martina-if: hm, I did it on purpose since it is a consequence of the previous item

martina-if: I can put it separate as well... what do you prefer?

cPu1: Oh, I didn't notice they were related. I agree with the current change then.

PR #1689, opened 2019-12-20
Add extra checks to build pipeline

martina-if:
Add two extra checks:
  - All generated files are up to date
  - The build image manifest and tag files are up to date


Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apk add --no-cache \
     libsass-dev \
     make \
     musl-dev \
+    jq \
```
marccarre: Why do we need `jq`? Is it currently missing and therefore failing because of it? Or is it for future work?

martina-if: I use it here [build-image-manifest.sh](https://github.com/weaveworks/eksctl/blob/master/build-image-manifest.sh#L13) and we use it in circleci to check if the manifest is up to date.

marccarre: Thanks for the confirmation! ✨

PR #1691, opened 2019-12-20
Remove repeated test and build steps from release pipelines

martina-if:
The release pipelines already call the build-and-test workflows so it shouldn't be
needed to call make build test again


martina-if:
@cPu1 merging this to fix the pipeline

PR #1710, opened 2020-01-10
Add --tags parameter to eksctl create iamserviceaccount.

PyScott:
### Description

Allow tagging iamserviceaccount CloudFormation stack and IAM role through CLI options.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/create/iamserviceaccount.go
+++ b/pkg/ctl/create/iamserviceaccount.go
@@ -40,6 +40,8 @@ func createIAMServiceAccountCmd(cmd *cmdutils.Cmd) {
 
 		fs.BoolVar(&overrideExistingServiceAccounts, "override-existing-serviceaccounts", false, "create IAM roles for existing serviceaccounts and update the serviceaccount")
 
+		fs.StringToStringVarP(&cfg.Metadata.Tags, "tags", "", map[string]string{}, `A list of KV pairs used to tag the AWS resources (e.g. "Owner=John Doe,Team=Some Team")`)
+
 		cmdutils.AddIAMServiceAccountFilterFlags(fs, &cmd.Include, &cmd.Exclude)
 		cmdutils.AddApproveFlag(fs, cmd)
 		cmdutils.AddRegionFlag(fs, cmd.ProviderConfig)


pkg/ctl/create/iamserviceaccount.go
```diff
@@ -40,6 +40,8 @@ func createIAMServiceAccountCmd(cmd *cmdutils.Cmd) {
 
 		fs.BoolVar(&overrideExistingServiceAccounts, "override-existing-serviceaccounts", false, "create IAM roles for existing serviceaccounts and update the serviceaccount")
 
+		fs.StringToStringVarP(&cfg.Metadata.Tags, "tags", "", map[string]string{}, `A list of KV pairs used to tag the AWS resources (e.g. "Owner=John Doe,Team=Some Team")`)
```
cPu1: Modifying `ClusterConfig.Metadata.Tags` would have no effect for this command as it's for attaching tags to the cluster's CloudFormation stack (which happens during cluster creation), not the CloudFormation stack created for IRSAs.

PyScott:
Good day and thank you for your reply.

Based on the code, ClusterConfig.Metadata is used for applying general metadata to the CloudFormation template for each type of resource created by eksctl (cluster, nodegroup, iamserviceaccount, iamidentitymapping, fargateprofile), such as StackName, setting K8s labels as stack tags, and configuring the AWS region. I personally think the name _ClusterConfig_ is confusingly named, because it is ultimately just a structure that describes _SharedConfiguration_ across eksctl supported resource types. You may see all the resource types instantiate a ClusterConfig below:

https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/cluster.go#L26
https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/nodegroup.go#L28
https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/iamserviceaccount.go#L18
https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/iamidentitymapping.go#L16
https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/fargate.go#L23

You may see in several of the resource types other than Cluster where ClusterConfig.Metadata fields are being defined explicitly:

https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/nodegroup.go#L46
https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/ctl/create/iamserviceaccount.go#L35

Furthermore, even without this PR, you may already define a configuration file to tag all of your resource types. This config file method uses the same code flow to which this PR simply exposes a CLI argument for. Here is a working example with iamserviceaccount one may use today:

`eksctl create iamserviceaccount -f tmp.yml --approve`
```---
kind: ClusterConfig

metadata:
  name: test-cluster
  region: us-east-1
  tags:
    FIRSTKEY: FIRSTVALUE
    SECONDKEY: SECONDVALUE

iam:
  withOIDC: true
  serviceAccounts:
  - metadata:
      name: ekstestrole
    attachPolicyARNs:
    - "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess"
```

We also do not need to explicitly define tags for all resource types inside CloudFormation templates because of [Tag Propagation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html). Technically it is only necessary to define resource level tags inside a CloudFormation template when the resource contains tags in addition to the stack tags (for example, adding a Name tag to ec2 resources).

> All stack-level tags, including automatically created tags, are propagated to resources that AWS CloudFormation supports. Currently, tags are not propagated to Amazon EBS volumes that are created from block device mappings.

To describe how this tagging procedure works inside eksctl, ClusterConfig.Metadata.Tags are read inside NewStackCollection and stored inside a StackCollection.sharedTags:

https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/cfn/manager/api.go#L52-L65

These StackCollection.sharedTags are then converted to []*cloudformation.CreateStackInput.Tags prior to CreateStack being called:

https://github.com/weaveworks/eksctl/blob/110ac582364be3afb6c55e701b50ac66183f0651/pkg/cfn/manager/api.go#L73-L75

After opening this PR I noticed that NodeGroups are similarly lacking the --tags CLI argument.

pkg/ctl/create/iamserviceaccount.go
PyScott: Thank you for these details. I have resubmitted the PR under #1756.

PR #1712, opened 2020-01-12
Address lack of coverage for vpc.go

sayboras:
### Description
To address the lack of test coverage for `pkg/vpc/vpc.go` without any change in implementation. This is related to https://github.com/weaveworks/eksctl/pull/1477.

The current coverage of this file is improved 7.7% to 55.8%.

Noted: Some refractor needs to be done so that we can fully unit test other methods.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/vpc/vpc_test.go
+++ b/pkg/vpc/vpc_test.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/weaveworks/eksctl/pkg/utils/ipnet"
+	"net"
 
 	"github.com/aws/aws-sdk-go/service/eks"
 
@@ -16,6 +18,11 @@ import (
 	. "github.com/weaveworks/eksctl/pkg/testutils"
 	"github.com/weaveworks/eksctl/pkg/testutils/mockprovider"
 )
+
+type setSubnetsCase struct {
+	Cfg   *api.ClusterConfig
+	Error error
+}
 
 type importVPCCase struct {
 	Cfg               *api.ClusterConfig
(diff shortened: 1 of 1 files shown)


pkg/vpc/vpc_test.go
```diff
@@ -14,7 +19,27 @@ import (
 	"github.com/weaveworks/eksctl/pkg/testutils/mockprovider"
 )
 
+type setSubnetsCase struct {
+	Cfg   *api.ClusterConfig
```
martina-if: Can we make these test fields private?

pkg/vpc/vpc_test.go
```diff
@@ -55,40 +352,68 @@ var _ = Describe("VPC Endpoints", func() {
...
+			ClusterName: "notFoundCluster",
+			Private:     false,
+			Public:      false,
+			DCOutput:    nil,
+			Error:       nil,
+		}),
 	)
 })
+
+func stringPointer(s string) *string {
```
martina-if: For this functionality you can use `pkg/utils/strings/strings.go:Pointer()`

sayboras: oh cool, I was trying to find the existing function for the same. Thanks for pointing out.

pkg/vpc/vpc_test.go
```diff
@@ -25,21 +26,21 @@ type setSubnetsCase struct {
 }
 
 type importVPCCase struct {
-	Cfg               *api.ClusterConfig
+	cfg               *api.ClusterConfig
 	ID                string
```
martina-if: Oh, I meant all the fields in these test structs: ID, DescripveVPCOutput, etc. As far as I see these are used only in this file right?

sayboras: oh yeah, I was thinking about that after pushed changes, let me make another commit.

sayboras: another look if you don't mind, thanks :)

Issue #1617, opened 2019-11-30
More consistent handling of invalid input and printing of usage/help


marccarre:
**What happened?**

* For CLI options marked as required via cobra, the following happens when they are missing:

```console
$ EKSCTL_EXPERIMENTAL=true ./eksctl enable repo
Error: required flag(s) "git-email", "git-url" not set
Usage: eksctl enable repo [flags]

Enable repository flags:
      --git-url string                    SSH URL of the Git repository to be used for GitOps, e.g. git@github.com:<github_org>/<repo_name>
      --git-branch string                 Git branch to be used for GitOps (default "master")
      --git-user string                   Username to use as Git committer (default "Flux")
      --git-email string                  Email to use as Git committer
      --git-private-ssh-key-path string   Optional path to the private SSH key to use with Git, e.g. ~/.ssh/id_rsa
      --git-paths strings                 Relative paths within the Git repo for Flux to locate Kubernetes manifests
      --git-label string                  Git label to keep track of Flux's sync progress; this is equivalent to overriding --git-sync-tag and --git-notes-ref in Flux (default "flux")
      --git-flux-subdir string            Directory within the Git repository where to commit the Flux manifests (default "flux/")
      --namespace string                  Cluster namespace where to install Flux, the Helm Operator and Tiller (default "flux")
      --with-helm                         Install the Helm Operator and Tiller (default true)
      --amend                             Stop to manually tweak the Flux manifests before pushing them to the Git repository

General flags:
      --cluster string       name of the EKS cluster to enable gitops on
  -r, --region string        AWS region
  -f, --config-file string   load configuration from a file (or stdin if set to '-')
      --timeout duration     maximum waiting time for any long-running operation (default 20s)

AWS client flags:
  -p, --profile string   AWS credentials profile to use (overrides the AWS_PROFILE environment variable)

Common flags:
  -C, --color string   toggle colorized logs (valid options: true, false, fabulous) (default "true")
  -h, --help           help for this command
  -v, --verbose int    set log level, use 0 to silence, 4 for debugging and 5 for debugging with AWS debug logging (default 3)

Use 'eksctl enable repo [command] --help' for more information about a command.


required flag(s) "git-email", "git-url" not set
```

* For input we merge from CLI and `ClusterConfig` and subsequently validate, no help is being printed:

```console
$ EKSCTL_EXPERIMENTAL=true ./eksctl enable repo --git-email foo@bar.com --git-url https://foo.bar
[...]

PR #1714, opened 2020-01-13
WIP: Aggregate gitops errors

teaguecole:
### Description

First time contributor here!  WIP for easier feedback. This is a proposed fix for https://github.com/weaveworks/eksctl/issues/1617.

Using hashicorp multierror to aggregate errors.

### Checklist
TODO: Should have a test that checks if both cluster and region are missing from `enable repo` then both errors should appear, not just one.

- [ N/A] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [X] Manually tested


pkg/ctl/cmdutils/gitops.go
```diff
@@ -112,27 +112,28 @@ func NewGitOpsConfigLoader(cmd *Cmd) ClusterConfigLoader {
 		flagsIncompatibleWithoutConfigFile: sets.NewString(),
 	}
 
+	validateErrs := errors.New("")
```
marccarre: Using something like https://github.com/hashicorp/go-multierror would be potentially cleaner and more flexible, especially in case of no error, since we can then return `ErrorOrNil()`:
```go
var result *multierror.Error

// ... accumulate errors here

// Return the `error` only if errors were added to the multierror, otherwise
// return nil since there are no errors.
return result.ErrorOrNil()
```

teaguecole: Sounds good, I will make that change. Im always hesitant about adding new packages as a new contributor, but I'm not sure where that comes from.

teaguecole:
After using hashicrop/go-multierror the following command

`EKSCTL_EXPERIMENTAL=true eksctl enable repo --git-email foo@bar.com --git-url https://foo.bar`

Has the following output.

`Error: 2 errors occurred:
	* --cluster must be set
	* --region must be set
`

This does look a lot cleaner than what I had before.  Are there other examples similar to this that you are aware of?

pkg/ctl/cmdutils/gitops.go
marccarre: We already transitively have it as a dependency anyway, so it is just about adding the latest version to `go.mod` 🙂 
https://github.com/weaveworks/eksctl/blob/286caebb750beac401aac38548a39783e69f7c9e/go.sum#L540-L542

marccarre:
I think this looks very good 👍 

1. There are other configuration loaders in the below files which could benefit from similar changes (`git grep ClusterConfigLoader`):
    - https://github.com/weaveworks/eksctl/blob/master/pkg/ctl/cmdutils/configfile.go
    - https://github.com/weaveworks/eksctl/blob/master/pkg/ctl/cmdutils/fargate.go
2. Then, we may want to remove some `cobra.MarkFlagRequired` in the various entry points/commands, and move such validation to the configuration loaders instead, so that errors related to these flags are reported in a similar way as for 1. / what you did [here](https://github.com/weaveworks/eksctl/pull/1714#issuecomment-574371969).

teaguecole:
I think I went overboard with multierror usage, in configfile based off the tests failing. Wanted to push this up now so I can get feedback incase I decide to poke at it this weekend

pkg/ctl/cmdutils/configfile.go
```diff
@@ -45,6 +46,8 @@ var (
 	)
 )
 
+var multiErr *multierror.Error
```
sayboras: I think it's best `not` to have package variable, which might cause side effect. This also makes unit test harder as you need to clear the value of this multiErr before any test.

.gitignore
```diff
@@ -40,3 +40,6 @@ integration/kubeconfig-*
 
 # direnv's files:
 .envrc
+
+#vscode
```
martina-if: We try not to add environment specific rules here. You can ignore these files for all your repos by having your own `~/.gitignore_global` file ([docs](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)).

teaguecole: Thanks! I didn't know that. I was just going to remove it before it ever gets merged. Thanks for the tip. Will do

pkg/ctl/cmdutils/configfile.go
teaguecole: Any suggestions as to where to declare multiErr?  At the function level for each function? Or in the `commonClusterConfigLoader` struct?  I would think to avoid any side effects, isolating a multierror instance to each function would be best.  Hopefully I am making logical sense, as I am new to this.  I got help on how to debug go tests, what go testing framework is being used etc. so I have a place to go from here.

sayboras: I feel like if it could be local function variable for public method, and argument for private method (so that you can consolidate all checks :+1: )

teaguecole:
Hi everyone.  I have been caught up with work so I haven't been able to put much time in to this for the past couple months. I am more than happy to keep working on this but it'll be slow.  I left off with an error that was a little difficult to debug.

marccarre:
@teaguecole, thanks for the heads up! 👍 
It'd be nice to have it done, but no emergency, feel free to work on it at your own pace 😊

martina-if:
Hi @teaguecole I will close this since we are doing a bit of spring cleaning but please reopen it whenever you want! :)

PR #1726, opened 2020-01-16
Implement new versioning

### LINKED ISSUE #1282: eksctl version displays 0.2.0-xxx when built from the master branch

martina-if:
This new versioning embeds the version of eksctl in the code so that regardless of where it's built 
the binary will always print the correct version.

The versioning format follows semver (except for the timestamp which contains `:`).

```
X.Y.Z[-preReleaseId][+gitCommit.timestamp]
```

Some examples:

  - development snapshot version: `0.13.0-dev+3d6107c3.2020-01-15T15:23:39Z`
  - a release candidate: `0.13.0-rc.0`
  - a release: `0.13.0`

The version is kept in `pkg/version/release.go` and `release_generate.go` is used to automate
certain operations like: preparing a release or release candidate, move to the next development version.

`release_generate.go` is used by the release scripts to automate the manipulation of the `pkg/version/release.go` file.

It also uses the version to produce artifacts in circleci. Check it here https://circleci.com/gh/weaveworks/eksctl/6102#artifacts/containers/0

fixes #1282

- [x] Manually tested


pkg/version/generate/release_generate.go
```diff
@@ -0,0 +1,107 @@
...
+			log.Fatalf("cannot parse rc version from pre-release id %s", version.PreReleaseID)
+		}
+		newRC :=  rcNumber + 1
+		return version.Version, fmt.Sprintf("rc.%d", newRC)
+	}
+	return version.Version, defaultReleaseCandidate
+}
+
+func nextDevelopmentIteration() (string, string) {
+	ver, err := semver.Make(version.Version)
```
cPu1: There's a variant of `semver.Make` that panics if the version is invalid. That might be more suitable here as we do not expect `version.Version` to ever be invalid.

```suggestion
	ver := semver.MustParse(version.Version)
```

tag-release-candidate.sh
```diff
@@ -36,29 +12,27 @@ function current_branch() {
...
-  # Ensure local master is up-to-date by pulling its latest version from origin
-  # and fast-forwarding local master:
-  git fetch origin master
-  git merge --ff-only origin/master
-  # Create the release branch:
-  git push origin master:"${release_branch}"
+#if [[ ! "$(git remote get-url origin)" =~ ^git@github.com:weaveworks/eksctl(\.git)?$ ]] ; then
+#  echo "Invalid origin: $(git remote get-url origin)"
+#  exit 3
+#fi
```
cPu1: Can we remove these commented lines if they're not required?

tag-release.sh
```diff
@@ -35,57 +12,70 @@ function current_branch() {
...
-  git fetch origin master
-  git merge --ff-only origin/master
-  # Create the release branch:
-  git push origin master:"${release_branch}"
+#if [[ ! "$(git remote get-url origin)" =~ ^git@github.com:weaveworks/eksctl(\.git)?$ ]] ; then
+#  echo "Invalid origin: $(git remote get-url origin)"
+#  exit 3
+#fi
+
+v=$(go run pkg/version/generate/release_generate.go print-version)
```
cPu1: `go run pkg/version/generate/release_generate.go` has been referenced a few times in this script. It might be worth considering extracting it out into a function that takes the "command" as an argument.

tag-release.sh
```diff
@@ -35,57 +12,70 @@ function current_branch() {
...
 git tag --annotate --message "${m}" --force "latest_release"
 git tag --annotate --message "${m}" "${v}"
+git push origin "${v}"
+
+# Update the site by putting everything from master into the docs branch
+git push --force origin "${release_branch}":docs
+
+# Check if need to bump version in master
+git checkout master
+if [ ! "$(current_branch)" = master ] ; then
```
cPu1: Do we need this check? When can `git rev-parse --abbrev-ref @` return something other than the checked out branch?

pkg/version/generate/release_generate.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const versionFilename = "pkg/version/release.go"
+const defaultPreReleaseId = "dev"
+const defaultReleaseCandidate = "rc.0"
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("missing argument")
+		return
```
cPu1: Nit: redundant `return` as `log.Fatal` also calls `os.Exit`.

pkg/version/generate/release_generate.go
```diff
@@ -0,0 +1,107 @@
...
+		newVersion, newPreRelease = prepareReleaseCandidate()
+	case "development":
+		newVersion, newPreRelease = nextDevelopmentIteration()
+	case "print-version":
+		// Print simplified version X.Y.Z
+		fmt.Println(version.Version)
+		return
+	default:
+		log.Fatalf("unknown option %q. Expected 'release', 'release-candidate', 'development' or 'print-version'", command)
+		return
```
cPu1: Nit: redundant `return`.

pkg/version/generate/release_generate.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func prepareRelease() (string, string) {
+	return version.Version, ""
+}
+
+func prepareReleaseCandidate() (string, string) {
+	if strings.HasPrefix(version.PreReleaseID, "rc.") {
+		// Next RC
+		rcNumber, err := strconv.Atoi(strings.TrimPrefix(version.PreReleaseID, "rc."))
```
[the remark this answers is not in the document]
martina-if: Hm, I think `TrimPrefix` is more clear and less error prone.

pkg/version/generate/release_generate.go
martina-if: Good idea

tag-release-candidate.sh
martina-if: Oh, we need these, I commented them out for testing. Will uncomment

tag-release.sh
martina-if: I think if for some reason you have uncommited changes the checkout can fail. I kept this from the existing script https://github.com/weaveworks/eksctl/blob/master/tag-release.sh#L39

Do you think it's better to remove it? I'm thinking it could be a good fail safe.

cPu1: >I kept this from the existing script https://github.com/weaveworks/eksctl/blob/master/tag-release.sh#L39
I see. Let's keep it then.

PR #1737, opened 2020-01-20
WIP Divide release in two steps

martina-if:
With this PR we can split the release workflow into two parts:
- First, build all the binaries so we can test them
- Second, approve the publishing of the builds in circleci so that they get released.

Note 1: goreleaser can build a project without releasing but I couldn't find a way to make it publish the already built binaries from the `dist/` folder. so this pipeline would actually build the binaries twice, but with no code changes in between so the binaries should be the same except for the embedded `buildDate`.

Note 2: the reason I leave the pipeline partially commented is that if I add the second step as `type:approval` it will require us to always manually approve before it can actually perform the release.

@cPu1 @marccarre I'm ok also with not merging this if we think it's not valuable as it is today.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,6 +138,31 @@ jobs:
               --volume=$HOME/go/pkg/mod/:/go/pkg/mod \
               weaveworks/eksctl-build:$(cat .docker/image_tag) ./do-release-candidate.sh
           no_output_timeout: 21m
+  prepare-release:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - restore-cache
+      - docker-pull-build-image
+      - docker-run:
+          cmd: go mod download
+      - run:
+          name: Create release binaries for all plaforms and upload to GitHub
+          command: |
+            docker run \
+              --env=GITHUB_TOKEN \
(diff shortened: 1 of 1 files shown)


marccarre:
Thanks a lot @martina-if! 🙇 ✨ 

Regarding _Note 1_, don't you think changes could eventually, accidentally creep in in the future?
If `goreleaser` cannot do the job, how do you feel about:

1. forking/improving `goreleaser` to be able to do what we want?
2. performing the GitHub release like done for Weave Net? See also:

    - https://github.com/weaveworks/weave/blob/master/docs/release-process.md,
    - https://github.com/weaveworks/weave/blob/master/bin/release, and
    - https://github.com/weaveworks/github-release

  N.B.: [Flux also uses the aforementioned scripts](https://github.com/fluxcd/flux/blob/ca0cf9ceec0b68f774da973de1ad8b3555c6c3aa/.circleci/config.yml#L132-L134). However, I'm not sure how the Homebrew release is performed.

martina-if:
I decided to leave the workflow as it is (all in one go) and add some Makefile commands to be able to do a release in two steps if necessary only in cases where it is absolutely needed. See #1798  and #1799

PR #1741, opened 2020-01-21
Add region of Beijing cn-north-1

martina-if:
- [ ] Add generated AMIs
- [ ] Manually tested


cPu1:
>- [ ] Add generated AMIs

We can ignore this for now since we have switched to SSM Parameter as the default AMI resolution strategy. Although static resolution won't work for the China regions, but only users who explicitly set `nodeGroup.ami: static` will be affected and there's no good reason to be using it.

PR #1742, opened 2020-01-21
snap: use 'git tag --sort' for more reliability

dholbach:
### Description

Sort the git tags the right way.

### Checklist
- [x] Manually tested

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,7 +25,7 @@ parts:
     artifacts: [ eksctl ]
     override-pull: |
       snapcraftctl pull
-      EKSCTL_TAG="$(git tag -l | egrep -v '^(latest_release)' | sort --version-sort | tail -n1)"
+      EKSCTL_TAG="$(git tag --sort=taggerdate | egrep -v '^(latest_release)' | tail -n 1)"
       set +e
       git describe --exact-match --tags $(git log -n1 --pretty='%h')
       retVal=$?


snap/snapcraft.yaml
```diff
@@ -25,7 +25,7 @@ parts:
     artifacts: [ eksctl ]
     override-pull: |
       snapcraftctl pull
-      EKSCTL_TAG="$(git tag -l | egrep -v '^(latest_release)' | sort --version-sort | tail -n1)"
+      EKSCTL_TAG="$(git tag --sort=taggerdate | egrep -v '^(latest_release)' | tail -n 1)"
```
martina-if: Do you need the tag or the version? shouldn't we get the version from the binary itself? 

I have a PR up for embedding the version in the code. With that one can get it with the binary by running `eksctl version` or from the code with `go run pkg/version/generate/release_generate.go print-version` or shorter `make print-version`.

Do you think it would make sense to use any of those here ^

dholbach: My idea was for a stable release to 
- git checkout 0.13.0 && snapcraft
- use 0.13.0 as snap version string

For any other non-tagged ref: use "LAST_TAG+GIT_SHA" as snap version string, e.g. 0.13.0+123abcd.

Does that work for you?

If we can do the snap version string using your code, I'm all for it. 😎

martina-if: Ok, let's use the tag since for "dirty" versions the string contains metadata (it would look like this `0.13.0-dev+3d6107c3.2020-01-15T15:23:39Z`) which is nice locally but probably not for a snap version. We can always change it later if we want.

dholbach: We can either go with this, or use something like https://github.com/fluxcd/flux/pull/2772/files (Hidde and Stefan seemed to prefer the  curl GH API | jq  approach). In the case of `eksctl` it points to `latest_release` though. :man_shrugging:

dholbach: So something like this maybe....?

```console
curl -s https://api.github.com/repos/weaveworks/eksctl/releases | jq .[].tag_name | sed 's/\"//g' | grep -v ^latest_release | head -n 1
```

martina-if: I think if we are building the binaries locally we should use the local tag. It shouldn't expect that it has been released to GH already so I prefer your approach :+1:

Issue #1671, opened 2019-12-16
Tag CloudFormation stacks with eksctl version


cPu1:
The CloudFormation stacks created by eksctl do not have any information pertaining to the eksctl version used to create the cluster or nodegroups.

Without versioning information, determining if a cluster stack supports certain features requires examining multiple CloudFormation resources and outputs. This is evident in this changelist: https://github.com/weaveworks/eksctl/pull/1665/files#diff-2e4434f1f52be0a86d04a807716efce3R25-R48. This can get unwieldy and results in a lot of code that'd need to be maintained, not to mention it also requires fetching the stack template to examine missing resources.

Instead, if the CloudFormation stacks were tagged with some version information (like the eksctl version used to create the cluster), the aforementioned changelist would've reduced to just comparing the stack's version with the eksctl version that added the feature, and would also avoid the additional calls to `DescribeStack` or `GetTemplate`.

Additionally, tagging the CloudFormation stacks with a version would also aid in debugging.

cPu1:
Fixed in #1757

PR #1757, opened 2020-01-26
Add EKSCTL version tag in cloudformation

sayboras:
### Description

Address https://github.com/weaveworks/eksctl/issues/1671

The tag name is `alpha.eksctl.io/eksctl-version` as per below screen shots. This tag will be updated upon create stack and update.

image

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/types.go
+++ b/pkg/apis/eksctl.io/v1alpha5/types.go
@@ -133,7 +133,7 @@ const (
 	// using an SSM GetParameter query
 	NodeImageResolverAutoSSM = "auto-ssm"
 
-	// EksctlVersionTag defines the tag of the cluster name
+	// EksctlVersionTag defines the version of eksctl which is used to provision or update EKS cluster
 	EksctlVersionTag = "alpha.eksctl.io/eksctl-version"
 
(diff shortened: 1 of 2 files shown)


sayboras:
@weaworks/eskctl please let me know if you have any comments, happy to address

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -133,6 +133,9 @@ const (
 	// using an SSM GetParameter query
 	NodeImageResolverAutoSSM = "auto-ssm"
 
+	// EksctlVersionTag defines the version of eksctl which is used to provision or update EKS cluster
+	EksctlVersionTag = "alpha.eksctl.io/eksctl-version"
```
martina-if: The API version is `v1alpha5` but eksctl's version is `0.13.0`... If we are using eksctl's version, should we still use tags that say `alpha`? :confused: It's probably good to keep all of them with the same convention but I'm wondering if we should graduate them to proper versions (not necessarily in this PR)... What do you guys think?

sayboras:
> LGTM. Could you please rebase so we can merge it?

Thanks for your help. It's ready now

Issue #1459, opened 2019-10-18
Update Flux packages


stefanprodan:
Use Flux and Helm Operator install modules that don't depend on Kubernetes.

stefanprodan:
In order to update helm-op the Kubernetes packages must be upgrade to 1.14:

```
build github.com/weaveworks/eksctl/cmd/eksctl: cannot load k8s.io/cli-runtime/pkg/genericclioptions/printers: module k8s.io/cli-runtime@latest (v0.0.0-20191016113839-5e0efc75cd33) found, but does not contain package k8s.io/cli-runtime/pkg/genericclioptions/printers
```

Depends on #1372

stefanprodan:
This is currently blocked by eksctl being stuck on Kubernetes v1.12.6 deps.

stefanprodan:
@martina-if @errordeveloper please consider the above packages when updating Kubernetes in eksctl

PR #1761, opened 2020-01-27
Update Flux packages and migrate to Helm v3

stefanprodan:
Changes:
- update `flux/pkg/install` to v1.18.0
- update `helm-operator/pkg/install` to helm-operator 1.0.0-rc8
- remove Flux client dependencies  
- make the enable repo command non-destructive by updating the Flux manifests in-place
- prevent Flux cache deletion on upgrades
- replace Tiller and Helm v2 with Helm v3 (Flux Helm Operator)

Tested on EKS 1.14 by running `enable repo` on the same cluster with different git branches and users. Before this PR, re-running `enable repo` had a disruptive effect causing downtime, since eksctl was deleting the HelmRelease CRD and all Helm releases before applying the new manifest.

Fix: #1459
Fix: #1762


martina-if:
Looks good. It needs checking that the IT are still passing.

stefanprodan:
Thanks you @martina-if I'll do the docs PR next week.

PR #1778, opened 2020-01-30
Add error checking

gliptak:
### Description

Correct ineffassign warning


martina-if:
Hi @gliptak thank you for the fix! I think this is being fixed as part of #1786 so if that's ok I will close this

PR #1758, opened 2020-01-27
Correct few static code check

sayboras:
### Description

Simple PR to address a few static code checking. All changes are highlighted by IDE, sorry I used the sample settings from work (which is quite strict sometimes) :cry: 

No change in functionality, just the below points:

- Remove redudant type 
- Remove un-necessary type conversion
- Remove un-reachable code
- Remove un-necessary error check

### Checklist
- [ ] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [ ] Manually tested
- [x] Added labels for change area (e.g. `area/nodegroup`) and target version (e.g. `version/0.12.0`)
- [ ] Added note in `docs/release_notes/draft.md` (or relevant release note)

PR #1786, opened 2020-02-05
Improve linting rules

sayboras:
### Description

Address comment https://github.com/weaveworks/eksctl/pull/1758#pullrequestreview-350002836

Enable few more linters:

- gosimple
- ineffassign
- staticcheck
- structcheck
- typecheck
- varcheck

Please find the below warning/issues by the golangci-lint for your reference:

```
pkg/gitops/flux/installer.go:383:2              ineffassign  ineffectual assignment to `manifests`
pkg/ctl/enable/profile.go:96:16                 ineffassign  ineffectual assignment to `err`
pkg/apis/eksctl.io/v1alpha5/types.go:707:26     staticcheck  SA5008: unknown JSON option "omitEmpty"
pkg/apis/eksctl.io/v1alpha5/types.go:711:29     staticcheck  SA5008: unknown JSON option "omitEmpty"
pkg/apis/eksctl.io/v1alpha5/types.go:713:44     staticcheck  SA5008: unknown JSON option "omitEmpty"
pkg/apis/eksctl.io/v1alpha5/vpc.go:58:23        staticcheck  SA5008: unknown JSON option "false"
pkg/apis/eksctl.io/v1alpha5/vpc.go:59:23        staticcheck  SA5008: unknown JSON option "true"
pkg/elb/cleanup.go:110:29                       staticcheck  SA4009: argument ctx is overwritten before first use
pkg/vpc/cleanup.go:78:3                         staticcheck  SA4004: the surrounding loop is unconditionally terminated
pkg/kubernetes/manifests.go:34:2                gosimple     S1011: should replace loop with `objects = append(objects, list.Items...)`
pkg/cfn/manager/api.go:73:2                     gosimple     S1011: should replace loop with `input.Tags = append(input.Tags, c.sharedTags...)`
pkg/ctl/cmdutils/filter.go:62:9                 gosimple     S1025: the argument is already a string, there's no need to use fmt.Sprintf
pkg/ctl/cmdutils/filter.go:71:9                 gosimple     S1025: the argument is already a string, there's no need to use fmt.Sprintf
pkg/ssh/loader.go:16:31                         gosimple     S1002: should omit comparison to bool constant, can be simplified to `!*sshConfig.Allow`
pkg/utils/ipnet/ipnet.go:42:2                   gosimple     S1023: redundant `return` statement
pkg/cfn/template/types.go:77:9                  gosimple     S1034: assigning the result of this type assertion to a variable (switch raw := raw.(type)) could eliminate the following type assertions:
                                                        pkg/cfn/template/types.go:79:20
                                                        pkg/cfn/template/types.go:82:20
                                                        pkg/cfn/template/types.go:84:21
                                                        pkg/cfn/template/types.go:86:25
                                                        pkg/cfn/template/types.go:88:27
pkg/ctl/utils/update_cluster_logging.go:160:19  gosimple  S1009: should omit nil check; len() for nil slices is defined as zero
pkg/ctl/utils/update_cluster_logging.go:161:20  gosimple  S1009: should omit nil check; len() for nil slices is defined as zero
```

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/types.go
+++ b/pkg/apis/eksctl.io/v1alpha5/types.go
@@ -707,15 +707,15 @@ type (
 	// NodeGroupInstancesDistribution holds the configuration for spot instances
 	NodeGroupInstancesDistribution struct {
 		//+required
-		InstanceTypes []string `json:"instanceTypes,omitEmpty"`
+		InstanceTypes []string `json:"instanceTypes,omitempty"`
 		// +optional
 		MaxPrice *float64 `json:"maxPrice,omitempty"`
 		//+optional
-		OnDemandBaseCapacity *int `json:"onDemandBaseCapacity,omitEmpty"`
+		OnDemandBaseCapacity *int `json:"onDemandBaseCapacity,omitempty"`
 		//+optional
-		OnDemandPercentageAboveBaseCapacity *int `json:"onDemandPercentageAboveBaseCapacity,omitEmpty"`
+		OnDemandPercentageAboveBaseCapacity *int `json:"onDemandPercentageAboveBaseCapacity,omitempty"`
 		//+optional
-		SpotInstancePools *int `json:"spotInstancePools,omitEmpty"`
+		SpotInstancePools *int `json:"spotInstancePools,omitempty"`
 	}
 )
 
--- a/pkg/apis/eksctl.io/v1alpha5/vpc.go
+++ b/pkg/apis/eksctl.io/v1alpha5/vpc.go
@@ -55,8 +55,8 @@ type (
 
 	// ClusterEndpoints holds cluster api server endpoint access information
 	ClusterEndpoints struct {
-		PrivateAccess *bool `json:"privateAccess,omitempty,false"`
-		PublicAccess  *bool `json:"publicAccess,omitempty,true"`
+		PrivateAccess *bool `json:"privateAccess,omitempty"`
+		PublicAccess  *bool `json:"publicAccess,omitempty"`
 	}
 )
 
--- a/pkg/ctl/enable/profile.go
+++ b/pkg/ctl/enable/profile.go
@@ -94,6 +94,9 @@ func doEnableProfile(cmd *cmdutils.Cmd, opts *ProfileOptions) error {
 		return err
 	}
 	usersRepoDir, err := ioutil.TempDir("", usersRepoName+"-")
+	if err != nil {
+		return err
+	}
 	logger.Debug("Directory %s will be used to clone the configuration repository and install the profile", usersRepoDir)
 	profileOutputPath := filepath.Join(usersRepoDir, "base")
 
--- a/pkg/ctl/enable/profile.go
+++ b/pkg/ctl/enable/profile.go
@@ -95,7 +95,7 @@ func doEnableProfile(cmd *cmdutils.Cmd, opts *ProfileOptions) error {
 	}
 	usersRepoDir, err := ioutil.TempDir("", usersRepoName+"-")
 	if err != nil {
-		return err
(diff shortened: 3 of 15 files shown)


pkg/vpc/cleanup.go
```diff
@@ -75,6 +75,8 @@ func CleanupNetworkInterfaces(ec2API ec2iface.EC2API, spec *api.ClusterConfig) e
 			return errors.Wrapf(err, "unable to delete network interface %q", eniID)
 		}
 		logger.Debug("deleted network interface %q", eniID)
+		// why we are removing only the first eni id
```
sayboras: @marccarre need your help to clarify this one :)

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -55,8 +55,8 @@ type (
 
 	// ClusterEndpoints holds cluster api server endpoint access information
 	ClusterEndpoints struct {
-		PrivateAccess *bool `json:"privateAccess,omitempty,false"`
-		PublicAccess  *bool `json:"publicAccess,omitempty,true"`
+		PrivateAccess *bool `json:"privateAccess,omitempty"`
```
sayboras: false or true is invalid `json` tag. The default values are set in defaults.go `ClusterEndpointAccessDefaults`

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -704,15 +704,15 @@ type (
 	// NodeGroupInstancesDistribution holds the configuration for spot instances
 	NodeGroupInstancesDistribution struct {
 		//+required
-		InstanceTypes []string `json:"instanceTypes,omitEmpty"`
+		InstanceTypes []string `json:"instanceTypes,omitempty"`
```
sayboras: due to this misspelled, these fields are considered as `required` in the schema. I believe that the intention is to have them as optional. Hence, I have updated schema for the same. Let me know if you have other idea.

pkg/ctl/enable/profile.go
```diff
@@ -94,6 +94,9 @@ func doEnableProfile(cmd *cmdutils.Cmd, opts *ProfileOptions) error {
 		return err
 	}
 	usersRepoDir, err := ioutil.TempDir("", usersRepoName+"-")
+	if err != nil {
+		return err
```
martina-if: It's good to add some more context to the error like:
```
        return errors.Wrapf(err, "unable to create temporary directory for %q", usersRepoName)
```

pkg/vpc/cleanup.go
sayboras: or @martina-if if you can clarify this concern :)

pkg/ctl/enable/profile.go
sayboras: Yeah right, thanks for your input. I have updated the same.

pkg/vpc/cleanup.go
martina-if: It looks like a bug to me. Actually I have seen deletion issues sometimes... it could be because of this...

martina-if: @cPu1 do you know anything about this?

sayboras: seems like this one requires some more analysis, let me raise one issue for tracking purpose. We can fix it in subsequent PR if you are fine :)

martina-if: I think that's a good idea, since it will be changing behavior :+1: then we can track it in a separate PR and if something breaks we know where to look

sayboras: cool, let me know if you have other comments, happy to address :)

martina-if:
Thank you! :sparkles:

Issue #1672, opened 2019-12-17
Option to set the client-id-list for the OIDC provider for use with STS VPC endpoints


john-auld:
**Feature request**
Add an option in the OIDC settings section to set the audience (called client-id in the cli) to a value provided by the user.

**Why do you want this feature?**
The OIDC provider created by eksctl is configured with the global endpoint for the STS service (sts.amazonaws.com), which does not work for nodegroups in a private subnet having no direct Internet connection and a VPC endpoint for the STS service.

The VPC endpoint can only be configured with regional STS endpoints, such as sts.eu-west-1.amazonaws.com and not for the global endpoint sts.amazonaws.com.

If a VPC Endpoint is used for STS, the Audience setting in the console should be changed from sts.amazonaws.com to sts.aws-region.amazonaws.com.

**What feature/behavior/change do you want?**

Add an option to configure the client-id-list when creating new clusters. This affects the success of AssumeRoleWithWebIdentity calls to the STS service.

```
iam:
  withOIDC: true
  oidcClientIdList:
     - sts.eu-west-1.amazonaws.com
```

For reference purposes, the cli commands for configuring the OIDC provider are shown below.

```
aws iam create-open-id-connect-provider \
          --url $ISSUER_URL \
          --thumbprint-list $ROOT_CA_FINGERPRINT \
          --client-id-list sts.<region>.amazonaws.com
```

aclevername:
Closing due to staleness. Feel free to re-open :smile:

PR #1788, opened 2020-02-06
Add support to set client-id-list for the OIDC

almariah:
### Description

Add support to set OIDC client ID list, #1672 

### Checklist
- [x] Manually tested


marccarre:
You may want to rebase on latest `master`, and then we can merge this PR. 🙂

site/content/usage/13-iamserviceaccounts.md
```diff
@@ -34,6 +34,8 @@ The IAM OIDC Provider is not enabled by default, you can use the following comma
 eksctl utils associate-iam-oidc-provider --cluster=<clusterName>
 ```
 
+By default the Client ID list (audience) of IAM OIDC Provider will be set to `sts.amazonaws.com`. To override it you can specify `--oidc-client-id-list` multiple times.
```
martina-if: I think if the flag is a list you should be able to just provide it once like: `--oidc-client-id-list sts1.amazon.com,sts.us-east-1.amazonaws.com,sts.eu-west-1.amazonaws.com` 

Can you do a little manual testing to check this and update the comment?

pkg/ctl/utils/associate_iam_oidc_provider.go
```diff
@@ -22,6 +22,9 @@ func associateIAMOIDCProviderCmd(cmd *cmdutils.Cmd) {
 	}
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+
+		fs.StringSliceVar(&cfg.IAM.OIDCClientIDList, "oidc-client-id-list", []string{}, "List of Client ID for OIDC provider (audience)")
```
martina-if: perhaps rename `oidc-client-id-list` to `oidc-client-ids`

pkg/apis/eksctl.io/v1alpha5/iam.go
```diff
@@ -25,6 +25,8 @@ type ClusterIAM struct {
 	// +optional
 	WithOIDC *bool `json:"withOIDC,omitempty"`
 	// +optional
+	OIDCClientIDList []string `json:"oidcClientIDList,omitempty"`
```
martina-if: To follow the same convention of the other fields I would use `OIDCClientIDS` 
(Remember to update https://github.com/weaveworks/eksctl/pull/1788/files#diff-4572c83eda74aa47ee807019c6900162R45)

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -41,6 +41,10 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 		return fmt.Errorf("iam.withOIDC must be enabled explicitly for iam.serviceAccounts to be created")
 	}
 
+	if IsDisabled(cfg.IAM.WithOIDC) && len(cfg.IAM.OIDCClientIDList) > 0 {
```
marccarre: We may want to move this validation to [`NewUtilsAssociateIAMOIDCProviderLoader`](https://github.com/weaveworks/eksctl/blob/f9e384a6efaa120a25624049c79053c560d413a5/pkg/ctl/cmdutils/configfile.go#L438). WDYT @cPu1, @martina-if?

@almariah, to provide some context, `eksctl` accepts input from both:

- CLI flags/args,
- `ClusterConfig` YAML file,

and then is supposed to merge this input, and then validate the end-result.

pkg/ctl/utils/associate_iam_oidc_provider.go
martina-if: Please check how other util commands add flags. Normally, the value of the flag is not stored directly into the `cfg`(`ClusterConfig`) as I think it can be overwritten later by the loader. It should be stored in a variable first and then passed to the command or to the loader.

marccarre: If a user was to provide a `ClusterConfig` with some `OIDCClientIDList` values, then providing some via the CLI flag would override these silently. 

1. Values provided via this flag should probably be stored in a different temporary slice,
2. This slice should then be passed to [`NewUtilsAssociateIAMOIDCProviderLoader`](https://github.com/weaveworks/eksctl/blob/f9e384a6efaa120a25624049c79053c560d413a5/pkg/ctl/cmdutils/configfile.go#L438).

Then, we have a choice:

- A. either fail if values are provided using _both_ CLI flags and `ClusterConfig`, in which case we need something like:
  ```go
  	l.flagsIncompatibleWithConfigFile.Insert(
  		"oidc-client-id-list",
  	)
  ```
  This is what `eksctl` typically does at the moment, for better or worse, but at least this approach would be consistent with the current behaviour.
  And if no values are in the `ClusterConfig` set it to the provided slice.
- B. merge the two lists of IDs, de-duplicate them (e.g. using a set), and then set the final result to `cfg.IAM.OIDCClientIDList` for it to be used by `eksctl`.

Disclaimer: the way these two sources of input are used together is currently, arguably, confusing, and something which we would like to decouple & make clearer.

marccarre: Nit: what would you think of having `oidc-client-ids` instead of `oidc-client-id-list`?

site/content/usage/13-iamserviceaccounts.md
almariah: I have tested it using a comma-separated list and it works, but I have seen similar explanation in the docs:
```
NOTE: you can specify `--attach-policy-arn` multiple times to use more then one policy.
``` 

should I change it?

pkg/apis/eksctl.io/v1alpha5/validation.go
almariah: @marccarre I figured it, in case of `validateWithConfigFile` we don't need this validation because

https://github.com/weaveworks/eksctl/blob/5817e7632b33af9852d87a1e8ff4be231613ff21/pkg/ctl/cmdutils/configfile.go#L465-L470

pkg/apis/eksctl.io/v1alpha5/iam.go
almariah: I will use `OIDCClientIDs`

pkg/ctl/utils/associate_iam_oidc_provider.go
almariah: I guess the second option of merging and de-duplicating make sense.

martina-if:
Regarding the release notes, there's no need to add anything to a release notes file anymore. Just make sure the title of the PR is descriptive enough to be added into the notes automatically later.

cPu1:
> Sorry for the late feedback.
> 
> I would like one more change. I think we shouldn't do the merging between the values of the flags and the config file. Even though it is a good idea I think it's important that we follow the same convention as in other use cases. In this case, the convention would be that the flag is incompatible with the homonym field in the config file so one can only specify one of them at the same time.
> 

I agree with this. All eksctl commands prohibit the usage of flags that map to a `ClusterConfig` field if a config file is specified. One reason is because it can be confusing to users whether the flags override the `ClusterConfig` field or are merged with it, but fields that have scalar values cannot be merged.

pkg/ctl/utils/associate_iam_oidc_provider.go
```diff
@@ -16,12 +16,16 @@ func associateIAMOIDCProviderCmd(cmd *cmdutils.Cmd) {
...
+	var oidcClientIDs []string
 	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
 		cmd.NameArg = cmdutils.GetNameArg(args)
-		return doAssociateIAMOIDCProvider(cmd)
+		return doAssociateIAMOIDCProvider(cmd, oidcClientIDs)
 	}
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+
+		fs.StringSliceVar(&oidcClientIDs, "oidc-client-ids", []string{}, "List of Client ID for OIDC provider (audience)")
```
cPu1: Minor nit: 
```suggestion
		fs.StringSliceVar(&oidcClientIDs, "oidc-client-ids", []string{}, "List of Client IDs for OIDC provider (audience)")
```

pkg/ctl/cmdutils/configfile.go
```diff
@@ -454,18 +454,34 @@ func NewUtilsEnableEndpointAccessLoader(cmd *Cmd, privateAccess, publicAccess bo
...
+		l.ClusterConfig.IAM.OIDCClientIDs = oidcClientIDs
 		return l.validateMetadataWithoutConfigFile()
 	}
 
 	l.validateWithConfigFile = func() error {
 		if api.IsDisabled(l.ClusterConfig.IAM.WithOIDC) {
 			return fmt.Errorf("'iam.withOIDC' is not enabled in %q", l.ClusterConfigFile)
 		}
+
+		oidcClientIDs = append(oidcClientIDs, l.ClusterConfig.IAM.OIDCClientIDs...)
```
cPu1: As mentioned in my comment, we should disallow passing `oidc-client-ids` if a ClusterConfig file is passed (`-f config.yaml`).

marccarre:
> @marccarre are you ok with this?

Yes, I am @martina-if. I was open to both approaches but keeping things consistent is a strong argument against merging values. We can always merge values later if we want to, for all commands.

michaelbeaumont:
@almariah hi, would you have time to rebase and update the PR?

PR #1792, opened 2020-02-07
Add integration tests for Fargate

### LINKED ISSUE #1736: Add integration tests for Fargate

Integration tests for Fargate integration could make it to #1623 in time before the deadline.
Some ought to be added in order to avoid any regression on Fargate integration and automatically ensure this feature still works, release after release.

Add tests to cover the following scenarii:

- Create cluster with `--fargate`.
- Create an unmanaged cluster and then enable Fargate integration.
- Creating a managed cluster with Fargate integration enabled as well.
- Delete cluster.

cPu1:
### Description

Adds integration tests for Fargate. 

Closes #1736 

The tests follow this sequence:
- Create a new cluster, with each test using a different set of flags
- Create a new Fargate profile
- Deploy pods matching the Fargate profile selector
- Assert that the pods run on Fargate

### Checklist
- [x] Manually tested

### DIFF
--- a/integration/fargate_test.go
+++ b/integration/fargate_test.go
@@ -137,6 +137,23 @@ var _ = Describe("(Integration) Fargate", func() {
 		})
 	})
 
+	Context("Creating a cluster with --fargate and --managed", func() {
+		ft := &fargateTest{}
+
+		BeforeEach(func() {
+			setup(ft, "--fargate", "--managed")
+		})
+
+		It("should support Fargate", func() {
+			assertFargateDefaultProfile(ft.clusterName, ft.kubeTest)
+			assertFargateNewProfile(ft.clusterName, ft.kubeTest)
+		})
+
+		AfterEach(func() {
+			deleteCluster(ft.clusterName)
+		})
+	})
+
 	Context("Creating a cluster without --fargate", func() {
 		ft := &fargateTest{}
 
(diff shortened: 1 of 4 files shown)


integration/fargate_test.go
```diff
@@ -0,0 +1,173 @@
...
+		)
+		Expect(cmd).To(RunSuccessfully())
+	}
+
+	type fargateTest struct {
+		clusterName string
+		kubeTest    *harness.Test
+	}
+
+	setup := func(ft *fargateTest, createArgs ...string) {
```
martina-if: Is there a reason for keeping these functions as variables instead of extracting them as functions at the end of this file?

integration/fargate_test.go
```diff
@@ -0,0 +1,173 @@
...
+			"--cluster", clusterName,
+			"--name", "fp-default",
+			"--region", region,
+			"--wait",
+			"--verbose", "4",
+		)
+		Expect(cmd).To(RunSuccessfully())
+	}
+
+	assertFargateNewProfile := func(clusterName string, kubeTest *harness.Test) {
```
martina-if: since it's not only asserting what do you think about renaming this to `testCreateFargateProfile`?

integration/fargate_test.go
cPu1: Mainly to avoid polluting the package scope with these functions. All integration tests are part of the same `integration_test` package.

integration/fargate_test.go
cPu1: Sounds good. I'll rename the other to `testDefaultFargateProfile`

integration/fargate_test.go
martina-if: Right, lets leave it like that then and when we move it to its own package we extract them

Issue #1643, opened 2019-12-06
Log `--approve` warning for `update cluster` command


martina-if:
## Description

Today when `eksctl update cluster` command is run, the user needs to add the flag `--approve` to have eksctl actually perform the upgrade. That's because eksctl acts in "planning" mode. This is logged very subtlety:

```console
$ eksctl update cluster $CLUSTER_NAME
[ℹ]  eksctl version fb1e0f53d42f6ee88b56d6bb0e14ab2c2465fc36
[ℹ]  using region ap-northeast-1
[ℹ]  re-building cluster stack "eksctl-martina-test-f16-cluster"
[ℹ]  (plan) updating stack to add new resources [FargatePodExecutionRole] and outputs [FargatePodExecutionRoleARN]
[ℹ]  checking security group configuration for all nodegroups
[ℹ]  all nodegroups have up-to-date configuration
```

The key is in the log line number 4 `(plan) ...`.

## Expected

I think this command should print a warning as the last log line like in the case of deleting a nodegroup:

```console
$ eksctl delete nodegroup -f fargate-managed-normal.yaml --include ng-2
[ℹ]  eksctl version 0.2.0-1019-gdc6f0367
[ℹ]  using region ap-northeast-1
[ℹ]  comparing 2 nodegroups defined in the given config ("fargate-managed-normal.yaml") against remote state
[ℹ]  combined include rules: ng-2
[ℹ]  1 nodegroup (ng-2) was included (based on the include/exclude rules)
[ℹ]  2 nodegroups (managed-1, ng-1) were excluded (based on the include/exclude rules)
[ℹ]  (plan) would delete 1 nodegroups from auth ConfigMap in cluster "martina-test-all"
[ℹ]  (plan) would drain 1 nodegroup(s) in cluster "martina-test-all"
[ℹ]  (plan) would delete 1 nodegroups from cluster "martina-test-all"
[ℹ]  (plan) 1 task: { delete nodegroup "ng-2" [async] }
[✔]  (plan) would have deleted 1 nodegroup(s) from cluster "martina-test-all"
[!]  no changes were applied, run again with '--approve' to apply the changes
```

So it would look like:

```console
$ eksctl update cluster $CLUSTER_NAME
[ℹ]  eksctl version fb1e0f53d42f6ee88b56d6bb0e14ab2c2465fc36
[ℹ]  using region ap-northeast-1
[ℹ]  re-building cluster stack "eksctl-martina-test-f16-cluster"
[...]

PR #1795, opened 2020-02-08
Ability to print the `--approve` warning message in `update` command

spaghettifunk:
### Description

This PR fixes the condition for printing the `--approve` warning message. The condition has been loosen up and delegating the responsibility to the `cmd.Plan` variable for displaying the message.

Issue #1643

### Checklist
- [x] Manually tested


martina-if:
@spaghettifunk  thank you!

Normally this `cmdutils.LogPlanModeWarning()` is used with an extra condition on whether any changes were supposed to be applied but I think in this case it makes sense to relax the condition.

Can you please rebase this branch so we can merge it?

spaghettifunk:
> Actually I think we should modify the return value [here](https://github.com/weaveworks/eksctl/blob/master/pkg/cfn/manager/cluster.go#L174) to `true` instead. What do you think?

Yes it makes sense! Even if the user wants to `plan` the `AppendNewClusterStackResource` should `return true` when it is actually possible to append new resources. I update my branch and push. Thanks for the feedback 😄

pkg/ctl/update/cluster.go
```diff
@@ -132,7 +132,7 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 		logger.Critical("failed checking nodegroups", err.Error())
 	}
 
-	cmdutils.LogPlanModeWarning(cmd.Plan && (stackUpdateRequired || versionUpdateRequired))
+	cmdutils.LogPlanModeWarning(cmd.Plan)
```
martina-if: Thanks for making the changes :) 

I think that to keep the plan mode consistent across commands this should be left with the full condition as it was before. And because we fixed the other return value it should now print the message when it needs to.

martina-if:
@spaghettifunk thanks! Can you please squash the 11 commits into one?

martina-if:
superseded by #2069

Issue #1639, opened 2019-12-05
Ability to cusomize name of nodegroup instance


mshivaramie22:
Looks like currently there is no way to customise the instances that are created as part of nodegroup for EKS. This is making the EKS worker nodes look differently than the other instances in our organisation. We are trying to follow a naming pattern for these nodes and failing to do so. 

Please let the users choose the naming pattern of instances created by Nodegroups.

*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
To make the node group instance name in-line with the existing instances' naming convention.

**What feature/behavior/change do you want?**
An configuration item to set as a prefix or a name pattern for the node group instances.

Do not hesitate, when appropriate, to share the exact commands or API you would like, and/or to share a diagram (e.g.: asciiflow.com): "a picture is worth a thousand words".

spaghettifunk:
Hi @mshivaramie22 

Thanks for opening this issue! I was wondering if it would make sense to ad extra `configuration` parameter in the specs for a `nodeGroup`. Given the example below, we could add the extra flag `instancePrefix: ....`

```yaml
nodeGroups:
  - name: worker-ng-1
    labels:
      role: workers
      autoscaling: enabled
      purpose: ci-worker
    instanceType: m5.large
    instancePrefix: my-prefix
    desiredCapacity: 1
    minSize: 1
    maxSize: 3
    tags:
      environment: development
```
Each instance would then be named`my-prefix-cluster-1-worker-ng-1-Node`

spaghettifunk:
I made a PR about this issue. The picture below shows the result when overriding the nodes' name

<img width="1087" alt="Screenshot 2020-02-08 at 18 06 45" src="https://user-images.githubusercontent.com/1409994/74089275-d9a5d200-4a9f-11ea-906f-3397a335aab6.png">

foot:
Nice one! @mshivaramie22 does this cover your use case?

mshivaramie22:
Yes ! This was what I was looking for ! Thanks alot for taking this up...

jansony1:
+1. 
Maybe we could just add a switch to determine whether we need inherit name from cluster.

just turn on or turn off the "true" button

`{
			"Key":               "Name",
			"Value":             fmt.Sprintf("%s-%s-Node", n.clusterSpec.Metadata.Name, n.nodeGroupName),
[...]

PR #1796, opened 2020-02-08
Ability of changing the node instance name

spaghettifunk:
### Description

This PR introduce the possibility to add a prefix or override, the node instance's name. To use a prefix in front of the automatically-generated name, it is now possible to use the `instancePrefix` parameter under the `NodeGroups` object. The same for the `instanceName`.

**How it works**
When using the `instancePrefix`, the value will be appended in front of the generated name. If only `instanceName` is used, then this parameter will override the default naming to each node. 
It is possible to use both `instancePrefix` and `instanceName` at the same time.

**Example**
In folder `examples/18-override-nodes-name.yaml` there is an example of how it is possible to use these new parameters.

Issue #1639 

### Checklist
- [x] Manually tested


pkg/ctl/cmdutils/configfile_test.go
```diff
@@ -81,7 +82,7 @@ var _ = Describe("cmdutils configfile", func() {
 			examples, err := filepath.Glob(examplesDir + "*.yaml")
 			Expect(err).ToNot(HaveOccurred())
 
-			Expect(examples).To(HaveLen(17))
+			Expect(examples).To(HaveLen(numOfExamples))
```
foot: 👍

site/content/usage/20-schema.md
```diff
@@ -347,6 +347,10 @@ NodeGroup:
     iam:
       $ref: '#/definitions/NodeGroupIAM'
       $schema: http://json-schema.org/draft-04/schema#
+    instanceName:
+      type: string
+    instancePrefix:
+      type: string
```
foot: Do we have to any schema versions bumps after adding new properties?

pkg/cfn/builder/nodegroup.go
```diff
@@ -185,6 +186,21 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
 	return nil
 }
 
+// generateNodeName formulates the name based on the configuration in input
+func (n *NodeGroupResourceSet) generateNodeName() string {
+	name := []string{}
+	if n.spec.InstancePrefix != "" {
+		name = append(name, n.spec.InstancePrefix, "-")
```
foot: Omitting the `-` here could add a small amount of flexibility, `acme_` / `acme:`

spaghettifunk: True! However, we may end-up with a "strange" name. For example, let's say that you chose a prefix `foo` and a name `bar`. Without it it will look like `foobar` altogether. Is this something you have in mind? The reason why I put it there was to avoid having the user to write something like

```yaml
instancePrefix: foo-     # the extra - on the end
instanceName: bar
```

I also thought about including a `separator` among the NodeGroup configuration. I wasn't sure if it was actually necessary such a flexibility 😄 . 
Anyway, what do you think we should do in this case:
- add an extra flag called `instanceNameSeparator`
- remove the `-` and leave it without any prefixed separator
- use another separator instead
- ...

site/content/usage/20-schema.md
spaghettifunk: Now that I think about, it may be necessary. However, I don't know how to do it 😄

spaghettifunk:
Totally forgot about the flags! I just updated the branch @foot  😄

pkg/cfn/builder/nodegroup.go
foot: For sure, thats makes sense, and the prefix would mostly be coupled w/ the generated name which uses `-`s. `instanceNameSeparator` might be overkill :).

spaghettifunk: Shall I leave the code as it is?

foot: Yes, I would say so. 👍

site/content/usage/20-schema.md
foot: Looking at similar changes to this schema it doesn't look like a version bump is required, so, all good.

michaelbeaumont:
@spaghettifunk do you think you could rebase this on master?

spaghettifunk:
@michaelbeaumont I will rebase tonight after work hours 😄

michaelbeaumont:
@spaghettifunk ended up taking care of the CI issue, thanks for the contribution!

spaghettifunk:
@michaelbeaumont great! thanks :)

neilharris123:
@michaelbeaumont great work, does this only work with unmanaged nodegroups?

cPu1:
> @michaelbeaumont does this only work with unmanaged nodegroups?

Correct. The AWS Managed Nodegroups API does not support setting the instance name.

PR #1798, opened 2020-02-10
Add makefile command to manually publish a release

martina-if:
This PR makes it possible to decouple the building of binaries with the publishing of 
a release.

First, one can compile eksctl for all platforms using `make build-all` (which calls goreleaser
without publishing any release). Then one can do 
  - `make publish-github` or `make publish-rc-github`
  - and `make publish-homebrew`

*Note*: This relies on having a draft release already in GH with the release notes.

This code, can also be used from CircleCI if we want to automate the process.

- [ ] Manually tested


Makefile
```diff
@@ -192,6 +192,35 @@ prepare-release-candidate:
...
+.PHONY: publish-github
+publish-github: upload-github
+	github-release publish --user $(git_org) --repo $(git_repo) --tag $(eksctl_version)
+
+.PHONY: publish-rc-github
+publish-rc-github: upload-github
+	github-release release --user $(git_org) --repo $(git_repo) --tag $(shell eksctl version) --pre-release
+
+.PHONY: publish-homebrew
+publish-homebrew:
```
marccarre: Should we call this as part of `publish-github`? (In that case we could maybe rename the targets `publish-release` and `publish-rc`? Or add these which would call `*-github` and `publish-homebrew`?)

martina-if: I thought about it. I prefer to leave it de-coupled for now until we start using it. Then we can decide how to use it locally or from CircleCI. We can always refactor these later when we know more about how we want to use them.

marccarre: Sounds good to me! 👍

PR #1803, opened 2020-02-11
Update gitops docs

stefanprodan:
Update the docs to reflect the changes in #1761

Docs changes:
- mention that Flux Helm Operator comes with Helm v3 support
- remove Tiller references


stefanprodan:
@martina-if I've fixed the profile description

PR #1814, opened 2020-02-13
Default to SSM Parameter for AMI resolution

### LINKED ISSUE #1793: Default to SSM Parameter Store for resolving AMIs
eksctl defaults to obtaining the latest AMI for an EKS version and AMI family by looking up a [static list](https://github.com/weaveworks/eksctl/blob/d10fe715ca5caf7a67c226e683e57d864073fb97/pkg/ami/static_resolver_ami.go). There is no automated process to generate this list and it can quickly become obsolete as new AMIs are released. Not to mention, the AWS account used to generate this list needs access to all regions supported by eksctl.

Defaulting to the SSM Parameter Store for obtaining the latest AMI solves the shortcomings of the static resolution process. However, it is not supported for Ubuntu yet, so in order to continue supporting all node image families, eksctl can try SSM Parameter Store first, and if the query isn't supported, fall back to querying the latest AMI using `ec2:DescribeImages` and sorting (`NodeImageResolverAuto`) during cluster creation. Static resolution can also be used as a fallback but it runs the risk of giving a stale AMI ID. This will also help eliminate support for static resolution entirely due to its aforementioned shortcomings.

This process will be transparent to users not explicitly specifying an AMI resolution strategy (`NodeGroup.AMI = ""`), and static resolution will continue to be supported by explicitly setting `NodeGroup.AMI = "static"`.

cPu1:
### Description

Closes #1793 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -266,34 +266,38 @@ func (c *ClusterProvider) CheckAuth() error {
 }
 
 // EnsureAMI ensures that the node AMI is set and is available
-func (c *ClusterProvider) EnsureAMI(version string, ng *api.NodeGroup) error {
+func EnsureAMI(provider api.ClusterProvider, version string, ng *api.NodeGroup) error {
 	if api.IsAMI(ng.AMI) {
-		return ami.Use(c.Provider.EC2(), ng)
+		return ami.Use(provider.EC2(), ng)
 	}
 
 	var resolver ami.Resolver
 	switch ng.AMI {
 	case api.NodeImageResolverAuto:
-		resolver = ami.NewAutoResolver(c.Provider.EC2())
+		resolver = ami.NewAutoResolver(provider.EC2())
 	case api.NodeImageResolverAutoSSM:
-		resolver = ami.NewSSMResolver(c.Provider.SSM())
+		resolver = ami.NewSSMResolver(provider.SSM())
+	case api.NodeImageResolverStatic:
+		resolver = ami.NewStaticResolver()
 	default:
-		resolver = ami.NewDefaultResolver()
+		resolver = ami.NewMultiResolver(
+			ami.NewSSMResolver(provider.SSM()),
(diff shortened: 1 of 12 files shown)


pkg/eks/api.go
```diff
@@ -266,34 +266,38 @@ func (c *ClusterProvider) CheckAuth() error {
...
 	case api.NodeImageResolverAutoSSM:
-		resolver = ami.NewSSMResolver(c.Provider.SSM())
+		resolver = ami.NewSSMResolver(provider.SSM())
+	case api.NodeImageResolverStatic:
+		resolver = ami.NewStaticResolver()
 	default:
-		resolver = ami.NewDefaultResolver()
+		resolver = ami.NewMultiResolver(
+			ami.NewSSMResolver(provider.SSM()),
+			ami.NewAutoResolver(provider.EC2()),
```
martina-if: should we have the static one as the last delegate?

pkg/eks/api_test.go
```diff
@@ -141,19 +127,76 @@ var _ = Describe("eksctl API", func() {
...
+			provider = mockprovider.NewMockProvider()
+			mockDescribeImages(provider, "ami-123", func(input *ec2.DescribeImagesInput) bool {
+				return len(input.ImageIds) == 1
+			})
+
+		})
+
+		testEnsureAMI := func(expectedAMI string) {
+			err := EnsureAMI(provider, "1.14", ng)
+			ExpectWithOffset(1, err).ToNot(HaveOccurred())
```
martina-if: Nice one, I didn't know about `ExpectWithOffset()`

pkg/eks/api.go
cPu1: Auto resolution uses the same set of calls used by the static generation process, so if auto resolution fails, there's a good chance the static lookup will also fail.
Also, the plan is to deprecate static resolution and eventually remove it.

Issue #1703, opened 2020-01-09
deleting nodegroup while creating new ng leads to empty authconfigmap


lachlanbb:
**What happened?**
Running eksctl delete nodegroup -f nodegroup.yaml causes node/bootstrap role to be removed from AWS-AUTH config map... This instantly breaks every application and node in the cluster. 

**What you expected to happen?**
The cause of this is cancelling out a nodegroup creation before its finished and running the delete command straight after. It has caused outages for me twice now and I'm never using EKSCTL because of this. 

Fix both the AWS API throttling logic and aws-auth change validation. 

**How to reproduce it?**
eksctl delete nodegroup -f file

**Anything else we need to know?**
The nodegroup definition IAM key is as follows
```
iam:
      instanceRoleARN: "arn:aws:iam::123456789012:role/EKS-NodesRole"
      instanceProfileARN: "arn:aws:iam::123456789012:instance-profile/EKS-NodesInstanceProfile"
```

**Versions**
**eksctl:** 0.10.2

**Kubectl:**
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-14T04:24:34Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-c0eccc", GitCommit:"c0eccca51d7500bb03b2f163dd8d534ffeb2f7a2", GitTreeState:"clean", BuildDate:"2019-12-22T23:14:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}

**Logs**
eksctl delete nodegroup -f nodegroup-spot-nvme-b.yaml --approve
eksctl version 0.10.2
using region ap-southeast-2
comparing 1 nodegroups defined in the given config ("nodegroup-spot-nvme-b.yaml") against remote state
nodegroup "spot-a" present in the cluster, but missing from the given config
nodegroup "nodes-b" present in the cluster, but missing from the given config
nodegroup "spot-c" present in the cluster, but missing from the given config
nodegroup "nodes-c" present in the cluster, but missing from the given config
nodegroup "nodes-a" present in the cluster, but missing from the given config
nodegroup (spot-b) was included (based on the include/exclude rules)
[...]

Issue #1368, opened 2019-09-27
Remove node role from auth configmap in the end while deleting node groups


dhanvi:
**What happened?**

eksctl is deleting the role from auth configmap when starting the node group deletion process.

**What you expected to happen?**

eksctl should be removing the role from auth configmap only after the nodes in the node groups are perfectly drained.

Removing the role from auth before draining nodes is destructive and shouldn't be done. Attached logs.

**Versions**
```
  ➜  ~ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.5.3"}
  ➜  ~ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.10-eks-5ac0f1", GitCommit:"5ac0f1d9ab2c254ea2b0ce3534fd72932094c6e1", GitTreeState:"clean", BuildDate:"2019-08-20T22:39:46Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
```
[ℹ]  using region ap-southeast-1
[ℹ]  comparing 5 nodegroups defined in the given config ("eks_config.yaml") against remote state
[ℹ]  nodegroup "ng-1" present in the cluster, but missing from the given config
[ℹ]  combined include rules: ng-1
[ℹ]  1 nodegroup (ng-1) was included (based on the include/exclude rules)
[ℹ]  combined exclude rules: ng-1-v2,ng-2,ng-3,ng-4,ng-5
[ℹ]  5 nodegroups (ng-1-v2, ng-2, ng-3, ng-4, ng-5) were excluded (based on the include/exclude rules)
[ℹ]  will delete 1 nodegroups from auth ConfigMap in cluster "ap-southeast-1-prod-eks"
[!]  retryable error (Throttling: Rate exceeded
	status code: 400, request id: 97) from cloudformation/DescribeStacks - will retry after delay of 540ms
[...]

PR #1818, opened 2020-02-13
WIP: Identity removal from nodegroup deletion postponed to after.

d-luu:
The identity removal of the role ARN from the auth configmap was moved
to after the nodegroup deletion when the wait flag is used.
This was done to avoid issues where a nodegroup fails to delete for
whatever reason and be left with orphaned nodes without an identity
if it were to be deleted up front before the deletion of the nodegroup(s).

Deletion of identities from the auth configmap map via the delete
nodegroup call have been moved to tasks for more verbose planning.

Also for nodegroup deletion, stack descriptions have now been cached to
avoid duplicate calls to the Cloudformation API.

Issue #1703, #1368

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/authconfigmap/authconfigmap.go
+++ b/pkg/authconfigmap/authconfigmap.go
@@ -8,7 +8,6 @@ package authconfigmap
 
 import (
 	"fmt"
-
 	"github.com/aws/aws-sdk-go/aws/awsutil"
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
@@ -182,6 +181,7 @@ func (a *AuthConfigMap) RemoveIdentity(arnToDelete string, all bool) error {
 	if !all {
 		return fmt.Errorf("instance identity ARN %q not found in auth ConfigMap", arnToDelete)
 	}
+
 	return a.setIdentities(newidentities)
 }
 
@@ -244,7 +244,6 @@ func (a *AuthConfigMap) Save() (err error) {
 		a.cm, err = a.client.Create(a.cm)
 		return err
 	}
-
 	a.cm, err = a.client.Update(a.cm)
 	return err
 }
@@ -299,5 +298,22 @@ func RemoveNodeGroup(clientSet kubernetes.Interface, ng *api.NodeGroup) error {
 		return errors.Wrap(err, "updating auth ConfigMap after removing role")
 	}
 	logger.Debug("updated auth ConfigMap for %s", ng.Name)
+	return nil
+}
+
+// RemoveARNIdentity removes an ARN identity from the ConfigMap and
+// does a client update.
+func RemoveARNIdentity(clientSet kubernetes.Interface, arnToDelete string, all bool) error {
+	acm, err := NewFromClientSet(clientSet)
+	if err != nil {
+		return err
+	}
+	if err := acm.RemoveIdentity(arnToDelete, all); err != nil {
+		return errors.Wrap(err, "removing identity from auth ConfigMap")
+	}
+	if err := acm.Save(); err != nil {
+		return errors.Wrap(err, "updating auth ConfigMap after removing role")
+	}
+	logger.Debug("updated auth ConfigMap for %s", arnToDelete)
 	return nil
 }
--- a/pkg/authconfigmap/authconfigmap.go
+++ b/pkg/authconfigmap/authconfigmap.go
@@ -8,6 +8,7 @@ package authconfigmap
 
 import (
 	"fmt"
+
 	"github.com/aws/aws-sdk-go/aws/awsutil"
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
@@ -281,23 +282,6 @@ func AddNodeGroup(clientSet kubernetes.Interface, ng *api.NodeGroup) error {
 		return errors.Wrap(err, "saving auth ConfigMap")
 	}
 	logger.Debug("saved auth ConfigMap for %q", ng.Name)
-	return nil
-}
-
-// RemoveNodeGroup removes a nodegroup from the ConfigMap and
-// does a client update.
-func RemoveNodeGroup(clientSet kubernetes.Interface, ng *api.NodeGroup) error {
-	acm, err := NewFromClientSet(clientSet)
-	if err != nil {
-		return err
-	}
-	if err := acm.RemoveIdentity(ng.IAM.InstanceRoleARN, false); err != nil {
-		return errors.Wrap(err, "removing nodegroup from auth ConfigMap")
-	}
-	if err := acm.Save(); err != nil {
-		return errors.Wrap(err, "updating auth ConfigMap after removing role")
-	}
-	logger.Debug("updated auth ConfigMap for %s", ng.Name)
 	return nil
 }
 
--- a/pkg/cfn/manager/delete_tasks.go
+++ b/pkg/cfn/manager/delete_tasks.go
@@ -2,10 +2,9 @@ package manager
 
 import (
 	"fmt"
-
 	"github.com/aws/aws-sdk-go/service/cloudformation"
-
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/authconfigmap"
 	iamoidc "github.com/weaveworks/eksctl/pkg/iam/oidc"
 	"github.com/weaveworks/eksctl/pkg/kubernetes"
 )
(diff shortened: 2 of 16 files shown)


d-luu:
There is an issue on the AWS side when creating managed node groups that share the same IAM instance role with the unmanaged node groups. Deletion of the managed node groups (either through eksctl or through the AWS portal/cli) will remove the identity (all entries of it) from the auth configmap and orphan the unmanaged nodes.

pkg/cfn/manager/delete_tasks.go
```diff
@@ -67,7 +66,11 @@ func (c *StackCollection) NewTasksToDeleteNodeGroups(shouldDelete func(string) b
...
 		return nil, err
 	}
+	return c.NewTasksToDeleteNodeGroupsFromCFNDescriptions(shouldDelete, wait, cleanup, nodeGroupStacks)
+}
 
+// NewTasksToDeleteNodeGroupsFromCFNDescriptions defines tasks required to delete all of the nodegroups from a list of cloudformation stack descriptions
+func (c *StackCollection) NewTasksToDeleteNodeGroupsFromCFNDescriptions(shouldDelete func(string) bool, wait bool, cleanup func(chan error, string) error, nodeGroupStacks []*Stack) (*TaskTree, error) {
 	tasks := &TaskTree{Parallel: true}
 
 	for _, s := range nodeGroupStacks {
```
martina-if: > if !shouldDelete(name) {
			continue
		}

I think the filtering part should be moved into `NewTasksToDeleteNodeGroups()` so that the responsibility of this function is to just delete any task that is provided. That would be a better API IMO. What do you think?

pkg/cfn/manager/delete_tasks.go
```diff
@@ -181,3 +184,30 @@ func (c *StackCollection) NewTasksToDeleteIAMServiceAccounts(shouldDelete func(s
...
+	for _, n := range cfgMarkedForDeletion.NodeGroups {
+		tasks.Append(&authConfigMapTask{
+			info:      fmt.Sprintf("remove identity role arn %q from auth configmap", n.IAM.InstanceRoleARN),
+			clientSet: clientSet,
+			call: func(clientSet kubernetes.Interface) error {
+				return authconfigmap.RemoveARNIdentity(clientSet, n.IAM.InstanceRoleARN, false)
+			},
+		})
+	}
+	for _, n := range cfgMarkedForDeletion.ManagedNodeGroups {
```
palemtnrider: Please refactor as there is redundant code between these two blocks.

pkg/ctl/delete/nodegroup.go
```diff
@@ -74,16 +76,32 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 
 	stackManager := ctl.NewStackManager(cfg)
 
+	descriptions, err := stackManager.DescribeNodeGroupStacks()
```
palemtnrider: Any particular reason we are ignoring the error from this call?

pkg/ctl/delete/nodegroup_test.go
```diff
@@ -0,0 +1,364 @@
...
+			TemplateBody: aws.String(createTemplateBody(api.NodeGroupTypeUnmanaged, "ng2", "my-cluster", ng2.IAM.InstanceRoleARN)),
+		}, nil)
+
+	mockCFN.On("GetTemplate",
+		mock.MatchedBy(func(input *cloudformation.GetTemplateInput) bool {
+			return input.StackName != nil && *input.StackName == "mng1"
+		})).
+		Return(&cloudformation.GetTemplateOutput{
+			TemplateBody: aws.String(createTemplateBody(api.NodeGroupTypeManaged, "mng1", "my-cluster", ng1.IAM.InstanceRoleARN)),
+		}, nil)
```
cPu1: Ideally, a new mock provider is created for each test case, defining only the mock functions expected in that test. Then if an unexpected call is made, the mock library throws an error and lets us catch unexpected calls.

pkg/authconfigmap/authconfigmap.go
```diff
@@ -301,3 +300,20 @@ func RemoveNodeGroup(clientSet kubernetes.Interface, ng *api.NodeGroup) error {
 	logger.Debug("updated auth ConfigMap for %s", ng.Name)
 	return nil
 }
+
+// RemoveARNIdentity removes an ARN identity from the ConfigMap and
+// does a client update.
+func RemoveARNIdentity(clientSet kubernetes.Interface, arnToDelete string, all bool) error {
```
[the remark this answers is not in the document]
d-luu: I went ahead and removed `RemoveNodeGroup()` since it was being used anymore and made the other changes you requested

pkg/cfn/manager/delete_tasks.go
d-luu: To clarify, do you mean moved out of rather than into? if anything passed to NewTasksToDeleteNodeGroups is to be deleted

I do think that would be better since its more succinct

pkg/cfn/manager/delete_tasks.go
d-luu: I removed the ManagedNodeGroup loop and put in a comment mentioning the aws api doing the deletion

pkg/ctl/delete/nodegroup.go
d-luu: It's a miss on my part, I went ahead and added it in

pkg/cfn/manager/nodegroup.go
```diff
@@ -102,6 +102,11 @@ func (c *StackCollection) ListNodeGroupStacks() ([]NodeGroupStack, error) {
 	if err != nil {
 		return nil, err
 	}
+	return c.ListNodeGroupStacksFromDescriptions(stacks)
+}
+
+// ListNodeGroupStacksFromDescriptions returns a list of NodeGroupStacks when passed in an existing list of stack descriptions
+func (c *StackCollection) ListNodeGroupStacksFromDescriptions(stacks []*Stack) ([]NodeGroupStack, error) {
```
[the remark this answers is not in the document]
d-luu: I agree with the naming change, I wasn't too keen on the one I put originally to begin with

pkg/ctl/delete/nodegroup.go
```diff
@@ -155,7 +159,59 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
...
+	for _, n := range cfgMarkedForDeletion.NodeGroups {
+		if n.IAM == nil || n.IAM.InstanceRoleARN == "" {
+			if err := ctl.GetNodeGroupIAMFromCFNDescriptions(stackManager, cfgMarkedForDeletion, n, descriptions); err != nil {
+				// we want to return the error instead of logging as we don't want to delete something prematurely
+				return err
+			}
+		}
+		numToDelete++
+	}
+	for _, n := range cfgMarkedForDeletion.ManagedNodeGroups {
```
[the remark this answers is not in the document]
d-luu: went ahead and removed and placed a comment

pkg/cfn/manager/tasks.go
```diff
@@ -258,6 +258,22 @@ func (t *kubernetesTask) Do(errs chan error) error {
 	return err
 }
 
+type authConfigMapTask struct {
+	info      string
+	clientSet kubernetes.Interface
```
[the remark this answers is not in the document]
d-luu: I agree this is better
I tried to keep with the pattern that I was seeing originally

pkg/ctl/delete/nodegroup.go
```diff
@@ -143,7 +144,10 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 			return false
 		}
 
-		tasks, err := stackManager.NewTasksToDeleteNodeGroups(shouldDelete, cmd.Wait, nil)
+		// NOTE: there is an issue in AWS where if a managed node group is using an IAM instance role also found in an
+		//   unmanaged node group, it will delete all identity entries in the auth configmap and possibly orphaning the
+		//   workers without having us call the removeARN function below to do so
```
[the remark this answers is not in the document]
d-luu: This is what I experienced with the added note that it removes all entries of the instance role ARN if its being shared

pkg/ctl/delete/nodegroup_test.go
d-luu: I moved it to a function and call it in each test now

Issue #1608, opened 2019-11-27
eksctl cannot be built while offline


marccarre:
**What happened?**

```console
$ make build
go-bindata version 3.15.0
env GOBIN=${GOPATH}/bin time go generate ./pkg/nodebootstrap ./pkg/addons/default ./pkg/addons
2019/11/22 17:05:00 Get https://raw.github.com/awslabs/amazon-eks-ami/master/files/eni-max-pods.txt: dial tcp: lookup raw.github.com: no such host
exit status 1
pkg/nodebootstrap/userdata.go:19: running "go": exit status 1
pkg/addons/default/generate.go:3: running "curl": exit status 6
        1.32 real         0.79 user         0.57 sys
make: *** [generate-always] Error 1
```

**What you expected to happen?**

We ideally shouldn't need to be online to run `make build`, or at least we should be able to "fail gracefully"/skip these steps so that the build can actually happen with whatever has already been pulled from the web.

michaelbeaumont:
Can be built offline now

PR #1819, opened 2020-02-14
Support for offline build and test

d-luu:
There are files used in the build/test to go generate that make
calls to retrieve files from online resources.
This change would allow using artifacts already downloaded.

Artifacts needed:
- pkg/nodebootstrap/maxpods.go
- pkg/addons/default/assets/aws-node.yaml
    
Usage:
export EKSCTL_BUILD_OFFLINE=true

Note that when submitting a PR, you should unset the environment
variables so that you can get the latest files.

Issue #1608

### Checklist
- [x] Manually tested


martina-if:
Hi @d-luu , thank you for contributing to eksctl :)

I checked the PR and although it's a good idea I think we should solve this in a different way.

eksctl has many build time dependencies. Some of these are generated code (which doesn't need any connection to the internet) and others are more of a "downloading" operation.

What I would like to have is a way to separate these two kinds of resources so that the generated code is always generated when we run `make build` but the resources that are downloaded from the internet are only updated when we want to do that (for example, with a `make upgrade-assets` or similarly named command). This would also help having reproducible build under our control.

Would you like to work on that?

d-luu:
@martina-if I can make the changes you requested

d-luu:
@martina-if I have updated the makefile to include a download-assets rule

I have also added a conditional in the generate where if any files are missing it will try and download them (I was going to try this using the `wildcard` function with `ifeq` but the shell conditionally was cleaner in my opinion)...this was done so that it doesnt break current build processes

martina-if:
Superseded by https://github.com/weaveworks/eksctl/pull/2039

PR #1821, opened 2020-02-14
Fix bug in update-cluster-endpoint #1820

### LINKED ISSUE #1820: update-cluster-endpoint not using correct values from flags
When we merged #1693 we introduced a bug.

The fields in the `ClusterConfig` object are always set by the loader with default values which means https://github.com/weaveworks/eksctl/pull/1693/files#diff-3d2e69c16896f6dcabb52f22a3fdac50R88 was not a good way to determine whether we are dealing with a config file or not.

As a result, the command using argument flags always tries to set the default values.

This is breaking the integration tests.

martina-if:
fixes #1820

- [x] Manually tested
- [x] Make sure Integration tests pass

## Integration tests passing

```

Ran 10 of 10 Specs in 3830.525 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestSuite (3830.53s)
PASS
```

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -95,10 +95,6 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 		}
 	}
 
-	if !cfg.HasClusterEndpointAccess() {
-		return ErrClusterEndpointNoAccess
-	}
-
 	if cfg.VPC != nil && len(cfg.VPC.PublicAccessCIDRs) > 0 {
 		cidrs, err := validateCIDRs(cfg.VPC.PublicAccessCIDRs)
 		if err != nil {
@@ -111,6 +107,9 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 
 // ValidateClusterEndpointConfig checks the endpoint configuration for potential issues
 func (c *ClusterConfig) ValidateClusterEndpointConfig() error {
+	if !c.HasClusterEndpointAccess() {
+		return ErrClusterEndpointNoAccess
+	}
 	endpts := c.VPC.ClusterEndpoints
 	if NoAccess(endpts) {
 		return ErrClusterEndpointNoAccess
--- a/pkg/ctl/utils/update_cluster_endpoint_access.go
+++ b/pkg/ctl/utils/update_cluster_endpoint_access.go
@@ -32,7 +32,7 @@ func updateClusterEndpointsCmd(cmd *cmdutils.Cmd) {
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 	})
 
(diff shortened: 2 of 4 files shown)


pkg/ctl/utils/update_cluster_endpoint_access.go
```diff
@@ -85,17 +72,16 @@ func doUpdateClusterEndpoints(cmd *cmdutils.Cmd, newPrivate bool, newPublic bool
...
+		if !flagSet("private-access", cmd) {
 			newPrivate = curPrivate
 		}
-		if !publicSet {
+		if !flagSet("public-access", cmd) {
 			newPublic = curPublic
 		}
+	} else {
+		newPublic = *cfg.VPC.ClusterEndpoints.PublicAccess
+		newPrivate = *cfg.VPC.ClusterEndpoints.PrivateAccess
```
cPu1: If a ClusterConfig file is supplied with a blank `vpc.clusterEndpoints` field, the defaults set in `ClusterEndpointAccessDefaults()` can override the current settings.

cPu1:
Apologies, I left a couple comments but just saw that the PR has been updated to WIP.

pkg/ctl/utils/update_cluster_endpoint_access.go
martina-if: You are right. Good catch! I'll redo this part as I did initially by changing the loader and moving the validation to the loader instead of having it in `NewCtl()`

martina-if:
@cPu1 I implemented it in a different way. I ran a few manual tests and I think it should be fine now, but please take a look and tell me if you find any potential issues. 

The current way of defining defaults, loading and validation is not good for these cases... we should refactor all this.

In the meantime I'm checking if the endpoint access integration tests are working.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -111,6 +107,9 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 
 // ValidateClusterEndpointConfig checks the endpoint configuration for potential issues
 func (c *ClusterConfig) ValidateClusterEndpointConfig() error {
+	if !c.HasClusterEndpointAccess() {
```
sayboras: @martina-if this check is fine, however the implementation of `HasClusterEndpointAccess` is a little bit weird for me

```
func (c *ClusterConfig) HasClusterEndpointAccess() bool {
	if c.VPC != nil && c.VPC.ClusterEndpoints != nil {
		pubAccess := c.VPC.ClusterEndpoints.PublicAccess
		privAccess := c.VPC.ClusterEndpoints.PrivateAccess
		hasPublicAccess := pubAccess != nil && *pubAccess
		hasPrivateAccess := privAccess != nil && *privAccess
		return hasPublicAccess || hasPrivateAccess
	}
	return true // so if nothing is specficified, it returns true, which is has cluster endpoint configured.
}
```

PR #1857, opened 2020-02-27
Remove integration tests timeout

martina-if:
Removing it since they still take a very long time to run and they timeout:

```
panic: test timed out after 3h30m0s.
```


martina-if:
@marccarre @cPu1 I am merging this small PR. Let me know if you disagree but I want to see if the tests work with this change.

PR #1746, opened 2020-01-22
Added the ability to override the kube config cluster server.

treacher:
### Description

This change allows users with a private cluster to be able to use SSH
tunnelling to access the cluster from their local machine via a
port other than 443.

The KUBECONFIG_CLUSTER_ENDPOINT environment variable can now be
set to override the cluster endpoint in the kube config object
that's generated via kubeconfig.go.

This change references #1552 and #1618

### Manual Testing Performed

Ran the following commands with and without setting the `KUBECONFIG_CLUSTER_ENDPOINT`

- eksctl get iamidentitymapping
- eksctl create iamidentitymapping
- eksctl delete iamidentitymapping
- eksctl create iamserviceaccount
- eksctl delete iamserviceaccount
- eksctl create nodegroup
- eksctl delete nodegroup
- eksctl drain nodegroup
- eksctl delete cluster

I observed that the above commands were still using the EKSEndpoint output from the EKS Cluster stack when not setting the KUBECONFIG_CLUSTER_ENDPOINT environment variable.

I observed that the above commands used my specified endpoint when setting the KUBECONFIG_CLUSTER_ENDPOINT environment variable and that the command succeeded. 

### Checklist
- [x] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, and `examples` directory)
- [x] Manually tested
- [ ] Added labels for change area (e.g. `area/nodegroup`) and target version (e.g. `version/0.12.0`)
- [x] Added note in `docs/release_notes/draft.md` (or relevant release note)

PR #1878, opened 2020-03-03
Revert "Merge pull request #1746 add ability to override kubeconfig cluster endpoint"

treacher:
### Description

After testing 0.14.0 this morning to create a new node group I noticed that the nodes were not coming up. After further investigation into the issue it turned out that the `KUBECONFIG_CLUSTER_ENDPOINT` was being used in the servers kubeconfig file as well.

We're using something along the lines of this as we're using SSH tunnelling:
```
export KUBECONFIG_CLUSTER_ENDPOINT=https://$KUBERNETES_API:8001
```

My initial assumption was that the kubeconfig file generated locally for performing eksctl commands was created separately from the one generated on the Kubernetes host but it turns out both of these are created from the same process.

Because of this, the change I've added causes the nodes not to be able to talk to the API as they're trying to access it on port 8001 instead of 443.

I clearly didn't test this change thoroughly enough with creating node groups and for that I'm sorry.

### DIFF
--- a/pkg/utils/kubeconfig/kubeconfig.go
+++ b/pkg/utils/kubeconfig/kubeconfig.go
@@ -63,12 +63,6 @@ func New(spec *api.ClusterConfig, username, certificateAuthorityPath string) (*c
 		CurrentContext: contextName,
 	}
 
-	clusterEndpoint, hasClusterEndpoint := os.LookupEnv("KUBECONFIG_CLUSTER_ENDPOINT")
-
-	if hasClusterEndpoint {
-		c.Clusters[clusterName].Server = clusterEndpoint
-	}
-
 	if certificateAuthorityPath == "" {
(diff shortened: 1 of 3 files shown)


treacher:
I'm not sure whether to put this change in 0.15.0 or whether it should be a hotfix. Let me know and I'll update what is required accordingly :)

martina-if:
Hi @treacher thanks a lot :) 

If I understand correctly, this only affects users who use this override, correct? If that is the case, I think it's enough to merge it into master and into the branch `release-0.15`. Can you open a "clone" PR to that branch as well?

martina-if:
@treacher also can you please rebase so I can merge this?

treacher:
@martina-if all rebased and I created a clone PR here: https://github.com/weaveworks/eksctl/pull/1889 =)

Issue #1586, opened 2019-11-20
set update-auth-configmap false by default


jithendra89:
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
A clear and concise description of why you want this feature. Include use-cases and any specific constraints you may have.
Im updating nodegroups in the existing cluster. I have created new node groups using the same cluster role. Then I'm trying to delete and drain node groups as described in https://eksctl.io/usage/managing-nodegroups/. But when I do cluster breaks and authentication failed due to eksctl updating configmap and it has set cluster roles to []. 
I think its best user decided whether to update configmap or not. 

**What feature/behavior/change do you want?**
A clear and concise description of the change you would like to see in eksctl.

Make --update-auth-configmap default false so config map don't get updates by default.

rndstr:
closing this as duplicate of #1703 since the newer one has more information and discussion going on now, thanks for submitting!

PR #1881, opened 2020-03-03
WIP Change default value of update-config-map flag for ng deletion

sayboras:
### Description
Fixes https://github.com/weaveworks/eksctl/issues/1703
Fixes https://github.com/weaveworks/eksctl/issues/1586

This flag is only used in `delete nodegroup` command.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/cmdutils.go
+++ b/pkg/ctl/cmdutils/cmdutils.go
@@ -179,7 +179,7 @@ func AddWaitFlagWithFullDescription(fs *pflag.FlagSet, wait *bool, description s
 
 // AddUpdateAuthConfigMap adds common --update-auth-configmap flag
 func AddUpdateAuthConfigMap(fs *pflag.FlagSet, updateAuthConfigMap *bool, description string) {
-	fs.BoolVar(updateAuthConfigMap, "update-auth-configmap", true, description)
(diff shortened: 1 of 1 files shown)


martina-if:
This LGTM, @cPu1 what do you think?

sayboras:
> in any case, this PR also changes it for `create ng` which must be changed:
> https://github.com/weaveworks/eksctl/blob/57171882dd10b6a0ad28b5cd4827953520b23a33/pkg/ctl/create/nodegroup.go#L52

I overlooked this file as I read `Remove nodegroup IAM role from aws-auth configmap` description, and assumed that it's the same deletion scenario. Meanwhile, during nodegroup creation, we will be ### **adding** entry into config map. 

With this information, changing the default value to `false` will cause the newly created nodegroup unable to join the cluster. Thanks for pointing it out.

I am marking this PR WIP and we will continue discussing about approach in https://github.com/weaveworks/eksctl/issues/1703.

martina-if:
@rndstr @sayboras what should we do about this PR then?

sayboras:
I think we can close this PR. 

Just a question, is it a good time to address approach mentioned in https://github.com/weaveworks/eksctl/issues/544? Its covering all of scenarios that I can think of right now.

martina-if:
@sayboras I replied in #544

Issue #1191, opened 2019-08-22
unknown field "spotAllocationStrategy"


yahavb:
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
aws released capacity-optimized pools. According to https://aws.amazon.com/blogs/compute/introducing-the-capacity-optimized-allocation-strategy-for-amazon-ec2-spot-instances/ `"SpotAllocationStrategy": "capacity-optimized"` is required to support it. 

**What feature/behavior/change do you want?**
when creating a mixed-instance node group, the pool/s to be launched will use "capacity-optimized" so the spot instances will use the least volatile pools. 

Do not hesitate, when appropriate, to share the exact commands or API you would like, and/or to share a diagram (e.g.: asciiflow.com): "a picture is worth a thousand words".
```
nodeGroups:
    - name: mixed-instances-1
      minSize: 2
      maxSize: 5
      ssh:
        publicKeyName: id_rsa.pub
      instancesDistribution:
        maxPrice: 0.5
        instanceTypes: ["c4.large","c4.xlarge","c4.2xlarge","c4.4xlarge","c4.8xlarge","c5.large","c5.xlarge","c5.2xlarge","c5.4xlarge"]
        onDemandPercentageAboveBaseCapacity: 50
        spotAllocationStrategy: capacity-optimized
        spotInstancePools: 2
        onDemandBaseCapacity: 0

ruecarlo:
Just one tiny correction in the  example provided  above in regards to `capacity-optimized`.

`capacity-optimized` and `spotInstancePools` are incompatible. `spotInstancePools` is only relevant to `lowest-price`

Doc link : https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_InstancesDistribution.html

yahavb:
good call @ruecarlo

martina-if:
fixed via https://github.com/weaveworks/eksctl/pull/1883

PR #1883, opened 2020-03-03
support spotAllocationStrategy on instances distribution config

bwagner5:
### Description
https://github.com/weaveworks/eksctl/issues/1191

Support the spotAllocationStrategy field in the instancesDistribution configuration block for node groups. This allows for spot configurations to choose to use lowest-price (defaulted in the API if not specified) or capacity-optimized. The capacity-optimized strategy will choose the deepest capacity pools from the instance-types specified.

### Checklist
- [x] Manually tested
- [] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)


pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -430,6 +430,20 @@ func validateInstancesDistribution(ng *NodeGroup) error {
 		return fmt.Errorf("spotInstancePools should be between 1 and 20")
 	}
 
+	if distribution.SpotInstancePools != nil && distribution.SpotAllocationStrategy != nil {
+		return fmt.Errorf("spotInstancePools cannot be specified when also specifying spotAllocationStrategy")
```
ruecarlo: Not entirely sure the check correlates with the message, or what's the intention of the error message: "spotInstancePools cannot be specified when also specifying spotAllocationStrategy"

`spotInstancePools` cannot be used when the `spotAllocationStrategy` is set to `capacity-optimized`. This has been captured right in the documentation (change in this PR over 08-spot-instances.md )

spotInstancePools is only intended to be used with lowes-price allocation strategy.  That, in my opinion should be a check to do at this stage, but: Is that what the conditional or the error message reflect ?

bwagner5: I'll update with a more specific check and modify the error message returned.

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -352,9 +352,12 @@ var _ = Describe("ClusterConfig validation", func() {
 
 	Describe("ssh flags", func() {
 		var (
-			testKeyPath = "some/path/to/file.pub"
-			testKeyName = "id_rsa.pub"
-			testKey     = "THIS IS A KEY"
+			testKeyPath                                 = "some/path/to/file.pub"
```
martina-if: Nit: we didn't have this function before but now we can [use this](https://github.com/weaveworks/eksctl/blob/master/pkg/utils/strings/strings.go#L4) instead of declaring a string variable here and then using it to make a pointer out of it:

```
			It("It fails when the spotAllocationStrategy is not a supported strategy", func() {
				ng.InstancesDistribution.SpotAllocationStrategy = strings.Pointer("unsupported-strategy")

				err := validateInstancesDistribution(ng)
				Expect(err).To(HaveOccurred())
			})
```

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -469,6 +472,29 @@ var _ = Describe("ClusterConfig validation", func() {
 				err = validateInstancesDistribution(ng)
 				Expect(err).ToNot(HaveOccurred())
 			})
+
+			It("It fails when the spotAllocationStrategy is not a supported strategy", func() {
+				ng.InstancesDistribution.SpotAllocationStrategy = &testUnsupportedSpotAllocationStrategy
```
martina-if: don't you need `ng.InstancesDistribution.SpotInstancePools = newInt(2)` here too?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -296,6 +302,14 @@ func SupportedNodeVolumeTypes() []string {
 	}
 }
 
+// SupportedSpotAllocationStrategies are the spot allocation strategies supported by ASG
+func SupportedSpotAllocationStrategies() []string {
```
martina-if: Since the only use we have for this is to check if the allocation strategy string is in the list of known values  I would instead implement a function that check's that:

```
func isSpotAllocationStrategySupported(allocationStragegy string) bool {
  switch (allocationStrategy) {
    case SpotAllocationStrategyLowestPrice, SpotAllocationStrategyCapacityOptimized:
      return true
    default:
      return false
  }
}
```

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -430,6 +430,20 @@ func validateInstancesDistribution(ng *NodeGroup) error {
 		return fmt.Errorf("spotInstancePools should be between 1 and 20")
 	}
 
+	if distribution.SpotInstancePools != nil && distribution.SpotAllocationStrategy != nil && *distribution.SpotAllocationStrategy == SpotAllocationStrategyCapacityOptimized {
+		return fmt.Errorf("spotInstancePools cannot be specified when also specifying spotAllocationStrategy: %s", SpotAllocationStrategyCapacityOptimized)
+	}
+
+	if distribution.SpotAllocationStrategy != nil {
+		spotAllocationStrategies := make(map[string]bool)
```
martina-if: Check my earlier comment. I think for two strings the benefit of constructing a map for the look up is not worth it compared with the extra complexity of the code.

pkg/apis/eksctl.io/v1alpha5/validation_test.go
bwagner5: nice, yeah "strings" makes it much nicer looking

pkg/apis/eksctl.io/v1alpha5/validation_test.go
bwagner5: it would actually check whether the allocation strategy is defined with or without spotInstancePools. I'd like to leave it out though in case the validation logic would ever get changed to check if spotInstancePools is specified and spotAllocationStrategy is not "lowest-price" (since that's the only allocation strategy where it is valid to specify spotInstancePools).

pkg/apis/eksctl.io/v1alpha5/types.go
bwagner5: I like having the isSupported helper, but I still like the supported strategies slice. It's currently used in the error message to print the support strategies. I used slice.Contains to implement the isSupported helper rather than the switch. When other spot allocation strategies are added to the ASG API, it will make it really easy to add the new constant to the slice and not having to touch anything else. Let me know if you agree or not.

pkg/apis/eksctl.io/v1alpha5/validation.go
bwagner5: 👍

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: Sounds great :+1:

pkg/apis/eksctl.io/v1alpha5/validation_test.go
martina-if: Good point :+1:

martina-if:
Hi @bwagner5 Thank you for the changes :) if you rebase this branch on master we can merge it

ruecarlo:
Awesome. Thanks for this one! @martina-if, @bwagner5 !

ranshn:
hi @martina-if , we noticed that this PR wasn't included in eksctl 0.15.0-rc.2 . can it be included in the next release please?

bwagner5:
looks like it was added but not included in release notes. Opened a PR to add to 0.15.0 release notes: https://github.com/weaveworks/eksctl/pull/1928

ruecarlo:
Hmmm, Not sure I might be missing something, but it looks to me that 0.15.0-rc2 comes from a [branch](https://github.com/weaveworks/eksctl/tree/release-0.15) where this was not merged. https://github.com/weaveworks/eksctl/blob/0.15.0-rc.2/pkg/apis/eksctl.io/v1alpha5/types.go does not have the new `SpotAllocationStrategy`

martina-if:
@ranshn @bwagner5 @ruecarlo This feature will be released in `0.16.0`.

lloydchang:
@bwagner5 @ruecarlo @martina-if @ranshn You may be interested in bug https://github.com/weaveworks/eksctl/issues/3516 that I just reported.  Thank you.

PR #1884, opened 2020-03-04
Parallelise integration tests

### LINKED ISSUE #1192: integration test structure
We currently have a fairly ad-hoc way of defining tests in `integration/`. As the number of tests grows, we should come up with a better structure for these that would allows us to add more tests easily while keeping the house tidy.

I am thinking we should consider doing the following:

- top-level package/file
  - has main function, parses flag and environment vars, initialises any shared structures
  - defines helpers that provides cluster name and region helpers and holds other related state (like kubeconfig path etc)
  - has ability of invoking sub-tests
- sub-test package(s)/file(s)
   - one for each of the areas of functionality
   - avoid global variables, name and other parameters should be provided from top-level
- test utilities
   - turn-up/tear-down
   - gomega matchers, command runner etc

Tasks:
- explore ginkgo best-practice for structuring tests
  - determine if we can use package or it'd have to be one large package with multiple files
- start implementing it and define more tasks as needed

marccarre:
### Description

- Parallelise integration tests (260 min. -> 100 min.).
- Try to improve package structure where possible/easy.
- Unique cluster names for all integration tests (and prefixed with `it-` to be more easily identifiable).

Fixes #1192.

### Checklist

- [x] Manually tested

### Manual testing

```console
$ time make integration-test
[...]
[2] Running: ginkgo -tags integration -v --progress integration/tests/cluster_api/... --
[4] Running: ginkgo -tags integration -v --progress integration/tests/fargate/... --
[1] Running: ginkgo -tags integration -v --progress integration/tests/before_active/... --
[5] Running: ginkgo -tags integration -v --progress integration/tests/managed/... --
[3] Running: ginkgo -tags integration -v --progress integration/tests/crud/... --
[0] Running: ginkgo -tags integration -v --progress integration/tests/backwards_compat/... --
[6] Running: ginkgo -tags integration -v --progress integration/tests/quickstart_profiles/... --
[7] Running: ginkgo -tags integration -v --progress integration/tests/quickstart_profiles_gen/... --
[...]
2020/03/04 18:33:12 ================== Summary ================================
[0] Ran 0 of 1 Specs in 0.002 seconds
[0] SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 1 Skipped
[0] PASS
[0]
[0] Ginkgo ran 1 suite in 12.931377473s
[0] Test Suite Passed
[1] Ran 4 of 4 Specs in 746.526 seconds
[1] SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
[1] PASS
[1]
[1] Ginkgo ran 1 suite in 12m41.132345413s
[1] Test Suite Passed
[7] Ran 1 of 1 Specs in 1193.230 seconds
[7] SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
[7] PASS
[7]
[7] Ginkgo ran 1 suite in 20m6.603725551s
[7] Test Suite Passed
[6] Ran 2 of 2 Specs in 1238.106 seconds
[6] SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
[6] PASS
[6]
[6] Ginkgo ran 1 suite in 20m51.902667694s
[6] Test Suite Passed
[5] Ran 11 of 11 Specs in 1846.401 seconds
[5] SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
[5] PASS
[5]
[5] Ginkgo ran 1 suite in 31m1.819156924s
[5] Test Suite Passed
[3] Ran 35 of 35 Specs in 2384.933 seconds
[3] SUCCESS! -- 35 Passed | 0 Failed | 0 Pending | 0 Skipped
[3] PASS
[3]
[3] Ginkgo ran 1 suite in 40m6.71300507s
[3] Test Suite Passed
[2] Ran 10 of 10 Specs in 4054.279 seconds
[2] SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
[2] PASS
[2]
[2] Ginkgo ran 1 suite in 1h7m48.511913539s
[2] Test Suite Passed
[4] Ran 3 of 3 Specs in 5927.064 seconds
[4] SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 0 Skipped
[4] PASS
[4]
[4] Ginkgo ran 1 suite in 1h39m2.083167267s
[4] Test Suite Passed
2020/03/04 18:33:12 ================== See ya Space Cowboy! ===================
real	99m29.418s
user	4m25.758s
sys	5m0.799s
```


integration/tests/params.go
```diff
@@ -0,0 +1,163 @@
...
+
+// GenerateKubeconfigPath generates a path in ${TEMP} based on these params' cluster name.
+func (p *Params) GenerateKubeconfigPath() {
+	p.KubeconfigPath = path.Join(os.TempDir(), fmt.Sprintf("%s.yaml", p.ClusterName))
+}
+
+// GenerateCommands generates eksctl commands with the various options & values
+// provided to this `Params` object.
+func (p *Params) GenerateCommands() {
+	p.EksctlCmd = runner.NewCmd(p.EksctlPath).
```
martina-if: Just out of curiosity, are the Command objects something that should be reused for some reason? Why create them once and reuse them instead of always returning a new one with something like:

```
func (p *Params) newCreateCommand() (p *runner.Command)  {
  return runner.NewCmd(p.EksctlPath).
      WithArgs("create").
      WithTimeout(25 * time.Minute)
}
```

integration/tests/params.go
```diff
@@ -0,0 +1,163 @@
...
+// FormatClusterName creates a new cluster name using this Params struct's
+// cluster name prefix, the provided name, and adds the cluster to the list of
+// clusters to eventually delete, once the test suite has run.
+func (p *Params) FormatClusterName(name string) string {
+	return p.formatClusterName(p.clusterNamePrefix, name)
+}
+
+func (p *Params) formatClusterName(prefix string, name string) string {
+	clusterName := fmt.Sprintf("it-%s-%s", prefix, name)
+	p.addToDeleteList(clusterName)
```
martina-if: Can we move the addition to the list to line 96?

integration/tests/params.go
```diff
@@ -0,0 +1,163 @@
...
+	return clusterName
+}
+
+// addToDeleteList adds the provided cluster name to the list of clusters to eventually delete.
+func (p *Params) addToDeleteList(clusterName string) {
+	p.clustersToDelete = append(p.clustersToDelete, clusterName)
```
martina-if: Even though there's not a lot of contention here, isn't this a potential race condition since `append()` is not thread safe?

integration/tests/params.go
```diff
@@ -0,0 +1,163 @@
...
+// Params groups all test parameters.
+type Params struct {
+	EksctlPath string
+	Region     string
+	Version    string
+	// Flags to help with the development of the integration tests
+	clusterNamePrefix string
+	ClusterName       string
+	DoCreate          bool
+	DoDelete          bool
```
martina-if: nit: how about renaming this to `deleteClusterAfterTest`? too long?

integration/tests/params.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (p Params) DeleteClusters() {
+	if !p.DoDelete {
+		return
+	}
+	for _, clusterName := range p.clustersToDelete {
+		cmd := p.EksctlDeleteClusterCmd.WithArgs(
+			"--name", clusterName,
+		)
+		session := cmd.Start()
```
martina-if: Since deletion takes long, should we call these in parallel or it becomes too complicated?

integration/tests/quickstart_profiles/quickstart_profiles_test.go
```diff
@@ -0,0 +1,110 @@
...
+			AssertFluxManifestsPresentInGit(branch, params.PrivateSSHKeyPath)
+			AssertFluxPodsPresentInKubernetes(params.KubeconfigPath)
+			// Clean-up:
+			err := git.DeleteBranch(branch, cloneDir, params.PrivateSSHKeyPath)
+			Expect(err).NotTo(HaveOccurred()) // Deleting the branch should have succeeded.
+		})
+	})
+})
+
+var _ = AfterSuite(func() {
```
martina-if: I am curious, does this get called when one kills the tests with SIGTERM?

integration/main.go
```diff
@@ -0,0 +1,160 @@
...
+		// processes, letting them run AfterSuite to destroy all clusters, and
+		// then letting them naturally terminate.
+		cancel()
+	}()
+	return ctx, interrupts
+}
+
+func listModules() []string {
+	testsDir := path.Join("integration", "tests")
+	files, err := ioutil.ReadDir(testsDir)
```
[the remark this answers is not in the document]
marccarre: From a "practical" perspective, that should be fine in the sense that I'd hope `ginkgo` would just ignore them as nothing to run. However, from a "design" perspective, I'd rather avoid that, and instead move such code to some utility package under `./integration/*`.

integration/tests/params.go
marccarre: I am actually not sure as this precedes me, and I didn't look too deep into what they do.

integration/tests/params.go
marccarre: We could, but we need to expose some extra method to add a cluster to the list of clusters to delete, and call that method from somewhere in the test. (I actually don't really like the way this all works, but haven't really invested time into making it better. For now, I wanted it to roughly work, and to focus on parallelising things.)

integration/tests/params.go
marccarre: It could be used for more than that. What about one of these?
```go
SetUp bool
TearDown bool
```
```go
DoSetUp bool
DoTearDown bool
```
```go
ShouldSetUp bool
ShouldTearDown bool
```

integration/tests/params.go
marccarre: Before being smarter about deletion, I'd rather parallelise more test cases.
For example, currently, the 1. `fargate`, 2. `cluster_api`, 3. `crud`, and 4. `managed` suites are the ones taking the longest to run, but 1./2./4. have tests which are mostly independent of each other, so we could start with parallelising this more. Then "untangle" 3. to make it more parallelisable as well. And then look into running deletion in parallel where possible. WDYT?

integration/tests/quickstart_profiles/quickstart_profiles_test.go
marccarre: Testing I did with dummy tests did show that, yes, as I propagate the signal to child `ginkgo` processes and then, these stop test suites and call `AfterSuite`:

> Since the AfterSuite typically includes code to clean up persistent state ginkgo will also run AfterSuite when you send the running test suite an interrupt signal (^C).

Source: https://onsi.github.io/ginkgo/#global-setup-and-teardown-beforesuite-and-aftersuite

integration/tests/params.go
martina-if: I'm not sure I understand why. What I mean is something like this:

```func (p *Params) NewClusterName(prefix string) string {
	formattedName := p.formatClusterName(prefix, names.ForCluster("", ""))
        p.addToDeleteList(clusterName)
        return formattedName
}

func (p *Params) FormatClusterName(name string) string {
	return p.formatClusterName(p.clusterNamePrefix, name)
}

func (p *Params) formatClusterName(prefix string, name string) string {
	clusterName := fmt.Sprintf("it-%s-%s", prefix, name)
	return clusterName
}
```
So that `FormatClusterName()` does only one thing. Right now it says format name but it has a side effect of changing the state of `p` without the caller knowing unless they read it's comment.

Then a caller can a) only format a name or b) format the name and add it to the list.

integration/tests/params.go
martina-if: Yeah, that also works. I think something saying "create", "destroy", "remove", "delete" is perhaps more clear but if you think there are more things to do in those steps then those names look good

integration/tests/params.go
martina-if: Sounds like a good plan :+1:

integration/tests/quickstart_profiles/quickstart_profiles_test.go
martina-if: NICE!

integration/tests/params.go
marccarre: It could also be "reversed", e.g.:
```go
SkipSetUp bool
SkipTearDown bool
```
or
```go
SkipCreate bool
SkipDelete bool
```
Any preference?

integration/tests/params.go
marccarre: There could be the day we start to parallelise tests within a test suite, something `ginkgo` doesn't do by default though, so no race at the moment.

integration/tests/params.go
martina-if: Yeah, I think I like the last ones more :)

integration/tests/params.go
marccarre: > I'm not sure I understand why.

I agree and also usually tend to have methods which do "one thing" and have no side effect, but in this specific case, I wanted to avoid "polluting" the integration tests, to avoid having extra "test infrastructure" related calls in them, as it makes it harder to understand the essence of what they do, of what they test, so was happy with that ~side-effect~ "magic".

integration/tests/params.go
marccarre: To be clearer, this is how things are parallelised at the moment:
```
  /- create cluster names array -> run tests -> delete all clusters -\
---- create cluster names array -> run tests -> delete all clusters ---->
  \- create cluster names array -> run tests -> delete all clusters -/
```

integration/tests/params.go
marccarre: Fixed. 👍

PR #1887, opened 2020-03-04
Add node-taints flag to add taints when registering the nodes in the nodegroup

niraizik:
### Description
Add --node-taints flag. To add taint fields when registering the nodes in the nodegroup

The field is already the in the Nodegroup struct but there is no command line flag
Only needed to add the flag and add to the Taints field

### Checklist
- [X] Manually tested


martina-if:
HI @niraizik  thank you for contributing to eksctl. We generally try not to add CLI flags for all the options. To simplify the experience we actually keep a reduce group of flags and we expect users to use config files when in need of more advanced features.

For that reason, and because taints specially have characters that don't play so nicely in a cli flag, we decided to keep them just for config files. 

I am closing this PR but if you think there's a use case we haven't considered please feel free to comment and reopen it.

PR #1889, opened 2020-03-04
Revert "Merge pull request #1746 add ability to override kubeconfig cluster endpoint"

treacher:
### Description

After testing 0.14.0 this morning to create a new node group I noticed that the nodes were not coming up. After further investigation into the issue it turned out that the `KUBECONFIG_CLUSTER_ENDPOINT` was being used in the servers kubeconfig file as well.

We're using something along the lines of this as we're using SSH tunnelling:
```
export KUBECONFIG_CLUSTER_ENDPOINT=https://$KUBERNETES_API:8001
```

My initial assumption was that the kubeconfig file generated locally for performing eksctl commands was created separately from the one generated on the Kubernetes host but it turns out both of these are created from the same process.

Because of this, the change I've added causes the nodes not to be able to talk to the API as they're trying to access it on port 8001 instead of 443.

I clearly didn't test this change thoroughly enough with creating node groups and for that I'm sorry.

Creating clone PR here @martina-if original pr: https://github.com/weaveworks/eksctl/pull/1878

### DIFF
--- a/pkg/utils/kubeconfig/kubeconfig.go
+++ b/pkg/utils/kubeconfig/kubeconfig.go
@@ -63,12 +63,6 @@ func New(spec *api.ClusterConfig, username, certificateAuthorityPath string) (*c
 		CurrentContext: contextName,
 	}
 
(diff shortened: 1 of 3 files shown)


martina-if:
Thanks!

PR #1916, opened 2020-03-10
Add support for kubernetes 1.15 in release 0.15.0-rc.2

martina-if:
### Description

eks just launched kubernetes vs 1.15: https://github.com/aws/containers-roadmap/issues/380#issuecomment-597351516

This PR adds support for creating clusters with 1.15 and upgrading to it from 1.14.

**note** this PR merges against the branch `release-0.15` so that we can release it in `0.15.0-rc.2`

### Checklist
- [ ] Manually tested


pkg/eks/eks.go
```diff
@@ -82,6 +83,14 @@ func (c *ClusterProvider) SupportsManagedNodes(clusterConfig *api.ClusterConfig)
 
 // ClusterSupportsManagedNodes reports whether the EKS cluster supports managed nodes
 func ClusterSupportsManagedNodes(cluster *awseks.Cluster) (bool, error) {
+	supportsManagedNodes, err := utils.IsMinVersion(api.Version1_15, *cluster.Version)
```
ajayk: should the minimum version for managed groups be `api.Version1_14` instead of 
`utils.IsMinVersion(api.Version1_15, *cluster.Version)`

cluster-1.15.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: martina-1-15-9
```
marccarre: Was this file checked in accidentally? If not, you may want to move it to `examples` and to change the cluster's name.

pkg/ctl/update/cluster.go
```diff
@@ -85,7 +85,7 @@ func doUpdateClusterCmd(cmd *cmdutils.Cmd) error {
 	case api.Version1_13:
 		cfg.Metadata.Version = api.Version1_14
 	case api.Version1_14:
-		cfg.Metadata.Version = api.Version1_14
+		cfg.Metadata.Version = api.Version1_15
```
marccarre: Given we had:
```go
	case api.Version1_14:		case api.Version1_14:
		cfg.Metadata.Version = api.Version1_14
```
before, do we want an additional case saying the following?
```go
	case api.Version1_15:		case api.Version1_15:
		cfg.Metadata.Version = api.Version1_15
```

site/content/introduction/01-getting-started.md
```diff
@@ -25,7 +25,7 @@ eksctl create cluster --name=cluster-1 --nodes=4
 
 ```
 
-EKS supports versions `1.12`, `1.13` and `1.14` (default).
+EKS supports versions `1.12`, `1.13` and `1.14` (default) and `1.15`.
```
marccarre: Two "_and_"s may read a bit odd?
```suggestion
EKS supports versions `1.12`, `1.13`, `1.14` (default) and `1.15`.
```

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -283,6 +286,7 @@ func SupportedVersions() []string {
 		Version1_12,
```
sayboras: do you want to deprecate 1_12 in same PR as well ?

pkg/eks/eks_test.go
```diff
@@ -408,6 +408,16 @@ var _ = Describe("EKS API wrapper", func() {
 			expectError:       false,
 			supports:          false,
 		}),
+		Entry("with 1.15", &managedNodesSupportCase{
+			kubernetesVersion: "1.15",
+			expectError:       false,
+			supports:          true,
+		}),
+		Entry("with 1.16", &managedNodesSupportCase{
```
sayboras: so 1.16 is coming soon as well ? :100:

pkg/eks/eks.go
cPu1: The minimum required version is 1.14 but 1.14 requires a platform version check also. Both of this happen below in line 94. This check is for returning early to avoid doing a platform version check.

pkg/eks/eks_test.go
cPu1: Not that I know of, this is just to ensure the validation function works for all newer versions.

PR #1923, opened 2020-03-11
Fix Fargate validation for EKS version 1.15 and above

cPu1:
### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/fargate/versions.go
+++ b/pkg/fargate/versions.go
@@ -2,7 +2,6 @@ package fargate
 
 import (
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
-	"github.com/weaveworks/eksctl/pkg/utils"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/fargate/versions.go
```diff
@@ -15,11 +15,20 @@ const (
 	MinKubernetesVersion = api.Version1_14
 )
 
-// IsSupportedBy reports whether the control plane version can support Fargate.
-func IsSupportedBy(controlPlaneVersion string) (bool, error) {
-	supportsFargate, err := utils.IsMinVersion(MinKubernetesVersion, controlPlaneVersion)
+// IsSupportedBy reports whether the control plane version can support Fargate and if checking the EKS platform version
+// is required
+func IsSupportedBy(controlPlaneVersion string) (bool, bool, error) {
```
martina-if: Instead of returning two booleans I think it's better if this function receives the controlPlaneVersion and the plarformVersion and it makes a decision based on both. I think it would be simpler and it would be more readable/mantainable. WDYT?

cPu1: Yes, it was a bit awkward to return two booleans, but I was trying to keep the existing pattern.

In that case, we don't really need this function and the consts exported in this file. I'll move this check to the caller `ClusterSupportsFargate` and remove `IsSupportedBy`.

Issue #1822, opened 2020-02-14
eksctl edit command


joshkurz:
**Why do you want this feature?**
I'd like to be able to run `eksctl edit cluster foobar` to be able to edit a `eksctl` resource. This would do a simple `eksctl get` on the current resource as yaml output, and put into the users editor of choice. Then once they save, the correct `eksctl` update command would be ran.

**What feature/behavior/change do you want?**
To add an `edit` command to `eksctl`. This would follow suit with how `kubectl` works.

sayboras:
I like this idea, hope that it is aligned with eksctl team.

There is a catch on this, currenlty update command didn't actually update all resouces though :cry:

joshkurz:
The edit command could just support what is supported with update. I wouldn't want to add any extra logic to 'edit' that is not just a pass through to the update command. So if that means 'edit' only supports 'Clusters' for now, then that would have to do. 

Then whatever functionality is added to 'update' would automatically be available in 'edit'. Also we would have a single source of truth for any bugs or issues that updating a resource my have.

Are their plans to add 'update' functionality to other supported resources?

PR #1935, opened 2020-03-13
[WIP] Add eksctl edit command

sarataha:
### Description
Hello, this is a basic functionality for issue #1822 which suggests having `eksctl edit cluster foobar` command.

Workflow:
* Call `eksctl get` and save the output in `resourceData` variable
* Add the resulting data to a temp file
* Open file in text editor fetched from `$EDITOR` environment variable
* Call `eksctl update` using the temp file with the new changes

Future Work:
* Based on the operating system, set default text editor if `$EDITOR` doesn't have a value 

### Checklist
- [ ] Manually tested


martina-if:
Hi @sarataha , thank you for this PR. This is a good idea, although I the update command is a little misleading and this feature doesn't fit in eksctl at the moment. 

The `eksctl update cluster` command, doesn't work like `kubectl update`. This command is used to upgrade the control plane to the next kubernetes version, and update the stacks if needed, but it doesn't modify the properties of the cluster.

I think if we build such a command in the future this kind of functionality will make a lot of sense but right now the update command doesn't read much of the config file: just the name and the region. So the edit functionality doesn't really fit right now unfortunately :/

martina-if:
Hi @sarataha , if that's ok I will close this PR, but if you think it should be reopened please feel free to do so. Again, thanks for contributing! Even if we close some issues or PRs we still want to encourage everyone to contribute; we really appreciate it :slightly_smiling_face:

sarataha:
Hi @martina-if , thank you very much for letting me know, and even though the edit functionality is not applicable right now; it was a great opportunity for me knowing more about the project structure. I will definitely keep working more on other issues 🙂

PR #1986, opened 2020-03-27
Set kubeReserved defaults according to instance type

### LINKED ISSUE #1981: Set defaults for kubeReserved according to instance type
The `kubeReserved` setting for the kubelet is currently not depending on the instance type. AWS, GKE, and AKS, are all setting defaults wrt the node beefiness:

**AWS**
Performs the calculations in their [AMI bootstrap.sh](https://github.com/awslabs/amazon-eks-ami/blob/51105608a55f096724fafcca0012dfa4fe930ca7/files/bootstrap.sh#L164-L181) See PR https://github.com/awslabs/amazon-eks-ami/pull/367 for more insights.

(This script is replaced by our own [bootstrap.al2.sh](https://github.com/weaveworks/eksctl/blob/master/pkg/nodebootstrap/assets/bootstrap.al2.sh) which follows these guiding principles: https://github.com/weaveworks/eksctl/blob/master/docs/current-001-node_bootstrap.md#guiding-principles)

**GKE**
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu

**AKS**
https://docs.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations

To have sane defaults for users of eksctl.

- precompile a list of instance types with their vcpu count, available memory, and storage.
- use this list to calculate kubeReserved flags according to a formula depending on the instance type
- put it into the /etc/eksctl/kubelet.yaml to have defaults that can be overwritten with eksctl's `kubeletExtraConfig` settings

(ref #1823)

rndstr:
Replaces #1823
Closes #1981 

### Description

This change implements default values for `kube-reserved` the same way
as the [official AMIs] do which is dependent on the instance type.

Reserved memory and cpu are set according to a formula while the storage
is always set to 1GiB. See the source code for details.

Instance type information (such as available storage, memory, and cpu)
may not only change in the future but also new instances will be added.
There is a make target `generate-kube-reserved` which fetches all instance
types by querying the AWS API `ec2:DescribeInstanceTypes` in each
region. It then saves it to `pkg/nodebootstrap/reserved_data.go` ready
to be taken as input for computing appropriate default values.

[official AMIs]: https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/nodebootstrap/reserved_generate.go
+++ b/pkg/nodebootstrap/reserved_generate.go
@@ -1,0 +1,101 @@
+//+build ignore
+
+// Fetches AWS instance type data such as memory, cpus, and
+// (ephemeral) storage. Writes it to a Go file to be included
+// by eksctl.
+//
+// Since each region may support different instance types we
+// need to query all regions. But the same instance type in
+// different regions has consistent specs.
+package main
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	. "github.com/dave/jennifer/jen"
+
+	"github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/nodebootstrap"
+)
+
+const outputFilename = "reserved_data.go"
+
+func main() {
+	infos := map[string]nodebootstrap.InstanceTypeInfo{}
+	for _, region := range v1alpha5.SupportedRegions() {
+		fmt.Printf("Fetching instance types in %q\n", region)
+		client := ec2.New(newSession(region))
+		if err := updateRegionInstanceTypes(client, infos); err != nil {
+			checkError(region, err)
+		}
+	}
+	if err := render(infos); err != nil {
+		panic(err)
+	}
+	fmt.Println("Done.")
+}
+
+func checkError(region string, err error) {
+	if err == nil {
+		return
+	}
+	// Allow failing for Chinese regions
+	if aerr, ok := err.(awserr.Error); ok {
+		if aerr.Code() == "AuthFailure" && strings.HasPrefix(region, "cn-") {
+			fmt.Println("  AuthFailure, skipping.")
+			return
+		}
+	}
+	fmt.Println("Make sure your environment provides credentials to run ec2:DescribeInstanceTypes in supported regions")
+	panic(err)
+}
+
+func updateRegionInstanceTypes(client *ec2.EC2, infos map[string]nodebootstrap.InstanceTypeInfo) error {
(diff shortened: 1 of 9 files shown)


pkg/nodebootstrap/reserved_generate.go
```diff
@@ -0,0 +1,101 @@
...
+	. "github.com/dave/jennifer/jen"
+
+	"github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/nodebootstrap"
+)
+
+const outputFilename = "reserved_data.go"
+
+func main() {
+	fmt.Println("!!! Your environment needs to provide credentials to run ec2:DescribeInstanceTypes in supported regions")
```
martina-if: Perhaps we should print this message only if the authentication fails?

pkg/nodebootstrap/userdata.go
```diff
@@ -100,7 +100,7 @@ func makeKubeletConfigYAML(spec *api.ClusterConfig, ng *api.NodeGroup) ([]byte,
 
 	// use a map here, as using struct will require us to add defaulting etc,
 	// and we only need to add a few top-level fields
-	obj := map[string]interface{}{}
+	obj := api.InlineDocument{}
```
martina-if: Should the comment be updated?

martina-if: out of curiosity why was this changed?

rndstr: `api.inlineDocument` is a typedef of `map[string]interface{}` and is used throughout the source. (e.g., i saw it in `userdata_bottlerocket.go`).

i changed this line too because in the code i added below i use it instead of the generic `map[string]interface{}`, to keep it consistent in this file

site/content/usage/11-customizing-the-kubelet.md
```diff
@@ -53,7 +53,8 @@ nodeGroups:
 In this example, given instances of type `m5a.xlarge` which have 4 vCPUs and 16GiB of memory, the `Allocatable` amount
 of CPUs would be 3.4 and 15.4 GiB of memory. In addition, the `DynamicKubeletConfig` feature gate is also enabled. It is
 important to know that the values specified in the config file for the the fields in `kubeletExtraconfig` will 
-completely overwrite the default values specified by eksctl.
+completely overwrite the default values specified by eksctl. However, omitting one or more `kubeReserved` parameters
+will cause the missing parameters to be defaulted to sane values based on the aws instance type being used.
```
InAnimaTe, line 57: @rndstr It would be amazing if there was a chart of these for the most popular instances right in the docs instead of having to look at the code. At the very least, possibly link to the part in the Code where these are defined? Also, you should certainly mention instance types which aren't supported by this.

martina-if: Hi @InAnimaTe , thanks, I think this is a good idea. I created the following issue https://github.com/weaveworks/eksctl/issues/2091 to track this

PR #1997, opened 2020-03-31
Make kubernetes 1.15 the default version

martina-if:
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/types.go
+++ b/pkg/apis/eksctl.io/v1alpha5/types.go
@@ -105,10 +105,10 @@ const (
 	Version1_15 = "1.15"
 
 	// DefaultVersion represents default Kubernetes version supported by EKS
-	DefaultVersion = Version1_14
+	DefaultVersion = Version1_15
 
 	// LatestVersion represents latest Kubernetes version supported by EKS
-	LatestVersion = Version1_14
+	LatestVersion = Version1_15
 
 	// DefaultNodeType is the default instance type to use for nodes
 	DefaultNodeType = "m5.large"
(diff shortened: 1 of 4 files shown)


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -105,7 +105,7 @@ const (
 	Version1_15 = "1.15"
 
 	// DefaultVersion represents default Kubernetes version supported by EKS
-	DefaultVersion = Version1_14
+	DefaultVersion = Version1_15
 
 	// LatestVersion represents latest Kubernetes version supported by EKS
 	LatestVersion = Version1_14
```
sayboras: do we need to update this value as well ?

martina-if: oh, this should have changed when we released 1.15. Thanks for the catch! Fixed

site/content/introduction/01-getting-started.md
```diff
@@ -25,7 +25,7 @@ eksctl create cluster --name=cluster-1 --nodes=4
 
 ```
 
-EKS supports versions `1.12`, `1.13`, `1.14` (default) and `1.15`.
+EKS supports versions `1.12`, `1.13`, `1.14` and `1.15` (default).
 With `eksctl` you can deploy either version by passing `--version`.
```
cPu1: I think we should also change this text since we have more than 2 versions to select from.

Something along these lines:
```suggestion
With `eksctl` you can deploy any of the supported versions by passing `--version`.
```

alanoberto:
Is this implemented?

martina-if:
Hi @alanoberto this will be released in version 0.17.0

PR #1999, opened 2020-04-01
Fix IDE build issue: add same build annotations to test file as prod file

marccarre:
### Description

Prior to this change, some IDEs were failing to build the project because methods from:

```
pkg/version/generate/release_generate.go
```

annotated with `// +build release` could NOT be found by its tests:

```
pkg/version/generate/release_generate_test.go
```

Indeed, both files were in the same `main` package so that tests should be able to call functions from `release_generate.go` directly but the fact the test file didn't have `// +build release` made the compiler fail to find these, leading to errors like:

```
undeclared name: prepareRelease
```

### Checklist

- [x] Manually tested: fixes build errors under VisualCode

### DIFF
--- a/pkg/version/generate/generate_suite_test.go
+++ b/pkg/version/generate/generate_suite_test.go
@@ -1,3 +1,5 @@
+// +build release
+
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


marccarre:
Closing this for now. May be fixing a non-issue/something local to my machine.

PR #2002, opened 2020-04-03
New default IP mapping settings for public subnets

martina-if:
### Description

This PR makes eksctl leverage the property `MapPublicIpOnLaunch` for subnets (Also known as "Assign IP v4 address") to assign ip addresses for nodes on public subnets.

At the same time we disable the property `AssociatePublicIpAddress` on the Network interface settings of the ASG.

To keep eksctl compatible with existing cluster the following functionality has changed:

- `eksctl create cluster` will enable `MapPublicIpOnLaunch` in public subnets and will unset `AssociatePublicIpAddress` on the ASG
- `eksctl create nodegroup` will check if the new nodes will be able to talk to the cluster endpoint by checking:
    - if the nodes are in a public subnet, and the subnets don't have `MapPublicIpOnLaunch` enabled, and the cluster doesn't have the access to the private endpoint enabled either -> it will throw an error because the nodes won't be able to join the cluster
- `eksctl update cluster` will not upgrade the subnets in the stack because this operation has to be done consciously by the user
- a new command `utils update-legacy-subnet-settings` has been added to update the subnet's settings with `MapPublicIpAddress` enabled either via CF if the VPC was created by eksctl or via the EC2 API if the VPC was imported.

 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/nodegroup.go
+++ b/pkg/cfn/builder/nodegroup.go
@@ -242,7 +242,8 @@ func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_La
 		ImageId:  gfn.NewString(n.spec.AMI),
 		UserData: n.userData,
 		NetworkInterfaces: []gfn.AWSEC2LaunchTemplate_NetworkInterface{{
-			AssociatePublicIpAddress: gfn.NewBoolean(!n.spec.PrivateNetworking),
+			// Explicitly un-setting this so that it doesn't get defaulted to true
+			AssociatePublicIpAddress: nil,
 			DeviceIndex:              gfn.NewInteger(0),
 			Groups:                   n.securityGroups,
 		}},
--- a/pkg/cfn/manager/compat.go
+++ b/pkg/cfn/manager/compat.go
@@ -1,11 +1,17 @@
 package manager
 
 import (
+	"fmt"
+
+	gfn "github.com/awslabs/goformation/cloudformation"
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
+
 	"github.com/weaveworks/eksctl/pkg/cfn/builder"
 	"github.com/weaveworks/eksctl/pkg/cfn/outputs"
+	"github.com/weaveworks/eksctl/pkg/vpc"
 )
 
 // FixClusterCompatibility adds any resources missing in the CloudFormation stack in order to support new features
@@ -60,6 +66,7 @@ func (c *StackCollection) FixClusterCompatibility() error {
 	if !stackSupportsFargate {
 		logger.Info("cluster stack is missing resources for Fargate")
 	}
+
 	logger.Info("adding missing resources to cluster stack")
 	_, err = c.AppendNewClusterStackResource(false, true)
 	return err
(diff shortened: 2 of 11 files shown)


pkg/cfn/builder/nodegroup.go
```diff
@@ -242,7 +242,7 @@ func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_La
 		ImageId:  gfn.NewString(n.spec.AMI),
 		UserData: n.userData,
 		NetworkInterfaces: []gfn.AWSEC2LaunchTemplate_NetworkInterface{{
-			AssociatePublicIpAddress: gfn.NewBoolean(!n.spec.PrivateNetworking),
+			AssociatePublicIpAddress: nil,
```
sayboras, line 245: this one is for new CF only, so can we just remove this line, unless it's used for updating CF as well ?

martina-if: This is so new stacks don't get this set. I think it has been set by default so now we have to explicitly unset it.

At least, the current functionality enables this and I couldn't find where that was done so I assumed that it is done by default.

cPu1: I think we should add a short comment stating that this is now set at the subnet level.

pkg/cfn/manager/compat.go
```diff
@@ -73,3 +80,65 @@ func (c *StackCollection) hasManagedToUnmanagedSG() (bool, error) {
 	stackResources := gjson.Get(stackTemplate, resourcesRootPath)
 	return builder.HasManagedNodesSG(&stackResources), nil
 }
+
+// EnsureMapPublicIPOnLaunchEnabled sets this subnet property to true when it is not set or is set to false
+// returns the modified template and the list of modified subnets
```
cPu1, line 85: This comment seems outdated as it only returns the error.

pkg/vpc/vpc.go
```diff
@@ -219,9 +222,98 @@ func ImportSubnetsFromList(provider api.ClusterProvider, spec *api.ClusterConfig
...
+				return err
+			}
+			subnetsToValidate.Insert(subnetIDs...)
+		} else {
+			// This ng doesn't have AZs defined so we need to check all public subnets
+			for _, subnet := range spec.VPC.Subnets.Public {
+				subnetsToValidate.Insert(subnet.ID)
+			}
+			break
+		}
```
cPu1, line 271: This block is same as the one for eksctl-managed nodegroups. Might want to extract it into a function defined within `ValidateLegacySubnetsForNodeGroups`  so it has access to `subnetsToValidate`.

pkg/vpc/vpc.go
```diff
@@ -219,9 +222,98 @@ func ImportSubnetsFromList(provider api.ClusterProvider, spec *api.ClusterConfig
...
+	}
+
+	if err := ValidateExistingPublicSubnets(provider, subnetsToValidate.List()); err != nil {
+		logger.Critical(err.Error())
+		return errors.Errorf("subnets for one of more new nodegroups don't meet requirements. "+
```
cPu1, line 276: Minor typo:
```suggestion
		return errors.Errorf("subnets for one or more new nodegroups don't meet requirements. "+
```

pkg/vpc/vpc.go
```diff
@@ -276,3 +368,38 @@ func cleanupSubnets(spec *api.ClusterConfig) {
 		}
 	}
 }
+
+func validatePublicSubnet(subnets []*ec2.Subnet) error {
+	for _, sn := range subnets {
+		if sn.MapPublicIpOnLaunch == nil || !*sn.MapPublicIpOnLaunch {
+			return fmt.Errorf("found mis-configured subnet %q. Expected public subnet with property "+
+				"\"MapPublicIpOnLaunch\" enabled. Without it new nodes won't get an IP assigned", *sn.SubnetId)
```
cPu1, line 376: Should we collect all public subnets that have `MapPublicIpOnLaunch` disabled and show them in the error message?

pkg/vpc/vpc.go
```diff
@@ -276,3 +368,38 @@ func cleanupSubnets(spec *api.ClusterConfig) {
...
+		}
+	}
+	return nil
+}
+
+func SelectPublicNodeGroupSubnets(nodegroupAZs []string, clusterSpec *api.ClusterConfig) ([]string, error) {
+	numNodeGroupsAZs := len(nodegroupAZs)
+	if numNodeGroupsAZs == 0 {
+		return make([]string, 0), nil
```
cPu1, line 385: Minor nit: can also be just
```suggestion
		return nil, nil
```

and the calling code will still work.

martina-if: thanks!

pkg/vpc/vpc.go
martina-if: good idea

site/content/usage/06-vpc-networking.md
```diff
@@ -15,6 +15,9 @@ not insecure in principle, but some compromised workload could risk an access vi
 
 If that functionality doesn't suit you, the following options are currently available.
 
+>**_IMPORTANT_**: From `eksctl` version `0.17.0` and onwards public subnets will have the property `MapPublicIpOnlaunch` enabled, and the property `AssociatePublicIpAddress` disabled in the auto-scaling groups."
```
cPu1, line 18:
```suggestion
>**_IMPORTANT_**: From `eksctl` version `0.17.0` and onwards public subnets will have the property `MapPublicIpOnLaunch` enabled, and the property `AssociatePublicIpAddress` disabled in the Auto Scaling Group for the nodegroups."
```

visunali:
few comments for description:
* regarding `eksctl create nodegroup`
  * even though cluster has enabled private endpoint access, the nodes launched in public subnets with no public IP assigned will still need internet access to pull CNI/kubeproxy to join the cluster.
  * so if the nodes are in a public subnet, and the subnets don't have MapPublicIpOnLaunch enabled, it will throw error.
* regarding other operations on nodegroup, we also explicitly throw the same error during `update nodegroup` and `upgrade nodegroup` if public subnet has such invalid configuration. Since these operations might involve scaling up nodegroup, those newly-launched nodes will have no internet connectivity if such configuration is present.

PR #2033, opened 2020-04-14
Publish RCs to the homebrew tap

### LINKED ISSUE #2010: Publish RCs in Homebrew tap
Check https://docs.brew.sh/Manpage#install-options-formula

martina-if:
With this change when we publich a release candidate it will also be published in Weaveworks homebrew tap.

closes #2010 

- [x] Manually tested

### DIFF
--- a/.goreleaser.brew.yml
+++ b/.goreleaser.brew.yml
@@ -18,3 +18,6 @@ brew:
   dependencies:
     - kubernetes-cli
     - aws-iam-authenticator
+
+  custom_block: |
+    head "https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_Darwin_amd64.tar.gz"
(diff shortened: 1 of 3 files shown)


martina-if:
cc @kalbir tap urls  https://github.com/weaveworks/homebrew-tap/tree/master/Formula

kalbir:
I was thinking something like this:

kalbir:
I was thinking something like this:

kalbir:
I was thinking something like this:

https://docs.brew.sh/Formula-Cookbook#advanced-formula-tricks

I think if we can do the thing in that link above then I should be able to run `brew upgrade eksctl --HEAD` and get the latest RC. For the way it is in this PR I think I need to have a 2nd version of eksctl, called `eksctl-private` which is the RC. 

Does that make sense?

kalbir:
I think that `goreleaser` has a way of doing this `head` specification here: https://goreleaser.com/homebrew/ 

It looks like you can use the `custom_block` to specify a location for `--HEAD`.

.goreleaser.brew.yml
```diff
@@ -18,3 +18,6 @@ brew:
   dependencies:
     - kubernetes-cli
     - aws-iam-authenticator
+
+  custom_block: |
+    head https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_Darwin_amd64.tar.gz
```
sayboras, line 23: you might need to add double quote here as well :sob:

martina-if: Thanks Tam. I thought I did but I forgot to push the changes.

PR #2034, opened 2020-04-14
Document new priority labels

martina-if:
Document the priority categories we will be using when triaging bug issues.


martina-if:
@kalbir yeah, exactly, that's what I wrote initially but I am not sure we are ready to commit to it just yet. I'll leave it as it is now, and we can tweak it after a few weeks if we think we are ready :) Thanks!

kalbir:
Ha perfect! Great minds :)

Issue #2037, opened 2020-04-15
Build failure in master


sayboras:
**What happened?**
`make test` failed in master branch

**What you expected to happen?**
`make test` should be running successfully

**How to reproduce it?**
Run `make test`

**Logs**
diff --git a/pkg/nodebootstrap/maxpods.go b/pkg/nodebootstrap/maxpods.go
index 6c8abf22..0e4497d9 100644
--- a/pkg/nodebootstrap/maxpods.go
+++ b/pkg/nodebootstrap/maxpods.go
@@ -90,6 +90,10 @@ var maxPodsPerNodeType = map[string]int{
        "i3en.6xlarge":  234,
        "i3en.large":    29,
        "i3en.xlarge":   58,
+       "inf1.24xlarge": 437,
+       "inf1.2xlarge":  38,
+       "inf1.6xlarge":  234,
+       "inf1.xlarge":   38,
        "m1.large":      29,
        "m1.medium":     12,
        "m1.small":      8,
HINT: to fix this, run 'git commit pkg/nodebootstrap/assets.go pkg/nodebootstrap/maxpods.go pkg/addons/default/assets.go pkg/addons/default/assets/aws-node.yaml pkg/addons/assets.go pkg/ami/static_resolver_ami.go site/content/usage/20-schema.md pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go pkg/eks/mocks/EKSAPI.go pkg/eks/mocks/ELBAPI.go pkg/eks/mocks/SSMAPI.go pkg/eks/mocks/IAMAPI.go pkg/eks/mocks/EC2API.go pkg/eks/mocks/CloudFormationAPI.go pkg/eks/mocks/CloudTrailAPI.go pkg/eks/mocks/STSAPI.go pkg/eks/mocks/ELBV2API.go --message "Update generated files"'
make[1]: *** [Makefile:161: check-all-generated-files-up-to-date] Error 1

PR #2039, opened 2020-04-15
Separate downloaded files from generated ones in Makefile

martina-if:
This only forces generated files to be up to date before building. Files
that are downloaded should be updated consciously and not in every PR.
This makes sure that the build pipeline doesn't block PRs when new
instances are published (for maxpods), for example.

Triggered by https://github.com/weaveworks/eksctl/issues/2037
Possible fix for https://github.com/weaveworks/eksctl/issues/1608
Supersedes https://github.com/weaveworks/eksctl/pull/1819

- [x] Manually tested
- [ ] close #1819 if this is approved

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,6 @@ gocache := $(shell go env GOCACHE)
 GOBIN ?= $(gopath)/bin
 
 
-always_generate_in_packages := ./pkg/nodebootstrap ./pkg/addons/default ./pkg/addons
-
 generated_code_deep_copy_helper := pkg/apis/eksctl.io/v1alpha5/zz_generated.deepcopy.go
 
 generated_code_aws_sdk_mocks := $(wildcard pkg/eks/mocks/*API.go)
@@ -18,13 +16,15 @@ conditionally_generated_files := \
   userdocs/src/usage/schema.md \
   $(generated_code_deep_copy_helper) $(generated_code_aws_sdk_mocks)
 
+all_downloaded_files := \
+  pkg/ami/static_resolver_ami.go \
+  pkg/nodebootstrap/maxpods.go \
+  pkg/addons/default/assets/aws-node.yaml \
+
 all_generated_files := \
   pkg/nodebootstrap/assets.go \
-  pkg/nodebootstrap/maxpods.go \
   pkg/addons/default/assets.go \
-  pkg/addons/default/assets/aws-node.yaml \
   pkg/addons/assets.go \
-  pkg/ami/static_resolver_ami.go \
   $(conditionally_generated_files)
 
 .DEFAULT_GOAL := help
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -146,12 +146,13 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
 ##@ Code Generation
 
 .PHONY: generate-always
-generate-always: ## Generate code (required for every build)
```
sayboras, line 149: just want to confirm if you intentionally want to remove this make target details in `make help`.

Makefile
```diff
@@ -18,13 +16,15 @@ conditionally_generated_files := \
   site/content/usage/20-schema.md \
   $(generated_code_deep_copy_helper) $(generated_code_aws_sdk_mocks)
 
+all_downloaded_files := \
```
sayboras, line 19: this variable is used anywhere or just for tracking purpose ?

Makefile
```diff
@@ -146,12 +146,13 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
 ##@ Code Generation
 
 .PHONY: generate-always
-generate-always: ## Generate code (required for every build)
+generate-always:  pkg/ami/static_resolver_ami.go pkg/nodebootstrap/maxpods.go pkg/addons/default/assets/aws-node.yaml
```
sayboras, line 149: these make targets are not  `.PHONY`, and all these files are available in repo, hence I am struggling to find out in which cases these target will be triggered. Can you help to clarify a little bit ?

sayboras: (testing the comment from golang)

sayboras: We could do that as well. However, do you feel that we need to run some targets (e.g. _update-ami_ or _update-maxpods_, etc) in CI ? Could be in PR or master ? or before release ?

martina-if: No, I don't want them to run in CI at this point. That's why I'm moving them out of the main building path. I don't want to block PRs on updating external files. That should go into its own PR and should be a conscious decision.

It would be ok, though, to have CI check if these need updating periodically and open a PR, I just don't want development slowed down when an unrelated external dependency needs updating.

sayboras: Oh i got it. LGTM :100:

Makefile
```diff
@@ -165,9 +164,23 @@ check-all-generated-files-up-to-date: generate-all
 	@# generate-groups.sh can't find the lincense header when using Go modules, so we provide one
 	printf "/*\n%s\n*/\n" "$$(cat LICENSE)" > $@
 
-.PHONY: generate-ami
-generate-ami: ## Generate the list of AMIs for use with static resolver. Queries AWS.
-	time go generate ./pkg/ami
+### Update AMIs in ami static resolver
+.PHONY: update-ami
+update-ami: ## Generate the list of AMIs for use with static resolver. Queries AWS.
+	go generate ./pkg/ami
```
cPu1, line 170: Any reason why we're no longer timing this?

martina-if: Do you look at it? I find all those annoying and I'm not really looking at them. I can add it if we are going to use it

martina-if: I'm actually thinking about removing all of the `time` commands....

cPu1: >Do you look at it? 

Not really. I asked because the other ones are being timed.

PR #2071, opened 2020-04-22
Allow private clusters to be created

### LINKED ISSUE #2074: Private-only endpoint fails for pure Fargate setup with transit gateways
**What happened?**
We have a configuration where we use transit gateways to communicate from otherwise private subnets. This allows the EKS control plane to communicate with the Fargate scheduler, and so we can spin up EKS clusters that have private-only endpoints visible only on our intranet.

For our network topology, and a Fargate-only configuration, the following ClusterConfig options should pass validation:

```
vpc:
  clusterEndpoints:
    privateAccess: true
    publicAccess: false
```

**Versions**
- eksctl version: 0.17.0
- kubectl version: v1.18.2

**Logs**
We get a validation failure, which is triggered from validation.go, line 119:

```
if PrivateOnly(endpts) {
  return ErrClusterEndpointPrivateOnly
}
```

When we patch the code to ignore that validation check, the cluster is created successfully and actually works perfectly. Our patch keeps the validation check, but only fails if there are node groups present. Thus it works for our (pure Fargate) setup. Updated validation.go, line 119:

```
if PrivateOnly(endpts) && len(c.NodeGroups) > 0 {
  return ErrClusterEndpointPrivateOnly
}
```

saada:
### Description

Allow private clusters to be created now that the [feature](https://github.com/aws/containers-roadmap/issues/221#issuecomment-565596179) is available.

closes #2074

### Checklist
- [x] Manually tested

### DIFF
--- a/examples/02-custom-vpc-cidr-no-nodes.yaml
+++ b/examples/02-custom-vpc-cidr-no-nodes.yaml
@@ -1,7 +1,7 @@
 # An example of ClusterConfig object with custom VPC IPv4 CIDR,
 # and auto-allocated IPv6 CIDRs for all subnets; also without
 # any nodegroups:
---- 
+---
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
 
@@ -12,5 +12,9 @@ metadata:
 vpc:
   cidr: 10.10.0.0/16
   autoAllocateIPv6: true
+  # disable public access to endpoint and only allow private access
+  clusterEndpoints:
+    publicAccess: false
+    privateAccess: true
 
 nodeGroups: []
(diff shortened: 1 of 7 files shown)


saada:
The tests failed for some odd reason even though they work locally.
I'll rerun the circleci job.

sayboras:
@saada just curious if you have tried to provision the cluster manually, if yes, you can tick the checkbox as well :100:.

examples/02-custom-vpc-cidr-no-nodes.yaml
```diff
@@ -12,5 +12,9 @@ metadata:
 vpc:
   cidr: 10.10.0.0/16
   autoAllocateIPv6: true
+  # by default clusters are publicly exposed and disabled for private access
```
sayboras, line 15: the comment here and values are not in sync I think

saada: right... I struggled writing this comment ...
I'm trying to explain the default and here we show how we can override the default.
Let me improve that

AMALCGCG:
I just verified that this fixes https://github.com/weaveworks/eksctl/issues/2074

PR #2072, opened 2020-04-22
Enable slack notifications for broken master builds

martina-if:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,10 +69,14 @@ commands:
             - "/home/circleci/go/pkg/mod/"
   notify-slack:
     description: Notify failures via Slack
+    parameters:
+      fail_only:
+        type: boolean
+        default: true
     steps:
       - run: exit 0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -69,10 +69,14 @@ commands:
...
     description: Notify failures via Slack
+    parameters:
+      failures_only:
+        type: boolean
+        default: true
     steps:
       - run: exit 0
       - slack/status:
-          fail_only: false
+          fail_only: << parameters.failures_only >>
```
rndstr, line 79: out of curiosity, was there a reason for having the parameter naming different than the flag name we are setting?

martina-if: Not really, I though it could potentially be less problematic but it probably doesn't matter. I can make them the same if you think it's clearer.

martina-if: Actually since I have to rebase anyway I'll rename it :+1:

Issue #2026, opened 2020-04-11
Expand linting rules to all go files


sayboras:
Related to comment https://github.com/weaveworks/eksctl/pull/2019#issuecomment-611266352, we can expand linting rules for all `_test.go` files. Moreover, we can enforce the same for different build tags such as `release` and `integration`

This will surely help with standardizing the code base.

sayboras:
@alrs as you already do some works with this issue, please let me know if you have time to take it forward.  Otherwise, I can pick it up as well :100:

alrs:
I plan to keep going once https://github.com/weaveworks/eksctl/pull/2019 lands.

sayboras:
Thanks. In my opinion, the below configuration seems reasonable, feel free to make a change along the way.

```
diff --git a/.golangci.yml b/.golangci.yml
index ccd3d8ec..4b01b975 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,11 +10,12 @@ run:
   issues-exit-code: 1
 
   # include test files or not, default is true
-  tests: false
+  tests: true
 
   # list of build tags, all linters use it. Default is empty list.
-  #build-tags:
-  #  - mytag
+  build-tags:
+    - release
+    - integration
 
   # which dirs to skip: they won't be analyzed;
   # can use regexp here: generated.*, regexp is applied on full path;
@@ -25,9 +26,6 @@ run:
     - ^vendor$
     - ^build$
     - ^pkg\/eks\/mocks$
-    - \/usr\/local\/go
-    - ^pkg\/testutils
-    - ^pkg\/drain
 
   # which files to skip: they will be analyzed, but issues from them
   # won't be reported. Default value is empty list, but there is
@@ -37,11 +35,6 @@ run:
   #  - ".*\\.my\\.go$"
   #  - lib/bad.go
     - ^pkg\/nodebootstrap\/assets.go
-    - ^pkg\/testutils\/(ginkgo|client).go
-    - ^pkg\/testutils\/mockprovider\/mock_provider.go
-    - ^pkg\/cfn\/template\/matchers\/matchers.go
-    - ^integration\/matchers\/matchers.go
-    - ^integration\/runner\/runner.go
 
 
 # output configuration options
@@ -148,6 +141,15 @@ linters-settings:
     range-loops: true # Report preallocation suggestions on range loops, true by default
     for-loops: false # Report preallocation suggestions on for loops, false by default
 
+issues:
[...]

PR #2104, opened 2020-04-28
Enable linters for all go files in the codebase

sayboras:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/2026

### Note
_No_ change in actual behavior or logic in eksctl.

### Checklist
- [ ] Manually tested


.golangci.yml
```diff
@@ -148,6 +141,15 @@ linters-settings:
     range-loops: true # Report preallocation suggestions on range loops, true by default
     for-loops: false # Report preallocation suggestions on for loops, false by default
 
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    - linters: [errcheck]
```
sayboras, line 147: I feel like skipping the below rules are making sense. Please let me know if you have different idea.

michaelbeaumont: i'm curious, why ignore the first two?

sayboras: Oh I just removed the first rule, second one is to avoid the below
```
integration/main.go:26:42         staticcheck  SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?)
```

However, as syscall.SIGTERM is already trapped, I can remove the second rule as well. Please let me know if you have any other concern.

Issue #2105, opened 2020-04-28
Flaky unit test - ClusterConfig validation FargateProfile


sayboras:
I noticed the below test is quite flaky, will be better if this test is deterministic. 

```shell scripts
Running Suite: /src/pkg/apis/eksctl.io/v1alpha5/api_suite_test.go
 
=================================================================
 
Random Seed: 1588071843
 
Will run 69 of 69 specs
 
••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••
------------------------------
 
• Failure [0.007 seconds]
 
ClusterConfig validation 
/src/pkg/apis/eksctl.io/v1alpha5/validation_test.go:10 
  FargateProfile 
  /src/pkg/apis/eksctl.io/v1alpha5/validation_test.go:587 
    Validate 
    /src/pkg/apis/eksctl.io/v1alpha5/validation_test.go:588 
      passes on randomly generated fields [It] 
      /src/pkg/apis/eksctl.io/v1alpha5/validation_test.go:654 
      Unexpected error: 
          <*errors.errorString | 0xc000276aa0>: { 
              s: "invalid Fargate profile \"UavZYBNlGPcoQwkbkggaBnhpj\": no profile selector", 
          } 
          invalid Fargate profile "UavZYBNlGPcoQwkbkggaBnhpj": no profile selector 
      occurred
      /src/pkg/apis/eksctl.io/v1alpha5/validation_test.go:659
 
------------------------------
 
••
 
JUnit report was created: /src/test-results/__src__pkg__apis__eksctl__io__v1alpha5__api__suite__test__go_1.xml
 
```

Few occurences:

https://app.circleci.com/pipelines/github/weaveworks/eksctl/1702/workflows/d2986de7-4091-4c3e-944e-f8ff319ebf4b/jobs/7729
https://app.circleci.com/pipelines/github/weaveworks/eksctl/1633/workflows/7aae9f05-1fa1-4486-81b7-2419c4b74456/jobs/7658

PR #2112, opened 2020-04-29
Fix flaky unit test due to random data

sayboras:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/2105

### Approach

Underlying issue is due to https://github.com/bxcodec/faker/blob/1c6d6edb0123edda0348adc3ccb7e3480b9aaf07/faker.go#L488, which might cause the slice length as zero. I can enforce the length of slice with tag e.g. `faker:"len=1"`, but it's not looking good in struct. 

The default value https://github.com/bxcodec/faker/blob/1c6d6edb0123edda0348adc3ccb7e3480b9aaf07/faker.go#L29 is 100, so this test case will probably fail one in hundred times. We can also increase default values to reduce the odds, but it still fails some time :)

In short, I just add the check in case generated slice is empty.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation_test.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation_test.go
@@ -1,7 +1,6 @@
 package v1alpha5
 
 import (
-	"github.com/bxcodec/faker"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


martina-if:
Hi @sayboras thanks for fixing this. I've seen this error before too. Let's get rid of this test and add a deterministic test instead with 2 or 3 profile selectors. What do you think?

sayboras:
> Hi @sayboras thanks for fixing this. I've seen this error before too. Let's get rid of this test and add a deterministic test instead with 2 or 3 profile selectors. What do you think?

@martina-if it's the best option I can say. Let me remove such randomness in unit test. Maybe dependency _fake_ might not be required anymore.

Issue #2125, opened 2020-05-03
Order of operations when creating a cluster, finish control plane first?


whereisaaron:
**What happened?**
Creating a new a cluster from scratch with a config file (existing VPC) it created the control plane but then failed to create a managed nodegroup for a (semi) legitimate reason, AWS apparently won't let you scale node groups to zero. But that wasn't the issue.

I deleted, modified. and created the node group with the config file and that was fine. However I noticed several other things hadn't happen which weren't at all dependent on the node group, e.g.
- Control plane logging was not enabled
- IAM Service Acccounts were not created, and 
- Fargate profiles in the config file were not created

Those things are dependent only on the control plane working. It seems like at least the logging and IAM Service Accounts could have been done first before attempting node groups?

```
[ℹ]  building managed nodegroup stack "eksctl-fooo-nodegroup-group-1"
[ℹ]  deploying stack "eksctl-foo-nodegroup-group-1"
[✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-foo-nodegroup-group-1"
[ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
[✖]  AWS::EKS::Nodegroup/ManagedNodeGroup: CREATE_FAILED – "Property validation failure: [Value for property {/ScalingConfig/MinSize} is less than minimum allowed value {1.000000}]"
[!]  1 error(s) occurred and cluster hasn't been created properly, you may wish to check CloudFormation console
```

**What you expected to happen?**
All of the control plane settings to be applied before attempting node group creation.

**How to reproduce it?**
Create a cluster with control plane logging and IAM Service Accounts and a node group config that will fail

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
0.19.0-rc.0

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-eks-e16311", GitCommit:"e163110a04dcb2f39c3325af96d019b4925419eb", GitTreeState:"clean", BuildDate:"2020-03-27T22:40:13Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #2129, opened 2020-05-04
Do some nodegroup-independent tasks in parallel with nodegroup creation

michaelbeaumont:
### Description

Extra cluster tasks don't appear to need a node group at all to finish, only the cluster, so they can be done in parallel with nodegroup creation.

Fixes part of #2125 

### Checklist
- [x] Manually tested


pkg/cfn/manager/create_tasks.go
```diff
@@ -22,6 +22,16 @@ func (c *StackCollection) NewTasksToCreateClusterWithNodeGroups(nodeGroups []*ap
 			supportsManagedNodes: supportsManagedNodes,
 		},
 	)
+	var postClusterTasks *TaskTree
+	if len(postClusterCreationTasks) > 0 {
+		postClusterTasks = &TaskTree{Parallel: true}
```
martina-if, line 27: I'm thinking that since the tasks come from the caller, perhaps it should be the caller's responsibility to know whether they  can run in parallel or not. WDYT?

pkg/cfn/manager/create_tasks.go
```diff
@@ -32,7 +42,7 @@ func (c *StackCollection) NewTasksToCreateClusterWithNodeGroups(nodeGroups []*ap
 
 	if nodeGroupTasks.Len() > 0 {
 		nodeGroupTasks.IsSubTask = true
-		tasks.Append(nodeGroupTasks)
+		postClusterTasks.Append(nodeGroupTasks)
```
martina-if, line 45: Don't these need to be appended to `tasks` instead of `postClusterTasks` so they don't get lost given that they get added to `tasks` on line 31?

pkg/ctl/create/cluster.go
```diff
@@ -319,8 +319,8 @@ func doCreateCluster(cmd *cmdutils.Cmd, ng *api.NodeGroup, params *cmdutils.Crea
 		if err != nil {
 			return err
 		}
-		tasks := stackManager.NewTasksToCreateClusterWithNodeGroups(cfg.NodeGroups, cfg.ManagedNodeGroups, supportsManagedNodes)
-		ctl.AppendExtraClusterConfigTasks(cfg, params.InstallWindowsVPCController, tasks)
```
martina-if, line 323: Was it intentional to lose this line?

pkg/cfn/manager/create_tasks.go
michaelbeaumont: i think it's ok since `postClusterTasks` is a pointer to my `Tree`, I can `Append` to it even after I've `Append`ed it to `tasks` on line 31. and in the other branch  line 32-34, `postClusterTasks == tasks`.
but it's actually hard to read as it is, I'll improve it

pkg/ctl/create/cluster.go
michaelbeaumont: yeah because `AppendExtraCluster...` became `CreateExtraCluster...` and I call that before `NewTasksToCreate...`

pkg/cfn/manager/create_tasks.go
michaelbeaumont: hmm actually perhaps it's a reasonable assumption that the `postClusterCreationTasks` can run in parallel with the node tasks?

michaelbeaumont: hopefully it's a bit easier to see what's going on now

pkg/cfn/manager/create_tasks.go
michaelbeaumont: hopefully it's a bit easier to see what's going on now

martina-if: Right. For some reason I assumed that the tasks.Append() would append the tasks from one tree into the other but it appends the tree directly obviously. Thanks.

martina-if: Both ways work for me, leave it as you prefer :+1:

PR #2138, opened 2020-05-05
Move Gitops guides below the Usage section in the site

martina-if:
I think the usage section is more important and now that the usage submenu is collapsed it should be bumped up compared to the gitops guide.

@kalbir what do you think?

### Checklist


kalbir:
Ha yes, I was thinking the same thing just the other day! Agree with this :)

PR #2156, opened 2020-05-08
Tag new clusters with Metadata.Tags

michaelbeaumont:
### Checklist
- [ ] Manually tested


pkg/eks/update.go
```diff
@@ -162,6 +162,31 @@ func (c *ClusterProvider) UpdateClusterVersion(cfg *api.ClusterConfig) (*awseks.
...
+	tags := make(map[string]*string)
+	for key, value := range clusterConfig.Metadata.Tags {
+		tags[key] = &value
+	}
+	if err := c.RefreshClusterStatus(clusterConfig); err != nil {
+		return err
+	}
+	input := &awseks.TagResourceInput{
+		ResourceArn: c.Status.clusterInfo.cluster.Arn,
+		Tags:        tags,
```
sayboras, line 176: hope you dont mind if I chime in.

eksctl is having some string util method such as
https://github.com/weaveworks/eksctl/blob/master/pkg/utils/strings/strings.go#L28 
https://github.com/weaveworks/eksctl/blob/master/pkg/utils/strings/strings.go#L39
https://github.com/weaveworks/eksctl/blob/master/pkg/utils/strings/strings.go#L48

pkg/eks/update.go
```diff
@@ -162,6 +162,31 @@ func (c *ClusterProvider) UpdateClusterVersion(cfg *api.ClusterConfig) (*awseks.
...
+	}
+	_, err := c.Provider.EKS().TagResource(input)
```
sayboras, line 178: just curious if this will override any existing tags creatd by Cloudformation if there is any.

michaelbeaumont: it doesn't

sayboras: right, just check that EKS CF now didn't suppor tags, and no tag propagration happened i.e. EKS tags are empty after provision

pkg/eks/update.go
```diff
@@ -162,6 +163,27 @@ func (c *ClusterProvider) UpdateClusterVersion(cfg *api.ClusterConfig) (*awseks.
...
 }
 
+// UpdateClusterTags calls eks.TagResource and tags the cluster
+func (c *ClusterProvider) UpdateClusterTags(clusterConfig *api.ClusterConfig) error {
+	if err := c.RefreshClusterStatus(clusterConfig); err != nil {
+		return err
+	}
+	input := &awseks.TagResourceInput{
+		ResourceArn: c.Status.clusterInfo.cluster.Arn,
+		Tags:        utilsstrings.ToPointersMap(clusterConfig.Metadata.Tags),
```
sayboras, line 173: Here is the definition for this field in eks/api.go, it requires at at least 1 element 

```
     Tags map[string]*string `locationName:"tags" min:"1" type:"map" required:"true"`
```

Understand that you already have a length check for config.Metadata.Tags before calling UpdateClusterTags method, however this method is public, there might be the casse going the line, it will be called directly with empty tags (which eventually causes issue). 

What do think about utilsstring.NilPointersMapIfEmpty(utilsstrings.ToPointersMap(clusterConfig.Metadata.Tags)) ?

michaelbeaumont: True, but doesn't `required` mean `nil` would be a problem too?

sayboras: oh you are right. Grrr I should change to bigger Macbook (maybe 15 or 16 inch one), missed that part first time :)

michaelbeaumont: I will include the length check inside the update function

sayboras:
LGTM :heavy_check_mark:

PR #2062, opened 2020-04-21
Add link checking

dholbach:
closes: #958

PR #2157, opened 2020-05-09
Update CONTRIBUTING.md with user docs step

sayboras:
Add a small section how to build user docs locally. Then I feel like
re-organize make target a little bit

### Checklist
- [ ] Manually tested

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -40,8 +40,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
-        run: |
-          pip install -r userdocs/requirements.txt
+        run: make install-site-deps
       - name: Build docs for link check
         run: make build-pages
         # Using liche action to check generated HTML site
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -36,7 +36,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,7 +100,7 @@ make build
 To build the user docs locally, please refer to the below steps:
 
 ```bash
-# Requires python3 and pip installed in your local system
+# Requires python3 and pip3 installed in your local system
 make install-site-deps
 make build-pages
 make serve-pages
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,9 @@ This project is written in Go. To be able to contribute you will need:
 2. Make sure that `$(go env GOPATH)/bin` is in your shell's `PATH`. You can do so by
    running `export PATH="$(go env GOPATH)/bin:$PATH"`
 
+3. (Optional) [User documentation](https://eksctl.io) is built and generated with [mkdocs](https://www.mkdocs.org/).
+   Please make sure you have python3 and pip installed on your local system.
+
 #### 2. Fork and clone the repo
 
 Make a fork of this repository and clone it by running:
@@ -94,6 +97,15 @@ make test
 make build
 ```
 
+To build the user docs locally, please refer to the below steps:
+
+```bash
+# Requires python3 and pip installed in your local system
+make install-site-deps
+make build-pages
+make serve-pages
+```
+
 To run integration test you will need an AWS account.
 ```bash
 make integration-test-container TEST_V=1
@@ -109,7 +121,7 @@ make integration-test-container TEST_V=1
 > ```
 
 If you prefer to use Docker, the same way it is used in CI, you can use the
-following comands:
+following commands:
 
 ```
 make -f Makefile.docker test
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -26,11 +26,16 @@ all_generated_files := \
 .DEFAULT_GOAL := help
 
 ##@ Dependencies
+.PHONY: install-build-deps
```
michaelbeaumont, line 29: I'm not a `make` expert by any means so I could be wrong but should this be `install-all-deps`?

dholbach: `install-site-deps` looking further down?

Makefile
```diff
@@ -26,11 +26,16 @@ all_generated_files := \
...
+install-all-deps: install-build-deps install-site-deps ## Install all dependencies for building both binary and user docs)
 
 .PHONY: install-build-deps
 install-build-deps: ## Install dependencies (packages and tools)
 	./install-build-deps.sh
-	pip3 install -r userdocs/requirements.txt
+
+.PHONY: install-site-deps
+install-site-deps: ## Install dependencies for user docs
+	pip install -r userdocs/requirements.txt
```
dholbach, line 38: Hum. At least in the case of Ubuntu, the tool is called `pip3` - not sure about other distributions or MacOS.

michaelbeaumont: good point, `pip3` is safer.

.github/workflows/check-links.yaml
```diff
@@ -40,8 +40,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
-        run: |
-          pip install -r userdocs/requirements.txt
+        run: make install-site-deps
```
dholbach, line 43: Looking at https://github.com/weaveworks/eksctl/pull/2157/checks?check_run_id=658968162 it seems this pulls in a bunch of Go dependencies too?

Makefile
sayboras: It was my mistake, corrected :)

.github/workflows/check-links.yaml
sayboras: wow good spot. I have taken a closer look, it is due to two attributes _eksctl_version_ and _godeps_cmd_ in Makefile.common and Makefile.

Please refer the commit (used for testing only) https://github.com/weaveworks/eksctl/pull/2157/commits/c82d5794225edd7689a168d856e45e1af283a689

Here is the respective github action run
https://github.com/weaveworks/eksctl/pull/2157/checks?check_run_id=665342849

Makefile
sayboras: Good point, I have changed to pip3 to be safe.

Makefile
```diff
@@ -249,7 +254,6 @@ eksctl-image: ## Build the eksctl image that has release artefacts and no build
 	$(MAKE) -f Makefile.docker $@
 
 ##@ Site
-
 .PHONY: serve-pages
 serve-pages: ## Serve the site locally
```
dholbach, line 258:
```suggestion
serve-pages: install-site-deps ## Serve the site locally
```

dholbach: If you add this here (and for the `build-pages` target too), the user won't have to run `make install-site-deps` manually. Just an idea.

sayboras: Thanks for your idea, however, I feel like the number of times you are running _build_ or _serve_ are significantly higher than _install-site-deps_. Additionally, updating depencies are kind of concious action. So even this step is fast, I still would love to keep it sepaprately.

This is kind of consistent with our binary build process as well.

dholbach: Ok, wfm.

.github/workflows/check-links.yaml
dholbach: Mh. Looks like the Go deps are still being pulled. So yeah. Not sure if we really need this here? No big deal in any case.

BTW we had the same conversation here already: https://github.com/weaveworks/eksctl/pull/2062#discussion_r412518621

sayboras: Oh sorry, I reverted the testing commit, and github didn't keep the run if there was forced push. Should have capture the screen shot of github action before :)

michaelbeaumont: i like the idea of keeping these commands all under `make`.
at the moment, it's not _really_ a problem to pull in the go deps, am I understanding that right?
if it becomes a problem, I would fix it in the `Makefile`.

sayboras: You are right. If required, we can change the Makefile to avoid running any go command running outside of make target :)

dholbach: yeah, no problem

CONTRIBUTING.md
```diff
@@ -94,6 +97,15 @@ make test
 make build
 ```
 
+To build the user docs locally, please refer to the below steps:
+
+```bash
+# Requires python3 and pip3 installed in your local system
+make install-site-deps
```
martina-if, line 104: @sayboras is this step needed when deploying the site? If so I think we need to make sure this is called from `make build-pages` so it's run on netlify too before deploying the site

dholbach: No, it's in the netlify settings. We define the docs directory there and the build step. It looks at runtime.txt and requirements.txt in there (https://docs.netlify.com/configure-builds/manage-dependencies/#python).

sayboras: @martina-if oh I didn't have access to netlify to confirm, but I think @dholbach comment helps to answer your question.

Related changs mainly for running and bringing up the pages locally to make sure look and feel as expected.

PR #2162, opened 2020-05-11
Install neuron plugin for inf1 instances

michaelbeaumont:
### Checklist
- [x] Add integration test
- [x] Manually tested


pkg/addons/assets/neuron-device-plugin.yaml
```diff
@@ -0,0 +1,45 @@
...
+        - key: aws.amazon.com/infa
+          operator: Exists
+          effect: NoSchedule
+      # Mark this pod as a critical add-on; when enabled, the critical add-on
+      # scheduler reserves resources for critical add-on pods so that they can
+      # be rescheduled after a failure.
+      # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+      priorityClassName: "system-node-critical"
+      containers:
+        - image: 790709498068.dkr.ecr.us-east-1.amazonaws.com/neuron-device-plugin:latest
```
martina-if, line 32: I think we need to add some templating here like we do for the other addons since this id won't work for all the regions/partitions.

pkg/ctl/cmdutils/create_cluster.go
```diff
@@ -13,6 +13,7 @@ type CreateClusterCmdParams struct {
 	SetContext                  bool
 	AvailabilityZones           []string
 	InstallWindowsVPCController bool
+	InstallNeuronDevicePlugin   bool
```
martina-if, line 16: We should ask Kalbir if this should be a CLI flag or a field in the config file?

pkg/eks/tasks.go
```diff
@@ -104,6 +125,27 @@ func (c *ClusterProvider) NewTasksRequiringControlPlane(cfg *api.ClusterConfig)
...
+func (c *ClusterProvider) ClusterTasksForNodeGroups(cfg *api.ClusterConfig, installNeuronDevicePluginParam bool) *manager.TaskTree {
+	tasks := &manager.TaskTree{
+		Parallel:  false,
+		IsSubTask: true,
+	}
+	var reallyInstallNeuronDevicePlugin bool
+	for _, ng := range cfg.NodeGroups {
+		reallyInstallNeuronDevicePlugin = reallyInstallNeuronDevicePlugin || HasInstanceType(ng, utils.IsNeuronInstanceType)
+	}
+	reallyInstallNeuronDevicePlugin = reallyInstallNeuronDevicePlugin && !installNeuronDevicePluginParam
```
martina-if, line 138: > `&& !installNeuronDevicePluginParam` 

was this intended or should it be `reallyInstallNeuronDevicePlugin && installNeuronDevicePluginParam`

michaelbeaumont: yes it should be! I had it `skipInstall` before. thanks

pkg/addons/assets/neuron-device-plugin.yaml
michaelbeaumont: do you know where the account id comes from in this case/how it compares to the account ids we use for the vpc-controller images? 
It seems like the instructions for vpc-controller depend on the region, but for the neuron plugin there's just this account id + us-east:
https://github.com/aws/aws-neuron-sdk/blob/master/docs/neuron-container-tools/tutorial-k8s.md

pkg/eks/nodegroup.go
```diff
@@ -174,6 +174,21 @@ func hasAmazonLinux2Node(nodeGroups []KubeNodeGroup) bool {
 	return false
 }
 
+// HasInstanceType returns whether some node in the group fulfils the type check
+func HasInstanceType(nodeGroup *api.NodeGroup, hasType func(string) bool) bool {
```
martina-if, line 178: I think this function would be more reusable if it was in the api package because it is imported from anywhere (and it only needs information that belongs there). If we leave it in the eks package we won't be able to import it while validating the config file, for example.

(the eks package is mainly for making requests to AWS)

pkg/ctl/create/utils.go
```diff
@@ -53,26 +53,17 @@ func checkVersion(cmd *cmdutils.Cmd, ctl *eks.ClusterProvider, meta *api.Cluster
...
-	if hasType(nodeGroup.InstanceType) {
-		return true
-	}
-	if nodeGroup.InstancesDistribution != nil {
-		for _, instanceType := range nodeGroup.InstancesDistribution.InstanceTypes {
-			if hasType(instanceType) {
-				return true
-			}
+func showDevicePluginMessageForNodeGroup(nodeGroup *api.NodeGroup, installNeuronPlugin bool) {
+	if eks.HasInstanceType(nodeGroup, utils.IsNeuronInstanceType) {
```
martina-if, line 57: nit: the instance type is inferentia so I would rename `IsNeuronInstanceType` to `IsInfInstanceType`,`IsInferentiaType`or similar

michaelbeaumont:
@martina-if @cPu1 integration tests pass.

deliahu:
@michaelbeaumont is there a way to prevent the device plugin from being installed via the [eksctl config file](https://eksctl.io/usage/schema/), or is `eksctl create cluster --install-neuron-plugin=false` the only way? It's not a big deal, I just have a slight preference for the config file since all my other configuration is there, so I wanted to check if it was possible.

michaelbeaumont:
@deliahu Hi, no there isn't at the moment.

deliahu:
@michaelbeaumont ok sounds good, thanks for the quick response!

PR #2195, opened 2020-05-15
Add bash and zsh completion to homebrew formula

eddiezane:
### Description

This PR adds a [custom install block](https://goreleaser.com/customization/homebrew/) to the goreleaser brew config to install bash and zsh completion.

I tested these changes manually in a direct edit of the brew formula. I think goreleaser updated their config format here to be arrays (brews: - / repos: -) instead of the current singular (brew: / repo:) so this may need a larger patch to make the config compliant.

### Checklist
- [x] Manually tested

/area build-and-release


.goreleaser.brew.yml
```diff
@@ -20,4 +20,15 @@ brew:
...
+
+  install: |
+      bin.install "eksctl"
+
+      # Install bash completion
+      output = Utils.popen_read("#{bin}/eksctl completion bash")
+      (bash_completion/"eksctl").write output
+
+      # Install zsh completion
+      output = Utils.popen_read("#{bin}/eksctl completion zsh")
```
sayboras, line 33: just curious how it's working in general and in the case that user didn't have zsh ?

eddiezane: Homebrew convention is to always install bash/zsh completion when available regardless of whether the user has the shell installed or not. Default location for zsh is `/usr/local/share/zsh/site-functions`.

This is copied from the formula for kubectl https://github.com/Homebrew/homebrew-core/blob/a8e1b997192a49fc974eeeb914028abea1c6bb80/Formula/kubernetes-cli.rb#L32-L38.

sayboras: thanks for your clarification, it's good to know :100:

.goreleaser.brew.yml
```diff
@@ -20,4 +20,15 @@ brew:
     - aws-iam-authenticator
 
   custom_block: |
-    head "https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_Darwin_amd64.tar.gz"
+    head "https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_Darwin_amd65.tar.gz"
```
michaelbeaumont, line 23: `65`

eddiezane: 👍 rebased

michaelbeaumont: It's still changed to `amd65`, that can't be right. accidental vim ctrl-a?

eddiezane: Ahhh sorry. Totally misunderstood your original comment. 

Good catch - must have been ctrl-a (which is also my tmux bind).

Fixed.

PR #2207, opened 2020-05-20
Add utils command to output JSON schema

### LINKED ISSUE #2059: Please publish or provide extraction of the JSON Schema for the config file
# Why do you want this feature?

As far as I can tell, the only way to validate eksctl configs in VSCode is by providing a JSON Schema file to the [redhat.vscode-yaml](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) extension.

The YAML published to https://eksctl.io/usage/schema/ is a [subset of the JSON schema](https://github.com/weaveworks/eksctl/blob/58fda687361e90555af96e1589cabdf679f9154f/cmd/schema/generate.go#L31) and I can't work out how to recreate the missing elements such that it'll validate my eksctl config file.

I suspect I was close, but then #2020 might be blocking me.

```
{
    "title": "eksctl",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "$ref": "#/definitions/ClusterConfig",
    "definitions": 
```
prepended to the JSON conversion of the [YAML schema reference](https://eksctl.io/usage/schema/) validated some things, and failed against things like the elements of `TypeMeta`.

Note: This would make #2052 more important, if people are actually validating their files against the schema.

# What feature/behavior/change do you want?

Some options come to mind:
## Publish a JSON Schema file at a known location

e.g., https://eksctl.io/usage/schema/v1alpha5.json. This has the issue that older eksctl users can't get the schema for their version from this URL. Since eksctl doesn't yet have schema migration, there's nothing in the repo to generate older versions. Perhaps !1020 will give us that, for some scope of "older", i.e. if the current release supports migration from some version, we have the spec for it and can generate the schema too.

## Publish a JSON Schema file to http://json.schemastore.org/

That would make the schema immediately available to editors, but I'm not sure if publishing alpha config file schemas there is appropriate. I'm also not sure how a schema without a fileMatch is matched to a YAML file, it might need a `$schema` key.

## Include in eksctl a command to actually dump its own jsonschma.

That means a little more work locally, but would ensure that you get exactly the schema for the eksctl release you're working with.

And the code is already there in the `generate` cli.

michaelbeaumont:
Closes #2059

### Description

I added it as `eksctl utils schema` but I'm definitely not attached to that path.

### Checklist
- [ ] Manually tested


martina-if:
LGTM

PR #2208, opened 2020-05-20
Increase cluster update timeout to 45 minutes

michaelbeaumont:
### Checklist
- [ ] Manually tested


cPu1:
In my tests for `update cluster`, the update operation completed within 30 minutes. But looks like that duration can change depending on the load.

michaelbeaumont:
> In my tests for `update cluster`, the update operation completed within 30 minutes. But looks like that duration can change depending on the load.

Yeah and I just want to increase it to some number so that it always finishes for our integration tests, for now. 45 minutes might be too high in the end.

PR #2179, opened 2020-05-13
Add unit test data for 1.14, 1.15 and 1.16

sayboras:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/2178

**Note**: these config files are extracted from existing cluster for respective version. Some minor changes in version (e.g. coredns, cni, etc) as I might upgrade them along the way. Currently, the default versions for all versions are same as per https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html

### Testing

<details>
<summary>Create cluster</summary>

```shell scripts
$ ./eksctl create cluster --managed --node-private-networking --version 1.16
[ℹ]  eksctl version 0.20.0-dev+880af000.2020-05-14T22:40:08Z
[ℹ]  using region us-east-2
[ℹ]  setting availability zones to [us-east-2a us-east-2c us-east-2b]
[ℹ]  subnets for us-east-2a - public:192.168.0.0/19 private:192.168.96.0/19
[ℹ]  subnets for us-east-2c - public:192.168.32.0/19 private:192.168.128.0/19
[ℹ]  subnets for us-east-2b - public:192.168.64.0/19 private:192.168.160.0/19
[ℹ]  using Kubernetes version 1.16
[ℹ]  creating EKS cluster "floral-wardrobe-1589460163" in "us-east-2" region with managed nodes
[ℹ]  will create 2 separate CloudFormation stacks for cluster itself and the initial managed nodegroup
[ℹ]  if you encounter any issues, check CloudFormation console or try 'eksctl utils describe-stacks --region=us-east-2 --cluster=floral-wardrobe-1589460163'
[ℹ]  CloudWatch logging will not be enabled for cluster "floral-wardrobe-1589460163" in "us-east-2"
[ℹ]  you can enable it with 'eksctl utils update-cluster-logging --region=us-east-2 --cluster=floral-wardrobe-1589460163'
[ℹ]  Kubernetes API endpoint access will use default of {publicAccess=true, privateAccess=false} for cluster "floral-wardrobe-1589460163" in "us-east-2"
[ℹ]  2 sequential tasks: { create cluster control plane "floral-wardrobe-1589460163", 2 parallel sub-tasks: { no tasks, create managed nodegroup "ng-c8e176f4" } }
[ℹ]  building cluster stack "eksctl-floral-wardrobe-1589460163-cluster"
[ℹ]  deploying stack "eksctl-floral-wardrobe-1589460163-cluster"
[...]

PR #2222, opened 2020-05-21
Fix kubernetes client serializer

michaelbeaumont:
Broken in #2179
See also https://github.com/kubernetes/kubernetes/pull/76098

This should probably have shown up in tests, which I will write next.
### Checklist
- [ ] Added tests that cover your change (if possible) #2225 
- [x] Manually tested

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -109,7 +109,7 @@ func (c *RawClient) new() (*RawClient, error) {
 		c.config.APIPath = "/api"
 	}
 	if c.config.NegotiatedSerializer == nil {
-		c.config.NegotiatedSerializer = scheme.Codecs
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/client.go
```diff
@@ -109,7 +109,7 @@ func (c *RawClient) new() (*RawClient, error) {
 		c.config.APIPath = "/api"
 	}
 	if c.config.NegotiatedSerializer == nil {
-		c.config.NegotiatedSerializer = scheme.Codecs
+		c.config.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
```
martina-if, line 112: Good catch...

sayboras: I introduced another bug :cry:

martina-if: No Worries :)

PR #2227, opened 2020-05-22
Disallow setting node-role.kubernetes.io label 

### LINKED ISSUE #2197: node-role.kubernetes.io/ in labels leads to errors for k8s>=1.16
**What happened?**
Under eksctl nodeGroups config a label with `node-role.kubernetes.io` as described here https://github.com/weaveworks/eksctl/pull/582 leads to new nodegroup creation hanging on `[ℹ]  waiting for at least 1 node(s) to become ready in`
```
    labels:
      node-role.kubernetes.io/user: user
```

An error message saying `node-role.kubernetes.io/` is not a permitted label for kubernetes versions>=1.16 as described here https://github.com/kubernetes/kubernetes/issues/85073

**How to reproduce it?**
See config above

The use case for this was to easily see node roles in the output of `kubectl get nodes` A workaround is just to use a generic label:
```
    labels:
      role: user
```
The use a flag `kubectl get nodes -L role`

I suspect there may be other open issues that this will fix (https://github.com/weaveworks/eksctl/issues/2124). This was really annoying to debug since there weren't really any informative error messages or warnings, just nodes visable in ec2 console but not via kubectl. Adding some logic to check for disallowed labels would be nice.

**Versions**
```
$ eksctl version 1.19
$ kubectl version 1.16
```

cPu1:
### Description

It's no longer permitted by the kubelet: kubernetes/kubernetes#85073

Fixes #2197 

### Checklist
- [x] Manually tested

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,6 +57,7 @@ require (
 	k8s.io/code-generator v0.16.8
 	k8s.io/kops v1.15.2
 	k8s.io/kubelet v0.0.0
+	k8s.io/kubernetes v1.15.3
 	k8s.io/legacy-cloud-providers v0.0.0
 	sigs.k8s.io/aws-iam-authenticator v0.5.0
 	sigs.k8s.io/yaml v1.2.0
--- a/go.sum
+++ b/go.sum
@@ -1480,6 +1480,7 @@ k8s.io/kube-scheduler v0.16.8/go.mod h1:Yq1uX8tvuzb/dxjx0nKQE88XBRDRyhW75drSDHLT
 k8s.io/kubectl v0.16.8/go.mod h1:s6lhPwTkZjHBVHH6BRVRUA1SAqqZF8osaWSYTvR1H1U=
 k8s.io/kubelet v0.16.8 h1:737GeCz3Bu4NxGivTj910vPaL6UYkxaXoOGuMarzCI0=
 k8s.io/kubelet v0.16.8/go.mod h1:mzDpnryQg2dlB6V3/WAgb1baIamiICtWpXMFrPOFh6I=
+k8s.io/kubernetes v1.16.8 h1:AQb20svioSN1foO9LOdZiUOM8zRmNua2PnYB8bvO48w=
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -56,7 +56,8 @@ require (
 	k8s.io/cloud-provider v0.16.8
 	k8s.io/code-generator v0.16.8
 	k8s.io/kops v1.15.2
-	k8s.io/kubelet v0.0.0
+	k8s.io/kubelet v0.18.3
```
michaelbeaumont, line 59: does it matter that we use `0.18` here, `1.15` below and `0.16` elsewhere?

cPu1: `k8s.io/kubelet` doesn't need to be updated. This change only needs `k8s.io/kubernetes v1.15.3`. I was updating some k8s.io packages locally and forgot to revert the kubelet dependency.

go.sum
```diff
@@ -1480,6 +1480,7 @@ k8s.io/kube-scheduler v0.16.8/go.mod h1:Yq1uX8tvuzb/dxjx0nKQE88XBRDRyhW75drSDHLT
 k8s.io/kubectl v0.16.8/go.mod h1:s6lhPwTkZjHBVHH6BRVRUA1SAqqZF8osaWSYTvR1H1U=
 k8s.io/kubelet v0.16.8 h1:737GeCz3Bu4NxGivTj910vPaL6UYkxaXoOGuMarzCI0=
 k8s.io/kubelet v0.16.8/go.mod h1:mzDpnryQg2dlB6V3/WAgb1baIamiICtWpXMFrPOFh6I=
+k8s.io/kubernetes v1.16.8 h1:AQb20svioSN1foO9LOdZiUOM8zRmNua2PnYB8bvO48w=
```
martina-if, line 1483: Why is it 1.16 here but 1.15 in go.mod?

cPu1: That's because of the replace directive in `go.mod`:

```
k8s.io/kubernetes => k8s.io/kubernetes v1.16.8
```

martina-if: Aaah, thanks!

PR #2229, opened 2020-05-22
Add netlify config as code

michaelbeaumont:
https://docs.netlify.com/configure-builds/file-based-configuration/

### Checklist
- [ ] Manually tested

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,7 @@
+[build]
+  base = "userdocs"
+  publish = "userdocs/site"
+  command = "cd .. && make -f Makefile.docs build-pages"
+
+[build.environment]
+  PYTHON_VERSION = "3.7"


netlify.toml
```diff
@@ -0,0 +1,7 @@
+[build]
+  base = "userdocs/"
```
martina-if, line 2: Why not just leave `base` empty and `command = make -f Makefile.docs build-pages` ?

michaelbeaumont: Netlify tried to be magic and installs `requirements.txt` because it finds it in base. Ill try installing explicitly instead

martina-if: ah I see

michaelbeaumont: I just saw it also needs `runtime.txt` in the base dir so I'll leave it as `userdocs`

PR #2241, opened 2020-05-26
Support URLs as template data when creating stack

rndstr:
(This PR does not add any features to eksctl itself but extends stack manipulation functions to accept template data in various forms which expands applicability for other code using the stack manager.)

### Description

This allows `StackCollection.DoCreateStackRequest()` and
`StackCollection.UpdateStack()` to accept an URL instead of the template
body.


michaelbeaumont:
Perhaps good to be clear though that with this PR we are implicitly supporting eksctl as a library.

PR #2246, opened 2020-05-27
Fix bug in supporting non-default branches for gitops repos

### LINKED ISSUE #2240: Flux / GitOps setup fails for non-default git.repo.branch
**What happened?**

Eksctl fails to setup Flux (and the Helm Operator) when a custom (non-`master` / default) `git.repo.branch` is configured.

Eksctl creating a new orphan branch as configured, and committing the Flux and Helm Operator manifests to this branch.

**How to reproduce it?**

```sh
cat << EOF | eksctl create cluster -f -
kind: ClusterConfig
metadata:
  name: custom-flux-branch
  region: eu-north-1
git:
  repo:
    url: "git@github.com:org/example.git"
    branch: custom
EOF
```

yields

```log
...
[!]  flux deployment was not found
[ℹ]  Generating manifests
[ℹ]  Cloning git@github.com:org/example.git
Cloning into '/tmp/eksctl-install-flux-clone-188933638'...
remote: Enumerating objects: 31, done.        
remote: Counting objects: 100% (31/31), done.        
remote: Compressing objects: 100% (28/28), done.        
remote: Total 31 (delta 2), reused 31 (delta 2), pack-reused 0        
Receiving objects: 100% (31/31), 26.37 KiB | 999.00 KiB/s, done.
Resolving deltas: 100% (2/2), done.
error: pathspec 'custom' did not match any file(s) known to git
Error: unable to install flux: cannot clone repository git@github.com:org/example.git: exit status 1
```

**Versions**

```console
$ eksctl version
0.20.0
```

hiddeco:
### Description

Fixes #2240 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -5,7 +5,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
 	"github.com/kris-nova/logger"
@@ -107,11 +106,7 @@ func (git *Client) cloneRepoInPath(clonePath string, options CloneOptions) error
 		// Switch to target branch
 		args := []string{"checkout", options.Branch}
 		if options.Bootstrap {
-			empty, err := git.isRepoEmpty()
-			if err != nil {
-				return err
-			}
-			if empty {
+			if !git.isRemoteBranch(options.Branch) {
 				args = []string{"checkout", "-b", options.Branch}
 			}
 		}
@@ -123,13 +118,9 @@ func (git *Client) cloneRepoInPath(clonePath string, options CloneOptions) error
 	return nil
 }
 
-func (git *Client) isRepoEmpty() (bool, error) {
-	// A repository is empty if it doesn't have branches
-	files, err := ioutil.ReadDir(filepath.Join(git.dir, ".git", "refs", "heads"))
-	if err != nil {
-		return false, err
-	}
-	return len(files) == 0, nil
+func (git *Client) isRemoteBranch(branch string) bool {
+	err := git.runGitCmd("ls-remote", "--heads", "--exit-code", "origin", branch)
+	return err == nil
 }
 
 // Add performs can perform a `git add` operation on the given file paths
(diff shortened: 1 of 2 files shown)


pkg/git/git.go
```diff
@@ -104,11 +103,7 @@ func (git *Client) cloneRepoInPath(clonePath string, options CloneOptions) error
 		// Switch to target branch
 		args := []string{"checkout", options.Branch}
 		if options.Bootstrap {
-			empty, err := git.isRepoEmpty()
-			if err != nil {
-				return err
-			}
-			if empty {
+			if !git.isRemoteBranch(options.Branch) {
```
michaelbeaumont, line 106: It's not clear to me why it's important that it not be a remote branch in order to execute `checkout -b`? What happens if it isn't a remote branch but still exists? Why not check existence for the condition?

hiddeco: :100:

pkg/git/git.go
```diff
@@ -182,6 +173,9 @@ func (git Client) Commit(message, user, email string) error {
 
 // Push pushes the changes to the origin remote
 func (git Client) Push() error {
+	if err := git.runGitCmd("config", "push.default", "current"); err != nil {
```
martina-if, line 176: Is it possible to avoid having to touch the local configuration for this? I was planning on avoiding that for the user and the email as well so it would be nice to have. (Although no need to block the pr for that)

martina-if:
Thanks for the fix Hidde!

pkg/git/git.go
hiddeco: After a couple of experiments this seemed to be the easiest way to cause the push to work without complaints about a missing upstream branch.

The other alternative would be to set the remote tracking branch after checking out, but this did not seem inline with the tactic used for the commit, which applies configuration at the time it is actually required for an action.

martina-if:
@hiddeco can you rebase?

hiddeco:
@martina-if done!

Issue #1183, opened 2019-08-22
add `--config-file` flag to `eksctl scale nodegroup`


errordeveloper:
xref #1126

PR #2022, opened 2020-04-09
Add nodes-min and nodes-max flags in ng scale

sayboras:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/1893
Fixes https://github.com/weaveworks/eksctl/issues/809

### Approach

I am having below assumptions/scopes while implementing this feature.
- `--nodes` flag is mandatory flag
- following `fail fast` approach i.e. if there is anything wrong with input, just throw error instead of waiting for error from CF.
- supporting for configuration file will be done later in separate PR. I would like to hear your inputs if I am going with right direction firstly.

### Testing

<details>
<summary>invalid flag values</summary>

```shell script
$ ./eksctl scale ng --cluster amazing-sculpture-1586756685 ng-89ba46f7 --nodes 2 --nodes-max 1 --nodes-min 1
Error: number of nodes must be within range of min nodes and max nodes

$ ./eksctl scale ng --cluster amazing-sculpture-1586756685 ng-89ba46f7 --nodes 2 --nodes-max 1              
Error: maximum number of nodes must be greater than or equal to number of nodes

$ ./eksctl scale ng --cluster amazing-sculpture-1586756685 ng-89ba46f7 --nodes 2 --nodes-min 3
Error: minimum number of nodes must be less than or equal to number of nodes

```

</details>

<details>
<summary>no change in current value</summary>

```shell script
$ ./eksctl scale ng --cluster amazing-sculpture-1586756685 ng-89ba46f7 --nodes 2                            
[ℹ]  scaling nodegroup stack "eksctl-amazing-sculpture-1586756685-nodegroup-ng-89ba46f7" in cluster eksctl-amazing-sculpture-1586756685-cluster
[ℹ]  no change for nodegroup "ng-89ba46f7" in cluster "eksctl-amazing-sculpture-1586756685-cluster": nodes-min 2, desired 2, nodes-max 2

$ ./eksctl scale ng --cluster amazing-sculpture-1586756685 ng-89ba46f7 --nodes 2 --nodes-min 2              
[ℹ]  scaling nodegroup stack "eksctl-amazing-sculpture-1586756685-nodegroup-ng-89ba46f7" in cluster eksctl-amazing-sculpture-1586756685-cluster
[ℹ]  no change for nodegroup "ng-89ba46f7" in cluster "eksctl-amazing-sculpture-1586756685-cluster": nodes-min 2, desired 2, nodes-max 2

[...]

PR #2257, opened 2020-05-28
Add support for config file in scale nodegroup

sayboras:
### Description

Relates https://github.com/weaveworks/eksctl/pull/2022
Fixes https://github.com/weaveworks/eksctl/issues/1183

_**PS**_: I am running out of AWS credit currently in my personal AWS account, will do manual testing later. So just mark this for early review and feedback.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -27,9 +27,11 @@ type ClusterConfigLoader interface {
 type commonClusterConfigLoader struct {
 	*Cmd
 
-	flagsIncompatibleWithConfigFile, flagsIncompatibleWithoutConfigFile sets.String
-
-	validateWithConfigFile, validateWithoutConfigFile func() error
+	flagsIncompatibleWithConfigFile    sets.String
+	flagsIncompatibleWithoutConfigFile sets.String
+	validateWithConfigFile             func() error
+	validateWithoutConfigFile          func() error
+	nameArgumentAllowed                bool
 }
 
 var (
@@ -96,8 +98,10 @@ func (l *commonClusterConfigLoader) Load() error {
 		}
 	}
 
-	if l.NameArg != "" {
-		return ErrCannotUseWithConfigFile(fmt.Sprintf("name argument %q", l.NameArg))
+	if !l.nameArgumentAllowed {
+		if l.NameArg != "" {
+			return ErrCannotUseWithConfigFile(fmt.Sprintf("name argument %q", l.NameArg))
+		}
 	}
 
 	if meta.Name == "" {
--- a/pkg/ctl/cmdutils/scale.go
+++ b/pkg/ctl/cmdutils/scale.go
@@ -1,0 +1,103 @@
+package cmdutils
+
+import (
+	"fmt"
+
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+)
+
+// NewScaleNodeGroupLoader will load config or use flags for 'eksctl scale nodegroup'
+func NewScaleNodeGroupLoader(cmd *Cmd, ng *api.NodeGroup) ClusterConfigLoader {
+	l := newCommonClusterConfigLoader(cmd)
+	l.nameArgumentAllowed = true
+
+	l.flagsIncompatibleWithConfigFile.Insert(
+		"nodes",
+		"nodes-min",
+		"nodes-max",
+	)
+
+	l.validateWithConfigFile = func() error {
+		if err := validateNameArgument(cmd, ng); err != nil {
+			return err
+		}
+
+		ngFilter := NewNodeGroupFilter()
+		ngFilter.AppendIncludeNames(ng.Name)
+
+		matchedNodeGroup := ngFilter.FilterMatching(l.ClusterConfig.NodeGroups)
+		if len(matchedNodeGroup) == 0 {
+			return fmt.Errorf("node group %s not found", ng.Name)
+		}
+
+		if err := validateNumberOfNodes(matchedNodeGroup[0]); err != nil {
+			return err
+		}
+		ng = matchedNodeGroup[0]
+		l.Plan = false
+		return nil
+	}
+
+	l.validateWithoutConfigFile = func() error {
+		if l.ClusterConfig.Metadata.Name == "" {
+			return ErrMustBeSet(ClusterNameFlag(cmd))
+		}
+
+		if err := validateNameArgument(cmd, ng); err != nil {
+			return err
+		}
+
+		if err := validateNumberOfNodes(ng); err != nil {
+			return err
+		}
+		l.Plan = false
+		return nil
+	}
+
+	return l
+}
+
+func validateNameArgument(cmd *Cmd, ng *api.NodeGroup) error {
+	if ng.Name != "" && cmd.NameArg != "" {
+		return ErrFlagAndArg("--name", ng.Name, cmd.NameArg)
+	}
+
+	if cmd.NameArg != "" {
+		ng.Name = cmd.NameArg
+	}
+
+	if ng.Name == "" {
+		return ErrMustBeSet("--name")
+	}
+
+	return nil
+}
+
+func validateNumberOfNodes(ng *api.NodeGroup) error {
+
+	if ng.DesiredCapacity == nil || *ng.DesiredCapacity < 0 {
+		return fmt.Errorf("number of nodes must be 0 or greater")
+	}
+
+	if ng.MaxSize != nil && *ng.MaxSize < 0 {
+		return fmt.Errorf("maximum number of nodes must be 0 or greater")
+	}
+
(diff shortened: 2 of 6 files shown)


pkg/ctl/cmdutils/configfile.go
```diff
@@ -426,6 +426,90 @@ func NewDeleteNodeGroupLoader(cmd *Cmd, ng *api.NodeGroup, ngFilter *NodeGroupFi
 	return l
 }
 
+// NewScaleNodeGroupLoader will load config or use flags for 'eksctl scale nodegroup'
+func NewScaleNodeGroupLoader(cmd *Cmd, ng *api.NodeGroup, ngFilter *NodeGroupFilter) ClusterConfigLoader {
```
martina-if, line 430: Can we move this to its own file?

pkg/ctl/cmdutils/configfile_test.go
```diff
@@ -301,6 +303,79 @@ var _ = Describe("cmdutils configfile", func() {
...
+			}
+
+			ngFilter := NewNodeGroupFilter()
+			ng := &api.NodeGroup{}
+			if err := NewScaleNodeGroupLoader(cmd, ng, ngFilter).Load(); err != nil {
+				Expect(err.Error()).To(Equal(params.error.Error()))
+			} else {
+				Expect(params.error).NotTo(HaveOccurred())
+				Expect(ng.DesiredCapacity).To(Not(Equal(0)))
+			}
```
martina-if, line 338: This needs to be written in a different way so that in case of error the messages make sense and don't panic (the argument of `Expect()` should always be the actual value):

```
err := NewScaleNodeGroupLoader(cmd, ng, ngFilter).Load()
if params.error != nil {
    Expect(err).To(HaveOccurred())
	Expect(err.Error()).To(ContainSubstring(params.error.Error()))
} else {
}
```

pkg/ctl/scale/nodegroup.go
```diff
@@ -31,6 +31,8 @@ func scaleNodeGroupWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *cmdutils.Cmd
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
 		fs.StringVar(&cfg.Metadata.Name, "cluster", "", "EKS cluster name")
 		fs.StringVarP(&ng.Name, "name", "n", "", "Name of the nodegroup to scale")
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
+		cmdutils.AddNodeGroupFilterFlags(fs, &cmd.Include, &cmd.Exclude)
```
martina-if, line 35: If a command is only going to be applied to one nodegroup and doesn't support using globs I think it would be a better UX to use `--name` instead of `--include` and `--exclude`. It is also less complex code-wise.

@mb @cPu1 What do you think?

pkg/ctl/cmdutils/configfile.go
sayboras: sure :100:

sayboras: what do you have in mind for file name ? I am thinking about group all node related loader into one file (e.g. nodegroup_configfile.go)

Or just specific file for scale node group e.g. scale_nodegroup_confilefile.go ?

martina-if: There are similar files in that package like `fargate.go` so perhaps `scale.go`

sayboras: thanks, I was a little bit confused with some other files like _create_cluster.go_. Let me rename it to scale.go :tada:

pkg/ctl/cmdutils/scale.go
```diff
@@ -0,0 +1,91 @@
...
+
+		if ng.DesiredCapacity == nil || *ng.DesiredCapacity < 0 {
+			return fmt.Errorf("number of nodes must be 0 or greater")
+		}
+
+		if ng.MaxSize != nil && *ng.MaxSize < 0 {
+			return fmt.Errorf("maximum number of nodes must be 0 or greater")
+		}
+
+		if ng.MaxSize != nil && ng.MinSize != nil && (*ng.MinSize > *ng.DesiredCapacity || *ng.MaxSize < *ng.DesiredCapacity) {
```
martina-if, line 29: I think this check can be removed for simplification since it's contained in the checks of lines 33 and 41, isn't it?

pkg/ctl/cmdutils/scale.go
```diff
@@ -0,0 +1,91 @@
...
+func NewScaleNodeGroupLoader(cmd *Cmd, ng *api.NodeGroup, ngFilter *NodeGroupFilter) ClusterConfigLoader {
+	l := newCommonClusterConfigLoader(cmd)
+
+	l.flagsIncompatibleWithConfigFile.Insert(
+		"nodes",
+		"nodes-min",
+		"nodes-max",
+	)
+
+	validateNodeGroup := func(ng *api.NodeGroup) error {
```
martina-if, line 19: This could be reused, lets extract it to a function called something like `validateNumberOfNodes()` or `validateCapacityValues()` or something like that.

pkg/ctl/cmdutils/scale.go
sayboras: yes, you are right. Let me do it once we confirm the above point about --name flag if you don't mind :)

pkg/ctl/scale/nodegroup.go
sayboras: Please let me know if the above proposal is confirmed, thanks :)

michaelbeaumont: Confirmed as far as I'm concerned, I think @cPu1 too since he also :+1:ed it

sayboras: cool, I can continue with this PR soon, hope that `@mb` will give a go ahead as well :pray:

michaelbeaumont: @sayboras I think I'm mb :grin:

sayboras: didn't have time to wrap it up this week, will plan sometime this weekend or next week if you don't mind

martina-if: Sure! No hurries!

pkg/ctl/cmdutils/configfile.go
```diff
@@ -27,9 +27,11 @@ type ClusterConfigLoader interface {
 type commonClusterConfigLoader struct {
 	*Cmd
 
-	flagsIncompatibleWithConfigFile, flagsIncompatibleWithoutConfigFile sets.String
-
-	validateWithConfigFile, validateWithoutConfigFile func() error
+	flagsIncompatibleWithConfigFile    sets.String
```
sayboras, line 30: for readability and alignment only

pkg/ctl/cmdutils/configfile.go
```diff
@@ -27,9 +27,11 @@ type ClusterConfigLoader interface {
...
+	flagsIncompatibleWithoutConfigFile sets.String
+	validateWithConfigFile             func() error
+	validateWithoutConfigFile          func() error
+	nameArgumentAllowed                bool
```
sayboras, line 34: new flag to bypass the existing error with name argument is given

sayboras:
the git history is a little bit messy, so I rebased and squashed into one commit, hope it will not caush so much trouble reviewing.

pkg/ctl/scale/nodegroup_test.go
```diff
@@ -75,6 +86,26 @@ var _ = Describe("scale", func() {
 				args:  []string{"nodegroup", "ng", "--cluster", "dummy", "--nodes", "2", "--nodes-min", "1", "--nodes-max", "1"},
 				error: fmt.Errorf("number of nodes must be within range of min nodes and max nodes"),
 			}),
+			Entry("with config file and no name flags", invalidParamsCase{
+				args:  []string{"nodegroup", "-f", "../cmdutils/test_data/scale-ng-test.yaml"},
+				error: fmt.Errorf("--name must be set"),
+			}),
+			Entry("with config file and nodes flags", invalidParamsCase{
+				args:  []string{"nodegroup", "-f", "../cmdutils/test_data/scale-ng-test.yaml", "--nodes", "2"},
```
martina-if, line 94: The convention is that if one uses the config file the other CLI flags are incompatible so I would make all these flags incompatible with the config file and expect the user to set them in the appropriate field in the yaml.

sayboras: @martina-if Is this one https://github.com/weaveworks/eksctl/pull/2257/files#diff-4976b4e216739d1a8ce8733f6c089b52R14 good enough, or there is something else that I could have missed ?

netiperher:
I'm a bit confused about how this new future should be used. Neither of this commands work for me:

```
eksctl scale nodegroup -f cluster.yaml
Error: --name must be set

eksctl scale nodegroup -f cluster.yaml --name core
Error: cannot use --name when --config-file/-f is set

eksctl version
0.23.0
```

PR #2266, opened 2020-05-29
Mark metadata required, fix Makefile.docs

michaelbeaumont:
### Checklist
- [ ] Manually tested


Makefile.docs
```diff
@@ -1,5 +1,5 @@
-userdocs/src/usage/schema.json:
-	cp ./pkg/apis/eksctl.io/v1alpha5/assets/schema.json ./userdocs/src/usage/schema.json
+userdocs/src/usage/schema.json: pkg/apis/eksctl.io/v1alpha5/assets/schema.json
+	cp $< $@
```
martina-if, line 2: What is this magic?

michaelbeaumont: `make` built-in variables! `$<` is the first dependency thingy and `$@` is the target.

martina-if: Can we have the actual names instead? I prefer it to be explicit than something I won't understand in one month from now :/

PR #2274, opened 2020-05-30
Unattended Flux installation for GitHub repos (automatic add of deploy key)

### LINKED ISSUE #2273: Unattended Flux installation by automated deploy key management

As of today, `eksctl create cluster` and `eksctl enable repo` can setup a GitOps config repo and install Flux manifests into the cluster automatically. But you still need to manually register a deploy key emitted by eksctl to e.g. GitHub to allow the installed Flux to successfully pull manifests from the config repo. This manual step prevents me from fully automating cluster creation.

Given that GitHub has a dedicated API for creating and deleting deploy keys, there can be two options.

1. Enhance eksctl to allow exporting the public key generated by Flux on `eksctl create cluster` and `eksctl enable repo`, so that we can possibly use external tool(s) to create a deploy key from the exported flux ssh public key.

2. Enhance eksctl to create the deploy key on your behalf.

I believe the latter is nicer for U/X.

mumoshu:
You don't need to manually add GitHub deploy keys anymore.

This feature enables you to install Flux via eksctl in an unattended way, by automatically creating GitHub deply key on cluster creation and on `eksctl enable repo`, and by automatically deleting the deploy key on cluster deletion.

All you need to use this feature is providing `GITHUB_TOKEN` that has access to your repository's deploy keys, and a standard cluster.yaml that contains a `git` configuration for installing Flux.

### eksctl create cluster

eksctl automatically creates a deploy key named  `eksctl-REGION-NAME` from the public ssh key generated by Flux

```
$ eksctl create cluster -f cluster.yaml
*snip*

[ℹ]  created "flux:Deployment.apps/memcached"
[ℹ]  created "flux:Service/memcached"
[ℹ]  created "flux:Secret/flux-git-deploy"
[ℹ]  created "flux:ServiceAccount/helm-operator"
[ℹ]  created "ClusterRole.rbac.authorization.k8s.io/helm-operator"
[ℹ]  created "ClusterRoleBinding.rbac.authorization.k8s.io/helm-operator"
[ℹ]  Waiting for Helm Operator to start
[ℹ]  Helm Operator started successfully
[ℹ]  see https://docs.fluxcd.io/projects/helm-operator for details on how to use the Helm Operator
[ℹ]  Waiting for Flux to start
[ℹ]  fetching public SSH key from Flux
[ℹ]  Flux started successfully
[ℹ]  see https://docs.fluxcd.io/projects/flux for details on how to use Flux
[ℹ]  Flux will only operate properly once it has write-access to the Git repository
[ℹ]  Creating GitHub deploy key from Flux SSH public key
[ℹ]  eksctl-flux-us-east-2-gitops1-br8qufsllhcifvnfv0k0 configured with Flux SSH public key
2020-05-30T10:12:45.276+0900 [DEBUG] plugin.terraform-provider-eksctl: ssh-rsa AAAAB3NzaC1yc2EAAAADAQA*OMITTED*
```

### eksctl delete cluster

eksctl automatically deletes the deploy key named `eksctl-REGION-NAME` by calling GitHub API.

```
$ eksctl delete cluster -f cluster.yaml
[ℹ]  using region us-east-2
[ℹ]  deleting EKS cluster "gitops1-br8qufsllhcifvnfv0k0"
[ℹ]  deleted 0 Fargate profile(s)
[✔]  kubeconfig has been updated
[ℹ]  3 sequential tasks: { delete nodegroup "ng1", delete IAM OIDC provider, delete cluster control plane "gitops1-br8qufsllhcifvnfv0k0" }
[ℹ]  will delete stack "eksctl-gitops1-br8qufsllhcifvnfv0k0-nodegroup-ng1"
[ℹ]  waiting for stack "eksctl-gitops1-br8qufsllhcifvnfv0k0-nodegroup-ng1" to get deleted
[ℹ]  will delete stack "eksctl-gitops1-br8qufsllhcifvnfv0k0-cluster"
[ℹ]  waiting for stack "eksctl-gitops1-br8qufsllhcifvnfv0k0-cluster" to get deleted
[✔]  all cluster resources were deleted
[ℹ]  Deleting GitHub deploy key
[ℹ]  Deleted GitHub deploy key eksctl-flux-us-east-2-gitops1-br8qufsllhcifvnfv0k0
```

### eksctl enable repo

eksctl automatically creates a deploy key named  `eksctl-REGION-NAME` from the public ssh key generated by Flux

```
$ eksctl enable repo -f cluster.yaml
```

Please also note that this feature has an option to make the deploy key "read-only". With the read-only deploy key, you can effectively block Flux from ever pushing commits to the repository.

~~This can be enabled by setting `git.readOnly` to `true` or passing `--readonly` to `eksctl enable repo`.~~

This can be enabled by setting `git.operator.readOnly` to `true` or passing `--read-only` to `eksctl enable repo`.

`git.operator.readOnly` affects Flux only and doesn't prevent eksctl from writing to the git repo for Flux manifests. Use `git.commitOperatorManifests: false` for that.

Resolves #2273

### Checklist
- [x] Manually tested


pkg/gitops/deploykey/deploykey.go
```diff
@@ -0,0 +1,155 @@
...
+	"regexp"
+
+	"github.com/google/go-github/v31/github"
+	"github.com/kris-nova/logger"
+	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/gitops/flux"
+	"golang.org/x/oauth2"
+)
+
+type Manager struct {
```
stefanprodan, line 16: I think this file should be called `github.go` and the Manager renamed to `GitHubProvider`. To allow different integrations e.g. [GitLab API](https://docs.gitlab.com/ee/api/deploy_keys.html) we could create an interface `GitProvider` with the current Manager functions.

pkg/gitops/deploykey/deploykey.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	logger.Info("Deleted GitHub deploy key %s", title)
+
+	return nil
+}
+
+func (km *Manager) getGitHubAPIClient(ctx context.Context) (*github.Client, bool) {
+	githubToken := os.Getenv("GITHUB_TOKEN")
```
stefanprodan, line 122: Can you please create a constant for `GITHUB_TOKEN` and check its existence in `ForCluster`. If the env var is not set then all functions should be `no-op`. The `GitHubProvider` could have a `token` field with the value being set in `ForCluster`.

pkg/gitops/deploykey/deploykey.go
```diff
@@ -0,0 +1,155 @@
...
+	m := r.FindStringSubmatch(km.repoURL)
+
+	if len(m) != 3 {
+		return "", "", false
+	}
+
+	return m[1], m[2], true
+}
+
+func (km *Manager) getDeployKeyTitle(cluster *api.ClusterMeta) string {
```
stefanprodan, line 153: km has a `cluster` filed,  the signature could be `getDeployKeyTitle()`.

mumoshu:
@stefanprodan Hey! Thanks for reviewing. Just added e212ee10 to hopefully address every point.

pkg/gitops/deploykey/deploykey.go
```diff
@@ -0,0 +1,38 @@
...
+				km.repoURL = repo.URL
+			}
+
+			km.readOnly = git.ReadOnly
+		}
+
+		return km
+	}
+
+	logger.Info("GITHUB_TOKEN is not set. Please set it so that eksctl is able to create and delete GitHub deploy key from Flux SSH public key")
```
stefanprodan, line 35: We should log this only if the URL contains `github.com`.

pkg/gitops/deploykey/github.go
```diff
@@ -0,0 +1,127 @@
...
+	if _, err := gh.Repositories.DeleteKey(ctx, owner, repo, keyID); err != nil {
+		return err
+	}
+
+	logger.Info("deleted GitHub deploy key %s", title)
+
+	return nil
+}
+
+func (p *GitHubProvider) getGitHubAPIClient(ctx context.Context) *github.Client {
```
stefanprodan, line 99: Does this matches `ssh://` URLs?

pkg/gitops/flux/installer.go
```diff
@@ -117,27 +116,29 @@ func (fi *Installer) Run(ctx context.Context) (string, error) {
...
+		return "", nil, err
 	}
 
 	logger.Info("Flux started successfully")
 	logger.Info("see https://docs.fluxcd.io/projects/flux for details on how to use Flux")
 
-	logger.Info("Committing and pushing manifests to %s", fi.opts.Repo.URL)
-	if err = fi.addFilesToRepo(); err != nil {
-		return "", err
+	if !fi.opts.ReadOnly {
```
stefanprodan, line 130: Why is this conditioned by readonly?

pkg/gitops/gitops.go
```diff
@@ -46,6 +48,10 @@ func Setup(k8sRestConfig *rest.Config, k8sClientSet kubeclient.Interface, cfg *a
 		return err
 	}
 
+	if fluxSSHKey != nil {
+		return deploykey.ForCluster(cfg).Put(context.Background(), *fluxSSHKey)
+	}
+
 	logger.Info(userInstructions)
```
stefanprodan, line 55: If the user doesn't set the token env var it should be prompted to set the key manually. But it will never get to it.

pkg/gitops/deploykey/github.go
mumoshu: No. Should we match `ssh://git@github.com:OWNER/REPO.git`-styled URLs as well?

pkg/gitops/flux/installer.go
mumoshu: I thought the intention of the user would be that they won't like eksctl to write to the repo as well if you want Flux to be read-only. Is that a right assumption?

If not, maybe we'd better split this to two configuration fields, the one for eksctl as e.g. `git.commitManifestsOnStart` and one for flux as e.g.`git.operator.readOnly`.

WDYT?

pkg/gitops/gitops.go
mumoshu: Addressed in 754fe0b3. Does it look good to you now?

pkg/gitops/deploykey/deploykey.go
mumoshu: Sounds good! Addressed in 169f2a1e

mumoshu: Sounds good! Addressed in 169f2a1e

pkg/gitops/deploykey/github.go
mumoshu: Added support for the full ssh urls in 77ebbd68

pkg/gitops/deploykey/deploykey.go
mumoshu: 4b1ceca7 should be better and correct.

pkg/gitops/flux/installer.go
mumoshu: Anyway, I've added an additional commit to implement what I think best now based on your review: 18f7d768

It's now:

```yaml

git:
  # NEW: Commit and push Flux manifests to the Git Repo on install
  # The default value is true for backward-compatbiility, so explicitly set to `false `to disable
  #commitOperatorManifests: false
  repo:
    url: "git@github.com:mumoshu/gitops-demo.git"
    branch: master
    fluxPath: "flux/"
    user: "gitops"
    email: "gitops@myorg.com"
    privateSSHKeyPath: /Users/c-ykuoka/.ssh/id_rsa
  operator:
    namespace: "flux"
    # NEW: Instruct Flux to read-only mode and create the deploy key as read-only
    readOnly: true
```

And commands like `eksctl enable repo` has been given respective common flux flags:

```
      --read-only                         Instruct Flux to read-only mode and create the deploy key as read-only
      --commit-operator-manifests         Commit and push Flux manifests to the Git Repo on install (default true)
```

pkg/gitops/gitops.go
stefanprodan: Have you tested this without providing a token? It should prompt the user to manually set the deploy key.

martina-if:
@mumoshu Thank you so much for taking the time to build this. We need to discuss in the team if we want to keep this functionality here or somewhere else. I will mark this PR as "changes requested" so we don't accidentally merge it before we have decided.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -712,6 +712,8 @@ type Git struct {
 	Operator Operator `json:"operator,omitempty"`
 	// +optional
 	BootstrapProfile *Profile `json:"bootstrapProfile,omitempty"` // one or many profiles to enable on this cluster once it is created
+	// +optional
+	CommitOperatorManifests *bool `json:"commitOperatorManifests,omitempty"` // Commit and push Flux manifests to the Git Repo on install
```
martina-if, line 716: Is there any particular reason to add this new field?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -749,6 +751,8 @@ type Operator struct {
 	Namespace string `json:"namespace,omitempty"` // e.g. flux
 	// +optional
 	WithHelm *bool `json:"withHelm,omitempty"` // whether to install the Flux Helm Operator or not
+	// +optional
+	ReadOnly bool `json:"readOnly,omitempty"` // Instruct Flux to read-only mode and create the deploy key as read-only
```
martina-if, line 755: Let's move this into the `Operator` field

pkg/ctl/enable/repo.go
```diff
@@ -88,11 +90,20 @@ func doEnableRepository(cmd *cmdutils.Cmd) error {
...
 	}
 
-	userInstructions, err := installer.Run(context.Background())
+	userInstructions, fluxSSHKey, err := installer.Run(context.Background())
 	if err != nil {
 		logger.Critical("unable to set up gitops repo: %s", err.Error())
 		return err
 	}
+
+	if fluxSSHKey != nil {
```
martina-if, line 99: Why not move this logic inside the installer instead?

mumoshu: @martina-if It's just that I wanted not to bloat the scope of the installer.

If you're fine, I can definitely move this inside the installer, either in this PR or in another PR.

WDYT?

pkg/apis/eksctl.io/v1alpha5/types.go
mumoshu: Yeah. I believe it has some relationship with the way how I setup Flux.

I usually point multiple clusters' Flux to the same config repository, so that I can swap the cluster for e.g. blue-green-style K8s version upgrade easily(I don't need to lock, recreate or copy the config repo for the new cluster. I can just reuse the config repo this way).

It's like you reuse AWS launch configuration when you want two identical ASGs. I want two identical clusters so I reuse the config repo.

Trying to generate, commit, and pushing Flux manifests for each cluster creation breaks this model.

I thought some percentage of people who would benefit from the unattended Flux installation would use the same model as mine, and therefore find this helpful. That's why I included it in this PR.

WDYT?

pkg/apis/eksctl.io/v1alpha5/types.go
mumoshu: I guess you've misread it, by any chance? It's already a field of the `Operator` struct, if that's what you mean. Please feel free to correct me if I'm missing something.

pkg/gitops/gitops.go
mumoshu: Yes. And it works as intended for me.

Here's the log from `eksctl create -f cluster.yaml` when GITHUB_TOKEN unset:

```
*snip*
[ℹ]  created "flux:Deployment.apps/memcached"
[ℹ]  created "CustomResourceDefinition.apiextensions.k8s.io/helmreleases.helm.fluxcd.io"
[ℹ]  created "flux:Deployment.apps/helm-operator"
[ℹ]  Waiting for Helm Operator to start
[ℹ]  Helm Operator started successfully
[ℹ]  see https://docs.fluxcd.io/projects/helm-operator for details on how to use the Helm Operator
[ℹ]  Waiting for Flux to start
[ℹ]  fetching public SSH key from Flux
[ℹ]  Flux started successfully
[ℹ]  see https://docs.fluxcd.io/projects/flux for details on how to use Flux
[ℹ]  Committing and pushing manifests to git@github.com:mumoshu/gitops-demo.git
[ℹ]  Nothing to commit (the repository contained identical files), moving on
Everything up-to-date
[ℹ]  Flux will only operate properly once it has write-access to the Git repository
[ℹ]  GITHUB_TOKEN is not set. Please set it so that eksctl is able to create and delete GitHub deploy key from Flux SSH public key
[ℹ]  please configure git@github.com:mumoshu/gitops-demo.git so that the following Flux SSH public key has write access to it
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC9u7F63s57eFD7mHmOKPMR2Czi76G4vAYExnlH0jT/SUnkPWaj2hH+OPzJgVPIKk2VaskEyLf4D8z5CRy1FVjtan+gf2L9yAY9KU2d+LXo7qdSyT/XauUfdnkWN5En30zDCSQtbY/lvoh3SXuj2GL6FB/QLqpPH3ULWpswOXeoIiLoig1SFPgtuFwy66whyOx8o0MDHHr9MVrQtiPQWtCP8/x51puuFqXQ25R7srXccEYhYo1bk09g03gk7o4shGaZx5eOSLliFgIH6C7nk7j4OGnlLtrOb/atCD79NzboOvUA6EjEzs/3j8D2DJbY08ahZJ/ryOIStT7BB6a6o+wOy3aHLs3wk2pOsHiQL6NyDLr/3X4wrR0DAUyeLBLqRBqa/oNL50EfyJ7h24BHi22BK8PMmAKxGQSDKaAtq6wQ4xKptohy37utROrvTmIcrc4NgwRCwvyy8svegv68HrPZlK4hUkjCPw6IPhvIlPdhDDER8qx8ziTgB0ayEXh1kiM= root@flux-56b96cc9fb-d5r28
```

mumoshu: I believe the last line `please configure git@github.com:mumoshu/gitops-demo.git so that the following Flux SSH public key has write access to it` is the prompt you asked.

It has been there before this change, and will remain there after this change.

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: you are right! I misread this, sorry about that

pkg/gitops/gitops.go
stefanprodan: Thanks for testing it

pkg/gitops/flux/installer.go
```diff
@@ -283,7 +284,7 @@ func getFluxManifests(opts *api.Git, cs kubeclient.Interface) (map[string][]byte
 		GitLabel:           opts.Operator.Label,
 		GitUser:            opts.Repo.User,
 		GitEmail:           opts.Repo.Email,
-		GitReadOnly:        false,
+		GitReadOnly:        opts.Operator.ReadOnly,
 		Namespace:          opts.Operator.Namespace,
 		ManifestGeneration: true,
 		AdditionalFluxArgs: []string{"--sync-garbage-collection"},
```
stefanprodan, line 290: When using readonly, it would be best to deploy Flux without memcached, this can be done by setting an additional flag `--registry-disable-scanning`.

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: After consulting with the flux team it seems like a valid use case for other people too so let's keep this flag. I think it makes more sense inside the Operator as `git.operator.commitManifests`, WDYT?

pkg/ctl/enable/repo.go
martina-if: I think for now it makes more sense inside the installer so let's move it there.

pkg/apis/eksctl.io/v1alpha5/types.go
mumoshu: @martina-if Thanks! I believe this should better be here as-is, or alternatively under a new field `git.installer` like `git.installer.commitManifests`.

I'm thinking about it this way because `git.operator` you've suggested seems to be dedicated for configuring the operator(flux), not the installer of the operator. And that's the installer who commit/push Flux manifests.

From what I can see from the current spec, anything directly under `git` seems to be for configuring the installer, which make me thing this can safely be `git.commitManifests`, not something new like e.g. `git.installer.commitManiefsts`. But I'm definitely open to more feedback/comments regarding this. WDYT?

mumoshu: Or perhaps it can be `git.operator.commitManifests` because it's for commiting manifests for the operator? If that's your intention, that makes sense. Would you mind confirming?

pkg/gitops/flux/installer.go
mumoshu: Sounds great! Added in 8e6afca3

pkg/apis/eksctl.io/v1alpha5/types.go
martina-if: Exactly, that's what I meant :)

pkg/ctl/enable/repo.go
mumoshu: Addressed in 2dba2c88. Please note that I had to replicate `flux.PublicKey` struct in the `deploykey` package to cut import cycle between `flux` and `deploykey` packages.

pkg/apis/eksctl.io/v1alpha5/types.go
mumoshu: Thanks! Addressed in d14cb1a6

Issue #2265, opened 2020-05-29
asgMetricsCollection without a Metrics list fails to deploy


TBBle:
**What happened?**

Attempting to create a NodeGroup that collects all metrics by default at 1Minute intervals failed:
```
[✖]  creating CloudFormation stack "eksctl-clusterName-nodegroup-groupName": ValidationError: [/Resources/NodeGroup/Type/MetricsCollection/0/Metrics] 'null' values are not allowed in templates
```

Given the following in my NodeGroup:
```
    asgMetricsCollection:
      - granularity: 1Minute
```

Which is consistent with the [eksctl config schema](https://eksctl.io/usage/schema/) that lists only `granularity` as required, and the [CloudFormation schema documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-metricscollection.html#cfn-as-metricscollection-metrics) that if you omit Metrics, all metrics are enabled.

**What you expected to happen?**

The NodeGroup created successfully, and all ASG metrics are collected by CloudWatch.

**How to reproduce it?**

Create a nodegroup using the `asgMetricsCollection` as defined above.

**Anything else we need to know?**

Probably just a simple fix to [`metricsCollectionResource`](https://github.com/weaveworks/eksctl/blob/c84ce3d172c5e4ceef75dbb237d9b5717644e237/pkg/cfn/builder/nodegroup.go#L368) to not add the `"Metrics"` entry if `m.Metrics` was omitted.

I suspect an empty array will in fact disable all metrics, so the only workaround is to use the full list of metrics.

**Versions**
Please paste in the output of these commands:
```
PS C:\Projects\AWSPerfect> eksctl version
0.20.0
PS C:\Projects\AWSPerfect> kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.11", GitCommit:"d94a81c724ea8e1ccc9002d89b7fe81d58f89ede", GitTreeState:"clean", BuildDate:"2020-03-12T21:08:59Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"windows/amd64"}
[...]

PR #2275, opened 2020-05-30
Fix CF template if asgMetricsCollection is having no metrics

sayboras:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/2265

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/cfn/builder/nodegroup.go
+++ b/pkg/cfn/builder/nodegroup.go
@@ -370,9 +370,9 @@ func metricsCollectionResource(asgMetricsCollection []api.MetricsCollection) []m
 	for _, m := range asgMetricsCollection {
 		newCollection := make(map[string]interface{})
 
-		var metrics []string
-		metrics = append(metrics, m.Metrics...)
-		newCollection["Metrics"] = metrics
+		if len(m.Metrics) > 0 {
+			newCollection["Metrics"] = m.Metrics
+		}
 		newCollection["Granularity"] = m.Granularity
 
 		metricsCollections = append(metricsCollections, newCollection)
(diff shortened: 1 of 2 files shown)


pkg/cfn/builder/nodegroup.go
```diff
@@ -370,9 +370,9 @@ func metricsCollectionResource(asgMetricsCollection []api.MetricsCollection) []m
 	for _, m := range asgMetricsCollection {
 		newCollection := make(map[string]interface{})
 
-		var metrics []string
-		metrics = append(metrics, m.Metrics...)
-		newCollection["Metrics"] = metrics
+		if len(m.Metrics) > 0 {
+			newCollection["Metrics"] = m.Metrics
```
michaelbeaumont, line 374: The bug is that we have `newCollection["Metrics"] = nil`, right? `[]` seems to be a valid value for CFN.

I think comparing to `nil` should be enough?

sayboras: Thanks for your input. I am not sure about the effect of passing [] in CF, will it fail for validation, or not ? So to be safe, I just proceed to with _len()_ (which internally check for nil slice as well).

If you have a chance to confirm the above, I can make a chance. Result will be the same from end user 's point of view.

michaelbeaumont: I guess the question is what does the user expect in the case where they pass `[]` explicitly, no metrics or all metrics?

sayboras: I can't test it out right now as I have no credit in my aws account. My intepretation on aws docs for this case is _all metrics_

michaelbeaumont: indeed, seems to be the case that `null` and `[]` are treated the same by cfn.

PR #2286, opened 2020-06-03
Improve config documentation generation

### LINKED ISSUE #2258: Improve config documentation generation (examples, docs for fields)
We investigated config docs in https://github.com/weaveworks/eksctl/issues/2166 and at the moment, we have the new [config documentation](https://deploy-preview-2229--eksctl.netlify.app/usage/schema/) coming up.
However, we would like descriptions, links and examples as well. We [don't have a great way to add these things](https://github.com/weaveworks/eksctl/issues/2166#issuecomment-628101476) yet though.
This issue tracks implementing this, which will [look similar to the skaffold config](https://github.com/weaveworks/eksctl/issues/2166#issuecomment-628099428), using comments on fields to add additional info.

Examples:
```
	// Format formats the date and time.
	// See [#Time.Format](https://golang.org/pkg/time/#Time.Format).
	// Defaults to `2006-01-02_15-04-05.999_MST`.
	Format string `json:"format,omitempty"`
```

michaelbeaumont:
### Description
Everything for the generation is in the first commit in `pkg/schema`! :wink: 
Closes #2258

~I also think this should go in the `0.21.0` release if possible.~
Nevermind, the only reason was because I'm updating the resulting json schema version from `draft-04` (from using `alecthomas/jsonschema`) to `draft-07` with this. But I doubt this will be an issue for anyone.

[Preview](https://deploy-preview-2286--eksctl.netlify.app/usage/schema/)
[Example with links in description](https://deploy-preview-2286--eksctl.netlify.app/usage/schema/#nodeGroups-asgMetricsCollection)

### Checklist
  * Could probably use some additional tests
- [x] Manually tested


pkg/utils/ipnet/ipnet.go
```diff
@@ -18,10 +18,14 @@ var nullString = "null"
...
 var emptyIPNet = net.IPNet{}
 
-// IPNet wraps net.IPNet to get CIDR serialization.
-type IPNet struct {
-	net.IPNet
-}
+type (
+	// IPNet wraps net.IPNet to get CIDR serialization.
+	// +jsonschema noderive
+	// +jsonschema { "type": "string" }
```
michaelbeaumont, line 24: We do custom marshaling here so the schema needs to be overridden somewhere, this seems like a nice declarative way of doing it.

pkg/apis/eksctl.io/v1alpha5/cloudwatch.go
```diff
@@ -1,16 +1,17 @@
...
-	ClusterLogging *ClusterCloudWatchLogging `json:"clusterLogging,omitempty"`
-}
-
-// ClusterCloudWatchLogging container config parameters related to cluster logging
-type ClusterCloudWatchLogging struct {
-	//+optional
-	EnableTypes []string `json:"enableTypes,omitempty"`
-}
+type (
+	// ClusterCloudWatch contains config parameters related to CloudWatch
```
michaelbeaumont, line 4: because of limitations of the go internal `ast` machinery, it appears to be much harder to get access to the comments on a struct if that struct is declared top level like `type X struct {...}`. If all structs are inside a `type ( X struct {...} )` the code can stay simpler.

cmd/schema/generate.go
```diff
@@ -1,59 +1,49 @@
...
 	}
-	schema := reflector.Reflect(&v1alpha5.ClusterConfig{})
-	// We have to manually add examples here, because we can't tag `TypeMeta`
-	// from the k8s package
-	if kind, ok := schema.Definitions["ClusterConfig"].Properties.Get("kind"); ok {
-		t := kind.(*jsonschema.Type)
-		t.Examples = []interface{}{"ClusterConfig"}
+
+	// We add some examples and blacklist some descriptions
+	if t, ok := schema.Definitions["ClusterConfig"].Properties["kind"]; ok {
```
michaelbeaumont, line 27: This might also be possible to override using the `// +jsonschema` directives on fields.

pkg/schema/definition/comment.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+var (
+	regexpDefaults = regexp.MustCompile("(.*)Defaults to `(.*)`")
+	regexpExample  = regexp.MustCompile("(.*)For example: `(.*)`")
+	pTags          = regexp.MustCompile("(<p>)|(</p>)")
+
+	// patterns for enum-type values
+	enumValuePattern     = "^[ \t]*`(?P<name>[^`]+)`([ \t]*\\(default\\))?: .*$"
+	regexpEnumDefinition = regexp.MustCompile("(?m).*Valid [a-z]+ are((\\n" + enumValuePattern + ")*)")
```
michaelbeaumont, line 19: the code is now so different from the original skaffold code but this is the biggest thing I ended up leaving in.
Do we think it could be useful? Do we have/potentially have this type of enum in our config?
See [how it would look](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L138-L143).

cmd/schema/generate.go
michaelbeaumont: Note this code is exactly the same as beforehand when we used the `alecthomas/jsonschema` package.

pkg/apis/eksctl.io/v1alpha5/cloudwatch.go
michaelbeaumont: Pasting some more info here:

> the type X struct variant is parsed differently (as a GenDecl). and at the point where I look in the package scope and find the TypeSpec for X, those comments are missing because they're on the GenDecl which contains that TypeSpec. with type ( X struct ) the comments are on the TypeSpec itself.
> i think I'd have to go back over the files belonging to the package, check all GenDecls, look into the Scope of the package and replace the comments on TypeSpecs with the comments on the GenDecls.

michaelbeaumont: hmm I fixed it, it's actually not too bad

pkg/utils/ipnet/ipnet.go
michaelbeaumont: Syntax changed

pkg/schema/definition/comment.go
martina-if: Yes, I think the `nat` for example has only a few values it can take. There are probably a few more examples.

PR #2306, opened 2020-06-08
Don't use generic helper functions from kops

michaelbeaumont:
### Description
IMO, no reason to use these functions from `kops` of all places. What do you think?

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/utils/slice/slice.go
+++ b/pkg/utils/slice/slice.go
@@ -1,0 +1,12 @@
+package slice
+
+// Contains checks if a slice contains an element
+func Contains(list []string, e string) bool {
+	for _, x := range list {
+		if x == e {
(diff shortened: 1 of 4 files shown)


pkg/utils/slice/slice.go
```diff
@@ -0,0 +1,12 @@
...
+// Contains checks if a slice contains an element
+func Contains(list []string, e string) bool {
+	for _, x := range list {
+		if x == e {
+			return true
+		}
+	}
+
+	return false
+}
```
cPu1, line 12: This can be addressed later, but I think we shouldn't have this utility function at all. This is a simple `for` loop that works only for strings and we have not been using this in other parts of the code.

PR #2308, opened 2020-06-08
Allow StackCollection.DeleteStackByName() to succeed

### LINKED ISSUE #2307: StackCollection.DeleteStackByName() cannot succeed
**What happened?**
_(note both of these functions are not in use in eksctl code)_

[`StackCollection.DeleteStackByName(string)`](https://github.com/weaveworks/eksctl/blob/1807fff6b60e8dcf633c13f6f58682714f619918/pkg/cfn/manager/api.go#L339) and `StackCollection.DeleteStackByNameSync(string, chan error)` allow to delete a stack by name.

The method it eventual calls is [`StackCollection.DeleteStackBySpec(*Stack)`](https://github.com/weaveworks/eksctl/blob/1807fff6b60e8dcf633c13f6f58682714f619918/pkg/cfn/manager/api.go#L378) passing in `&Stack{StackName: "…", StackId: "…"}`. That function only proceeds if the passed in `Stack` object has the cluster's name in its tags which we can see will never be the case because the initial function doesn't set that field.

The stack to be deleted.

**Versions**
current `master` at commit 1807fff6b60e8dcf633c13f6f58682714f619918

rndstr:
Fixes #2307

`StackCollection.DeleteStackBySpec(*Stack)` only deletes the stack if
its tags match the clustername. For that reason we not only need to pass
in the `StackName` and `StackId` but also `Tags`.

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -337,8 +337,7 @@ func defaultStackStatusFilter() []*string {
 
 // DeleteStackByName sends a request to delete the stack
 func (c *StackCollection) DeleteStackByName(name string) (*Stack, error) {
-	i := &Stack{StackName: &name}
-	s, err := c.DescribeStack(i)
+	s, err := c.DescribeStack(&Stack{StackName: &name})
 	if err != nil {
 		err = errors.Wrapf(err, "not able to get stack %q for deletion", name)
 		stacks, newErr := c.ListStacksMatching(fmt.Sprintf("^%s$", name), cloudformation.StackStatusDeleteComplete)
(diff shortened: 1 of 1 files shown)


pkg/cfn/manager/api.go
```diff
@@ -337,8 +337,7 @@ func defaultStackStatusFilter() []*string {
 
 // DeleteStackByName sends a request to delete the stack
 func (c *StackCollection) DeleteStackByName(name string) (*Stack, error) {
-	i := &Stack{StackName: &name}
-	s, err := c.DescribeStack(i)
+	i, err := c.DescribeStack(&Stack{StackName: &name})
```
martina-if, line 340: nit: can we rename `i` to `stack`?

rndstr: after talking on Slack, it was decided to leave it as `i` as it seems consistent with usage throughout eksctl that uses this var name to describe something like "_input_ stack to perform action on".

for actions that return a stack it is usually assigned to `s` which would denote a complete stack object having all the fields filled out.

Issue #2299, opened 2020-06-05
Support App Mesh Preview Environment


bcelenza:
**Why do you want this feature?**

AWS App Mesh provides a [Preview Channel](https://docs.aws.amazon.com/app-mesh/latest/userguide/preview.html) that allows customers to try out and provide feedback on features before we make them generally available. The preview endpoint is distinct from the standard production endpoint for security, availability, and data durability reasons (see FAQs in the first link).

**What feature/behavior/change do you want?**

Because the preview environment is separate, it also uses a different signing name and service ID: `appmesh-preview` instead of `appmesh`. This means that the current [addon policy](https://eksctl.io/usage/iam-policies/) for App Mesh will not work for the preview endpoint. The current policy (from [iam_helper.go](https://github.com/weaveworks/eksctl/blob/master/pkg/cfn/builder/iam_helper.go#L93-L116)):

```
"appmesh:*",
"servicediscovery:CreateService",
"servicediscovery:DeleteService",
"servicediscovery:GetService",
"servicediscovery:GetInstance",
"servicediscovery:RegisterInstance",
"servicediscovery:DeregisterInstance",
"servicediscovery:ListInstances",
"servicediscovery:ListNamespaces",
"servicediscovery:ListServices",
"servicediscovery:GetInstancesHealthStatus",
"servicediscovery:UpdateInstanceCustomHealthStatus",
"servicediscovery:GetOperation",
"route53:GetHealthCheck",
"route53:CreateHealthCheck",
"route53:UpdateHealthCheck",
"route53:ChangeResourceRecordSets",
"route53:DeleteHealthCheck",
```

To support the preview environment, the policy needs to contain `appmesh-preview:*` instead of `appmesh:*`.

My inclination is to create a new addon policy called `appMeshPreview` so the cluster operator could distinguish which App Mesh endpoint was authorized for use. We could also just add `appmesh-preview:*` to the existing addon policy, but it may cause confusion to customers on which endpoint is actually being used.

**Additional Context**

I'm currently working on adding preview support to the App Mesh controller for k8s: https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/75, https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/257

PR #2310, opened 2020-06-09
Add support for App Mesh Preview Environment

stefanprodan:
This PR adds support for App Mesh preview environment. Fix: #2299

Changes:
- add `appMeshPreview` addon policy
- add `appmesh-preview-access` flag to node groups

### Checklist
- [ ] Manually tested


stefanprodan:
running `make check-all-generated-files-up-to-date` locally look ok, no idea what's going on @martina-if

martina-if:
@stefanprodan I think it could be because we've been upgrading different tools. Can you run `make install-build-deps` and `make generate-all` again? (I can do it for you and commit the changes if you want but it would be good if you have it correctly set up as well)

stefanprodan:
@martina-if thanks! I had the old tooling installed, should be ok now.

pkg/cfn/builder/api_test.go
```diff
@@ -1136,6 +1137,49 @@ var _ = Describe("CloudFormation template builder API", func() {
...
+				"servicediscovery:ListNamespaces",
+				"servicediscovery:ListServices",
+				"servicediscovery:GetInstancesHealthStatus",
+				"servicediscovery:UpdateInstanceCustomHealthStatus",
+				"servicediscovery:GetOperation",
+				"route53:GetHealthCheck",
+				"route53:CreateHealthCheck",
+				"route53:UpdateHealthCheck",
+				"route53:ChangeResourceRecordSets",
+				"route53:DeleteHealthCheck",
```
martina-if, line 1177: Can we extract these into a common array for both tests so it's easier to see in a quick look what is the difference between appmesh and appmesh-preview?

pkg/cfn/builder/iam_helper.go
```diff
@@ -90,28 +90,33 @@ func createRole(cfnTemplate cfnTemplate, iamConfig *api.NodeGroupIAM, managed bo
...
-				"servicediscovery:GetOperation",
-				"route53:GetHealthCheck",
-				"route53:CreateHealthCheck",
-				"route53:UpdateHealthCheck",
-				"route53:ChangeResourceRecordSets",
-				"route53:DeleteHealthCheck",
-			},
+			append(appMeshActions, "appmesh:*"),
+		)
+	} else if api.IsEnabled(iamConfig.WithAddonPolicies.AppMeshPreview) {
```
martina-if, line 117: I think if both flags are allowed at the same time both actions should be included. If only one of them is going to be included in the policy the UX should fail when both flags or fields are provided (as they are incompatible).

stefanprodan: I've reconsidered this, both flags could be specified as one may want to switch between stable and preview on the same cluster.

pkg/cfn/builder/api_test.go
stefanprodan: Done!

stefanprodan:
@sayboras I've updated the yaml, thanks.

bcelenza:
Thanks for tackling this @stefanprodan !

Issue #1537, opened 2019-11-09
incorrect usage of k8s.io/kubernetes


carnott-snap:
**What happened?**
When importing `eksctl` as a library, I pull in `k8s.io/kubernetes` as a transitive dependency.

I would like to use the `k8s.io/kubernetes/pkg/cloudprovider/providers/aws` package version that is compatible with the other `kubernetes-1.12.6` libraries that I am importing. I am aware that we should not import `k8s.io/kubernetes` directly, per kubernetes/kubernetes#81878. I looked in [kubernetes/cloud-providers-aws](https://github.com/kubernetes/cloud-provider-aws), however I do not see any `kubernetes-1.12.6` tag and the code was removed at kubernetes/cloud-provider-aws@b377713a2f1de9d744a6f706b77b9d16f72012ed.

**What you expected to happen?**
Customers should not import `k8s.io/kubernetes` directly, see kubernetes/kubernetes#81878, and instead using component modules: `k8s.io/cloud-provider` and `k8s.io/cloud-provider-aws`.

**Anything else we need to know?**
I would have submitted this as a PR, since the `k8s.io/cloud-provider` fix is just a `find -exec sed`, unfortunately I am unclear what the best candidate for replacing `k8s.io/kubernetes/pkg/cloudprovider/providers/aws` is. [kubernetes/cloud-providers-aws](https://github.com/kubernetes/cloud-provider-aws) does not have any tags, let alone `kubernetes-1.12.6`, and the related logic was removed: kubernetes/cloud-provider-aws@b377713a2f1de9d744a6f706b77b9d16f72012ed. As such, I have created an upstream ticket to track the need: kubernetes/kubernetes#85022.

michaelbeaumont:
Duplicate of https://github.com/weaveworks/eksctl/issues/813

PR #2316, opened 2020-06-10
Remove k8s.io/kubernetes dependency

michaelbeaumont:
### Description

According to
https://github.com/golang/go/issues/32776#issuecomment-505607726, it's
not supposed to be used as a dependency.

Of course, hardcoding these labels isn't _great_ but they shouldn't change too
often.

Additionally, I think this code might end up being removed if we agree
with https://github.com/weaveworks/eksctl/issues/2197#issuecomment-635860961.

Closes https://github.com/weaveworks/eksctl/issues/1537

### Checklist
- [ ] Manually tested


michaelbeaumont:
@cPu1
yeah that's an issue but I can't see any other way. The label checking logic isn't exported in `k8s.io/kubelet` or anywhere else (as far as I can tell?)
ultimately i think that comment is saying we shouldn't be using kubelet labels to set node labels in the first place.

cPu1:
>ultimately i think that comment is saying we shouldn't be using kubelet labels to set node labels in the first place.

It is, but the labels feature (`nodeGroup.labels`) relies on support from the kubelet. So if the kubelet doesn't allow setting a label, supporting it in eksctl would be more complex to implement.

go.mod
```diff
@@ -61,7 +61,6 @@ require (
 	k8s.io/code-generator v0.16.8
 	k8s.io/kops v1.15.2
 	k8s.io/kubelet v0.16.8
-	k8s.io/kubernetes v1.16.8
```
sayboras, line 64: will this line be removed as well https://github.com/weaveworks/eksctl/blob/master/go.mod#L97 ?

michaelbeaumont: good question! when I remove it, it gets pulled into `go.sum` anyway at `v1.15.3`, I believe because of `kops`. Perhaps it's safer to leave it?

sayboras: Just my opinion. 

I just enable verbose mode in go build to actually find out _exactly_ what libraries will be used. I don't think k8s.io/kubernetes package is used anywhere. go.mod sometimes tries to be smart while deriving dependencies transitively.

```shell script
$ go clean --cache
$  CGO_ENABLED=0 go build -v ./cmd/eksctl > package.txt 2>&1 
$ wc -l package.txt 
775 package.txt
$ rg "k8s.io/kubernetes" package.txt   
```

martina-if:
Let's see if there is a better place for this code like the kubelet module and open a PR for it.

michaelbeaumont:
Opened a [PR to kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/pull/92527)

michaelbeaumont:
Waiting on the above PR

PR #2335, opened 2020-06-15
Add c5a and c5ad instances to maxpods.go

### LINKED ISSUE #2333: Failure to create instance types c5a.24xlarge or m5a.24xlarge
**What happened?**
Failure to deploy node group with machine type c5a.24xlarge or m5a.24xlarge.

A nodegroup of c24.xlarge is deployed

**How to reproduce it?**
Try and create any cluster with c5a.24xlarge set as instance type. 

I can note that type c5ad.24xlarge DOES work though. The issue isolated to just the “a” instance types. 

Running a workstation with Ubuntu 20.04

The command will eventually time out waiting for the actual instance to start up. When you check if there is an instance in the auto scaling group there actually is! 

My stab at debugging points to an issue with the aws-node deployment and then communication back to the control plane. 

**Versions**
Eksctl 0.21.0
Kubectl 1.18

blakestoddard:
### Description

Regenerate `maxpods.go` to pickup values for `c5a` and `c5ad`.

Fixes #2333 (as I hit this today trying to run `c5a` instances and was scratching my head as to why `MAX_PODS` was empty.)


michaelbeaumont:
Thanks!

PR #2339, opened 2020-06-17
Windows 1909 support

### LINKED ISSUE #2321: Windows 1909 AMI support for tag --node-ami-family
Currently, the eksctl code for SSM resolver does not support Windows 1909 AMI. It only supports resolving WindowsServer2019CoreContainer or WindowsServer2019FullContainer AMI.

michaelbeaumont:
### Description
Closes #2321 

### Checklist
- [ ] Manually tested


pkg/ami/auto_resolver.go
```diff
@@ -42,7 +45,9 @@ func OwnerAccountID(imageFamily, region string) (string, error) {
 	switch imageFamily {
 	case api.NodeImageFamilyUbuntu1804:
 		return ownerIDUbuntu1804Family, nil
-	case api.NodeImageFamilyWindowsServer2019CoreContainer, api.NodeImageFamilyWindowsServer2019FullContainer:
+	case api.NodeImageFamilyWindowsServer2019CoreContainer,
+		api.NodeImageFamilyWindowsServer2019FullContainer,
+		api.NodeImageFamilyWindowsServer1909CoreContainer:
```
cPu1, line 50: Since this `case` clause has grown a bit, I think we should now use `isWindowsImage`.

pkg/nodebootstrap/userdata.go
```diff
@@ -193,7 +193,9 @@ func NewUserData(spec *api.ClusterConfig, ng *api.NodeGroup) (string, error) {
 		return NewUserDataForUbuntu1804(spec, ng)
 	case api.NodeImageFamilyBottlerocket:
 		return NewUserDataForBottlerocket(spec, ng)
-	case api.NodeImageFamilyWindowsServer2019FullContainer, api.NodeImageFamilyWindowsServer2019CoreContainer:
+	case api.NodeImageFamilyWindowsServer2019FullContainer,
+		api.NodeImageFamilyWindowsServer2019CoreContainer,
+		api.NodeImageFamilyWindowsServer1909CoreContainer:
```
cPu1, line 198: Same as above, I think we should use `isWindowsImage` here.

pkg/ami/auto_resolver.go
michaelbeaumont: fixed, `IsWindowsImage` now in `default` case

michaelbeaumont:
@cPu1 I had to update the test but I changed them to not be dependent on the exact AMI names, what do you think?

cPu1:
>@cPu1 I had to update the test but I changed them to not be dependent on the exact AMI names, what do you think?


Makes sense. We are going to eventually deprecate static resolution for AMIs, so those tests will be removed as well.

PR #2362, opened 2020-06-22
Update tags when calling StackCollection.UpdateStack()

### LINKED ISSUE #2361: StackCollection.UpdateStack() does not update tags
**What happened?**
eksctl does not update tags of a stack (and its resources) when `StackCollection.UpdateStack()` is being called.

To have `StackCollection.UpdateStack()` apply modified clusterconfig's `metadata.tags` when being run.
Note that existing tags should be merged with eksctl's tags in order not to lose any tags applied by someone else.

**Versions**
eksctl commit 6538a82f (current master)

rndstr:
Closes #2361 

When calling StackCollection.UpdateStack() it will now also update the
value of its tags (shared tags as defined in clusterconfig metadata.tags).

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -157,6 +157,12 @@ func (c *StackCollection) CreateStack(name string, stack builder.ResourceSet, ta
 func (c *StackCollection) UpdateStack(stackName, changeSetName, description string, templateData TemplateData, parameters map[string]string) error {
 	logger.Info(description)
 	i := &Stack{StackName: &stackName}
+	// Read existing tags
+	s, err := c.DescribeStack(i)
+	if err != nil {
+		return err
+	}
+	i.SetTags(s.Tags)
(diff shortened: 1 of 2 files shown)


pkg/cfn/manager/api.go
```diff
@@ -493,7 +499,9 @@ func (c *StackCollection) doCreateChangeSetRequest(i *Stack, changeSetName strin
 		StackName:     i.StackName,
 		ChangeSetName: &changeSetName,
 		Description:   &description,
+		Tags:          i.Tags,
```
sayboras, line 502: what do think about having _Tags:          append(i.Tags, c.sharedTags...),_ here directly ?

martina-if:
@rndstr Can you see if there is a similar unit test that can be copied/modified to add a regression test for this?

rndstr:
@martina-if i didn't find any unit tests for the `StockCollection` (requires fair amount of mocking) but i managed to add some for the `UpdateStack()`

PR #2372, opened 2020-06-25
Auto format example files for consistency

sayboras:
### Description

Auto format example files for consistency

### Checklist
- [ ] ~~Added tests that cover your change (if possible)~~
- [ ] ~~Manually tested~~

### DIFF
--- a/examples/01-simple-cluster.yaml
+++ b/examples/01-simple-cluster.yaml
@@ -13,8 +13,8 @@ nodeGroups:
     desiredCapacity: 1
 
 cloudWatch:
-    clusterLogging:
-        # enable specific types of cluster control plane logs
-        enableTypes: ["audit", "authenticator", "controllerManager"]
-        # all supported types: "api", "audit", "authenticator", "controllerManager", "scheduler"
-        # supported special values: "*" and "all"
+  clusterLogging:
+    # enable specific types of cluster control plane logs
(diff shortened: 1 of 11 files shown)


sayboras:
> Hmm might be a good idea to just check `examples/` with `prettier` (or something else? alternatives?) too.

I am relying no auto format code from Goland, but not sure which tool it used. I have tried [prettier](https://prettier.io/docs/en/cli.html) before, however, it didn't align comment well. Any suggestion or recommendation on tool ? Thanks

Goland
image

Prettier 
image

michaelbeaumont:
Hmm I'd prefer something universally available so that everyone can replicate the formatting

sayboras:
> Hmm I'd prefer something universally available so that everyone can replicate the formatting

@michaelbeaumont What could be the right place to add this prettier tool ? We can have the installation step in makefile or just add it in eksctl-build. Either way is not idea in my opinion as it will polute code base (not to mention that it's javascript). 

Another alternative option is to use gihub action for this purpose. However, it might prevent developer to run and check before commit.

michaelbeaumont:
I could imagine a pre-commit hook. Any other tool would be fine, just should be one that can be run by everyone.

sayboras:
Seems like it might take longer to discuss and most likely we will go with different approach compared to changes in this PR.

Close this PR and create one issue for tracking purpose.

PR #2373, opened 2020-06-25
Correct cert manager IAM polices

### LINKED ISSUE #2077: cert-manager addon instance permissions has an unneeded permission
**What happened?**

The instance role created by the cert-manager addon includes `route53:ListHostedZones`, however the cert-manager documentation does not list this as required.

The addon permissions match the upstream documentation, or a bug is filed against the upstream documentation.

**How to reproduce it?**

Inspect [eksctl source](https://github.com/weaveworks/eksctl/blob/32027355678915365c68053e9293900b951cbc04/pkg/cfn/builder/iam_helper.go#L55) and compare with [the upstream documentation](https://cert-manager.io/docs/configuration/acme/dns01/route53/#set-up-a-iam-role).

I did this while I was attempting to recreate the same permissions using IRSA, to make sure I was reading/parsing it all correctly.

_It'd be a super-nice feature so support something like the instance addons for IRSA accounts, by the way. Although presumably in that case, only one addon per ServiceAccount... Anyway, that's a different feature request for later._

I suspect this was overlooked during refactoring to add cert-manager support to the existing external-dns support, as they are currently intertwined in the code. external-dns *does* [document this permission as needed](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#iam-policy), and so I guess [when cert-manager was added](https://github.com/weaveworks/eksctl/pull/885), it was copied over by accident.

**Versions**

N/A

**Logs**

N/A

sayboras:
### Description

To correct cert manager IAM polices
Fixes #2077

### Checklist
- [ ] Manually tested

### DIFF
--- a/userdocs/src/usage/iam-policies.md
+++ b/userdocs/src/usage/iam-policies.md
@@ -34,6 +34,9 @@ example, a CI server that needs to push images to ECR.
 
 The `ebs` policy enables the new EBS CSI (Elastic Block Store Container Storage Interface) driver.
 
+### Cert Manager Policy
+The `certManager` policy enable the ability add records to Route53 in order to solve the DNS01 challenge. More information can be found [here](https://cert-manager.io/docs/configuration/acme/dns01/route53/#set-up-a-iam-role).
+
 [comment]: <> (TODO: One section per addon and brief explanation of what it is )
 
 ## Adding a custom instance role
--- a/userdocs/src/usage/iam-policies.md
+++ b/userdocs/src/usage/iam-policies.md
@@ -35,7 +35,7 @@ example, a CI server that needs to push images to ECR.
 The `ebs` policy enables the new EBS CSI (Elastic Block Store Container Storage Interface) driver.
 
 ### Cert Manager Policy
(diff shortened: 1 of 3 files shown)


userdocs/src/usage/iam-policies.md
```diff
@@ -34,6 +34,9 @@ example, a CI server that needs to push images to ECR.
 
 The `ebs` policy enables the new EBS CSI (Elastic Block Store Container Storage Interface) driver.
 
+### Cert Manager Policy
+The `certManager` policy enable the ability add records to Route53 in order to solve the DNS01 challenge. More information can be found [here](https://cert-manager.io/docs/configuration/acme/dns01/route53/#set-up-a-iam-role).
```
cPu1, line 38: Minor nit: 
```suggestion
The `certManager` policy enables the ability to add records to Route 53 in order to solve the DNS01 challenge. More information can be found [here](https://cert-manager.io/docs/configuration/acme/dns01/route53/#set-up-a-iam-role).
```

sayboras: opps :), thanks :100:

Issue #1765, opened 2020-01-27
Provide a mechanism in eksctl to deploy a fully private EKS/VPC environment 


mreferre:
*Before creating a feature request, please search existing feature requests to see if you find a similar one. If there is a similar feature request please up-vote it and/or add your comments to it instead*

**Why do you want this feature?**
A lot of Enterprise customers (especially in the financial sector but not limited to that) want to deploy fully private EKS environments (with no Internet connectivity at-all). These environments (including the VPC hosting them) should not have external connectivity but should allow for hybrid scenarios (i.e. consuming both the EKS service as well as workloads running on it from on-prem via a VPN connection). 

**What feature/behavior/change do you want?**
Ideally a new flag (or a set of flags) that, when used together, achieve the result above. 

Today people are implementing this pattern using CloudFormation, bash scripts or a combination of those (e.g. [this is an example](https://github.com/jpbarto/private-eks-cluster)). 

There are a few existing issues that cover some of this (e.g. [1230](https://github.com/weaveworks/eksctl/issues/1230) and [1358](https://github.com/weaveworks/eksctl/issues/1358)) but the idea about this feature request is to go a step further and not only deploy an EKS cluster that is accessible from within the VPC but create a complete self-contained specific end-to-end setup (including the VPC itself) that does not have Internet connectivity at-all.

henrybravo:
+1 for this, will be very much appreciated by many users

sayboras:
definitely this feature will help to check all security boxes. Just want to share my experience, we eventually goes with custom resource to do all `post cluster` creation steps. However, one more issue still needs to be addressed https://github.com/aws/containers-roadmap/issues/607 (not related to eksctl though). Maybe this is something you could help with @mreferre ?

mreferre:
Thanks @sayboras for chiming in. Yes #607 is being worked on.

kleemae:
+1

maxbraun:
+1

martina-if:
Related information in #80 and #1230

martina-if:
Start by making a proposal on how we want to tackle this

michaelbeaumont:
https://github.com/weaveworks/eksctl/issues/747 is potentially relevant

mreferre:
> Start by making a proposal on how we want to tackle this

Sorry Martina was this a comment to me? I have provided some background in the ticket but let me know if you want/need me to expand.

martina-if:
[...]

PR #2384, opened 2020-06-30
Implement creation of fully-private clusters

cPu1:
### Description
Adds support for creating fully-private clusters.

To create a fully-private cluster, a top-level field `privateCluster` is added to `ClusterConfig`:

```yaml
privateCluster:
  enabled: true
  additionalEndpointServices:
  - "autoscaling"
```

`privateCluster.enabled` enables creation of a fully-private cluster. When this field is set, eksctl creates no public subnets, but only private subnets that have no outbound internet access (i.e., no NAT gateway or NAT instance). To enable worker nodes to access AWS services privately, eksctl creates VPC endpoints for the following services:
- Interface endpoints for ECR (both ecr.api and ecr.dkr) to pull container images (AWS CNI plugin etc)
- A gateway endpoint for S3 to pull the actual image layers
- An interface endpoint for EC2 because the aws-cloud-provider integration accesses it
- An interface endpoint for STS to support Fargate and IAM Roles for Services Accounts (IRSA)
- An interface endpoint for CloudWatch logging (`logs`) if CloudWatch logging is enabled

These VPC endpoints are essential for a functional private cluster, and as such, eksctl does not support configuring or disabling them. However, a cluster might need to access other AWS services privately (e.g., Autoscaling required by the Cluster Autoscaler). These services can be specified in `privateCluster.additionalEndpointServices`, which will instruct eksctl to create a VPC endpoint for each of them. The list of additional endpoint services that can be enabled for private access will be limited to a few services initially.

`privateNetworking` for both self-managed and EKS-managed nodegroups must be set explicitly. It's  an error to leave `privateNetworking` unset, as it's not possible to create public nodegroups in a fully-private cluster as there's no public subnet. Rather than enabling this field by default for a fully-private cluster, we require users to explicitly set it to make the behaviour explicit and avoid confusion.

```yaml
nodeGroups:
- name: ng1
  instanceType: m5.large
  desiredCapacity: 2
  # privateNetworking must be explicitly set for a fully-private cluster
  # Rather than defaulting this field to true for a fully-private cluster, we require users to explicitly set it
  # to make the behaviour explicit and avoid confusion.
  privateNetworking: true
```

Furthermore, it is an error to set either `clusterEndpoints.publicAccess` or `clusterEndpoints.privateAccess`, as a fully-private cluster can have private access only, and allowing modification of these fields can break the cluster.

```yaml
vpc:
  clusterEndpoints:
#    It is an error to set either of the following two fields, as a fully-private cluster defaults to private-only access
#    publicAccess: true
#    privateAccess: false
```

Combining the above snippets, a `ClusterConfig` for a fully-private cluster would look like this:

```yaml
kind: ClusterConfig

metadata:
  name: private-cluster
  region: us-west-2

privateCluster:
  enabled: true
  additionalEndpointServices:
  - "autoscaling"


nodeGroups:
- name: ng1
  instanceType: m5.large
  desiredCapacity: 2
  # privateNetworking must be explicitly set for a fully-private cluster
  # Rather than defaulting this field to true for a fully-private cluster, we require users to explicitly set it
  # to make the behaviour explicit and avoid confusion.
  privateNetworking: true

managedNodeGroups:
- name: m1
  instanceType: m5.large
  desiredCapacity: 2
  privateNetworking: true
```

It's important to note that the initial implementation will have a limitation where eksctl will initially create the cluster with both public and private endpoints access enabled, because eksctl needs access to the Kubernetes API server to allow self-managed nodes to join the cluster and to support GitOps and Fargate. After these operations have completed, eksctl will switch the cluster endpoint access to private-only. This additional update does mean that creation of a fully-private cluster will take longer than for a standard cluster. In the future, we may consider switching to a VPC-enabled Lambda function to perform these API operations.

https://github.com/weaveworks/eksctl/issues/1765#issuecomment-648161235

Closes https://github.com/weaveworks/eksctl/issues/1765

### Checklist
- [x] Manually tested


pkg/cfn/builder/vpc_endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+			endpoint.SubnetIds = e.subnetsForAZs(endpointDetail.AvailabilityZones)
+			endpoint.PrivateDnsEnabled = gfn.NewBoolean(true)
+			endpoint.SecurityGroupIds = []*gfn.Value{e.clusterSharedSG}
+		}
+
+		resourceName := fmt.Sprintf("VPCEndpoint%s", strings.ToUpper(
+			strings.ReplaceAll(endpointDetail.Service, ".", ""),
+		))
+
+		// TODO attach policy document
```
martina-if, line 83: Reminder comment

userdocs/src/usage/eks-private-cluster.md
```diff
@@ -0,0 +1,147 @@
...
+The only required field to create a fully-private cluster is `privateCluster.enabled`:
+
+```yaml
+privateCluster:
+  enabled: true
+```
+
+## Configuring private access to additional AWS services
+
+To enable worker nodes to access AWS services privately, eksctl creates VPC endpoints for the following services:
```
martina-if, line 20: Reminder to add the information about the cloudformation endpoint here

userdocs/src/usage/eks-private-cluster.md
```diff
@@ -0,0 +1,147 @@
...
+managedNodeGroups:
+- name: m1
+  instanceType: m5.large
+  desiredCapacity: 2
+  privateNetworking: true
+
+```
+
+## Cluster Endpoint Access
+A fully-private cluster does not support modifying `clusterEndpointAccess` during cluster creation.
```
martina-if, line 73: Is there a check in commands like `eksctl update/upgrade cluster` and `eksctl utils update-cluster-endpoints` that throw an error if this is set?

cPu1: There's no check in `eksctl utils update-cluster-endpoints`. I have left it like that because I think if users still want to change the endpoint access, it should be possible. 

Why do we need a check in `eksctl updqate/upgrade cluster`?

martina-if: in `upgrade cluster` eksctl reconstructs the stacks and looks for differences before applying changes so I am not sure what will happen if someone changes these values and then performs an upgrade. Have you checked?

userdocs/src/usage/eks-private-cluster.md
```diff
@@ -0,0 +1,158 @@
...
+It is an error to set either `clusterEndpoints.publicAccess` or `clusterEndpoints.privateAccess`, as a fully-private cluster
+can have private access only, and allowing modification of these fields can break the cluster.
+
+
+## User-supplied VPC and subnets
+eksctl supports creation of fully-private clusters using a pre-existing VPC and subnets. Only private subnets can be
+specified and it's an error to specify subnets under `vpc.subnets.public`.
+
+eksctl creates VPC endpoints in the supplied VPC and modifies route tables for the supplied subnets. Each subnet should
+have an explicit route table associated with it because eksctl does not modify the main route table.
```
martina-if, line 93: Should we add something like 
```
!!! warning
    Enabling the cloudformation endpoint service is required for managing the cluster using eksctl after its creation
```
?

pkg/ctl/create/nodegroup.go
```diff
@@ -256,7 +243,12 @@ func doCreateNodeGroups(cmd *cmdutils.Cmd, ng *api.NodeGroup, params createNodeG
 
 		for _, ng := range cfg.ManagedNodeGroups {
 			if err := ctl.WaitForNodes(clientSet, ng); err != nil {
-				return err
+				if cfg.PrivateCluster.Enabled {
+					logger.Info("error waiting for nodes to join the cluster; this command was likely run from outside the cluster's VPC as the API server is not reachable, nodegroup(s) should still be able to join the cluster, underlying error is: %v", err)
```
martina-if, line 247: should it be a warning instead?

userdocs/src/usage/eks-private-cluster.md
cPu1: The `clusterEndpointAccess` is set via an API call and not via CloudFormation. Also, the changes that are applied in `upgrade cluster` are only for adding new resources like the resources for supporting managed nodegroups.

pkg/ctl/create/nodegroup.go
cPu1: I went with `info` because we already check if it's a private cluster.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -560,7 +564,6 @@ func NewClusterVPC() *ClusterVPC {
 		},
 		NAT:              DefaultClusterNAT(),
 		AutoAllocateIPv6: Disabled(),
-		ClusterEndpoints: ClusterEndpointAccessDefaults(),
```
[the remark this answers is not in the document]
cPu1: Yeah, it broke some tests. I have pushed a fix.

pkg/cfn/builder/vpc_endpoint.go
cPu1: I'm planning to address this in a separate PR.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -486,6 +486,9 @@ type ClusterConfig struct {
 	// +optional
 	VPC *ClusterVPC `json:"vpc,omitempty"`
 
+	// +optional
+	PrivateCluster *PrivateCluster `json:"privateCluster,omitempty"`
```
martina-if, line 490: Would be nice to document this field right away

cPu1: Good point. Addressed in the latest commit.

JulianWgs:
Very cool feature! What is the ETA of the release? :)

cPu1:
> Very cool feature! What is the ETA of the release? :)

@JulianWgs it'll be out in this Friday's RC.

JulianWgs:
Thank you! I‘m looking forward to it :)

Issue #2368, opened 2020-06-24
Add T3 unlimited support


efim-a-efim:
**Why do you want this feature?**
AWS has a special pricing feature for T-type instances called "T3 Unlimited". It allows T3 instances to use CPU cycles outside of their ECU limit (of course, for additional cost). Eksctl does not add this option to instance groups with T3 instances.

**What feature/behavior/change do you want?**
Eksctl should have a special option in instance group configuration (something like `cpuCredits: "standard|unlimited"`). If this option is specified and nodegroup's instance types include `t3.*`, a [`CpuCredits` parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-creditspecification.html) should be added to LaunchTemplate

michaelbeaumont:
Closed by #2391

PR #2391, opened 2020-07-01
Added T3 Unlimited support

efim-a-efim:
### Description

Added support of [T3 Unlimited](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-unlimited-mode.html) flag for nodegroups. See #2368

This sets the flag:
```yaml
nodeGroups:
- name: ng1:
  instancesDistribution:
    instanceTypes:
      - t3.large
      - t3a.large
  cpuCredits: "unlimited"
```

To disable this functionality, use `cpuCredits: "standard"`.

### Checklist
- [ ] Manually tested


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -698,6 +698,9 @@ type NodeGroup struct {
 	// +optional
 	EBSOptimized *bool `json:"ebsOptimized,omitempty"`
 
+	// T3Unlimited enables [T3 Unlimited](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-unlimited-mode.html), valid only for T-type instances
+	T3Unlimited *bool `json:"t3Unlimited,omitempty"`
```
michaelbeaumont, line 702: I would definitely prefer to have it called `CPUCreditsUnlimited` or similar. @martina-if @cPu1

efim-a-efim: Yes, I called it `t3Unlimited` only because it will be easier to google.

martina-if: Or even `cpuCredits: string`

pkg/cfn/builder/api_test.go
```diff
@@ -2933,6 +2939,94 @@ var _ = Describe("CloudFormation template builder API", func() {
...
+		roundtrip()
+
+		It("should have correct resources and attributes", func() {
+			Expect(getLaunchTemplateData(ngTemplate).CreditSpecification).ToNot(BeNil())
+			Expect(getLaunchTemplateData(ngTemplate).CreditSpecification.CpuCredits).ToNot(BeNil())
+			Expect(getLaunchTemplateData(ngTemplate).CreditSpecification.CpuCredits).To(Equal("unlimited"))
+		})
+	})
+
+	Context("NodeGroup{T3Unlimited=true InstancesDistribution.InstanceTypes=m5.large}", func() {
```
martina-if, line 3005: We should prevent this combination earlier on during the validation of the yaml file so that there are no unexpected subtle changes like changing the value of a field internally.  Please take a look at `validation.go` and `validation_test.go` to see how to do this.

efim-a-efim: Thanks, will take a look.

efim-a-efim: Fixed in latest commit

pkg/apis/eksctl.io/v1alpha5/types.go
efim-a-efim: Yes, `cpuCredits: string` looks best of all. I'll change this after stabilization if main functionality.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -236,6 +236,12 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
 		return err
 	}
 
+	if ng.CPUCredits != nil {
```
michaelbeaumont, line 239: sorry big nitpick, this check should go inside `validateCPUCredits` IMO, so it's always called like `validateInstancesDistribution` above.

pkg/cfn/builder/nodegroup.go
```diff
@@ -267,15 +269,37 @@ func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_La
...
 	}
 
+	// Add T3 Unlimited setting only if nodegroup has T-type instances
+	if n.spec.CPUCredits != nil {
+		if isTInstance {
+			launchTemplateData.CreditSpecification = &gfn.AWSEC2LaunchTemplate_CreditSpecification{
+				CpuCredits: gfn.NewString(strings.ToLower(*n.spec.CPUCredits)),
+			}
+		} else {
+			logger.Warning("cpuCredits option ignored, nodegroup %s has no T3 instance types", n.nodeGroupName)
```
martina-if, line 299: At this point this configuration should have been validated so I would avoid checking this again here. Let's check it only during the validation phase and trust that if cpuCredits is set it should be set in the template too

pkg/cfn/builder/nodegroup.go
```diff
@@ -251,6 +251,8 @@ func (n *NodeGroupResourceSet) GetAllOutputs(stack cfn.Stack) error {
 }
 
 func newLaunchTemplateData(n *NodeGroupResourceSet) *gfn.AWSEC2LaunchTemplate_LaunchTemplateData {
+	isTInstance := false
```
martina-if, line 254: Let's skip the use of this variable at all. If `n.spec.CPUCredits` was set then here we just trust that it has been validated before and set `CpuCredits` accordingly.

efim-a-efim: OK, agree.

joergrech:
Should this feature already work in eksctl 0.24.0-rc.1 with AWS Kubernetes? I'm getting the following error 

`Error: cpuCredits option set for nodegroup, but it has no t2/t3 instance types`

when executing

`eksctl create nodegroup -f ./kubernetes/EKS/eksctl-nodegroups.yaml`

with the following content for ./kubernetes/EKS/eksctl-nodegroups.yaml:
 
```
kind: ClusterConfig

metadata:
  name: <CLUSTER_NAME>
  region: eu-central-1

nodeGroups:
  - name: small-spot-nodes2
    availabilityZones: ["eu-central-1b"]
    labels:
      pool:     small-spot-nodes2
      disksize: small
      nodesize: small
      nodetype: spot
    instancesDistribution:
      instanceTypes: ["t3.small","t3a.small"]
      onDemandBaseCapacity: 0
      onDemandPercentageAboveBaseCapacity: 0
      spotAllocationStrategy: "lowest-price"
      spotInstancePools: 2
    cpuCredits: "standard"
    minSize: 1
    maxSize: 20
    desiredCapacity: 1
    volumeSize: 20
    iam:
      withAddonPolicies:
        autoScaler: true
```

NOTE: my goal is to prevent t2/t3 spot instances to generate additional CPUCredits (the default seems to be "unlimited")
NOTE2: it does not matter if the nodegroup exists or not (tested by changing the name)
NOTE3: `cpuCredits` does not work for managedNodeGroups (`error unmarshaling JSON: while decoding JSON: json: unknown field "cpuCredits"`)
NOTE4: I upgraded my Cluster from 1.16 to 1.17 today via the AWS EKS control plane

michaelbeaumont:
The default certainly shouldn't have changed with this PR.

michaelbeaumont:
@joergrech Can you try adding `instanceType: mixed` to the nodegroup?

joergrech:
OK, that worked with nodeGroups - not sure how/if this can work with managedNodeGroups. I'll check tomorrow if this is really set in AWS.
Thanks!

joergrech:
Seems to work - unlimited is disabled in EC2
Btw. "unlimited" is set as default from AWS side - see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-standard-mode-concepts.html
`T3 and T3a instances are launched as unlimited by default. T2 instances are launched as standard by default`

michaelbeaumont:
Yeah I saw that too, thanks @joergrech!

PR #2383, opened 2020-06-30
Implement make command to merge release branches back to master

martina-if:
It will maintain the master version of the file pkg/version/release.go
which contains eksctl hardcoded version. This is done through the file
.gitattributes and configuring a dummy merge strategy for it.

- [x] Manually tested
- [ ] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)
- [x] Make sure the title of the PR is a good description that can go into the release notes

Issue #1284, opened 2019-09-05
automate cherry-picking commit


errordeveloper:
Now that we have release branches, we need to cherry-pick changes from master to such branches.

It's tedious and error-prone, plus there are different ways to go about it. We should automate it. It would be nice to say `/cherry-pick release-0.2` in a PR comment, and have that PR cherry-picked once it get merged.

errordeveloper:
Some of challenges that I face are:

- having to decide if you should cherry-pick the merge commit vs individual commits
- missing commits, or cherry-picking in wrong order
- merging instead of cherry-picking (I've done it at least once by accident)
- making changes on release branch and cherry-picking back to master

marccarre:
Cherry-picked changes are:
- harder to track
    - they have different commit hashes,
    - they aren't listed by GitHub (unlike merged changes), e.g. https://github.com/weaveworks/eksctl/commit/5a97e9a6ea18ea001b2281196d2d373d520004ce and https://github.com/weaveworks/eksctl/commit/94e3555b6327fe38bdd9d38fef87a26644cc967f only appear in their respective branches even though they're identical,
- easier to get wrong (easy to forget one of the commits, easy to cherry-pick in the wrong order, etc.),

so I'd rather keep cherry-pick for exceptional cases and use simple `git merge`s instead.
This is a lot cleaner, a lot more tractable, and a much more robust workflow in my opinion.

martina-if:
I agree with that, although once you have a RC you should only include urgent bug fixes and nothing 
else. So any other feature that was merged into master afterwards shouldn't be in the actual release.
For that case I see two options:
  - cherry picking the bug fixes into the release branch 
  - commit bug fixes to the rc branch and then merge that branch into master

What do you think?

marccarre:
> cherry picking the bug fixes into the release branch

This I'd rather avoid, because of the aforementioned downsides of `cherry-pick`.

> commit bug fixes to the rc branch and then merge that branch into master

This works too, although is less granular than what I suggested above.
Therefore we'd have to be sure we want to merge the entire `release-X.Y` back to `master`.

Note that an equivalent approach to what I suggested above, if preferred, could be:

- merging the PR to `release-X.Y`, 
- **not** deleting the feature branch immediately,
- merging the feature branch to `master`,
- deleted the feature branch.

martina-if:
[...]

PR #2392, opened 2020-07-01
Add github workflow to merge release branch to master

### LINKED ISSUE #2336: Don't commit changes to pkg/version/release.go in release branches
This prevents merging release branches back into master.

~We could embed version information into the executable with some other method~
Use merge driver to ignore changes to the file from the release branch

michaelbeaumont:
### Description

closes #2336

[Tested this a bunch already](https://github.com/michaelbeaumont/eksctl/pull/5), it will automatically open a PR for every push to `release-*`. There was talk of a command (`/cherry` in a comment) in https://github.com/weaveworks/eksctl/issues/1284 but I don't think this is necessary.

This would make the workflow (the hotfix, git flow based method):
1. fix in a `release` branch
1. open and merge a PR to the `release` branch
1. wait for automatic PR to `master`
    - it's rather the exception that we _don't_ want to merge it to `master` and in that case, we just close the PR.

What this doesn't support is fixing something on `master` and merging it to one or more `release` branches (like the `/cherry 0.2.0 0.3.0 0.4.0` comment method would support)

Do you @martina-if @cPu1 think this makes sense?

Only thing left is that since the PR is opened by a github action, it [won't trigger the github actions usually run on a PR](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token). The way around this is to use a real account to open the PR. There doesn't seem to be a `secrets.GITHUB_TOKEN` on this repo atm. 
I think the best way would be to create a Github App for the repo and use it similarly to [how this action suggests doing it](https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens).

This can use `make merge-release-branch-back` from #2383 as well.

### Checklist
- [x] Manually tested

### DIFF
--- a/.github/workflows/release-merge.yaml
+++ b/.github/workflows/release-merge.yaml
@@ -1,0 +1,38 @@
+name: Merge release changes to master
+on:
+  push:
+    branches:
+      - release-*
+env:
+  VERSION_FILE: pkg/version/release.go
+  DEFAULT_BRANCH: master
+jobs:
+  merge_release:
+    name: Merge release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Merge release branch into ${{env.DEFAULT_BRANCH}}
+        run: |
+          git checkout $DEFAULT_BRANCH
+          git checkout -b merge-${{ github.sha }}
+          git config user.name "weaveworksbot"
+          git config user.email "weaveworksbot@users.noreply.github.com"
+          git merge --no-commit --no-ff ${{ github.ref }}
+          git checkout ORIG_HEAD -- $VERSION_FILE
+          git add $VERSION_FILE
+          EDITOR=true git merge --continue
+          git push --set-upstream origin HEAD
+      - uses: actions/github-script@v2.0.0
+        name: Open PR to ${{env.DEFAULT_BRANCH}}
+        with:
+          github-token: ${{ secrets.WEAVEWORKSBOT_TOKEN }}
+          script: |
+            await github.pulls.create({
+              ...context.repo,
+              title: `Merge ${context.ref} to ${{ env.DEFAULT_BRANCH }}`,
+              head: `merge-${context.sha}`,
+              base: "${{ env.DEFAULT_BRANCH }}",
+            });


.github/workflows/release-merge.yaml
```diff
@@ -0,0 +1,39 @@
...
+          fetch-depth: 0
+      - name: Merge release branch into ${{env.DEFAULT_BRANCH}}
+        run: |
+          git checkout $DEFAULT_BRANCH
+          git checkout -b merge-${{ github.sha }}
+          # This info comes from https://api.github.com/users/github-actions%5Bbot%5D
+          # This way we make it look like github-actions[bot] made the commits
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git merge --no-commit --no-ff ${{ github.ref }}
```
sayboras, line 25: so it might be the case that we have multiple PRs to master and the later will contain changes of previous as well ?

sayboras: just want to clarify the flow, if there is the commit you don't want to merge to master, basically, as mentioned, you will close it. But then subsequent PR might also contain the changes of un-want commit.

michaelbeaumont: Yeah that's a good point. In that case we just have to rebase the branch ourselves from then on. But I think that should really be quite the exception, but maybe I'm missing a scenario

sayboras: @michaelbeaumont I am not sure either :sob:, sorry to bringing up subject which I don't fully understand.

One related note on releasing process (not related to this PR) is that I haven't seen any the patch fix (e.g. 0.22.1). Do we have supporting EOL policy as well (e.g. last 2 releases only or something like that)? One common approach is _backporting_, all commits will go to _master_, PR can be tagged with label for automating backport to specific release branch. There are plenty of available github actions and apps to support this workflow.

michaelbeaumont: Yeah, that's what I mentioned in the description. We don't really backport beyond the current RC/release at the moment. That's why this workflow works at all.

I personally prefer this since it keeps people up to date. I think it's something we have to consider when/if we introduce breaking changes.

PR #2409, opened 2020-07-09
Enable bodyclose linter in golangci-lint config

sayboras:
### Description

As per godoc, the caller is responsible to close body.

	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body.

bodyclose lint is also added to avoid similar cases in future.

Signed-off-by: Tam Mach <sayboras@yahoo.com>

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/iam/oidc/api.go
+++ b/pkg/iam/oidc/api.go
@@ -132,7 +132,7 @@ func (m *OpenIDConnectManager) getIssuerCAThumbprint() error {
 	if err != nil {
 		return errors.Wrap(err, "connecting to issuer OIDC")
 	}
-
+	defer response.Body.Close()
 	if response.TLS != nil {
 		if numCerts := len(response.TLS.PeerCertificates); numCerts >= 1 {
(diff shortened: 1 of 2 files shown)


pkg/iam/oidc/api.go
```diff
@@ -132,7 +132,11 @@ func (m *OpenIDConnectManager) getIssuerCAThumbprint() error {
 	if err != nil {
 		return errors.Wrap(err, "connecting to issuer OIDC")
 	}
-
+	defer func() {
+		if response != nil {
```
michaelbeaumont, line 136: just curious, can it be that `response` is `nil` _and_ `err` is `nil`?

sayboras: you are right, that scenario is not happened. Otherwise we will have nil pointer in next line already `if response.TLS != nil`

Let me remove this nil check :)

PR #2414, opened 2020-07-10
Fix cluster upgrading to default version always

martina-if:
The upgrade/update cluster commands where always trying to upgrade to the default version when
one was not specified by the user.

```
eksctl upgrade cluster  martina-test-1-13
[ℹ]  eksctl version 0.24.0-dev+2c97a450.2020-07-10T08:09:08Z
[ℹ]  using region eu-north-1
Error: upgrading more than one version at a time is not supported. Found upgrade from "1.13" to "1.16". Please upgrade to "1.14" first
```

```
eksctl upgrade cluster martina-test-1-13
[ℹ]  eksctl version 0.24.0-dev+19f9b371.2020-07-10T09:30:41Z
[ℹ]  using region eu-north-1
[ℹ]  (plan) would upgrade cluster "martina-test-1-13" control plane from current version "1.13" to "1.14"
[ℹ]  re-building cluster stack "eksctl-martina-test-1-13-cluster"
[✔]  all resources in cluster stack "eksctl-martina-test-1-13-cluster" are up-to-date
[ℹ]  checking security group configuration for all nodegroups
[ℹ]  all nodegroups have up-to-date configuration
[!]  no changes were applied, run again with '--approve' to apply the changes
```

The problem is that when we create a new ClusterConfig object we set the version to the default version. Then this
value is used as a default for the `--version` flag.

- [x] Manually tested

### DIFF
--- a/pkg/ctl/update/cluster.go
+++ b/pkg/ctl/update/cluster.go
@@ -14,6 +14,8 @@ import (
 
 func updateClusterCmd(cmd *cmdutils.Cmd) {
 	cfg := api.NewClusterConfig()
+	// Reset version before loading
+	cfg.Metadata.Version = ""
 	cmd.ClusterConfig = cfg
 
 	cmd.SetDescription("cluster", "DEPRECATED: use 'upgrade cluster' instead. Upgrade control plane to the next version. ",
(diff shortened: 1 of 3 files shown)


michaelbeaumont:
Argh I don't like that there's a massive list of "Check PR labels" statuses when looking at the commit statuses and that they don't overwrite each other and there's an X because the first one failed :unamused:

Issue #2395, opened 2020-07-02
Eksctl creating worker nodes with too much reserved memory for kubelet


avasquez614:
We've been using `eksctl` for a while and it has been working wonderfully. However, a couple of months ago we've noticed that the nodes that `eksctl` was creating had too much `kubeReserved` memory. For example, just yesterday I created a 1.14 EKS cluster, with a single `m5.xlarge` worker, and this is what it has for `kubeReserved`:

```
"kubeReserved": {
    "cpu": "80m",
    "ephemeral-storage": "1Gi",
    "memory": "2662Mi"
  }
```

I had made some research before and I thought it was because of this issue with the AWS AMIs: https://github.com/awslabs/amazon-eks-ami/issues/387, but according to one of the devs that issue has already been resolved. So now I don't really know what's causing so much memory to be reserved for the kubelet. Any ideas?

avasquez614:
By the way this is the info on the node:

```
Kernel Version:             4.14.181-140.257.amzn2.x86_64  
Kube Proxy Version:         v1.14.9-eks-49202c 
Kubelet Version:            v1.14.9-eks-49202c 
Operating System:           linux
Os Image:                   Amazon Linux 2
```

brianpursley:
Are you thinking that the memory is calculation is wrong, or that the calculation is performing correctly, but that the amount it returns is higher than you think it should be?

I didn't do the math, but 2662Mi reserved on a 16Gi instance seems to match how the calculation is designed.

See these comments I found explaining how eksctl calculates the value:
https://github.com/weaveworks/eksctl/blob/0afe5f1ff23a87eda004662cd278a1e5d5f6b20a/pkg/nodebootstrap/reserved.go#L94-L104

And these test cases showing the calculated values for a few memory sizes.  There isn't a test case for 16Gi, but there is one for 12Gi and it reserves 2252Mi:
https://github.com/weaveworks/eksctl/blob/0afe5f1ff23a87eda004662cd278a1e5d5f6b20a/pkg/nodebootstrap/reserved_test.go#L30-L37

It looks like the algorithm eksctl uses matches what is in [amazon-eks-ami](https://github.com/awslabs/amazon-eks-ami/blob/ff690788dfaf399e6919eebb59371ee923617df4/files/bootstrap.sh#L150-L163)

brianpursley:
Oh., I think I see what you are saying.
[...]

PR #2443, opened 2020-07-15
Updated default kubelet reserved memory calculation to match the calculation used by amazon-eks-ami

brianpursley:
### Description

The reserved memory calculation was changed in amazon-eks-ami to be based on pod density
(11*maxPodsPerNode+255) instead of total memory.
See https://github.com/awslabs/amazon-eks-ami/blob/21426e27e3845319dbca92e7df32e5c4b984a1d1/files/bootstrap.sh#L163

This commit introduces a new InstanceTypeInfo property to store MaxPodsPerNode and populates it
using the data from maxpods.go so that it will be available to use when calculing the default memory
to reserve.

Because the image type specs (cpu, memory) come from an API call and the max number of pods per node
comes from a file in amaxon-eks-ami, there is a chance that an image type is returned by the API for
which there is no corresponding record in maxpods.go. To handle this, the original memory
calculation remains as a fallback that will be used if max pods per node is unknown.

Issue #2395

### Checklist
- [x] Manually tested
- [ 
] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)


pkg/nodebootstrap/reserved.go
```diff
@@ -93,17 +98,17 @@ func (i InstanceTypeInfo) DefaultStorageToReserve() string {
...
-// 6% of the next 114688Mi of memory (up to 131072Mi)
-// 2% of any memory above 131072Mi
+// See https://github.com/awslabs/amazon-eks-ami/blob/21426e27e3845319dbca92e7df32e5c4b984a1d1/files/bootstrap.sh#L154-L165
 func (i InstanceTypeInfo) DefaultMemoryToReserve() string {
-	mib := memProgression.calculate(i.Memory, minimumMemoryToReserve)
+	var mib int64
+	if i.MaxPodsPerNode > 0 {
+		mib = 11*i.MaxPodsPerNode + minimumMemoryToReserve
+	} else {
+		// As a fallback, if MaxPodsPerNode is 0, use the original calculation, which is a progressive calculation based on total memory.
```
michaelbeaumont, line 107: Do we need the fallback? I don't see any types with `MaxPodsPerNode: 0`

brianpursley: You probably saw this, but in the diff, it is collapsed. Some examples of instance types with no max pods are:
- m5ad.8xlarge
- m5ad.16xlarge

I wasn't really quite sure what to make of this, as these seem to not be available when I created an EC2 instance via the console. They don't exist in the list.

I was hesitant to completely remove the old calc in case those instance types need to be supported.

As a test, I just tried to create a cluster using m5ad.8xlarge, and it failed, so I guess those instance types won't work anyway, so I think you're probably right, the fallback is unnecessary and can be removed.

I'll update the PR to remove it.

michaelbeaumont: Ah yeah true, I was checking for a missing line `MaxPodsPerNode:`.

michaelbeaumont: Ok, those instances don't exist apparently (anymore?). Now I'm wondering why they show up at all.

brianpursley: I pushed a new update that removes the fallback and related field which is no longer used.

I also discovered a problem (with this new PR) where there are mixed instance types and it takes the minimum across all types.  I had not updated that branch of the code, so it was using the fallback, and the unit test still passed because it was using the fallback calc.

ptal and see what you think now.

Issue #2172, opened 2020-05-12
`eksctl delete cluster` should delete cluster and all associated resources


kalbir:
**Why do you want this feature?**
There are [numerous issues](https://github.com/weaveworks/eksctl/labels/area%2Fdeletions) that track cases where someone has run `eksctl delete cluster` and it has not resulted in the cluster and all of the resources created by `eksctl` being deleted.

This means that people are sometimes left with surprise bills, which is a really bad outcome. It also slows people down when they have to go and investigate why a particular cluster did not delete. 

Some specific examples: 
closed issues
- https://github.com/weaveworks/eksctl/issues/755
- https://github.com/weaveworks/eksctl/issues/1856
- https://github.com/weaveworks/eksctl/issues/1325
- https://github.com/weaveworks/eksctl/issues/536
- https://github.com/weaveworks/eksctl/issues/1849
- https://github.com/weaveworks/eksctl/issues/1653

(more under the [area/deletions](https://github.com/weaveworks/eksctl/labels/area%2Fdeletions) label)

**What feature/behavior/change do you want?**

Rather that trying to chase after all of the cases, is there some pattern we could introduce that aimed for an eventually consistent delete? i.e. could we retry deletions or target specific areas in a particular order to ensure that all of the resources will eventually get deleted.

aclevername:
https://github.com/weaveworks/eksctl/issues/1653 can get removed from the "open issues" section as the OIDC provider does not get deleted during `eksctl delete cluster`. The two remaining NG/ENI issue are getting worked on #536

Himangini:
Closing this off since all the related issues mentioned above are completed

PR #2465, opened 2020-07-22
Cleanup ALBs created by AWS ALB Ingress Controller

programmer04:
### Description

Howdy!

Inspired how `Services` of type `LoadBalancer` are clean up I implemented automated deletion of `Ingresses` in the same way. The need is mentioned in e.g. issue #536 or #2172.

With this change during

```bash
eksctl delete cluster --config-file=./test.yml
```

instead of

```log
```

now there is an updated message

```log
[ℹ]  cleaning up AWS Load Balancers created by Kubernetes Services and Ingresses
```

The cleanup logic for `Services` works as used to and for `Ingress` similar logic is implemented:

- If not annotated with `"kubernetes.io/ingress.class": "alb"` skip (the same as non-AWS load balancers created by `Service` are skipped).
- If annotated with `"kubernetes.io/ingress.class": "alb"` try to delete via Kubernetes API and watch if will end successfully (by checking AWS API). If not report an error.
For instance, if [AWS ALB Ingress Controller](https://github.com/kubernetes-sigs/aws-alb-ingress-controller) was deleted before, Kubernetes object `Ingress` is deleted successfully but load balancer retains. So `eksctl` is checking if ALB was deleted from AWS and in this case, it wouldn't be. After a timeout, it will fail and print the message with AWS names of retained load balancers (both from `Ingresses`  and `Services`) like below.
    ```
    Error: deadline surpassed waiting for AWS load balancers to be deleted: 3d60a73f-2048game-2048ingr-6fa0
    ```
- If [AWS ALB Ingress Controller](https://github.com/kubernetes-sigs/aws-alb-ingress-controller) is not fully reconciled, clean may be skipped, and deleting the whole cluster fails in the old way like without this improvement.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/elb/cleanup.go
+++ b/pkg/elb/cleanup.go
@@ -84,8 +84,10 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
 		if lb == nil {
 			continue
 		}
-		logger.Debug("tracking deletion of load balancer %s of kind %d with security groups %v",
-			lb.name, lb.kind, convertStringSetToSlice(lb.ownedSecurityGroupIDs))
+		logger.Debug(
+			"tracking deletion of load balancer %s of kind %d with security groups %v",
+			lb.name, lb.kind, convertStringSetToSlice(lb.ownedSecurityGroupIDs),
+		)
 		awsLoadBalancers[lb.name] = *lb
 		logger.Debug("deleting 'type: LoadBalancer' Service %s/%s", s.Namespace, s.Name)
 		if err := kubernetesCS.CoreV1().Services(s.Namespace).Delete(s.Name, &metav1.DeleteOptions{}); err != nil {
@@ -102,8 +104,10 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
 		if lb == nil {
 			continue
 		}
-		logger.Debug("tracking deletion of load balancer %s of kind %d with security groups %v",
-			lb.name, lb.kind, convertStringSetToSlice(lb.ownedSecurityGroupIDs))
+		logger.Debug(
+			"tracking deletion of load balancer %s of kind %d with security groups %v",
+			lb.name, lb.kind, convertStringSetToSlice(lb.ownedSecurityGroupIDs),
+		)
 		awsLoadBalancers[lb.name] = *lb
 		logger.Debug("deleting 'kubernetes.io/ingress.class: alb' Ingress %s/%s", i.Namespace, i.Name)
 		if err := kubernetesCS.NetworkingV1beta1().Ingresses(i.Namespace).Delete(i.Name, &metav1.DeleteOptions{}); err != nil {
@@ -131,13 +135,11 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
 			delete(awsLoadBalancers, name)
 		}
 	}
-	numLB := len(awsLoadBalancers)
-	if numLB > 0 {
-		lbs := make([]string, numLB)
-		var i int
+
+	if numLB := len(awsLoadBalancers); numLB > 0 {
+		lbs := make([]string, 0, numLB)
 		for name := range awsLoadBalancers {
-			lbs[i] = name
-			i++
+			lbs = append(lbs, name)
 		}
 		return fmt.Errorf("deadline surpassed waiting for AWS load balancers to be deleted: %s", strings.Join(lbs, ","))
 	}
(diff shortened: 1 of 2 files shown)


pkg/elb/cleanup.go
```diff
@@ -130,6 +171,38 @@ func getServiceLoadBalancer(ctx context.Context, ec2API ec2iface.EC2API, elbAPI
...
+	// if does not pass ALB hadn't been provisioned so nothing to return.
+	if len(ingress.Status.LoadBalancer.Ingress) == 0 ||
+		len(ingress.Status.LoadBalancer.Ingress[0].Hostname) == 0 {
+		logger.Debug("%s is ALB Ingress, but probably not provisioned, skip", ingress.Name)
+		return nil
+	}
+	// Expected e.g. bf647c9e-default-appingres-350b-1622159649.eu-central-1.elb.amazonaws.com where AWS ALB name is
+	// bf647c9e-default-appingres-350b (cannot be longer than 32 characters).
+	hostNameParts := strings.Split(ingress.Status.LoadBalancer.Ingress[0].Hostname, ".")
+	if len(hostNameParts) == 0 {
```
michaelbeaumont, line 191: I think `strings.Split` only returns an empty list if both the string and the separator are empty, this will never be true.

pkg/elb/cleanup.go
```diff
@@ -51,39 +57,68 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
...
+			return errors.New(errStr)
+		}
+	}
+	// For k8s Kind Ingress
+	for _, i := range ingresses.Items {
+		lb := getIngressLoadBalancer(ctx, i)
+		if lb == nil {
+			continue
+		}
+		logger.Debug("tracking deletion of load balancer %s of kind %d with security groups %v",
```
michaelbeaumont, line 105: This formatting feels wrong to me, i think we would add a line break after `(`?

michaelbeaumont: and also put the final `)` on a new line

pkg/elb/cleanup.go
```diff
@@ -93,15 +128,21 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
...
 			// The load balancer and its security groups have been deleted
-			delete(loadBalancers, name)
+			delete(awsLoadBalancers, name)
 		}
 	}
-
-	if len(loadBalancers) > 0 {
-		return fmt.Errorf("deadline surpased waiting for load balancers to be deleted")
+	numLB := len(awsLoadBalancers)
+	if numLB > 0 {
```
michaelbeaumont, line 135: nitpick, could limit the scope more, `if numLB := ...; numLB > 0 {`

pkg/elb/cleanup.go
```diff
@@ -93,15 +128,21 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
...
+		lbs := make([]string, numLB)
```
michaelbeaumont, line 136: You could just use `append` in the below loop if you use `make([[]string, 0, numLB)` or just get `i` from `for i, name := range..` (or use `0, numLB` and `i` togegther) or just skip `make` completely and use `append`.
Pick one of those :slightly_smiling_face:

programmer04: I can't take `i` from `range` because it is `map`, but I can rewrite to `append` 🙂

pkg/elb/cleanup.go
programmer04: Yes, you're right good catch, I want to ensure that 
```go
if len(hostNameParts[0]) == 0 
```

pkg/elb/cleanup.go
michaelbeaumont: ah right didn't see that, thanks!

pkg/elb/cleanup.go
```diff
@@ -93,15 +132,19 @@ func Cleanup(ctx context.Context, ec2API ec2iface.EC2API, elbAPI elbiface.ELBAPI
...
 	}
 
-	if len(loadBalancers) > 0 {
-		return fmt.Errorf("deadline surpased waiting for load balancers to be deleted")
+	if numLB := len(awsLoadBalancers); numLB > 0 {
+		lbs := make([]string, 0, numLB)
+		for name := range awsLoadBalancers {
+			lbs = append(lbs, name)
+		}
+		return fmt.Errorf("deadline surpassed waiting for AWS load balancers to be deleted: %s", strings.Join(lbs, ","))
```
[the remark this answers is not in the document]
programmer04: Oh I didn't manage to improve it before the merge 😅

PR #2485, opened 2020-07-27
Add workflow to auto merge PRs

michaelbeaumont:
### Description

With this PR, we can add labels `automerge` or `autosquash` to merge the PR when it becomes ready. Additionally, `autoupdate` will merge changes from the base branch into the PR.
One small limitation is that the workflow isn't triggered when a PR becomes out of date, it will only update it if another trigger happens and it sees it's out of date.

### Checklist
- [ ] Manually tested


michaelbeaumont:
Hmm nevermind, the action is essentially useless. There needs to be a `GITHUB_TOKEN` set because workflow-triggered actions can't trigger other workflows (i.e. the post merge checks won't run) and `GITHUB_TOKEN` isn't available in forks. :(
https://github.com/pascalgn/automerge-action/issues/46#issuecomment-657220884

PR #2487, opened 2020-07-27
Added ASG SuspendProcesses support

### LINKED ISSUE #1355: Suspend ASG processes
When using a node group spanning multiple availability zones and using cluster autoscaler, there can be problems due to a native ASG process rebalancing nodes between AZ. AWS provides a way to suspend this process called [AZRebalance](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-suspend-resume-processes.html).

I'd like a way to provide a list of ASG processes to suspend when creating a node group. For example, here is how [kops](https://github.com/kubernetes/kops/blob/master/docs/instance_groups.md#suspending-scaling-processes-on-aws-autoscaling-groups) does it.

efim-a-efim:
### Description
Closes #1355 

If an ASG is created across multiple AZs, AWS by default tries to balance instance counts to be equal/optimal for each AZ. To do this, AWS stops instances and launches new ones. This behavior can be changed by suspending `AZRebalance` process in ASG.

This adds [ASG SuspendedProcesses](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_SuspendProcesses.html) nodegroup parameter support. This parameter allows to suspend any ASG process.

Sorry, but I could not create tests that will correctly cover this feature.

### Checklist
- [ ] Manually tested


martina-if:
Hi @efim-a-efim Thank you for the PR! I took a quick look and it look good. It would be good to have tests for it. I think they could be added to [api_test.go](https://github.com/weaveworks/eksctl/blob/master/pkg/cfn/builder/api_test.go).

bpaquet:
This feature only suspend the process during the UPDATE of the ASG. This does not set the Suspend process for the ASG. 
There is no way to do that through cloudformation, so it's not simple from Eksctl.

Cloud formation ASG: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html

PR #2527, opened 2020-08-10
Fix formatting for private cluster docs

weaveworksbot:


userdocs/src/usage/eks-private-cluster.md
```diff
@@ -63,22 +63,22 @@ any public subnets. The `privateNetworking` field (`nodeGroup[*].privateNetworki
...
- ```yaml
+```yaml
 nodeGroups:
 - name: ng1
   instanceType: m5.large
   desiredCapacity: 2
   # privateNetworking must be explicitly set for a fully-private cluster
-  # Rather than defaulting this field to `true` for a fully-private cluster, we require users to explicitly set it
-  # to make the behaviour explicit and avoid confusion.
+  # Rather than defaulting this field to `true`,
```
martina-if, line 72: there's a part missing "for a fully-private cluster"

michaelbeaumont: I removed it because it was redundant given the previous sentence and that it's in the private cluster docs

PR #2528, opened 2020-08-11
Handle delete cluster with failed stack

michaelbeaumont:
### Description
Part of #2172 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


pkg/eks/eks.go
```diff
@@ -195,7 +200,6 @@ func (c *ClusterProvider) CanOperate(spec *api.ClusterConfig) (bool, error) {
 	case awseks.ClusterStatusCreating, awseks.ClusterStatusDeleting, awseks.ClusterStatusFailed:
 		return false, fmt.Errorf("cannot perform Kubernetes API operations on cluster %q in %q region due to status %q", spec.Metadata.Name, spec.Metadata.Region, status)
 	default:
-		// all other states are considered operable, including UPDGRADING (which is missing from the SDK)
```
cPu1, line 198: The part about the `UPGRADING` state missing in the SDK was useful in this comment.

michaelbeaumont: Is it still missing? I see an Updating state

michaelbeaumont: https://github.com/aws/aws-sdk-go/blob/d85f343adb9b56f552867c2fa558d60f43a02242/service/eks/api.go#L6653-L6654

cPu1: It was about the `UPGRADING` state. I think the original comment might've actually meant the `UPDATING` state, as when a cluster is being upgraded, the status is `UPDATING`. So I think we're good to remove this comment.

Issue #813, opened 2019-05-22
Import eksctl as a library to integrate other utils


Jeffwan:
**Why do you want this feature?**
I am using eksctl to build an eks provision tool in kubeflow community. In my previous version, we use eksctl as a cli in shell scripts to create eks cluster. Now, community is moving to golang ctl and we need to migrate existing functionalities implemented by shell to golang.

I don't know if eksctl can be used for this purpose. I notice no one package import eksctl yet. 
https://www.grank.io/pkg/github.com/weaveworks/eksctl.html

**What feature/behavior/change do you want?**
Need eksctl SDK to create/delete cluster.

gemagomez:
This would be good to have, and we could use it for testing purposes also.

zWaR:
Is there perhaps an ETA for release of this feature?

austince:
Is there any opportunity to contribute to this effort? We'd love to use this for creating tooling for internal cluster setup that requires more than within the scope of eksctl (ie VPC peering), not just because Go is the community standard.

martina-if:
related issue: https://github.com/weaveworks/eksctl/issues/1538

jrake-revelant:
is there any plan to work on this topic or anyone actively working on it?

michaelbeaumont:
Remaining:
- [ ] [Incorrect use](https://github.com/golang/go/issues/32776#issuecomment-505607726) of `k8s.io/kubernetes` (see [previous attempt to fix](https://github.com/weaveworks/eksctl/pull/2316))

binchenX:
I would pick eksctl any time of the day if we have a sdk so that we can get ride of the scripts needed to integrate with the eksctl.

related to: 
https://github.com/weaveworks/eksctl/issues/3839

rverma-nsl:
@errordeveloper can you please help us here, this looks like a must have now.

cPu1:
> @errordeveloper can you please help us here, this looks like a must have now.

@rverma-nsl  which eksctl operations are you trying to invoke and from what language?

rverma-nsl:
almost all of them, top ones are create cluster, add nodegroup, add irsa roles, patch cluster upgrades and addons.

Skarlso:
Why not just call these and abstract the binary calling operations behind an interface for example? How would you benefit from calling these things from code vs calling out to the binary?

rverma-nsl:
[...]

PR #2537, opened 2020-08-13
Refactor nodegroup creation

michaelbeaumont:
### Description
Part of #813 

### Checklist
- [ ] Manually tested


cPu1 (approved):

pkg/actions/create/nodegroup.go
```diff
@@ -0,0 +1,307 @@
...
+					logger.Info("error waiting for nodes to join the cluster; this command was likely run from outside the cluster's VPC as the API server is not reachable, nodegroup(s) should still be able to join the cluster, underlying error is: %v", err)
+					break
+				} else {
+					return err
+				}
+			}
+		}
+
+		logger.Success("created %d managed nodegroup(s) in cluster %q", len(cfg.ManagedNodeGroups), cfg.Metadata.Name)
+	}
```
aclevername, line 225: this is a really long function, can some of the logic get separated out into funcs? The parts where we create a new scope (not sure what the proper term is?) in particular look like good places to move the logic into a descriptive function

pkg/actions/create/nodegroup.go
```diff
@@ -0,0 +1,307 @@
...
+	cmdutils.LogRegionAndVersionInfo(meta)
+
+	if err := ctl.CheckAuth(); err != nil {
+		return err
+	}
+
+	if ok, err := ctl.CanOperate(cfg); !ok {
+		return err
+	}
+
```
aclevername, line 67: I've seen these bits of logic all over the place, not suggesting as part of this PR but we should consider trying to move this into some shared step that occurs before a commands logic gets invoked. Something analogous to a [middleware](https://github.com/gorilla/mux#middleware)

pkg/actions/create/nodegroup.go
michaelbeaumont: done

pkg/actions/create/nodegroup.go
michaelbeaumont: yeah for sure, there's a lot of duplication here

aclevername (approved):
LGTM

PR #2610, opened 2020-09-01
Support for managed VPC controller

michaelbeaumont:
### Checklist
- [ ] Manually tested


pkg/cfn/builder/iam.go
```diff
@@ -71,16 +72,21 @@ func (c *ClusterResourceSet) addResourcesForIAM() {
 
 	c.rs.withIAM = true
 
+	managedPolicyArns := []string{
+		iamPolicyAmazonEKSClusterPolicy,
+	}
+	if !api.IsDisabled(c.spec.IAM.VPCResourceControllerPolicy) {
```
martina-if, line 78: nitpick: might read better as `if api.IsEnabled(...) {`

martina-if (approved):

pkg/cfn/builder/iam.go
michaelbeaumont: Yeah this always bothers me because they have different semantics for `nil`. We set defaults in `defaults.go` but if that's ever changed then this code breaks. (Definitely something that can be refactored since the code doesn't care about `nil` at this point but it's still exposed in the types). So I kind of played it safe here but I'm not sure.

cPu1 (approved):

PR #2661, opened 2020-09-22
Deprecate static AMI resolver

neha-viswanathan:
closes #2650 

### Description

This MR removes all references to static AMI resolver.

### Checklist
- [ ] Added tests that cover your change (if possible)
- [x] Added/modified documentation as required (such as the `README.md`, or the `userdocs` directory)
- [ ] Manually tested
- [ ] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)
- [x] Make sure the title of the PR is a good description that can go into the release notes

PR #2710, opened 2020-10-05
Deprecate static AMI resolver

neha-viswanathan:
### Description

This MR contains a warning in the command description of `create nodegroup` to warn about the deprecation of the static AMI resolver.
Related to #2661 

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -275,7 +275,7 @@ func ResolveAMI(provider api.ClusterProvider, version string, ng *api.NodeGroup)
 	case api.NodeImageResolverAutoSSM:
 		resolver = ami.NewSSMResolver(provider.SSM())
 	case api.NodeImageResolverStatic:
-		logger.Warning("'static' value for node-ami flag is deprecated and will be removed with release 0.33.0. Valid values will be 'auto-ssm' (default), 'auto' or an AMI id (advanced use)")
+		logger.Warning("'static' value for node-ami flag (nodeGroup.ami field) is deprecated and will be removed with release 0.33.0. Valid values will be 'auto-ssm' (default), 'auto' or an AMI id (advanced use)")
 		resolver = ami.NewStaticResolver()
 	default:
 		resolver = ami.NewMultiResolver(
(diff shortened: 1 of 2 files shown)


neha-viswanathan:
@michaelbeaumont @martina-if please review. Thanks!

pkg/ctl/create/nodegroup.go
```diff
@@ -44,7 +44,7 @@ func createNodeGroupCmdWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *cmdutils
 
 	cfg.Metadata.Version = "auto"
 
-	cmd.SetDescription("nodegroup", "Create a nodegroup", "", "ng")
```
michaelbeaumont, line 47: I don't think the warning should be shown here, because it's about a very specific option.

michaelbeaumont (changes requested):
I think a warning should be printed when we see the user is using the resolver here:
https://github.com/weaveworks/eksctl/blob/eff895ad956766480fb916de910e49036af15807/pkg/eks/api.go#L277-L278

pkg/ctl/create/nodegroup.go
michaelbeaumont: Please also mention release `0.33.0` as the deprecation point.

michaelbeaumont (approved):

pkg/eks/api.go
```diff
@@ -275,6 +275,7 @@ func ResolveAMI(provider api.ClusterProvider, version string, ng *api.NodeGroup)
 	case api.NodeImageResolverAutoSSM:
 		resolver = ami.NewSSMResolver(provider.SSM())
 	case api.NodeImageResolverStatic:
+		logger.Warning("'static' value for node-ami flag is deprecated and will be removed with release 0.33.0. Valid values will be 'auto-ssm' (default), 'auto' or an AMI id (advanced use)")
```
cPu1, line 278: We should also specify the field name:
```suggestion
		logger.Warning("'static' value for node-ami flag (nodeGroup.ami field) is deprecated and will be removed with release 0.33.0. Valid values will be 'auto-ssm' (default), 'auto' or an AMI id (advanced use)")
```

cPu1 (approved):

PR #2718, opened 2020-10-08
print formatted output for utils describe-stacks

### LINKED ISSUE #1495: Offer `utils describe-stacks` output in pure JSON
I'd like to be able to use `describe stacks` output to check or extract the state of `eksctl` cluster.
Right now the output is a hybrid/mix of plain text logging and JSON and so can't be parsed as JSON without pre-processing. And if you have to do that you might as well use the `aws cli` directly instead. So right now, for me, the `utils` comment is not very useful. But if it did output JSON, it would be quite useful (to me). JSON output would also enable JSON diff to track changes to a cluster.

I would like this to work:
`eksctl utils describe-stacks --name=mycluster | jq <expression>`
or this:
`eksctl utils describe-stacks --name=mycluster --output json | jq`

I can work around this by describing the stacks with the `aws cli` so this is just a wish list item. Just to say I think this utility would be more useful with a pure JSON output option.

neha-viswanathan:
Closes #1495 

### Description

This MR enables the use of a flag `output` for the `eksctl utils describe-stacks` command for JSON or YAML formatted output.

### How to run

`eksctl utils describe-stacks --name=mycluster --output json`

or

`eksctl utils describe-stacks --name=mycluster --output yaml`

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/utils/describe_stacks.go
+++ b/pkg/ctl/utils/describe_stacks.go
@@ -1,13 +1,17 @@
 package utils
 
 import (
+	"os"
+
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/kris-nova/logger"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/printers"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
-	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
 )
 
 func describeStacksCmd(cmd *cmdutils.Cmd) {
@@ -15,12 +19,32 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
 	cmd.ClusterConfig = cfg
 
 	var all, events, trail bool
+	var output printers.Type
 
 	cmd.SetDescription("describe-stacks", "Describe CloudFormation stack for a given cluster", "")
 
 	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
 		cmd.NameArg = cmdutils.GetNameArg(args)
-		return doDescribeStacksCmd(cmd, all, events, trail)
+		var printer printers.OutputPrinter
+		var err error
+		if output != "" {
+			if cmd.CobraCommand.Flags().Changed("all") ||
+				cmd.CobraCommand.Flags().Changed("events") ||
+				cmd.CobraCommand.Flags().Changed("trails") {
+				return errors.Errorf("since the output flag is specified, the flags `all`, `events` and `trail` cannot be used")
+			}
+		}
+		switch output {
+		case printers.TableType:
+			return errors.Errorf("output type %q is not supported", output)
+		case "":
+		default:
+			printer, err = printers.NewPrinter(output)
+			if err != nil {
+				return err
+			}
+		}
+		return doDescribeStacksCmd(cmd, all, events, trail, printer)
 	}
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
@@ -29,20 +53,23 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
 		fs.BoolVar(&all, "all", false, "include deleted stacks")
 		fs.BoolVar(&events, "events", false, "include stack events")
 		fs.BoolVar(&trail, "trail", false, "lookup CloudTrail events for the cluster")
+		fs.StringVarP(&output, "output", "o", "", "specifies the output formats (valid option: json and yaml)")
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
-func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
+func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, printer printers.OutputPrinter) error {
 	cfg := cmd.ClusterConfig
 
 	ctl, err := cmd.NewCtl()
 	if err != nil {
 		return err
 	}
-	cmdutils.LogRegionAndVersionInfo(cfg.Metadata)
+	if printer == nil {
+		cmdutils.LogRegionAndVersionInfo(cfg.Metadata)
+	}
 
 	if err := ctl.CheckAuth(); err != nil {
 		return err
@@ -69,6 +96,10 @@ func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
 
 	if len(stacks) < 2 {
 		logger.Warning("only %d stacks found, for a ready-to-use cluster there should be at least 2", len(stacks))
+	}
+
+	if printer != nil {
+		return printer.PrintObj(stacks, os.Stdout)
 	}
 
 	for _, s := range stacks {


pkg/ctl/utils/describe_stacks.go
```diff
@@ -1,26 +1,38 @@
...
 		cmd.NameArg = cmdutils.GetNameArg(args)
-		return doDescribeStacksCmd(cmd, all, events, trail)
+		switch output {
+		case "":
+			return doDescribeStacksCmd(cmd, all, events, trail, false)
+		case "json":
+			return doDescribeStacksCmd(cmd, all, events, trail, true)
+		default:
+			return fmt.Errorf("unknown output format: %s", output)
+		}
```
cPu1, line 35: It should be straightforward to add support for YAML as well (the `table` output type doesn't make sense here).

pkg/ctl/utils/describe_stacks.go
```diff
@@ -71,6 +86,14 @@ func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
 		logger.Warning("only %d stacks found, for a ready-to-use cluster there should be at least 2", len(stacks))
 	}
 
+	if json {
```
cPu1, line 89: We should either support the `all`, `events` and `trail` options for JSON (and YAML) output types, or add validation to disallow their usage when `--output` is specified.

cPu1 (commented):
I left a few minor comments but otherwise LGTM. 
Thanks for the PR!

pkg/ctl/utils/describe_stacks.go
neha-viswanathan: @cPu1 which one do you prefer?

pkg/ctl/utils/describe_stacks.go
```diff
@@ -1,26 +1,40 @@
...
+		switch output {
+		case "":
+			return doDescribeStacksCmd(cmd, all, events, trail, "")
+		case "yaml":
+			return doDescribeStacksCmd(cmd, all, events, trail, "yaml")
+		case "json":
+			return doDescribeStacksCmd(cmd, all, events, trail, "json")
+		default:
+			return fmt.Errorf("unknown output format: %s", output)
+		}
```
cPu1, line 37: You can construct a new `OutputPrinter` here and pass it to `doDescribeStacksCmd` where a `nil` value indicates default behaviour.  This validation is already performed in `printers.NewPrinter(output)`, but it accepts `table` so you'll have to validate for it. 

```suggestion

      var printer printers.OutputPrinter
		switch output {
		case printers.TableType:
			return errors.Errorf("output type %q is not supported", output)

		case "":

		default:
			var err error
			printer, err = printers.NewPrinter(output)
			if err != nil {
				return err
			}
		}
		return doDescribeStacksCmd(cmd, all, events, trail, printer)
```

pkg/ctl/utils/describe_stacks.go
```diff
@@ -1,26 +1,40 @@
...
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
-	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
 )
 
 func describeStacksCmd(cmd *cmdutils.Cmd) {
 	cfg := api.NewClusterConfig()
 	cmd.ClusterConfig = cfg
 
 	var all, events, trail bool
+	var output string
```
cPu1, line 22: To support the proposed changes, this should be:

```suggestion
	var output printers.Type
```

pkg/ctl/utils/describe_stacks.go
```diff
@@ -29,20 +43,28 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
...
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
-func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
+func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, output string) error {
+	if output != "" && (all || events || trail) {
+		logger.Warning("Since the output flag is specified, the flags `all`, `events` and `trail` (if set) will be ignored.")
```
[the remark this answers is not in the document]
neha-viswanathan: @cPu1 I was thinking of setting `DisableFlagParsing` to `true`. Do you think this is ok, or do you have a better suggestion?

go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	github.com/weaveworks/goformation/v4 v4.10.2-0.20200921152539-943d7e97b454
 	github.com/weaveworks/launcher v0.0.2-0.20200715141516-1ca323f1de15
 	github.com/whilp/git-urls v0.0.0-20191001220047-6db9661140c0
-	golang.org/x/tools v0.0.0-20200502202811-ed308ab3e770
+	golang.org/x/tools v0.0.0-20201013015553-ec925d8b1dc6
```
michaelbeaumont, line 56: any reason you've updated this? it will require a new docker base image

michaelbeaumont (dismissed):

go.mod
neha-viswanathan: @michaelbeaumont I updated the Go modules while testing locally, I will revert this change if it's a major change.

pkg/ctl/utils/describe_stacks.go
```diff
@@ -71,6 +97,13 @@ func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
 		logger.Warning("only %d stacks found, for a ready-to-use cluster there should be at least 2", len(stacks))
 	}
 
+	if printer != nil && output == printers.JSONType {
+		return printers.NewJSONPrinter().PrintObj(stacks, os.Stdout)
+	}
+	if printer != nil && output == printers.YAMLType {
+		return printers.NewYAMLPrinter().PrintObj(stacks, os.Stdout)
+	}
```
cPu1, line 105:
```suggestion
```suggestion
	if printer != nil {
		return printer.PrintObj(stacks, os.Stdout)
	}
```
```

File: pkg/ctl/utils/describe_stacks.go [hunk 8]
```diff
@@ -29,20 +46,29 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
...
+		fs.StringVarP(&output, "output", "o", "", "specifies the output formats (valid option: json and yaml)")
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
-func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
+func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, printer printers.OutputPrinter, output printers.Type) error {
+	if output != "" && (all || events || trail) {
```
cPu1, line 57: We should do the validation in `RunE`, and using `flag.Changed` (or `CobraCommand.Flags.Changed()`) is more accurate.

pkg/ctl/utils/describe_stacks.go
```diff
@@ -29,20 +46,29 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
...
+		logger.Warning("Since the output flag is specified, the flags `all`, `events` and `trail` (if set) will not be used.")
+		cmd.CobraCommand.DisableFlagParsing = true
+		os.Exit(3)
```
cPu1, line 60: `os.Exit` is not required here, just returning an error will handle exiting the command with a non-zero status.
 
(Also, as mentioned above, this should be in `RunE`)

neha-viswanathan:
@cPu1 thanks a lot for your suggestions. I appreciate your guidance.

pkg/ctl/utils/describe_stacks.go
```diff
@@ -29,20 +53,23 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
...
 		fs.BoolVar(&trail, "trail", false, "lookup CloudTrail events for the cluster")
+		fs.StringVarP(&output, "output", "o", "", "specifies the output formats (valid option: json and yaml)")
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
-func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool) error {
+func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, printer printers.OutputPrinter, output printers.Type) error {
```
cPu1, line 63: Since output is no longer used:
```suggestion
func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, printer printers.OutputPrinter) error {
```

pkg/ctl/utils/describe_stacks.go
```diff
@@ -1,26 +1,50 @@
...
 	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
 		cmd.NameArg = cmdutils.GetNameArg(args)
-		return doDescribeStacksCmd(cmd, all, events, trail)
+		var printer printers.OutputPrinter
+		var err error
+		if output != "" {
+			if cmd.CobraCommand.Flags().Changed("all") ||
+				cmd.CobraCommand.Flags().Changed("events") ||
+				cmd.CobraCommand.Flags().Changed("trails") {
+				return errors.Errorf("since the output flag is specified, the flags `all`, `events` and `trail` (if set) will not be used")
```
cPu1, line 34: Minor nit:
```suggestion
				return errors.Errorf("since the output flag is specified, the flags `all`, `events` and `trail` cannot be used")
```

cPu1 (commented):
Thanks for making the proposed changes :tada: . I've suggested some minor improvements and the PR is good to merge after they're resolved.

michaelbeaumont (approved):

Issue #2743, opened 2020-10-15
Add flag to enable Simple System Manager (SSM) for easy SSH access


aclevername:
**Why do you want this feature?**
I want to be able to easily SSH onto a worker node in order to be able to debug issues, for example when bootstrapping fails. Using [SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/what-is-systems-manager.html) is the recommended way to ssh onto nodes https://aws.github.io/aws-eks-best-practices/hosts/#minimize-access-to-worker-nodes. This removes the need for me to manually manage SSH keys like you currently do with the `--ssh-access` flag.

**What feature/behavior/change do you want?**
A new flag/option when creating the clusters to enable SSM on the nodes so that I can ssh onto the nodes using SSM.

E.g. `eksctl create cluster --enable-ssm`

Related issue: https://github.com/weaveworks/eksctl/issues/148

aclevername:
Closing as the desired functionality got merged as part of #2754

PR #2754, opened 2020-10-19
Add flag to enable Simple System Manager (SSM) for easy SSH access

aclevername:
### Description
This PR add the optional to enable [Simple Systems Manager (SSM)](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html) by passing the `--enable-ssm` flag. Once SSM is installed you can easily ssh onto nodes in the cluster by running `aws ssm start-session --target <node-id> --region <region>`

#### Manually Tested
- [X] AL2
- [X] Ubuntu
  - Ubuntu images already have SSM pre-installed, only the IAM policy needs to be added for it to work (see https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html)
- [x] Windows
    - Windows images already have SSM pre-installed, only the IAM policy needs to be added for it to work (see https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html)

Related issue https://github.com/weaveworks/eksctl/issues/2743

### Checklist
- [x] Manually tested


pkg/nodebootstrap/userdata_al2.go
```diff
@@ -54,6 +54,12 @@ func NewUserDataForAmazonLinux2(spec *api.ClusterConfig, ng *api.NodeGroup) (str
 
 	var scripts []string
 
+	if ng.SSH.EnableSSM != nil && *ng.SSH.EnableSSM {
```
michaelbeaumont, line 57: Are we sure this is run before the kubelet stuff?

aclevername: @michaelbeaumont I figured that as part of debugging failures that occur when bootstrapping having SSM installed first lets the users SSH on immediately, if for example the kubelet fails to start they won't be able to SSH on to debug

pkg/cfn/builder/iam.go
```diff
@@ -167,7 +169,11 @@ func (n *NodeGroupResourceSet) addResourcesForIAM() error {
 		n.rs.withNamedIAM = true
 	}
 
-	if err := createRole(n.rs, n.clusterSpec.IAM, n.spec.IAM, false); err != nil {
+	var enableSSM bool
+	if n.spec.SSH != nil && n.spec.SSH.EnableSSM != nil {
+		enableSSM = *n.spec.SSH.EnableSSM
+	}
```
cPu1, line 175:
```suggestion
  enableSSM := n.spec.SSH != nil && api.IsEnabled(n.spec.SSH.EnableSSM)
```

pkg/ctl/cmdutils/nodegroup_flags.go
```diff
@@ -37,6 +37,7 @@ func AddCommonCreateNodeGroupFlags(fs *pflag.FlagSet, cmd *Cmd, ng *api.NodeGrou
 
 	ng.SSH.Allow = fs.Bool("ssh-access", *ng.SSH.Allow, "control SSH access for nodes. Uses ~/.ssh/id_rsa.pub as default key path if enabled")
 	ng.SSH.PublicKeyPath = fs.String("ssh-public-key", "", "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
+	ng.SSH.EnableSSM = fs.Bool("enable-ssm", false, "Enable simple systems manager (SSM)")
```
cPu1, line 40: According to [this doc](https://docs.aws.amazon.com/systems-manager/latest/userguide/what-is-systems-manager.html), it was formerly known as Amazon Simple Systems Manager (SSM), now it's just AWS Systems Manager.

```suggestion
	ng.SSH.EnableSSM = fs.Bool("enable-ssm", false, "Enable AWS Systems Manager (SSM)")
```

userdocs/src/introduction.md
```diff
@@ -119,6 +119,14 @@ eksctl create cluster --ssh-access --ssh-public-key=my_kubernetes_key --region=u
 
 ```
 
+To use [AWS Simple Systems Manager (SSM)](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-sessions-start.html#sessions-start-cli) to SSH onto nodes, you can specify the `--enable-ssm` flag:
```
cPu1, line 122:
```suggestion
To use [AWS Systems Manager (SSM)](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-sessions-start.html#sessions-start-cli) to SSH onto nodes, you can specify the `--enable-ssm` flag:
```

cPu1 (changes requested):
I have left a few comments but otherwise LGTM.

One thing that's left out though is the SSM agent is not being installed for managed nodegroups. This should be handled in `pkg/cfn/builder/managed_launch_template.go` in `makeUserData`. Note that if a user specifies a custom launch template, we should disallow the `ssh.enableSsm` option.

pkg/nodebootstrap/userdata_al2.go
michaelbeaumont: Sorry I meant, is it clear whether or not the `addCommand` stuff ends up being run before the bootstrap script

aclevername: Oh sorry, I'm not sure. I saw that the `config.AddShellCommand` is used to add `PreBootstrapCommands`, so I assumed that meant it does run before the `bootstrap` script. Looking into it I do think adding it first does mean that the command runs before the bootstrap, as `addFilesAndScripts` just adds the file as an asset and appends the script as a command to the array of commands to run

pkg/cfn/builder/iam.go
aclevername: Much cleaner, Thanks @cPu1

pkg/ctl/cmdutils/nodegroup_flags.go
aclevername: good spot!

aclevername:
> I have left a few comments but otherwise LGTM.
Thanks for the comments @cPu1 , some great spots!
 
> One thing that's left out though is the SSM agent is not being installed for managed nodegroups. This should be handled in `pkg/cfn/builder/managed_launch_template.go` in `makeUserData`. Note that if a user specifies a custom launch template, we should disallow the `ssh.enableSsm` option.

Oops! Good spot. I'll update it to handle managed nodegroups and update the docs to note the behaviour for custom launch templates. Thanks 😄

pkg/nodebootstrap/userdata_al2.go
michaelbeaumont: OK, I see `runcmd` runs only on first boot. I personally think it's cleaner to add the commands as a self-contained script vs calling `AddShellCommand`x3. As far as I can tell, we just need to call `RunScript` and make sure it has a name that comes earlier than `bootstrap.al2.sh` (although I don't like how `RunScript` calls both `AddScript` and `AddCommand`).

aclevername: Sure, I was torn between a new script vs adding commands so happy to refactor it over to a script 😄

pkg/cfn/builder/managed_nodegroup.go
```diff
@@ -50,7 +50,12 @@ func (m *ManagedNodeGroupResourceSet) AddAllResources() error {
 
 	var nodeRole *gfnt.Value
 	if m.nodeGroup.IAM.InstanceRoleARN == "" {
-		if err := createRole(m.resourceSet, m.clusterConfig.IAM, m.nodeGroup.IAM, true); err != nil {
+		var enableSSM bool
+		if m.nodeGroup.SSH != nil && m.nodeGroup.SSH.EnableSSM != nil {
+			enableSSM = *m.nodeGroup.SSH.EnableSSM
+		}
```
cPu1, line 56:
```suggestion
		enableSSM := m.nodeGroup.SSH != nil && api.IsEnabled(m.nodeGroup.SSH.EnableSSM)
```

cPu1 (commented):
Support for managed nodegroups looks good :+1: 

We shouldn't ignore `enableSsm` if a launch template is specified, but rather disallow its usage.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -981,6 +982,8 @@ type (
 		PublicKeyName *string `json:"publicKeyName,omitempty"`
 		// +optional
 		SourceSecurityGroupIDs []string `json:"sourceSecurityGroupIds,omitempty"`
+		// +optional
```
michaelbeaumont, line 985: Here you could add a link to `introduction#ssh-access` similar to https://github.com/weaveworks/eksctl/blob/9e9e1ee8454fed0f8c2c343b4bb4b15243cad29d/pkg/apis/eksctl.io/v1alpha5/types.go#L567

cPu1 (approved):
:tada:

michaelbeaumont (approved):
:rocket:

Issue #2687, opened 2020-09-29
Failed nodegroup creation causes `get nodegroup` to fail


michaelbeaumont:
The same happens when failed creating the nodegroup.  if it's failed once you have to manually goto cloudformation to delete that stack. otherwise eksctl will keep looking into that failed stack which will preventing other normal operation.

```
eksctl get nodegroup --cluster abcd
Error: getting nodegroup stack summaries: mapping stack to nodegroup summary: error collecting Cloudformation outputs for stack eksctl-abcd-nodegroup-ng-workers-2: no output "InstanceRoleARN" in stack "eksctl-abcd-nodegroup-ng-workers-2"
```

_Originally posted by @chimerab in https://github.com/weaveworks/eksctl/issues/1967#issuecomment-700604257_

michaelbeaumont:
@michaelbeaumont 

at least below two versions have same problem.

eksctl version
0.27.0
0.28.1

First I try to create nodegorup with "eksctl create nodegroup -f xxxx-ng-workers-2.yaml" the stack failed because the volume size is smaller than image. 

[ℹ]  deploying stack "eksctl-abcd-nodegroup-ng-workers-2"
[✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-abcd-nodegroup-ng-workers-2"
[ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
[!]  AWS::EC2::LaunchTemplate/NodeGroupLaunchTemplate: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupEgress/EgressInterClusterAPI: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupEgress/EgressInterCluster: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupIngress/IngressInterClusterCP: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupIngress/IngressInterClusterAPI: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupIngress/IngressInterCluster: DELETE_IN_PROGRESS
[!]  AWS::EC2::SecurityGroupIngress/SSHIPv4: DELETE_IN_PROGRESS
[...]

PR #2762, opened 2020-10-22
Improving handling of failed nodegroups

aclevername:
### Description
Currently when a nodegroup fails to create you are unable to `get` or `delete` the nodegroup. Related issue: https://github.com/weaveworks/eksctl/issues/2687

This PR changes it so that:
- `get nodegroups` handles nodegroups that are unhealthy gracefully, by adding a new `Status` column to indicate the status of the cloudformation stack.
- `delete nodegroups` warns about errors that have occurred rather than exiting and hints where errors can be overcome

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/cfn/manager/nodegroup.go
+++ b/pkg/cfn/manager/nodegroup.go
@@ -28,6 +28,7 @@ type NodeGroupSummary struct {
 	StackName           string
 	Cluster             string
 	Name                string
+	Status              string
 	MaxSize             int
 	MinSize             int
 	DesiredCapacity     int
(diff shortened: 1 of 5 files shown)


pkg/ctl/delete/nodegroup.go
```diff
@@ -136,7 +142,10 @@ func doDeleteNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, updateAuthConfigMap
 		if !cmd.Plan {
 			for _, ng := range allNodeGroups {
 				if err := drain.NodeGroup(clientSet, ng, ctl.Provider.WaitTimeout(), maxGracePeriod, false); err != nil {
-					return err
+					if !force {
+						return err
+					}
```
cPu1, line 147: I think we should only ignore drain errors if the nodegroup's CloudFormation stack is not in a normal state. It's still possible for `drain.NodeGroup` to return an error if the nodegroup is healthy, and we should propagate that error and stop the deletion process.

aclevername: Makes sense! I'll change the logic

aclevername: Added a hint instead to suggest they can use the `--drain=false` flag to skip the error

pkg/ctl/create/nodegroup.go
```diff
@@ -198,7 +198,7 @@ func doCreateNodeGroups(cmd *cmdutils.Cmd, ng *api.NodeGroup, params createNodeG
 		errs := tasks.DoAllSync()
 		if len(errs) > 0 {
 			logger.Info("%d error(s) occurred and nodegroups haven't been created properly, you may wish to check CloudFormation console", len(errs))
-			logger.Info("to cleanup resources, run 'eksctl delete nodegroup --region=%s --cluster=%s --name=<name>' for each of the failed nodegroup", cfg.Metadata.Region, cfg.Metadata.Name)
+			logger.Info("to cleanup resources, run 'eksctl delete nodegroup --region=%s --cluster=%s --name=<name>' for each of the failed nodegroup. '--force' flag may be required", cfg.Metadata.Region, cfg.Metadata.Name)
```
michaelbeaumont, line 201: i think these two mentions of `--force` can be removed

pkg/cfn/manager/nodegroup.go
```diff
@@ -398,22 +404,11 @@ func (c *StackCollection) mapStackToNodeGroupSummary(stack *Stack, ngPaths *node
 		}
 		collectorSet := outputs.NewCollectorSet(collectors)
 		if err := collectorSet.MustCollect(*stack); err != nil {
-			return nil, errors.Wrapf(err, "error collecting Cloudformation outputs for stack %s", *stack.StackName)
+			logger.Info(errors.Wrapf(err, "error collecting Cloudformation outputs for stack %s", *stack.StackName).Error())
```
michaelbeaumont, line 407: maybe `logger.Warning`?

pkg/ctl/create/nodegroup.go
aclevername: oops! Thanks

pkg/cfn/manager/nodegroup.go
aclevername: 👍

michaelbeaumont (approved):

PR #2766, opened 2020-10-23
Refactor drain logic

aclevername:
### Description
- Backfill unit tests for drain logic
- Refactor complexity of drain function
- Make it clear when we are just undoing the cordoning

### Discussion
- I've kept the `retryDelay`, but am questioning wether its worth keeping or not. In the happy path scenari we will just continually hit the API to drain all nodes until it completes with no delay at all. Perhaps rather than having a `retryDelay` that only occurs when we error, we have a general delay between loops to prevent us from hammering the API too much?
- I've introduced a new tool for generating mocks [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter). If we are against introducing more tools I'm happy to modify this to use the Mockery tool, but I feel that `counterfeiter` is slightly more powerful in that it creates the mock functions for you, rather than having to set mock values using generic interfaces like in Mockery.
- Introduced `export_test.go` as a pattern for setting private fields. This is an alternative to making unnecessary fields public (or configurable via New constructor). https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd

### Checklist
- [ ] Manually tested


pkg/drain/evictor/pod.go
```diff
@@ -29,41 +29,41 @@ import (
 const (
 	daemonSetFatal      = "DaemonSet-managed Pods (use --ignore-daemonsets to ignore)"
 	daemonSetWarning    = "ignoring DaemonSet-managed Pods"
-	localStorageFatal   = "Pods with local storage (use --delete-local-data to override)"
+	localStorageFatal   = "Pods with local storage (use --Delete-local-data to override)"
```
michaelbeaumont, line 32: I think this slipped in

pkg/drain/evictor/pod.go
```diff
@@ -29,41 +29,41 @@ import (
...
 	localStorageWarning = "deleting Pods with local storage"
 	unmanagedFatal      = "Pods not managed by ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet (use --force to override)"
 	unmanagedWarning    = "deleting Pods not managed by ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet"
 
-	drainPodAnnotation       = "pod.alpha.kubernetes.io/drain"
+	drainPodAnnotation       = "Pod.alpha.kubernetes.io/drain"
```
michaelbeaumont, line 37: and here, might be others

pkg/drain/evictor/filters.go
```diff
@@ -158,7 +158,7 @@ func hasLocalStorage(pod corev1.Pod) bool {
 	return false
 }
 
-func (d *Helper) annotationFilter(pod corev1.Pod) podDeleteStatus {
+func (d Helper) annotationFilter(pod corev1.Pod) podDeleteStatus {
```
michaelbeaumont: What's the most idiomatic way to do this? I know only that most of the code uses pointers and using value here I think might cause unexpected bugs if someone decides to mutate the passed object later on?

pkg/drain/evictor/pod.go
aclevername: Goland rename variable leaking into some places it shouldn't, my bad!

pkg/drain/evictor/pod.go
aclevername: Goland rename variable leaking into some places it shouldn't, my bad!

pkg/drain/evictor/filters.go
aclevername: Good spot, changing to a pointer

pkg/eks/nodegroup.go
```diff
@@ -186,6 +186,7 @@ func LogWindowsCompatibility(nodeGroups []KubeNodeGroup, clusterMeta *api.Cluste
 	}
 }
 
+//go:generate /Users/jake/workspace/go/bin/mockery -name=KubeNodeGroup -output=mocks/
```
michaelbeaumont, line 189: Use `${GOBIN}` here

pkg/drain/evictor/evictor.go
```diff
@@ -127,51 +139,51 @@ func (d *Helper) EvictPod(pod corev1.Pod) error {
...
 
-// getPodsForDeletion lists all pods on a given node, filters those using the default
-// filters, and returns podDeleteList along with any errors. All pods that are ready
+// GetPodsForDeletion lists all pods on a given node, filters those using the default
+// filters, and returns PodDeleteList along with any errors. All pods that are ready
 // to be deleted can be obtained with .Pods(), and string with all warning can be obtained
 // with .Warnings()
-func (d *Helper) getPodsForDeletion(nodeName string) (*podDeleteList, []error) {
-	labelSelector, err := labels.Parse(d.PodSelector)
+func (d *Evictor) GetPodsForEviction(nodeName string) (*PodDeleteList, []error) {
```
michaelbeaumont, line 154: comment/name mismatch here

michaelbeaumont (approved):

PR #2775, opened 2020-10-27
Allow creation of iamserviceaccounts on clusters not created through eksctl

aclevername:
### Description
Currently you can use `eksctl utils associate-iam-oidc-provider` on non-eksctl created clusters (more specifficaly clusters not created through CloudFormation), but when it comes to utilising this functionality and creating iamserviceaccounts via `eksctl create iamserviceaccount` this only works on eksctl created clusters (clusters with CloudFormation stacks). This PR changes it so that you can use `eksctl create iamserviceaccount` against non-eksctl clusters.

related issue: https://github.com/weaveworks/eksctl/issues/2174

### Checklist
- [ ] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `userdocs` directory)
- [X] Manually tested
- [X] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)
- [X] Make sure the title of the PR is a good description that can go into the release notes

PR #2779, opened 2020-10-28
Get clusters states if a cluster was eksctl created

aclevername:
### Description
Some eksctl commands are supported against clusters that are created through eksctl, and some are not. Currenty using the CLI you cannot discover which clusters were created through eksctl. This PR adds a `EKSCTL CREATED` column to the `eksctl get clusters` output to indicate whether or not the cluster was created by eksctl. 

As part of https://github.com/weaveworks/eksctl/pull/2775 clusters that weren't created by eksctl can now have stacks associated to them, such as a stack for creating the iamserviceaccount. To check that a cluster is eksctl created it checks that 1. it has a stack with the `eksctl` tags and 2. It has the `ClusterStackName` as one of the outputs. Im open for changing 2. to check for something else that is unique to clusters, this was just the easiest to pick.

### Others
- Minor refactoring to seperate our get cluster vs get clusters
- backfill of tests

### Checklist
- [ ] Manually tested


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -458,6 +458,9 @@ type ClusterMeta struct {
 	// Annotations are arbitrary metadata ignored by `eksctl`.
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
+	// Whether the cluster was created by eksctl or not
+	// +optional
+	EKSCTLCreated string `json:"eksctlCreated"`
```
michaelbeaumont, line 463: I think this can be added to the `ClusterStatus`, does that make sense?

pkg/apis/eksctl.io/v1alpha5/assets/schema.json
```diff
@@ -307,6 +307,11 @@
           "x-intellij-html-description": "arbitrary metadata ignored by <code>eksctl</code>.",
           "default": "{}"
         },
+        "eksctlCreated": {
```
michaelbeaumont, line 310: I think this shouldn't appear here but I see that `status` is already in here so it's ok, though we should hide that from the docs too

pkg/ctl/get/cluster.go
```diff
@@ -65,5 +65,10 @@ func doGetCluster(cmd *cmdutils.Cmd, params *getCmdParams, listAllRegions bool)
 		return err
 	}
 
-	return ctl.ListClusters(cfg.Metadata.Name, params.chunkSize, params.output, listAllRegions)
+	if cfg.Metadata.Name == "" {
+		return ctl.ListClusters(params.chunkSize, params.output, listAllRegions)
```
michaelbeaumont, line 69: :100:

pkg/eks/eks.go
```diff
@@ -378,47 +376,30 @@ func (c *ClusterProvider) doListClusters(chunkSize int64, printer printers.Outpu
...
-func (c *ClusterProvider) doGetCluster(clusterName string, printer printers.OutputPrinter) error {
+func (c *ClusterProvider) GetCluster(clusterName string) (*awseks.Cluster, error) {
 	input := &awseks.DescribeClusterInput{
 		Name: &clusterName,
 	}
+
 	output, err := c.Provider.EKS().DescribeCluster(input)
 	if err != nil {
-		return errors.Wrapf(err, "unable to describe control plane %q", clusterName)
+		return &awseks.Cluster{}, errors.Wrapf(err, "unable to describe control plane %q", clusterName)
```
michaelbeaumont, line 386: The AWS API likes to return pointers so this is probably left over from switching but I think either return `nil` for the error case or return a value (which I'm partial to) and `awseks.Cluster{}` in the error case, wdyt?

aclevername: sounds good to me!

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -472,6 +472,7 @@ type ClusterStatus struct {
 	CertificateAuthorityData []byte `json:"certificateAuthorityData,omitempty"`
 	ARN                      string `json:"arn,omitempty"`
 	StackName                string `json:"stackName,omitempty"`
+	EKSCTLCreated            string `json:"eksctlCreated,omitempty"`
```
michaelbeaumont, line 475: Maybe we can have a new type for this with true false and unknown? Or at least `const string`s

aclevername: Good idea!

aclevername:
@michaelbeaumont @cPu1 any idea what this test is about?

Test: https://github.com/weaveworks/eksctl/pull/2779/commits/2785d9b49c781f8c98e33d382214b2af4921ffbb
Associated test data file: https://github.com/weaveworks/eksctl/pull/2779/commits/2e6bddf6f3e8149329903bc7b50607cb60f85856

It started failing when I moved the print logic into the clt/get package, but I don't understand the test enough to understand what it does or how to fix it, so I removed it.

aclevername:
> @michaelbeaumont @cPu1 any idea what this test is about?
> 
> Test: [2785d9b](https://github.com/weaveworks/eksctl/commit/2785d9b49c781f8c98e33d382214b2af4921ffbb)
> Associated test data file: [2e6bddf](https://github.com/weaveworks/eksctl/commit/2e6bddf6f3e8149329903bc7b50607cb60f85856)
> 
> It started failing when I moved the print logic into the clt/get package, but I don't understand the test enough to understand what it does or how to fix it, so I removed it.

Looking into it further it appears to be testing when the `--output json` flag is provided to the `eksctl get cluster`. I'll re-add the test and make it more descriptive

michaelbeaumont (approved):

Issue #2768, opened 2020-10-23
NodeInstanceProfile CREATE_FAILED: "The specified value for roleName is invalid"


michaelbeaumont:
**What happened?**
When a nodegroup contains `instanceRoleARN` but not `instanceProfileARN`, the nodegroup fails with the error:
```
CREATE_FAILED - "The specified value for roleName is invalid"
```

**What you expected to happen?**
The instance profile to be created successfully

**How to reproduce it?**
Config contains:
```
nodeGroups:
  - name: ng
    iam:
      instanceRoleARN: arn:aws:iam::123:role/path/rolename
```

Nodegroup stack includes:

```
    "NodeInstanceProfile": {
      "Type": "AWS::IAM::InstanceProfile",
      "Properties": {
        "Path": "/",
        "Roles": [
          "arn:aws:iam::123:role/rolename"
        ]
      }
    },
```
**Anything else we need to know?**
`0.30.0`

PR #2791, opened 2020-11-05
Use role names instead of ARN when creating instance profiles

aclevername:
### Description
When creating an instance profile you need to provide the name instead of the role for the ARN. Currently if your provide an `instanceRoleARN` value but not a `instanceProfileARN` we attempt to create an `InstanceProfile`, but we provide the value ARN value instead of the role name to the spec, causing an error. Related AWS doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-instanceprofile.html

Related issue: fixes https://github.com/weaveworks/eksctl/issues/2768

### Checklist
- [X] Manually tested


michaelbeaumont (approved):
This fixes #2768 right?

aclevername:
> This fixes #2768 right?

@michaelbeaumont yes

Issue #1359, opened 2019-09-23
Create more subnets when more than 3 AZs are selected


albertmichaelj:
**What happened?**
A description of actual behavior (with error messages).
I am trying to  create a cluster that spans all AZs in us-east-1 (a, b, c, d, e, f) since I have an application that only needs to be in the same region, but can run equally well in any AZ within that region. When I try to user more than four AZs, I get the following error message:
```
❯ eksctl create cluster -f eks-cluster-test.yaml
[ℹ]  using region us-east-1
[✖]  insufficient number of subnets (have 8, but need 12) for 6 availability zones
```

This is because `eksctl` only creates 8 subnets, and there needs to be a public and private subnet per AZ. I cannot figure out how to get `eksctl` to create more subnets, but it should automatically scale the subnets by AZs.

**What you expected to happen?**

**How to reproduce it?**
Include the steps to reproduce the bug
I use the following config file as `eks-cluster-test.yaml`:
```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: test-cluster
  region: us-east-1

nodeGroups:
  - name: control-nodes-1
    instanceType: t3.small
    desiredCapacity: 1

availabilityZones: ["us-east-1a", "us-east-1b", "us-east-1c", "us-east-1d", "us-east-1e", "us-east-1f"]
```
Then run `eksctl create cluster -f eks-cluster-test.yaml`.

**Anything else we need to know?**
What OS are you using, are you using a downloaded binary or did you compile eksctl, what type of AWS credentials are you using (i.e. default/named profile, MFA) - please don't include actual credentials though!

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
[ℹ]  version.Info{BuiltAt:"", GitCommit:"", GitTag:"0.6.0"}
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
error: You must be logged in to the server (the server has asked for the client to provide credentials)
```

**Logs**
Include the output of the command line when running eksctl. If possible, eksctl should be run with debug logs. For example:
`eksctl get clusters -v 4`
Make sure you redact any sensitive information before posting.
[...]

PR #2804, opened 2020-11-09
Support more than 4 azs

aclevername:
### Description
Currently EKSCTL limits you to specify 4 AZs, if you attempt to specify more it fails as we only split up the VPC CIDR block into 8 chunks (4x 1 private & 1 public subnet). This PR ups the capacity to 8 AZs (currently the biggest region has only 6 AZs) by adding the option to split the VPC CIDR block into 16.

related issue: https://github.com/weaveworks/eksctl/issues/1359

### Checklist
- [X] Manually tested


pkg/vpc/vpc.go
```diff
@@ -65,6 +76,31 @@ func SetSubnets(vpc *api.ClusterVPC, availabilityZones []string) error {
 	return nil
 }
 
+func SplitInto16(parent *net.IPNet) ([]*net.IPNet, error) {
```
michaelbeaumont, line 79: Ooh can we get rid of the `kops` dependency here and just include our own `SplitInto8`?

aclevername: will do!

michaelbeaumont (approved):
One potential improvement but either way, LGTM!

pkg/vpc/vpc.go
```diff
@@ -65,6 +74,56 @@ func SetSubnets(vpc *api.ClusterVPC, availabilityZones []string) error {
...
+			})
+		} else {
+			return nil, fmt.Errorf("Unexpected IP address type: %s", parent)
+		}
+	}
+
+	return subnets, nil
+}
+
+func SplitInto8(parent *net.IPNet) ([]*net.IPNet, error) {
```
michaelbeaumont, line 102: Sorry forgot to include comment, this could be reduced into `SplitIntoN`

aclevername: @michaelbeaumont I'm a bit hesitant to do that, while I managed to manipulate the first function to accurately produce splits of 16 instead of 8, I don't know how to ensure the same change for any given input, e.g. splitting into 10.

michaelbeaumont: Yes, it should work for any power of two (and only for powers of two).

PR #2815, opened 2020-11-10
Support upgrading non eksctl clusters

aclevername:
### Description
`eksctl upgrade cluster` can update clusters that aren't created via eksctl (e.g. created manually in the console)

The current upgrade flow for eksctl created cluster also checks that all of the [nodegroups are compatible](https://github.com/weaveworks/eksctl/blob/master/pkg/eks/compatibility.go#L45-L101). I have not added the same compatibility tests to the upgrade path for non-eksctl support as you can't currently create nodegroups for these clusters, so it can't contain any incompatible nodes. This is something we should be aware of in the future if we add support for creating nodegroups for non-eksclt created clusters.

### Checklist
- [X] Manually tested


pkg/ctl/upgrade/cluster.go
```diff
@@ -82,31 +84,39 @@ func DoUpgradeCluster(cmd *cmdutils.Cmd) error {
...
 	}
 
+	stackManager := ctl.NewStackManager(cfg)
+	stacks, err := stackManager.DescribeStacks()
+	if err != nil {
+		return err
+	}
+
+	if manager.IsClusterStack(stacks) {
+		return upgradeEKSCTLCluster(versionUpdateRequired, cmd.Plan, cfg, currentVersion, ctl, stackManager)
```
michaelbeaumont, line 101: Should we start a `Cluster interface` that provides `Upgrade` and have two concrete types implement it? Something we can expand as more non-eksctl created operations become possible.

aclevername: Great idea, I'll incorporate it into the PR

pkg/actions/cluster/owned.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (c *OwnedCluster) Upgrade(dryRun bool) error {
+	currentVersion := c.ctl.ControlPlaneVersion()
+	versionUpdateRequired, err := requiresVersionUpgrade(c.cfg.Metadata, currentVersion)
+	if err != nil {
+		return err
+	}
+
+	if err := c.ctl.LoadClusterVPC(c.cfg); err != nil {
```
Callisto13, line 34: did i peer pressure you into doing your errors like this? 😂

pkg/actions/cluster/owned.go
```diff
@@ -0,0 +1,70 @@
...
+	printer := printers.NewJSONPrinter()
+	if err := printer.LogObj(logger.Debug, "cfg.json = \\\n%s\n", c.cfg); err != nil {
+		return err
+	}
+
+	if versionUpdateRequired {
+		if err := updateVersion(dryRun, c.cfg, currentVersion, c.ctl); err != nil {
+			return err
+		}
+	}
```
Callisto13, line 47: do we want to move this up to below line 29? (or move 29 down a bit if something is needed for `updateVersion`) looks weird to do the check so far after getting the value

pkg/actions/cluster/owned.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	if err := c.ctl.ValidateExistingNodeGroupsForCompatibility(c.cfg, c.stackManager); err != nil {
+		logger.Critical("failed checking nodegroups", err.Error())
+	}
+
+	cmdutils.LogPlanModeWarning(dryRun && (stackUpdateRequired || versionUpdateRequired))
+	return nil
+
```
Callisto13, line 69:
```suggestion
```

pkg/actions/cluster/update.go
```diff
@@ -0,0 +1,97 @@
...
+	case api.Version1_14:
+		return api.Version1_15, nil
+	case api.Version1_15:
+		return api.Version1_16, nil
+	case api.Version1_16:
+		return api.Version1_17, nil
+	case api.Version1_17:
+		return api.Version1_18, nil
+	case api.Version1_18:
+		return api.Version1_19, nil
```
Callisto13, line 92: A couple of questions here:
1. How come you started at 1.12? We list versions as far back as 1.10 are listed as unsupported.
1. What is the UX for people whose next version is unsupported? I can see we give them this error, but do we suggest a next course of action? I can see people thinking "oh i'll just bump n versions then" which we also don't allow.

pkg/actions/cluster/owned.go
aclevername: it sort of feels in the right place, we error out early if we can't check if theres an update, and then afterwards we load cluster information, log what we've found and then proceed with the upgrade.

pkg/actions/cluster/update.go
aclevername: Great question, I'm not sure regarding 2). I know that we do manually add versions to eksctl when new versions of kubernetes are released. 

> I can see people thinking "oh i'll just bump n versions then" which we also don't allow.

Do you mean kubernetes versions? or CLI versions?

Callisto13: kubernetes

aclevername: I'm assuming the EKS API will reject it, although I haven't tried.

Callisto13: hmm i imagine that looks pretty. meh, we can wait until somebody tries it and gets confused

pkg/actions/cluster/unowned.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func NewUnownedCluster(cfg *api.ClusterConfig, ctl *eks.ClusterProvider) (*UnownedCluster, error) {
+	return &UnownedCluster{
+		cfg: cfg,
+		ctl: ctl,
+	}, nil
+}
+
+func (c *UnownedCluster) Upgrade(dryRun bool) error {
```
michaelbeaumont, line 22: this function seems to be duplicated in `owned.go` except for the `LoadClusterVPC` call, can that be reduced somehow?

aclevername: lmk what you think @michaelbeaumont https://github.com/weaveworks/eksctl/pull/2815/commits/a6e01da33ca345697881b96da1307b538e61b92f

aclevername:
Going to re-test the functionality once the PR is approved as a few minor refactors have been done since the last test

michaelbeaumont (approved):

aclevername:
> Going to re-test the functionality once the PR is approved before merging as a few minor refactors have been done since the last test

Tested works 👍

Issue #2776, opened 2020-10-27
create nodegroup creates 2 volumes, the root doesn't honor volumeSize


derekwlmscfa:
**What happened?**
`create nodegroup` is creating the EC2 nodes with two (not one) attached EBS volumes. The root volume is only 20GB, too small for use. The second volume is 100GB, honoring the `volumeSize` I specified.

**What you expected to happen?**
There should only be the one volume I specified in the config file, honoring the `volumeSize`. Or if two volumes are created, provide a way to control the root volume size.

**How to reproduce it?**
`eksctl create nodegroup --include='ng-2020-10-26b*' --config-file=cfg.yaml`

**Anything else we need to know?**
I did not see this behavior with other nodegroups I created in the past. Differences here are 1) different eksctl version, and 2) using Bottlerocket `amiFamily`.

Relevant config file portion (sensitive data redacted):
```
  - name: ng-2020-10-26br
    amiFamily: Bottlerocket
    ami: auto-ssm
    labels:
      "bottlerocket.aws/updater-interface-version": "1.0.0"
    iam:
      attachPolicyARNs:
        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy
      withAddonPolicies:
        appMesh: true
        autoScaler: true
        ebs: true
    instanceType: c5n.4xlarge
    maxSize: 75
    desiredCapacity: 8
    minSize: 8
    privateNetworking: true
    volumeSize: 100
    volumeType: gp2
```

**Versions**
Please paste in the output of these commands:
```
$ eksctl version
0.30.0
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-eks-065dce", GitCommit:"065dcecfcd2a91bd68a17ee0b5e895088430bd05", GitTreeState:"clean", BuildDate:"2020-07-16T01:44:47Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```

**Logs**
Include the output of the command line when running eksctl. If possible, eksctl should be run with debug logs. For example:
`eksctl get clusters -v 4`
[...]

PR #2843, opened 2020-11-16
Respect volume size for Bottlerocket nodegroups

aclevername:
### Description
Currently we increase the `/dev/xvda` disk by default for all nodegroups. When Bottlerocket images are used the disk setup is different:

Related issue: https://github.com/weaveworks/eksctl/issues/2776
> /dev/xvda is used to hold the operating system itself and contains the active & passive partitions, the bootloader, the dm-verity data, and the data store for the Bottlerocket API. /dev/xvdb is used for all of the things you run on top of Bottlerocket: container images, storage for running containers, and persistent storage/volumes.

This PR updates eksctl to increase the size of `/dev/xvdb` when using Bottlerocket

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/ami/api.go
+++ b/pkg/ami/api.go
@@ -53,10 +53,9 @@ func Use(ec2api ec2iface.EC2API, ng *api.NodeGroupBase) error {
 
 	case "ebs":
 		if !api.IsSetAndNonEmptyString(ng.VolumeName) {
+			ng.VolumeName = image.RootDeviceName
 			if ng.AMIFamily == api.NodeImageFamilyBottlerocket {
 				ng.VolumeName = aws.String("/dev/xvdb")
-			} else {
-				ng.VolumeName = image.RootDeviceName
 			}
 		}
 		rootDeviceMapping, err := findRootDeviceMapping(image)
--- a/pkg/ami/api.go
+++ b/pkg/ami/api.go
@@ -21,6 +21,9 @@ const (
 	ImageClassARM
 )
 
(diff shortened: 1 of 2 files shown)


pkg/ami/api.go
```diff
@@ -54,6 +54,9 @@ func Use(ec2api ec2iface.EC2API, ng *api.NodeGroupBase) error {
 	case "ebs":
 		if !api.IsSetAndNonEmptyString(ng.VolumeName) {
 			ng.VolumeName = image.RootDeviceName
+			if ng.AMIFamily == api.NodeImageFamilyBottlerocket {
+				ng.VolumeName = aws.String("/dev/xvdb")
```
Callisto13, line 58: is there a tidier way of setting this? feels like a bit of a "magic string" (if that is not already a thing

Callisto13 (approved):

pkg/ami/api.go
michaelbeaumont, line 58: Maybe better as a constant `BottlerocketDataDisk` or something? (So the name explains why we're setting it)

michaelbeaumont (approved):

DekusDenial:
@aclevername it seems like the `/dev/xvda` volume is not encrypted if we enable bottlerocket node group with `volumeEncrypted` and provide a `volumeKmsKeyID`, while `/dev/xvdb` is being encrypted just fine.

aclevername:
thanks @DekusDenial I've opened an issue https://github.com/weaveworks/eksctl/issues/2945

PR #2855, opened 2020-11-19
Add ISO regions (us-iso-east-1, us-isob-east-1)

Callisto13:
### TLDR 
- Add isolated regions and partition mappings
- Fetch account id from environment if isolated region is set (`ISO_EKS_ACCOUNT_ID`, name up for debate)
  - Try to fail fast when this is not set

### Detail
This PR adds the ISO regions and partitions, and also includes a change to the way we fetch [Resource/Account/Owner IDs](https://github.com/weaveworks/eksctl/blob/master/pkg/apis/eksctl.io/v1alpha5/types.go#L220-L244). This is because the account numbers for the AMIs which can be used in these secret isolated regions are also secret and cannot be made public.

All the other account IDs (for this particular AMI family) are simply stored in constants where they can be called from anywhere in the code... and they are, from just about everywhere. So adding a once-in-a-while user input flow is a little fiddly. I spiked on/considered several approaches:

1. The ["F the UX" plan](https://github.com/Callisto13/eksctl/commit/8ea941d2efb91c1049ab7244089677ec4138bcd2)
    - This involves checking the env in the same place we would normally just return a constant. It sucks because, if you had forgotten to set the env var, you would not know about it until _after_ your stack had been created.
1. The ["Passing it through everywhere" plan](https://github.com/Callisto13/eksctl/commit/8d312581d7ee78bec2c826296232226a0334a382)
    - This involves checking the region and validating the env var (or not) at the "top" of any command which would end up needing it (there are 5) and passing it down through the various threads. This is technically the "correct" thing to do, but since this account is used _everywhere_ and quite a long way down sometimes, it is super invasive. In one instance it may end up not being used at all, depending on whether a different image family was chosen for the nodegroup.
      - Since you ask, setting the "true" account id (based on both region _and_ AMI family) is not possible since each node group could ask for a different family, therefore there is no "true" id which can be applied everywhere.
1. The ["Java-style getter/setter" plan](https://github.com/Callisto13/eksctl/commit/da452579894e6cffe155635e472d488cb0aca78a)
    - In which the account ID is set once at the very beginning, stored in a private variable which can then be called via a public function when needed. This feels like an odd pattern in Go.
1. The "Sneaky init" plan (no spike)
    - Similar to the above, but done in an `init()` at the start. This would only be possible if I moved the whole flag/config loading into an init too, because I would need the region to decide whether or not to find the env var. I decided moving our already complicated cmd starting points into an init would be a deathtrap as inits are very hard to debug when something goes wrong.
1. **The "Just check twice" plan (this PR)**
    - Similar to plan 1, except we also check for the presence of the env var at the start of the call if an iso region is set.

I chose the approach in this PR (5) based on the fact I hated it the least. I am happy to change things if anyone has a better idea.

## Questions
1. Any thoughts on the name `ISO_EKS_ACCOUNT_ID` for the env var?
1. How do we want to document this? _Do_ we want to document it? Where?
1. Also: have I missed anything for adding the regions?

### Checklist
- [ ] Manually tested (has to be sent to aws)

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/account.go
+++ b/pkg/apis/eksctl.io/v1alpha5/account.go
@@ -1,0 +1,96 @@
+package v1alpha5
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	IsoEKSAccountIDEnv = "ISO_EKS_ACCOUNT_ID"
+
+	// eksResourceAccountStandard defines the AWS EKS account ID that provides node resources in default regions
+	// for standard AWS partition
+	eksResourceAccountStandard = "602401143452"
+
+	// eksResourceAccountAPEast1 defines the AWS EKS account ID that provides node resources in ap-east-1 region
+	eksResourceAccountAPEast1 = "800184023465"
+
+	// eksResourceAccountMESouth1 defines the AWS EKS account ID that provides node resources in me-south-1 region
+	eksResourceAccountMESouth1 = "558608220178"
+
+	// eksResourceAccountCNNorthWest1 defines the AWS EKS account ID that provides node resources in cn-northwest-1 region
+	eksResourceAccountCNNorthWest1 = "961992271922"
+
+	// eksResourceAccountCNNorth1 defines the AWS EKS account ID that provides node resources in cn-north-1
+	eksResourceAccountCNNorth1 = "918309763551"
+
+	// eksResourceAccountAFSouth1 defines the AWS EKS account ID that provides node resources in af-south-1
+	eksResourceAccountAFSouth1 = "877085696533"
+
+	// eksResourceAccountEUSouth1 defines the AWS EKS account ID that provides node resources in eu-south-1
+	eksResourceAccountEUSouth1 = "590381155156"
+
+	// eksResourceAccountUSGovWest1 defines the AWS EKS account ID that provides node resources in us-gov-west-1
+	eksResourceAccountUSGovWest1 = "013241004608"
+
+	// eksResourceAccountUSGovEast1 defines the AWS EKS account ID that provides node resources in us-gov-east-1
+	eksResourceAccountUSGovEast1 = "151742754352"
+)
+
+// EKSResourceAccountID provides worker node resources(ami/ecr image) in different aws account
+// for different aws partitions & opt-in regions.
+func EKSResourceAccountID(region string) (string, error) {
+	switch region {
+	case RegionUSIsoEast1:
+		return lookupIsoAccountID(region)
+	case RegionUSIsobEast1:
+		return lookupIsoAccountID(region)
+	default:
+		return publicResourceAccountID(region), nil
+	}
+}
+
+// ValidateEKSAccountID check that the ISO_EKS_ACCOUNT_ID env var has been set
+// when an isolated region is requested
+func ValidateEKSAccountID(region string) error {
+	if region == RegionUSIsoEast1 || region == RegionUSIsobEast1 {
+		if _, err := lookupIsoAccountID(region); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func lookupIsoAccountID(region string) (string, error) {
+	id := os.Getenv(IsoEKSAccountIDEnv)
+	if id == "" {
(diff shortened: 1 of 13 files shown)


Callisto13:
Any thoughts on this?

cmdallas:
1. Any thoughts on the name ISO_EKS_ACCOUNT_ID for the env var?
> +1. Currently, all 'special' aws regions (existing & roadmap) use the an 'iso' convention in their arn partitions.

2. How do we want to document this? Do we want to document it? Where?
> AWS has internal docs for customers of special regions that generally document these kinds of differences. This is a bit of a unique scenario, but I think it makes sense for us (AWS) to add this to our diff docs. I will take an action item to get this published in the internal docs.

Callisto13:
👍  thanks @cmdallas sounds good

Callisto13:
**Issue:** `eksctl get clusters --all-regions` errors (see below) for regular folks because they cannot run inside the ISO networks. The SDK's send request fails repeatedly but [will retry](https://github.com/aws/aws-sdk-go/blob/master/aws/corehandlers/handlers.go#L162-L164) because that is the standard behaviour for network errors. (We do not see this retry behaviour for Gov or CN regions; we get the expected `403` back in those cases.)

```
```

**Options:**
a) Change to EKSCTL: We exclude the ISO regions when calling `get clusters --all-regions` (while we are at it: the UX for getting Gov and CN regions when you do not have creds is a bit rubbish)
b) Change to aws-cdk: We set ISO-specific retry rules.

All thoughts and votes and new ideas welcome please!

mikestef9:
I prefer option a, would also like to get @cmdallas thoughts as well

cPu1:
Another option is to use `EC2:DescribeRegions` to get all the regions enabled for the account, and run tests on those regions only. Although it's unclear whether it would return isolated regions as well if an account has access to them.

Callisto13:
Ooh thanks @cPu1 ! That would solve all UX annoyances 😄

pkg/apis/eksctl.io/v1alpha5/account.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return nil
+}
+
+func lookupIsoAccountID(region string) (string, error) {
+	id, ok := os.LookupEnv(IsoEKSAccountIDEnv)
+	if !ok {
+		return "", errors.Errorf("%s not set, required for use of region: %s", IsoEKSAccountIDEnv, region)
+	}
```
cPu1, line 70: This will pass validation if the env var is set but is empty (e.g., `ISO_EKS_ACCOUNT_ID= eksctl create cluster`). I think we should validate that the value is non-empty.

cPu1 (approved):

pkg/apis/eksctl.io/v1alpha5/account.go
Callisto13: 👍  thanks

cmdallas:
@Callisto13 

Failing in region because a us-west-2 sts endpoint is being called:

```
export ISO_EKS_ACCOUNT_ID=<redacted>
./eksctl get cluster --name=<redacted>
Error: checking AWS STS access - cannot get role ARN for current session: RequestError: send request failed
cause by Post "https://sts.us-west-2.amazonaws.com"
```

https://github.com/weaveworks/eksctl/blob/d4b46b2c84a6175cd54d3038d8074413bdadfddc/pkg/eks/api.go#L264

Callisto13:
Thanks @cmdallas I will look into it and get a new build back to you asap 👍

Callisto13:
@cmdallas I think the region defaults to `us-west-2`, could you see what happens when you pass in the `--region` flag?

cmdallas:
My bad!

Cool, `get` commands are working fine. I attempted a `create` command, but we are missing SSM params in region. I will try and get that updated by Monday the 13th

cmdallas:
Sorry to slip!! I am out on vacation and will not be back for a few weeks. I will pass this on to my team for the time being.

Callisto13:
No worries @cmdallas, pretty much everyone on our side is also on vacation next week. I'll be back on the 29th.

Callisto13:
Gonna close this an re-implement when upstream APIs support these regions.

Issue #2591, opened 2020-08-28
Add additional fargate profile via eksctl configuration


cdenneen:
Add fp-default to config.

`eksctl create fargateprofile -f config.yaml`

fp-default is created

Add fp-new to config (second profile)

`eksctl create fargateprofile -f config.yaml `

```
Error: failed to create Fargate profile "fp-default" on EKS cluster "eks-cluster": failed to create Fargate profile "fp-default": ResourceInUseException: A Fargate Profile already exists with this name in this cluster.
{
  RespMetadata: {
    StatusCode: 409,
    RequestID: ""
  },
  Message_: "A Fargate Profile already exists with this name in this cluster."
}
```

`eksctl create fargateprofile -f config fp-new`
`Error: cannot use name argument when --config-file/-f is set`

PR #2871, opened 2020-11-23
Do not fail Fargate profile creation if it already exists

Callisto13:
### Description

**TLDR:** If a ResourceInUseException is thrown on subsequent creation attempts of an existing
profile, we should log rather than fail. This will let users run creates against updated config files.

**Notable changes:**
- As we discussed in a recent meeting about tech-debt and refactoring strategies, we want to try to keep packages tightly and clearly encapsulated, with clear entry/exit points and easily testable boundaries. I have made a small change so that, rather than calling the client code in `pkg/fargate` directly (and therefore indirectly testing that package as well), `pkg/eks/fargate` calls an interface. I only wrapped one func (`CreateProfile`) in this interface for now as use of the client is fairly scattered and I wanted to see how this would look/get thoughts on an overall strategy before we started to move everything over to a similar pattern. Any thoughts and comments are welcome.
- I have also used [Counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) for mocking as I think it gives a nicer UX. It would be cool to move to this fully over time, but strong opinions against this are welcome.

### Checklist
- [x] Manually tested

Closes https://github.com/weaveworks/eksctl/issues/2591


michaelbeaumont (approved):
Yes!

pkg/eks/fargate.go
```diff
@@ -3,29 +3,35 @@ package eks
...
 
-	"github.com/weaveworks/eksctl/pkg/fargate"
 	"github.com/weaveworks/eksctl/pkg/fargate/coredns"
 	"github.com/weaveworks/eksctl/pkg/utils/retry"
 	"github.com/weaveworks/eksctl/pkg/utils/strings"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 )
 
+//go:generate counterfeiter -o fakes/fargate_client.go . FargateClient
```
michaelbeaumont, line 18: This is already broken in `master` from https://github.com/weaveworks/eksctl/pull/2766/files, I missed it :cry:
`counterfeiter` has to be in `tools.go` and this should be:

```suggestion
//go:generate "${GOBIN}/counterfeiter" -o fakes/fargate_client.go . FargateClient
```

Callisto13: Gah! Nice spot, I will also throw in a fix for the other one for free.

Callisto13: Hmmm does not seem to like that :(, do you know if I need to set anything else up to make it work in CI?

michaelbeaumont: Ah nice one, forgot about `.requirements`!

michaelbeaumont (approved):

michaelbeaumont:
@Callisto13 WIP does this sometimes :shrug:, ready to merge!

PR #2872, opened 2020-11-23
Handle missing iam.serviceAccounts key when creating IAMServiceAccounts

mgreensmith:
### Description

Check for the presence of the `iam.serviceAccounts` key in the ClusterConfig file, and return an error if the key does not exist. Avoids a nil pointer panic when `create iamserviceaccount` is invoked with an insufficiently-specified config file.

### Checklist
- [x] Manually tested


pkg/addons/assets.go
```diff
@@ -27,15 +27,15 @@ import (
 func bindataRead(data []byte, name string) ([]byte, error) {
 	gz, err := gzip.NewReader(bytes.NewBuffer(data))
 	if err != nil {
-		return nil, fmt.Errorf("read %q: %w", name, err)
+		return nil, fmt.Errorf("read %q: %v", name, err)
```
michaelbeaumont, line 30: I think you have to update `golanci-lint`

mgreensmith: Good call, thanks. I verified that there are no generated file changes after I updated deps.

michaelbeaumont (approved):

Issue #2864, opened 2020-11-23
get cluster returns null instead of empty JSON array


mishra-swati:
**What happened?**
`eksctl get cluster --output json` returns `null` instead of `[]` when there are no clusters created with eksctl in the AWS account.

**What you expected to happen?**
`eksctl get cluster --output json` should return `[]`, an empty JSON array, to represent no clusters were found.

**How to reproduce it?**
1. Set up `eksctl` with an AWS account that has no EKS clusters created with eksctl
1. Run `eksctl get cluster --output json`

**Anything else we need to know?**
Version `0.31.0` was exhibiting the correct, expected behavior.

**Versions**
Please paste in the output of these commands:
`eksctl version`: `0.32.0`
`kubectl version --short --client`: `Client Version: v1.18.2`

**Logs**
```
$ eksctl get cluster -v 4 --output json
2020-11-22T18:49:10-08:00 [▶]  role ARN for the current session is *********
null%
```

hvydya:
Hello! I'd like to work on this. Can you assign it to me?

hvydya:
Looked at this today. Have a possible fix. Do I need to write any tests for this?

michaelbeaumont:
@hvydya Great! Feel free to open a PR and we'll see if there's any tests missing :slightly_smiling_face:

PR #2873, opened 2020-11-25
Print empty JSON array when `get clusters -o json` returns nothing

hvydya:
Issue: The printer for json and yaml type output will print "null" if the value supplied to it is nil and will print "[]" when the value supplied is an empty slice. The function responsible for returning the clusters output was sending nil when no clusters exist.

Fix: Initialised the variable to an empty slice in listClusters so that it returns an empty slice instead of nil. Wrote a test which will check if ListClusters returns the correct output i.e empty slice.

Issue #2864


hvydya:
Looks like I'll have to merge the latest master onto this. Is there a procedure defined for that? Or do I just `git pull origin master` in this branch and push it?

Also I'm unable to figure out how to add labels to this pr and link it to the issue. Need help on that (if that is to be done by me).

Callisto13:
Thanks for this!

> do I just git pull origin master in this branch and push it?

Yep 👍  that will be fine (with a `-r` may be nicer)

I will add some labels, think we have it set to maintainers 🤔 

And this should link the issue:

Closes https://github.com/weaveworks/eksctl/issues/2864

hvydya:
Thanks for the quick reply. Rebased it.

Happy Thanksgiving and a long weekend 😃 !

michaelbeaumont (approved):

PR #2886, opened 2020-12-01
Show status of fargate profile in get command.

### LINKED ISSUE #2845: Getting Fargate profiles should show profile statuses
Sometimes Fargate actions take a looong time to complete, to help with confusion it may be useful to display what state they are in.

Deletion is the worst. After running a delete command it can take up to 5 minutes for the profile to actually disappear which (if you are lazy like me and cnba to read anything or go to the console to check) can lead to this frustrating sequence of events:

```sh
# delete my extraneous profile
$ eksctl delete fargateprofile --cluster foo --name bar 

# think I'll have a look at the other profiles I have left
$ eksctl get fargateprofile --cluster foo -o yaml
...
- name: bar  # hmmm still there?
  selectors:
  - namespace: default
  subnets:
  - subnet-lolz
...

# I have no chill so try to kill it again
$ eksctl delete fargateprofile --cluster foo --name bar 
Error: failed to delete Fargate profile "bar": InvalidRequestException: Fargate Profile is currently not in Active State.
...

# rude
```

A bit more info when getting profiles:
```sh
$ eksctl get fargateprofile --cluster foo -o yaml
...
- name: bar  # hmmm still there?
  status: deleting/inactive/whatever # ah that's okay
  selectors:
  - namespace: default
  subnets:
  - subnet-lolz
...
```

hvydya:
Show status of fargate profile when `get fargateprofile` is called.

Closes #2845


aclevername (approved):
LGTM

hvydya:
So the conflicting file is auto-generated. Should I accept all changes from my branch to resolve them?

aclevername:
> So the conflicting file is auto-generated. Should I accept all changes from my branch to resolve them? Or do i need to re-generate those because accepting all my changes might affect the changes which were not made by me...Seems like latter one is the way to go. Let me know how to resolve this. 😄

Hey @hvydya, as you pointed out its an auto-generated file, so normally I would accept HEAD as part of the conflict, and then re-generate the files as this will always result in the desired outcome. Running `make test` will include the steps to re-generate the code 😄

michaelbeaumont (approved):

michaelbeaumont:
Thanks @hvydya !

PR #2892, opened 2020-12-02
Do not rebuild eksctl-build when creating releases

Callisto13:
When creating `eksctl-image` as part of the release process, we should not need to rebuild `eksctl-build`. Simply pulling the latest should be fine and will hopefully speed up the job for us.


aclevername (approved):
LGTM

michaelbeaumont (approved):

Issue #2538, opened 2020-08-13
Delete unnecessary code for multi-arch images


martina-if:
When launching support for ARM we added code that made sure that after creating a cluster the default addons where updated to the newest version with the multi-arch images (their label has the suffix `-eksbuild.1`). 

Once the multi-architecture images are used by EKS by default when creating clusters, that code can be removed from eksctl. 

The code is marked with a `// TODO`. Check the [ARM PR](https://github.com/weaveworks/eksctl/pull/2535) for more details.

Callisto13:
Thanks @martina-if for making this so straightforward ❤️

PR #2947, opened 2020-12-15
Delete UpdateAddonsTask

Callisto13:
### Description

All "addons" (yes, I will use quotation marks until we stop using that name for nearly everything as a default for when we cant think of anything better. /rant) are now created with multi-architecture images by default in all versions. This means that we no longer need to ensure that the "addons" are up-to-date before creating ARM nodes.

Closes https://github.com/weaveworks/eksctl/issues/2538

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/addons/default/addons.go
+++ b/pkg/addons/default/addons.go
@@ -1,33 +1,8 @@
 package defaultaddons
 
 import (
-	"github.com/pkg/errors"
-
 	"github.com/weaveworks/eksctl/pkg/kubernetes"
 )
-
-// EnsureAllAddonsUpToDate checks the default addons (aws-node, kube-proxy and coredns) are up to date
-// and if they are not it will update them to the latest version for the given control plane version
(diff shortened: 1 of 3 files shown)


pkg/addons/default/addons.go
```diff
@@ -1,34 +1,9 @@
...
-
-	_, err = UpdateCoreDNS(rawClient, region, controlPlaneVersion, false)
-	if err != nil {
-		return errors.Wrapf(err, "error updating coredns")
-	}
-
-	return nil
-}
-
 func DoAddonsSupportMultiArch(clientSet kubernetes.Interface, rawClient kubernetes.RawClientInterface, controlPlaneVersion string, region string) (bool, error) {
```
Callisto13, line 7: i did consider scrapping this as well, but it may be useful to leave it there in case users have messed with something and then decide they want arm 🤷‍♀️

michaelbeaumont (approved):

aclevername (approved):

PR #2948, opened 2020-12-15
add sub condition to vpc-cni addon policy

aclevername:
### Description

Policy will now look like:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::<redacted>:oidc-provider/oidc.eks.us-west-2.amazonaws.com/id/<redacted>"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "oidc.eks.us-west-2.amazonaws.com/id/<redacted>:aud": "sts.amazonaws.com",
          "oidc.eks.us-west-2.amazonaws.com/id/<redacted>:sub": "system:serviceaccount:kube-system:aws-node"
        }
      }
    }
  ]
}
```

before it did not have `"oidc.eks.us-west-2.amazonaws.com/id/<redacted>:sub": "system:serviceaccount:kube-system:aws-node"`

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/actions/addon/create.go
+++ b/pkg/actions/addon/create.go
@@ -39,6 +39,7 @@ func (a *Manager) Create(addon *api.Addon) error {
 	}
 
 	logger.Debug("addon: %v", addon)
+	namespace, serviceAccount := a.getKnownServiceAccountLocation(addon)
 
 	if a.withOIDC {
 		if addon.ServiceAccountRoleARN != "" {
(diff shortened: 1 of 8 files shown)


michaelbeaumont (approved):

pkg/actions/addon/create.go
```diff
@@ -176,8 +177,19 @@ func (a *Manager) getRecommendedPolicies(addon *api.Addon) []string {
 	}
 }
 
-func (a *Manager) createRoleUsingAttachPolicyARNs(addon *api.Addon) (string, error) {
-	resourceSet := builder.NewIAMRoleResourceSetWithAttachPolicyARNs(addon.Name, addon.AttachPolicyARNs, a.oidcManager)
+func (a *Manager) getKnownServiceAccountLocation(addon *api.Addon) (string, string) {
+	// API isn't case sensitive
+	switch strings.ToLower(addon.Name) {
```
michaelbeaumont, line 182: random thought, addons could have `Name() string` which auto lowercases the name

aclevername: great idea! I'll update

aclevername: I ended up going back a forth a bit on this (the smallest thing suddenly becomes so annoying!±). We can't add a `Name()` func because it clashes with the variable `Name`. I didn't want to change the variable Name or make it private. I was left with two alternatives:
- Mutate the `Name` at some point early on, e.g when loading config, to ensure its always lower cased
- Have a different name for the func, `NameLowerCase()`

The first introduces more hidden logic away from the core action on addons, the later keeps it a bit closer so I went with that.

michaelbeaumont: All not a big deal indeed, although i would have just made it private since we never need the upper case name.
But `NameLowerCase` isn't so descriptive maybe like `CanonicalName` or something that explains why you'd want to use the function (otherwise one might as well just use `ToLower(Name)`)

michaelbeaumont (approved):

PR #2962, opened 2020-12-17
Have kubelet use systemd cgroup driver on al2 and ubuntu

### LINKED ISSUE #2960: nodeboostrap should support changes to docker cgroups config
The EKS AMI should move to using systemd cgroup instead of cgroupfs. See https://github.com/awslabs/amazon-eks-ami/pull/521 for details. That change was reverted in https://github.com/awslabs/amazon-eks-ami/pull/587 because apparently eksctl nodebootsrap cannot handle docker daemon config changes.

eksctl should support work with the changes to the docker daemon config found in https://github.com/awslabs/amazon-eks-ami/pull/521

reegnz:
### Description

Kubernetes documentation indicates that for stability reasons
one should run kubernetes with the systemd cgroup driver if the
init system itself is systemd.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers

The same change is also be made in awslabs/amazon-eks-ami:

https://github.com/awslabs/amazon-eks-ami/pull/521

Fixes #2960

### Checklist
- [x] Manually tested


reegnz:
Verified manally:
`eksctl create cluster` with the new built binary
entered the node with kubectl `node-shell`  (https://github.com/kvaps/kubectl-node-shell)
checking `/etc/docker/daemon.json` -> ok, contains proper systemd cgroup config
checking `/etc/eksctl/kubelet.yaml` -> ok, contains proper systemd cgroup config
checking `/etc/systemd/system/kubelet.service.d/10-eksclt.al2.conf`  -> uses proper kubelet.yaml
`systemctl -t slice`  -> shows a bunch of kubepods slices
`systemd-cgls`  -> shows kubepods.slice below system.slice
`systemctl status containerd.service`  -> you can see runc processes spawned by it are using the -systemd-cgroup  flag
`systemd-cgtop` -> output contains the kubepods slices

Callisto13:
Thanks @reegnz the team will get to this asap 👍

Callisto13 (approved):
LGTM, thanks for catching this 👍

romeotheriault:
This change causes problems for us using ubuntu 16 (also doesn't work well on ubuntu 18) as our node ami with the included versions of systemd. It seems there are some bugs in systemd's cgroup driver in ubuntu 16 and 18 that cause docker to hang on creating containers, etc... and in turn it causes kubelet to hang and the nodes to become "NotReady". Downgrading eksctl to eksctl version 0.34.0 (the last version before this change) fixes the issue. 

It would be great if eksctl could provide a flag to allow the users to change (or choose which of) these types of options to use instead of eksctl changing a default underneath it's users. Especially for older OS releases.

Callisto13:
@romeotheriault could you open an issue for this please?

josephprem:
Hello folks,
Now that containerd can be enabled as cri, looks like we will need to handle ourselves to enable cgroup in /etc/containerd/config.toml

Issue #2988, opened 2020-12-24
make install-build-deps fails


pradeepnnv:
**What happened?**
When executing `make install-build-deps`, it failed with below error message.

```
make install-build-deps

install-build-deps.sh
make: install-build-deps.sh: No such file or directory
make: *** [install-build-deps] Error 1
```

**What you expected to happen?**
Make command should've completed successfully.

**How to reproduce it?**
Checkout master branch and execute `make install-build-deps` command.

**Anything else we need to know?**
N/A

PR #2989, opened 2020-12-24
Fix issue with makefile

pradeepnnv:
Update path of install-build-deps.sh in install-build-deps target

FIxes https://github.com/weaveworks/eksctl/issues/2988

Signed-off-by: pradeepnnv <pradeepnnv@gmail.com>


michaelbeaumont:
Hi, this is an issue with https://github.com/weaveworks/eksctl/blob/ae19dcf7791371b6be3c05c919775784d2e5876d/Makefile#L11 on OSX, can you please try instead to address setting `PATH` on OSX?

pradeepnnv:
Thanks @michaelbeaumont.

I do see that `build/scripts` is added to the PATH variable but it's not used when the script is being executed.

As per this SFO link, https://stackoverflow.com/questions/25506047/setting-path-within-makefile-on-mac-os-x-but-it-works-on-linux, looks like we need to add SHELL variable for it to work properly.

This is one option I found that's working.

```
install-build-deps: ## Install dependencies (packages and tools)
	SHELL=$(SHELL) build-image-manifest.sh
```

Please review and let know.

michaelbeaumont:
@pradeepnnv I think it'd be best if the solution works for all targets/uses of `build/scripts` files.

Callisto13:
@pradeepnnv are you still working on this? If not, one of the team can finish it off, it's a good fix to have 😃 .

pradeepnnv:
Hi @Callisto13 , yes I'll update the PR today. Hopefully the fix I proposed above can be used for all uses of the scripts files.

Callisto13:
Great 🎉  !

pradeepnnv:
@Callisto13 , @michaelbeaumont PR is updated. Please review.

pradeepnnv:
One more thing...I did not test other targets. I wasn't sure I can try those release related targets.

Callisto13:
👍  no problem, we can verify that they work

michaelbeaumont:
@pradeepnnv I meant adding something like `SHELL=/bin/bash` to the top of the file so that we don't have to worry about it at all when writing targets. Which according to https://stackoverflow.com/a/25506676 should work..

pradeepnnv:
Thanks @michaelbeaumont . Let me try that out. If that works, i'll adjust the PR accordingly.

pradeepnnv:
Setting SHELL at the beginning of the Makefile did not work.

pradeepnnv:
I checked couple of more examples and could not find any better way to fix this issue. 
Either we have to declare the full path to the script everywhere or have the SHELL variable in each target using the script.

Make file in kubernetes project seems to be using a similar fix. https://github.com/kubernetes/kubernetes/blob/master/build/root/Makefile#L204

Callisto13:
🤔  Is using the path less opaque to folks coming by? I have no strong opinions either way, @michaelbeaumont do you have a preference?

michaelbeaumont:
I don't think it's less opaque to use `PATH`, it's less noisy though. I personally like using `SHELL` everywhere, it keeps things uniform so that all targets are run with a shell, meaning fewer surprises. Unfortunately, that means remembering to add `SHELL` for all new targets. So we can just use the full path everywhere, please also remove the `PATH` assignment @pradeepnnv 
Thanks OSX...

Callisto13:
> I don't think it's less opaque to use PATH

Lol I meant using the full path, not the var. Happy with this 👍 

> Thanks OSX...

😆  linux FTW!

michaelbeaumont:
Err I of course meant I don't think it's _more_ opaque to use `PATH` :facepalm:

Callisto13:
Honestly I have lost track of what I was saying originally

pradeepnnv:
@michaelbeaumont , @Callisto13 I updated the makefile to use specific path.

Please review.

Callisto13 (approved):
👍  LGTM, thanks @pradeepnnv

Issue #2980, opened 2020-12-21
Better error messages when region not set


Callisto13:
**Why do you want this feature?**
Currently when you run various commands and forget to set the `--region` flag, the error is quite unhelpful

Example command:
```sh
$ eksctl get iamidentitymapping --cluster=foo
Error: fetching cluster status to determine operability: unable to describe cluster control plane: ResourceNotFoundException: No cluster found for name: foo.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "07485935-27ab-4839-a881-5ee6664cdaf3"
  },
  Message_: "No cluster found for name: foo."
}
```

**What feature/behavior/change do you want?**
It would be nice to wrap some of these errors with either:

a) a hint to try again with the region set
b) more elaborate error: `No cluster found for name: foo in region: <region>`

pradeepnnv:
I would like to work on this issue.

Callisto13:
Go for it @pradeepnnv  👍

pradeepnnv:
@Callisto13 , the error message "No cluster found for name: foo." is from AWS EKS API.

So I submitted a PR to update the error message in DescribeControlPlace function to include the region.

Btw, region is defaulting to us-west-2 if not set. Please review and let me know if any changes are required.

Callisto13:
Thanks @pradeepnnv 

> the error message "No cluster found for name: foo." is from AWS EKS API.

Makes sense.

> Btw, region is defaulting to us-west-2 if not set.

Indeed, hopefully we can help out people who do not know/forget this so they do not waste time debugging. 👍

PR #2990, opened 2020-12-24
Update error msg in DescribeControlPlane function

pradeepnnv:
Update the error message in
DescribeControlPlane function to
include region.

Fixes https://github.com/weaveworks/eksctl/issues/2980

Signed-off-by: pradeepnnv <pradeepnnv@gmail.com>

### Description

Included the region in the error message if cluster is not available in the region.

```
./eksctl get iamidentitymapping --cluster=foo
Error: unable to fetch cluster status to determine operability: Unable to describe cluster control plane in the region "us-east-1": ResourceNotFoundException: No cluster found for name: foo.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "f4b41b52-d86f-4005-94fa-ff35fb712e44"
  },
  Message_: "No cluster found for name: foo."
}
```

```
./eksctl get fargateprofile --cluster=foo
Error: Unable to describe cluster control plane in the region "us-east-1": ResourceNotFoundException: No cluster found for name: foo.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "5d6f934d-8121-480c-8d7f-5e288bb81549"
  },
  Message_: "No cluster found for name: foo."
}
```

### Checklist
- [ ] Manually tested


Callisto13:
Would you be able to also improve the other commands which do not return region in the error? `get cluster`, `get labels`, `get nodegroups` are some that I have found, there may be others.

Also what do you think about the UX of the errors/logs we return in these commands? Other commands return neater errors on lower logging levels, and most of them have a nice logging pattern where the region is mentioned, eg:

```sh
$ eksctl delete cluster foo
[ℹ]  eksctl version 0.34.0
[ℹ]  using region us-west-2
[ℹ]  deleting EKS cluster "foo"
Error: no eksctl-managed CloudFormation stacks found for "foo"
```

The logging pattern doesn't seem to be applied for `get` commands, I wonder if that was a deliberate decision 🤔 . It could be one way of communicating the region.

So 2 things to consider for this improvement:
1. Is just dumping out the error from these failures (`RespMetadata` etc) useful for users who have a low logging level set? Can/should we tidy that up?
2. Would it be nice, useful and consistent UX to add that nice logging to all commands?

pradeepnnv:
Sure @Callisto13. Let me review the other get commands and their error messages.

pradeepnnv:
@Callisto13 , I added the same logging available in delete to all get commands.

Commands `get addon`, `get cluster`, `get fargateprofile` and `get iamidentitymapping` are printing the region as expected.

```
./eksctl get iamidentitymapping  --cluster=foo

[ℹ]  eksctl version 0.37.0-dev+7a358ddf.2021-01-10T15:06:08Z
[ℹ]  using region us-west-2
Error: unable to fetch cluster status to determine operability: unable to describe cluster control plane: ResourceNotFoundException: No cluster found for name: foo.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "94a68e72-022a-4ebf-afbb-e50f34d05e72"
  },
  Message_: "No cluster found for name: foo."
}

   ./eksctl get cluster

[ℹ]  eksctl version 0.37.0-dev+7a358ddf.2021-01-10T15:06:08Z
[ℹ]  using region us-west-2
No clusters found
```

But the other get commands `get iamserviceaccount`, `get labels` and `get nodegroup` are not having the region value. Hence region is being printed as an empty string. Weirdly, this is the same case even if we pass the region as an argument.

```
 ./eksctl get iamserviceaccount  --cluster=foo --region=us-east-1
[ℹ]  eksctl version 0.37.0-dev+7a358ddf.2021-01-10T15:06:08Z
[ℹ]  using region
Error: unable to fetch cluster status to determine operability: unable to describe cluster control plane: ResourceNotFoundException: No cluster found for name: foo.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "a6663362-fdc0-4273-aa87-529087662386"
  },
  Message_: "No cluster found for name: foo."
}
```

Any idea what's happening here?

Callisto13:
Looking good 😃 .

🤔  It is probably because where you are doing `cmdutils.LogRegionAndVersionInfo(cmd.ClusterConfig.Metadata)` for those, the region has not been set in that Metadata yet (the flag has not been parsed into the config). Try moving it to after `cmd.NewCtl` and it should be on either the `ProviderConfig` or the `ClusterConfig`.

 Also, some commands (`get cluster` off the top of my head) have an "all" `-A` flag, in that case the logging behaviour may need to change a bit so that the region makes sense.

pradeepnnv:
Thanks. That helped. Now the region is printed in all get commands.

Also, I added a check to not print the region if all regions option is selected in get cluster command. That's the only one which accepts all regions flag.

Callisto13 (approved):
Awesome! Thanks @pradeepnnv 🎉

Issue #3005, opened 2020-12-30
kubelet.yaml set cgroupDriver: systemd instead of cgroupDriver: cgroupfs in AL2-GPU instances


DanielAmmar:
**What happened?**
launch unmanaged node group with p3.2xlarge gpu (ami-0f23f1b20f58cc97f)
however it failed to start -
```
systemctl status kubelet
● kubelet.service - Kubernetes Kubelet
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/kubelet.service.d
           └─10-eksclt.al2.conf
   Active: activating (auto-restart) (Result: exit-code) since Wed 2020-12-30 14:16:36 UTC; 4s ago
     Docs: https://github.com/kubernetes/kubernetes
  Process: 22376 ExecStart=/usr/bin/kubelet --node-ip=${NODE_IP} --node-labels=${NODE_LABELS},alpha.eksctl.io/instance-id=${INSTANCE_ID} --max-pods=${MAX_PODS} --register-node=true --register-with-taints=${NODE_TAINTS} --cloud-provider=aws --container-runtime=docker --network-plugin=cni --cni-bin-dir=/opt/cni/bin --cni-conf-dir=/etc/cni/net.d --pod-infra-container-image=${AWS_EKS_ECR_ACCOUNT}.dkr.ecr.${AWS_DEFAULT_REGION}.${AWS_SERVICES_DOMAIN}/eks/pause:3.3-eksbuild.1 --kubeconfig=/etc/eksctl/kubeconfig.yaml --config=/etc/eksctl/kubelet.yaml (code=exited, status=255)
  Process: 22365 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT -w 5 (code=exited, status=0/SUCCESS)
 Main PID: 22376 (code=exited, status=255)
```
error message:
`failed to run Kubelet: misconfiguration: kubelet cgroup driver: "systemd" is different from docker cgroup driver: "cgroupfs`

`cat /etc/eksctl/kubelet.yaml` points that `cgroupDriver: systemd` however I suspect it should be `cgroupDriver: cgroupfs`

docker cgroup in Amazon Linux 2 (GPU) is set to "cgroupfs" (vs. "systemd" in non GPU versions)

**How to reproduce it?**
launch gpu group node via eksctl v0.35.0

**Anything else we need to know?**
[...]

PR #3007, opened 2020-12-31
Set cgroupdriver to systemd on GPU nodes

Callisto13:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/3005. Maybe.

https://github.com/weaveworks/eksctl/pull/2962 set the cgroupdriver of both the docker daemon and the kubelet to systemd. Unfortunately, the GPU AMIs remove the `/etc/docker/daemon.json` file* and set things via flags. This meant that `0.35.0` broke things for people who want to create new node groups with GPU instances: the kubelet was set to use `systemd`, while the docker daemon (that file removed) defaulted to `cgroupfs`.

\* _which is a COMPLETE pain in the ass and I am VERY mad about it._

**There are 2 workarounds** (in the form of `preBootstrapCommands`) for users. This PR is a quick hack to get things working again, while we hopefully figure out something better OR wait for AWS to default to `systemd` in those accelerated AMIs (the standard ones are about to do this).

**Alternatives** to simply sedding the flag into that `OPTIONS` string:
- Provide our own `/etc/sysconfig/docker` file
- Provide our own `/etc/systemd/system/docker.service.d/override.conf`
- Write to that file a bit nicer than using `sed` 🤷‍♀️ 

The first 2 run the risk of getting out of step with the originals. The 3rd may be more trouble than it is worth.

I have asked AWS to ensure those accelerated AMIs set the cgroup driver to `systemd` by default, so we could just leave the hack in while we wait. Don't know how long that will be tho... (The standard AMIs are about to do `systemd` by default, but a different builder is used for GPU ones.)

**I am open to any and all suggestions.**

More information can be found here https://github.com/weaveworks/eksctl/issues/3005#issuecomment-752955957

I am gonna add an integration test as part of regression testing.

(I also simplified some other stuff around writing config files, but can take or leave that.)

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/nodebootstrap/userdata.go
+++ b/pkg/nodebootstrap/userdata.go
@@ -26,11 +26,11 @@ const (
 )
 
 type configFile struct {
-	content string
-	isAsset bool
+	dir      string
+	name     string
+	contents string
+	isAsset  bool
 }
-
-type configFiles = map[string]map[string]configFile
 
 func getAsset(name string) (string, error) {
 	data, err := Asset(name)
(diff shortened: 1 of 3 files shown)


aclevername (approved):
LGTM (I haven't tested the `sed` command myself)

An integration test would be a great way to ensure we don't break this again in the future! 😄

Callisto13:
👍  I will do the integration test separately bc I want to get a release out today

michaelbeaumont (approved):

cPu1 (approved):

PR #2969, opened 2020-12-18
Add support for deleting non-eksctl clusters

aclevername:
### Description
Related issue https://github.com/weaveworks/eksctl/issues/2812

### Checklist
- [ ] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `userdocs` directory)
- [ ] Manually tested
- [ ] Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)
- [ ] Make sure the title of the PR is a good description that can go into the release notes

PR #3023, opened 2021-01-04
Show progress of update

aclevername:
### Description

Show the status of the update every 20s (too frequent?)
- Lets the user see the status
- They know the cli hasn't hung

New output:
```
eksctl upgrade cluster --name jk-2 --approve
[ℹ]  eksctl version 0.36.0-dev+ce6cd153.2021-01-04T12:22:49Z
[ℹ]  using region us-west-2
[ℹ]  will upgrade cluster "jk-2" control plane from current version "1.17" to "1.18"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
[ℹ]  waiting for cluster "jk-2" to be updated, current status: "InProgress"
```

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/utils/waiters/waiters.go
+++ b/pkg/utils/waiters/waiters.go
@@ -43,7 +43,7 @@ func makeWaiter(ctx context.Context, name, msg string, acceptors []request.Waite
 		Delay:       makeWaiterDelay(),
 		Acceptors:   acceptors,
 		NewRequest: func(_ []request.Option) (*request.Request, error) {
-			logger.Debug(msg)
+			logger.Info(msg)
 			req := newRequest()
 			req.SetContext(ctx)
 			return req, nil


pkg/utils/waiters/waiters.go
```diff
@@ -12,6 +12,31 @@ import (
 	"github.com/pkg/errors"
 )
 
+func WaitForCondition(waitTimeout time.Duration, returnErr error, condition func() (bool, error)) error {
```
michaelbeaumont, line 15: do we need another way to do waiting? For example, the existing `Wait` uses a `Context` for the requests which this one doesn't as far as I can tell.

aclevername: Yeah I didn't feel great about adding another method for waiting. This implementation is actually copied from an existing place where we do this and I've already made this refactor as part of another Pr https://github.com/weaveworks/eksctl/pull/2969/files#diff-cbedd1f78dec611ba48d5aad52b557825e9997955c502748b9c9cb3d09979243L418-R429

The existing Wait doesn't provide me with a hook to be able to show progress during the wait. The only thing you could do is add inside your `NewRequest` func the getting of the status and logging, but it feels a bit hackey as thats not what the func is designed for. I can't add new functionality to the `Wait` as its part of `aws-sdk-go`.

pkg/eks/update.go
```diff
@@ -206,27 +205,28 @@ func (c *ClusterProvider) UpdateClusterVersionBlocking(cfg *api.ClusterConfig) e
...
 		input := &awseks.DescribeUpdateInput{
 			Name:     &clusterName,
 			UpdateId: update.Id,
 		}
-		req, _ := c.Provider.EKS().DescribeUpdateRequest(input)
-		return req
-	}
+		out, err := c.Provider.EKS().DescribeUpdate(input)
+		if err != nil {
+			return false, fmt.Errorf("failed to check update status: %v", err)
```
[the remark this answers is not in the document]
aclevername: 👍

pkg/utils/waiters/waiters.go
michaelbeaumont: The current `Wait()` already shows progress during the `NewRequest` func (albeit on debug level):

https://github.com/weaveworks/eksctl/blob/5a075b17acbbc2e81da1161c2fd557800880c007/pkg/utils/waiters/waiters.go#L46-L49

Can we just switch that to info?

aclevername: that only debugs a static message, the current implementation get the current status of the update and shows that. We could move that logic into the NewRequest but then we are overloading it more and making the same request twice, once in the `NewRequest` call and again afterwards when it uses the returned request to make the same call.

michaelbeaumont: I'm not sure if it's necessary to show the current status, isn't it mostly the same as the action we're taking unless it's failed (where we show an error) or successful (at which point it's done)? I.e. the message will always be something like "Waiting on {delete,create,update} of resource, current status: {deleting,creating,updating}"

aclevername: I think it's worth showing the status, also its possible new statuses will appear in the future

michaelbeaumont: We are showing the status in the message, we're just doing it twice. I don't think it's worth adding new code to `eksctl` (which isn't equal to the current code) for something we don't need right now. I also can't imagine a situation where there's a new status that wouldn't end the waiting (either through failure or success). Perhaps it's worth a PR to `aws-sdk-go`?

aclevername: Sorry I don't think I was super clear in my first comment, this wait logic already exists in the codebase in an unrefactored form, as part of https://github.com/weaveworks/eksctl/pull/2969 I refactored it out and re-used:
original usage https://github.com/weaveworks/eksctl/pull/2969/files#diff-cbedd1f78dec611ba48d5aad52b557825e9997955c502748b9c9cb3d09979243L418-R427

new usage:
- https://github.com/weaveworks/eksctl/pull/2969/files#diff-fa7f0dfa823f7f99001cd3d411d89673edec1d379968dda30880f7d13153feceR208-R224
- https://github.com/weaveworks/eksctl/pull/2969/files#diff-fa7f0dfa823f7f99001cd3d411d89673edec1d379968dda30880f7d13153feceR162-R180

This PR is just copying that as its not merged yet. 

> We are showing the status in the message, we're just doing it twice

we currently just printing the debug message that "something that isn't a failure is occurring", which I agree to a certain extent because we match on the other states is somewhat equivalent to saying its in progress, its a slight difference. Its probably worth removing the debug message if we stick with this instead

>  Perhaps it's worth a PR to aws-sdk-go

thats a good idea, but I still think we'd need this Wait logic for the above cases anyway

I'm also slightly biased that I prefer this way of waiting rather than using the aws-sdk-go wait functionality as this seems simpler to read and gives us more flexibility.

all that being said this isn't a hill I'm willing to die on 😄 happy for this PR just to be changing it from debug -> info

Callisto13: Is now a good time to derail this conversation and say a timestamp in each log line is way more useful than anything else :trollface:  ?

I do think this hook is cute tho. It may not be massively useful in the context of this PR (and i do agree that any status other than `InProgress` will likely be a failure) and it is expanding the use of that existing custom waiter (make it possible to do more that just hit the kube api) but where it is used in the [delete one](https://github.com/weaveworks/eksctl/pull/2969/files#diff-fa7f0dfa823f7f99001cd3d411d89673edec1d379968dda30880f7d13153feceR208-R225) is quite useful UX. That said, I would prefer it if we didn't have a mix of waiters. While this addition to the former kube clientset waiter is cool, it does now just look like the other sdk waiter, and we should pick one. 

TLDR: I am happy with the existing custom waiter being refactored somewhere sensible, and I like the addition (more in the context of the delete PR than here), but it has turned a different waiter communicating with a different thing into a similar waiter communicating with a variety of things. It is more flexible that the aws-sdk waiter, getting a PR into aws-sdk would be preferable long term at least 🤷‍♀️ .

Callisto13: Well, i can see I have been no help at all.

TIMESTAMPS 🤘 :trollface: 🙃

michaelbeaumont: yeah I agree with that in the general case. I think we should just use one or the other, but if we write it ourselves, it should be correct and there is a real cost to maintaining our own version.

However at the risk of repeating myself, showing the raw status (which will always be "Updating", correct me if I'm wrong) isn't useful to the user and is potentially confusing. Sometimes the "Updating" status doesn't even make sense in the context of the operation the user requested, it's an implementation detail, which is why it's important to show our own "x process is in progress".

aclevername: Sounds good 👍 

To summaries the actions I'm going to do (feel free to correct)
- This PR is going to change to just being an update from debug log -> info log. 1 line change
- The conversation around if we need this new Waiter can be moved into the delete clusters story, as its most valued there
- Depending on the outcome of the conversation in the above,  Add an issue to our backlog for opening a PR on the aws-sdk for having a customisable hook in the `Wait` functionality

aclevername:
new output:
```
./eksctl upgrade cluster --name jk --approve
[ℹ]  eksctl version 0.37.0-dev+2f52a681.2021-01-05T17:14:07Z
[ℹ]  using region us-west-2
[ℹ]  will upgrade cluster "jk" control plane from current version "1.17" to "1.18"
[ℹ]  waiting for requested "VersionUpdate" in cluster "jk" to succeed
[ℹ]  waiting for requested "VersionUpdate" in cluster "jk" to succeed
[ℹ]  waiting for requested "VersionUpdate" in cluster "jk" to succeed
```

michaelbeaumont (approved):
> new output:

:rocket: next step for everyone, remove `"VersionUpdate"` and similar user-hostile strings from the messages

Issue #3015, opened 2021-01-04
Add option to add cluster ingress rules to custom security groups


jalaziz:
**Why do you want this feature?**

eksctl already supports using existing security groups and IAM roles. This is great because we have a strict naming convention and this feature allows me to create the resources that cannot be changed after the cluster is created.

However, there are two security group ingress rules that eksctl normally creates that require the EKS cluster to be created first. These rules involve the cluster security group that is created by EKS. When specifying custom security groups, these ingress rules are not added automatically and must be added manually.

**What feature/behavior/change do you want?**

It would be great if eksctl could introduce a configuration option that would add the two ingress rules to an existing security group, if provided. 

The change seems to be simply moving the following block to work with a custom `sharedNodeSecurityGroup`: https://github.com/weaveworks/eksctl/blob/master/pkg/cfn/builder/vpc.go#L411-L431

I'd be happy to look into introducing the change if this is something that makes sense for the project.

Callisto13:
Thanks @jalaziz, this improvement makes sense to me.

If you want to PR something, that would be great 😃 . If not (or you end up not having time) let me know and the core team will handle it in the near future.

jalaziz:
Thanks @Callisto13!

I've started a draft PR at #3027. I still need to add tests and documentation, but hopefully the direction makes sense.

PR #3027, opened 2021-01-05
Add default security group rules to custom SGs

jalaziz:
### Description

Automatically add ingress rules to allow unmanaged nodes to communicate
with the cluster control plane and managed nodes when a custom shared
node security group is provided. Previously, these rules were only
created when the shared node security group was managed by eksctl.

To retain the previous behavior, we now provide a new
ManageDefaultSecurityGroupRules option which can be disabled.

Issue #3015

### Checklist
- [x] Manually tested


jalaziz:
I struggled to come up with a good name for the new option, so suggestions are welcome!

Also, enabling the option by default breaks the current behavior, but maybe it's acceptable since it should only affect new clusters?

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -97,6 +97,13 @@ type (
 		ExtraCIDRs []*ipnet.IPNet `json:"extraCIDRs,omitempty"`
 		// for pre-defined shared node SG
 		SharedNodeSecurityGroup string `json:"sharedNodeSecurityGroup,omitempty"`
+		// Automatically add security group rules to and from the default
+		// cluster security group and the shared node security group.
+		// This allows unmanagd nodes to communicate with the control plane
+		// and managed nodes.
+		// Defaults to `true`
+		// +optional
+		ManageDefaultSecurityGroupRules *bool `json:"manageDefaultSecurityGroupRules,omitempty"`
```
michaelbeaumont, line 106: I don't think users expect that `eksctl` adds ingress rules to their provided SGs by default. Could this be false by default? And can we rename it `ManagedSharedNodeSecurityGroup`?

michaelbeaumont (changes requested):
Thanks for the PR @jalaziz, just have one comment

pkg/apis/eksctl.io/v1alpha5/vpc.go
jalaziz: I thought about that, but faced a couple issues:

1. Without adding the rules, the provided security group is basically meaningless because you then have to manually find the created security group and add the rules yourself. The fundamental issue is that the cluster security group ID is not known until the EKS cluster is setup.

2. Making it false by default complicates the code a bit. We would effectively have three conditions instead of two. The default of generating a shared security group and adding the rules automatically, the default of providing the shared security group and NOT having the rules added automatically, and the non-default of providing the security group and enabling the setting to add the rules.

Since the default for the generated shared security group is to add the rules, it seemed like that should be the default for an explicitly specified group. Without adding the rules, specifying the security group doesn't do much other than provide access to unmanaged nodes.

I do agree that it's not great because it breaks current behavior, but it could be argued that the current behavior is faulty (especially when using managed nodes).

If you still think the default should be to not add the rules for an explicitly provided security group, I'd be happy to change that.

jalaziz: > I don't think users expect that `eksctl` adds ingress rules to their provided SGs by default. Could this be false by default? And can we rename it `ManagedSharedNodeSecurityGroup`?

As for renaming it. Are you suggesting that it should be a string of a boolean? I assume a bool since we already specify the security group ID elsewhere. If that's the case, the bool being true would indicate that the shared node security group would be managed by eksctl?

If so, that seems like it could be confusing to the user since the implicitly created shared node security group would be managed by default.

michaelbeaumont: Err wasn't 100% awake when I wrote that name apparently. Should be `ManageSharedNodeSecurityGroupRules` i.e. to match `SharedNodeSecurityGroup`

jalaziz: > Err wasn't 100% awake when I wrote that name apparently. Should be `ManageSharedNodeSecurityGroupRules` i.e. to match `VPC.SharedNodeSecurityGroup`
> 
> EDIT: Maybe something like `CreateManagedUnmanagedSecurityGroupRules`?

Ahh, makes more sense now, but still feels odd for that to be false by default when the current eksctl default is to manage the rules for the group it creates.

This is what I struggled with when trying to define the flag. Anything I came up with seemed long or confusing when I tried to differentiate it from the current default behavior.

It also seemed more symmetrical to have the flag work for both the implicitly created shared group and an explicitly provided one.

michaelbeaumont: No, I agree I think `true` by default makes sense here.

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -97,6 +97,13 @@ type (
 		ExtraCIDRs []*ipnet.IPNet `json:"extraCIDRs,omitempty"`
 		// for pre-defined shared node SG
 		SharedNodeSecurityGroup string `json:"sharedNodeSecurityGroup,omitempty"`
+		// Automatically add security group rules to and from the default
+		// cluster security group and the shared node security group.
+		// This allows unmanagd nodes to communicate with the control plane
+		// and managed nodes.
+		// Defaults to `true`
+		// +optional
+		ManageSharedNodeSecurityGroupRules *bool `json:"manageSharedNodeSecurityGroupRules,omitempty"`
```
cPu1, line 106: My preference would be to name this `DisableSharedNodeSecurityGroupRules` (or something else that signifies this disables a default feature) which is `false` by default, and only allow setting it if `VPC.SharedNodeSecurityGroup` is provided.

cPu1 (changes requested):
Thanks for the PR! I just have one comment:

This PR lets users disable attaching default security group rules by setting `ManageSharedNodeSecurityGroupRules` to false even when a custom security group is not provided. We shouldn't allow that because that'd prevent nodegroups from joining the cluster (unless there's a process in between cluster and nodegroup creation which adds the required SG rules).
Setting `ManageSharedNodeSecurityGroupRules` should only be allowed when `VPC.SecurityGroup` is provided.

jalaziz:
> Thanks for the PR! I just have one comment:
> 
> This PR lets users disable attaching default security group rules by setting `ManageSharedNodeSecurityGroupRules` to false even when a custom security group is not provided. We shouldn't allow that because that'd prevent nodegroups from joining the cluster (unless there's a process in between cluster and nodegroup creation which adds the required SG rules).
> 
> Setting `ManageSharedNodeSecurityGroupRules` should only be allowed when `VPC.SecurityGroup` is provided.

My thinking was that if you wanted to have eksctl create the shared security group but wanted to control the rules, you could use this option as well. It would be the same as if you specified your own security group today and had to manage the rules out of band. However, I do agree that it's a bit of a far fetched scenario.

pkg/apis/eksctl.io/v1alpha5/vpc.go
jalaziz: Yeah, that make sense! Would `DisableManagedSharedSecurityGroup Rules` be any better?

If we're saying that this should only matter if you're explicitly providing the shared node security group, then maybe it should actually remain as is and have a default of `false`? That would be more backwards compatible.

I realize I'm arguing against myself from earlier, but if we're breaking symmetry with the implicitly created group, we could keep things backwards compatible (we would have to test for both conditions anyway).

I still think it probably makes more sense to default to manage by default, but I'm a sample size of 1.

cPu1: >I realize I'm arguing against myself from earlier, but if we're breaking symmetry with the implicitly created group, we could keep things backwards compatible (we would have to test for both conditions anyway).


I kind of agree with your previous point about managing SG rules by default (when a custom SG is provided) at the expense of breaking backwards compatibility. If we're going this route, we should also log a message saying that the default behaviour has changed and SG rules will be added, and that it can be disabled by setting `disableManagedSharedSecurityGroup: true`

cPu1:
>My thinking was that if you wanted to have eksctl create the shared security group but wanted to control the rules, you could use this option as well. It would be the same as if you specified your own security group today and had to manage the rules out of band. However, I do agree that it's a bit of a far fetched scenario.

As you noted, it's not a common scenario, and by allowing more control over the SG rules eksctl creates, we'd be giving users more ways to break it, e.g., simply setting `ManageSharedNodeSecurityGroupRules: false` will prevent the nodegroups from joining.

jalaziz:
> >My thinking was that if you wanted to have eksctl create the shared security group but wanted to control the rules, you could use this option as well. It would be the same as if you specified your own security group today and had to manage the rules out of band. However, I do agree that it's a bit of a far fetched scenario.
> 
> As you noted, it's not a common scenario, and by allowing more control over the SG rules eksctl creates, we'd be giving users more ways to break it. For example, simply setting `ManageSharedNodeSecurityGroupRules: false` will prevent the nodegroups from joining. 

Yeah, entirely fair point. Will look into failing if the flag is set when not using the a custom shared security group.

pkg/apis/eksctl.io/v1alpha5/vpc.go
jalaziz: Any recommendation on where such a warning should live?

cPu1: I think it can go in the cluster task creation.

jalaziz:
@cPu1 I've added some validation but took a slightly different approach than discussed. The validation fails if `ManageSharedNodeSecurityGroupRules` is set to false and `SharedNodeSecurityGroup` is empty. This keeps the actual logic straightforward while preventing the bad configuration scenario.

I also added the warning if `ManageSharedNodeSecurityGroupRules` is enabled and `SharedNodeSecurityGroup` is set.

If everything looks good, I'll start looking into adding some tests.

jalaziz:
Just rebased onto master.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -123,6 +123,11 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
 		return errors.New("field secretsEncryption.keyARN is required for enabling secrets encryption")
 	}
 
+	// manageSharedNodeSecurityGroupRules cannot be disabled if using eksctl managed security groups
+	if cfg.VPC.SharedNodeSecurityGroup == "" && IsDisabled(cfg.VPC.ManageSharedNodeSecurityGroupRules) {
+		return errors.New("vpc.manageSharedNodeSecurityGroupRules must be enabled when using ekstcl-managed security groups")
```
cPu1, line 128: Minor nit:
```suggestion
		return errors.New("vpc.manageSharedNodeSecurityGroupRules must be enabled when using eksctl-managed security groups")
```

cPu1:
>If everything looks good, I'll start looking into adding some tests.

@jalaziz the changes look good to me! Please rebase and go ahead with adding tests.

aclevername:
@jalaziz is this PR ready for review? (its still in draft)

jalaziz:
> @jalaziz is this PR ready for review? (its still in draft)

@aclevername I was hoping to finish up the tests while on vacation but my laptop bricked on day 2. I'm finally back and getting everything restored, so I hope to get to this PR by the weekend.

jalaziz:
This PR should be largely ready. I've tested things manually and the various scenarios seem to work fine.

I added some basic defaults testing, but I could use some help/pointers for integration tests.

I looked for `sharedNodeSecurityGroup` tests, but didn't find any.

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -139,6 +139,14 @@ type (
 		ExtraCIDRs []*ipnet.IPNet `json:"extraCIDRs,omitempty"`
 		// for pre-defined shared node SG
 		SharedNodeSecurityGroup string `json:"sharedNodeSecurityGroup,omitempty"`
+		// Automatically add security group rules to and from the default
+		// cluster security group and the shared node security group.
+		// This allows unmanagd nodes to communicate with the control plane
```
cPu1, line 144:
```suggestion
		// This allows unmanaged nodes to communicate with the control plane
```

pkg/ctl/create/cluster.go
```diff
@@ -136,6 +136,11 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 		logger.Warning(api.ErrClusterEndpointPrivateOnly.Error())
 	}
 
+	// if using a custom shared node security group, warn that the rules are managed by default
+	if cfg.VPC.SharedNodeSecurityGroup != "" && api.IsEnabled(cfg.VPC.ManageSharedNodeSecurityGroupRules) {
+		logger.Warning("security group rules may be added by ekstcl; see vpc.manageSharedNodeSecurityGroupRules to disable this behavior")
```
cPu1, line 141:
```suggestion
		logger.Warning("security group rules may be added by eksctl; see vpc.manageSharedNodeSecurityGroupRules to disable this behavior")
```

userdocs/src/usage/vpc-networking.md
```diff
@@ -334,6 +334,33 @@ nodeGroups:
         clusterDNS: ["169.254.20.10","172.20.0.10"]
 ```
 
+## Custom Shared Node Security Group
+
+`eksctl` will create and manage a shared node security group that allows communication between
+unmanaged nodes and the cluster control place and managed nodes.
```
cPu1, line 340:
```suggestion
unmanaged nodes and the cluster control plane and managed nodes.
```

userdocs/src/usage/vpc-networking.md
```diff
@@ -334,6 +334,33 @@ nodeGroups:
...
+If you wish to provide your own custom security group instead, you may override the `sharedNodeSecurityGroup`
+field in the config file:
+
+
+```yaml
+vpc:
+  sharedNodeSecurityGroup: sg-0123456789
+```
+
+By default, when creating the cluster, `eksctl` will add rules to this security to allow communication to and
```
cPu1, line 351:
```suggestion
By default, when creating the cluster, `eksctl` will add rules to this security group to allow communication to and
```

cPu1 (changes requested):

cPu1 (approved):
LGTM :tada:

jalaziz:
Just rebased, squashed, and fixed the schema conflict.

Issue #2774, opened 2020-10-27
eksctl apply: cluster reconciliation


michaelbeaumont:
**Why do you want this feature?**
This is the umbrella issue concerning `eksctl apply`, which would support `eksctl apply -f config.yaml` to reconcile the current cluster with the given config (initially partially). Also sometimes known as `eksctl update`.

### Related issues:
#1497 
#462 (previous umbrella issue, closed because the history is confusing)
#20

Callisto13:
🎉  

This issue is going to track the delivery of a proposal for how we will implement gitops-style reconciliation with eksctl.

Deliverable is a proposal in `docs/` which details:
  1. The end state of `eksctl apply` (ie. ignoring everything else which currently exists, what would it look like if we could `apply` today?)
  1. What the rest of the `eksctl` UX looks like in a world with `apply` (ie. which flags, subcommands, etc would survive?)
  1. An implementation strategy: how we adapt the code to get from now to `apply` (design diagrams may be handy)
  1. A consideration of the risks and anything which may stop us from truly representing a cluster in config


Doing gitops reconciliation has been the goal of this project for a very long time, but the rollout was so gradual that the Plan was eventually forgotten as people moved off the project. We need to move incrementally, but let's aim to not be too incremental so that we forget the big picture.

Please add comments on what you would like to see in a proposal doc for this work, which questions you would like answered, etc

aclevername:
>1. The end state of eksctl apply (ie. ignoring everything else which currently exists, what would it look like if we could apply today?)

I think sticking to a relatively consistent state to the existing `kubectl apply` (so no sub commands afterwards) would be ideal, perhaps with some slight modifications (opinions/thoughts welcome):
- Keeping the dry run by default behaviour: using the existing `--approve` flag to state when you should actually create the resources. This differs from apply, but given the nature of eksctl and cost associated with creating clusters having a dry run by default feels nice. By default it could just show you the differ of the current world vs whats defined in the cluster config.
- Having the `--prune` functionality by default. Given the large majority of what we create is cloudformation stacks, finding all stacks that have been created that are no longer specified in the config isn't too complicated (regex name and/or tags). Might be some edge case resources where we don't use cloudformation.


> 2. What the rest of the `eksctl` UX looks like in a world with `apply` (ie. which flags, subcommands, etc would survive?)

I think this is an interesting one. Removing existing subcommands would presumably mean we are doing a breaking change, so a new major (1.0 🎉 ?)

[...]

PR #3037, opened 2021-01-06
Add first draft proposal for apply

michaelbeaumont:
### Description
See #2774 
[Rendered](https://github.com/weaveworks/eksctl/blob/apply/docs/proposal-007-apply.md)

### Checklist
- [ ] Manually tested


docs/proposal-007-apply.md
```diff
@@ -0,0 +1,137 @@
...
+and commands are unaffected. Continuing to support them shouldn't complicate
+the implementation too much as they are more or less subsets of
+operations performed by `apply`.
+
+## Questions to be answered
+
+-   Does the config represent the entire state of the config vs a partial
+    spec?
+    -   e.g. if I `apply` with a list of 1 nodegroups, are any other existing
+        nodegroups deleted?
```
michaelbeaumont, line 18: yes 😃 

Ideally by the end the config in general should be smarter, ie a user should be able to change any information and see it happen

_Originally posted by @Callisto13 in https://github.com/weaveworks/eksctl-private/pull/233#discussion_r552483539_

michaelbeaumont: I think so too like I wrote in the next section but it's not totally clear whether that's the right "default" (or should be the only option). For example [the `google` `terraform` provider supports IAM with 3 different levels of authoritativeness.](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_iam#google_storage_bucket_iam_policy#iam-policy-for-cloud-storage-bucket)

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,137 @@
...
+    -   do we also delete non-eksctl managed/created resources? i.e. does `eksctl`
+        own the cluster by default? (potential for configurability here)
```
michaelbeaumont, line 20: I don't think deleting would go well for us, but we will have find a new method of "adoption".

There would have to be a way to get an existing cluster's information and render it into a eksctl compatible config which then can be controlled via eksctl from then on.

It does get tricky if:
- Create non-eksctl cluster
- Get config, start controlling by eksctl
- Manually make "some change"
- What does eksctl do when it sees the new config? By then probably the same thing as it would if the cluster had been "owned" from the start.

_Originally posted by @Callisto13

michaelbeaumont: +1 being able to generate configs for unowned resources, I mentioned generating configs for imperatively created resources [here](https://github.com/weaveworks/eksctl-private/pull/233#discussion_r552489571). Expanding it to also get unowned is a great idea

_Originally posted by @aclevername_

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,137 @@
...
+    desiredCapacity: 1
+```
+
+we discover an existing nodegroup `ng-1` with the `eksctl.io/owned: true` tag (for example),
+do we delete it?
+
+The proposed answer is yes, it should be deleted. With `apply` we assume that all
+previous changes initiated by `eksctl` were initiated through the same config
+file/reconciliation process and that this nodegroup must have been
+deleted from this file.
```
michaelbeaumont, line 70: I think while the imperative commands are so easy and feature rich (you have flags for doing a lot of what you can do in the config file) this is probably going to lead to a lot of users accidentally deleting things because they made it imperatively. 

I think we can help reduce this by actively driving users towards using apply and away from the imperative commands by:
- Reducing the functionality of imperative commands. It reminds me of how kubectl's imperative commands are great for beginners and getting introduced to kubectl, but the moment you need to do anything more complex you are forced into a config file, which is great
- Being able to generate a config file for existing imperatively created resources. This might not be feasible, but for getting folks onboard to apply with existing clusters having a `eksctl get config` or something to that affect would be  great

I know you called out in the intro `Existing behavior and commands are unaffected`, but I couldn't help myself 😁

_Originally posted by @aclevername_

michaelbeaumont: This could be helped by not supporting imperative commands for a new config. Like once you specify nodegroups in v1alpha6 or whatever, you can no longer run create nodegroup. Or if we have a new kind for reconciliation.

I don't think removing functionality would go over well with anyone

docs/proposal-007-apply.md
michaelbeaumont: I have a vague memory of there being a risk here... didn't someone say in a call once that it is difficult (maybe impossible) to get a full cluster configuration?

_Originally posted by @Callisto13_

docs/proposal-007-apply.md
Callisto13: > This could be helped by not supporting imperative commands for a new config

+1 this

aclevername: > This could be helped by not supporting imperative commands for a new config

sounds good 💯 

> I don't think removing functionality would go over well with anyone

I'm definitely biased here, but using `apply` and config file to me is without a doubt the best way to use eksctl and something that we as a project should push users towards (the later might be a controversial opinion). "pushing" users towards that feels like something worth considering. A greater adoption of `apply` would also most likely also help get users into a space where using GitOps is more applicable and perhaps open a pathway for them using more of WW's tools. Reducing the imperative commands functionality is definitely a big thing and not something we should do lightly, but I don't think it should be dismissed.

Looking at it from an alternative point of view, if we view `v1.0` as a fresh start for eksctl, what would you want it to look like? I'm masively in favour of how kubectl handles this. It has barebone imperative commands for getting beginners active quickly and quickly pushes you towards config files for serious usage.

The above perhaps belongs in a different discussion, but I wanted to make sure my point made sense 😄

Callisto13: Completely agree. v1 let's us do away with all imperative commands in a socially acceptable way.

aclevername:
Once we have settled on this proposal do we want to open up a `v1` proposal to encapsulate this and other  (mainly breaking) changes we might want to make

Callisto13:
Can we add a note to explicitly call out that you should be able to define something in one part of the config, and use it in another (eg. define an iam role name, and use it elsewhere)? The goal is that you can define a full cluster everything will be created in the correct order, so this behaviour is implied, but probably good to call it out.

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,173 @@
...
+
+Proposed is creating a new API version, starting with `v1alpha6`, where, as we add
+`apply` support for cluster/nodegroup properties, we allow for aggressive reevaluation
+of the existing config structure that would hold these newly supported properties/aspects
+of the cluster/nodegroup.
+
+If we decide, _for example_, to add support for reconciling tags, we might
+decide it fits better as a top level key. We would then, with `v1alpha6`, remove
+`tags` under `metadata` and make it a top level field. It should be possible to
+automatically rewrite the previous version into the current version.
```
martina-if, line 157: Would the new eksctl apply work with clusters that were created with version v1alpha5?

michaelbeaumont: Yes, definitely!

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,173 @@
...
+the resources and won't be enumerated here.
+
+1. `apply` is put behind the explicit "experimental" flag (cf `enable profile/repo`)
+2. Take full ownership of `eksctl`-managed resources (i.e. delete missing resources by default)
+3. Options for non-`eksctl`-managed resources as well as resources where
+   ownership is unknowable are covered below
+4. Changing immutable fields through recreation is the default
+5. We gradually expand `apply` support to different parts of the cluster
+6. As we expand `apply` support, we reevaluate and update the config structure as
+   necessary in a new API version.
```
Callisto13, line 70: I think it would be good to spec out precisely what an entire cluster config looks like before we start. I feel one of the reasons the current config is incohesive is because it was patched together issue-by-issue

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,173 @@
...
+
+##### `eksctl` owns everything
+
+With this option, we assume that `eksctl` owns the set of such resources and
+delete any not listed in the config.
+
+##### `terraform`-style state
+
+Storing some kind of state/config "somewhere" is another option and has been discussed in
+https://github.com/weaveworks/eksctl/issues/642
```
Callisto13, line 124: I like the idea (from that linked thread) of representing the full cluster in a ConfigMap in the cluster itself (with a view to use a CRD in the future) and reconciling config changes against that diff. Would be good if we could get more information around the difficulties found in the original exploration and whether this needs to be a "thing" in and of itself:

> I'm removing this from the milestone, as MVP is not a simple as I hoped for. We should also create a design document for this actually.

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,173 @@
...
+decide it fits better as a top level key. We would then, with `v1alpha6`, remove
+`tags` under `metadata` and make it a top level field. It should be possible to
+automatically rewrite the previous version into the current version.
+
+#### Goals/justification for config refactors
+
+-   Ensure that the config makes sense in a world centered around reconciliation.
+-   Auto upgrade makes it easier for users to use an existing config
+-   **"Parse, don't validate"**: the code that uses the config should be insulated from how we
+    serialize/deserialize the config.
```
Callisto13, line 164: ❤️ ❤️ ❤️

docs/proposal-007-apply.md
michaelbeaumont: Before we start implementing perhaps, but IMO that's really the next step from this proposal. IMO the main reason the config is incohesive right now is simply because we can't/don't make breaking changes. Perhaps we could sketch out a tentative structure for the entire thing in terms of "let's think about which resources we're managing here" so we can think through cases for `apply` but I don't think we should wait for the whole entire config to be supported before releasing some version of `apply` and I think we shouldn't pretend we'll get it right the first time.

Callisto13: > we shouldn't pretend we'll get it right the first time.

👍 , totally. I was thinking more on "based on the resources we _know_ we're going to manage"

Callisto13:
could we have a table of contents at the top plz @michaelbeaumont ?

is it possible to also include a goals/non-goals bit at the top so it is super clear what the scope of this proposal and work is?

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,200 @@
...
+      * [non-eksctl-managed resources](#non-eksctl-managed-resources)
+   * [New config](#new-config)
+      * [Goals/justification for config refactors](#goalsjustification-for-config-refactors)
+* [Open questions](#open-questions)
+
+## Goals
+
+The purpose of `apply` is to subsume the various imperative commands
+`eksctl` provides now into one `apply` command that reconciles the actual state
+of an EKS cluster with the intended state as specified in the `eksctl` config file.
```
aclevername, line 31: do we want to call out a goal of `apply` being GitOpsable

docs/proposal-007-apply.md
```diff
@@ -0,0 +1,200 @@
...
+`apply` would reconcile desired with real-world state
+[as described in the motivation](#motivation).
+The exact details per resource will potentially differ depending on
+the resources and won't be enumerated here.
+
+1. `apply` is put behind the explicit "experimental" flag (cf `enable profile/repo`)
+2. Take full ownership of `eksctl`-managed resources (i.e. delete missing resources by default)
+3. Options for non-`eksctl`-managed resources as well as resources where
+   ownership is unknowable are covered below
+4. Changing immutable fields through recreation is the default
```
aclevername, line 94: This feels like something that would be quite big piece of work. An example would be disabling OIDC which isn't currently possible, do we also delete all of the IAMServiceAccounts as they wouldn't have been created in the first place if OIDC was false. 

A more explosive example would be if I specified a lower version of Kubernetes in my config would we delete the existing cluster and create a new cluster at the lower version, same for nodegroups etc

docs/proposal-007-apply.md
michaelbeaumont: first line

docs/proposal-007-apply.md
michaelbeaumont: you can disable OIDC (not through eksctl), but yes all dependent resources need to be altered/recreated. what else can we do? the user would have to fall back to imperative commands otherwise

docs/proposal-007-apply.md
michaelbeaumont: i don't think getting rid of imperative commands has much to do with this proposal/`eksctl apply`

docs/proposal-007-apply.md
michaelbeaumont: i think it follows from "reconciles the actual state of an EKS cluster with the intended state as specified in the `eksctl` config file." that it's "GitOpsable." and in fact I realize now even "gitops-compatible" is missing a good definition IMO, for it to be included here.

docs/proposal-007-apply.md
aclevername: We could have some immutable fields that we don't support changing through reaction. I don't think we should do it lightly, but some fields seem impractical to support.

docs/proposal-007-apply.md
aclevername: calling it out specifically in `goals` is worthwhile

docs/proposal-007-apply.md
michaelbeaumont: I think in that case the user will then have to just delete and recreate the resources themselves, I think the power of `apply` is that `eksctl` recognizes and handles that case. See [the example of nodegroups, a big motivator for `apply`](https://github.com/weaveworks/eksctl/issues/462) and [a duplicate issue](https://github.com/weaveworks/eksctl/issues/443). I don't think the user should "ever" have to fallback to imperative commands and I don't see any other recourse if they can't replace it. What should the user do if they can't recreate a resource because of a field.

aclevername: I think the `nodegroups` example makes complete sense to support, I'm not disagreeing, very few fields are worth restricting.

To be more specific about whats worth restricting I think config field changes that would invoke the deletion and recreation of the cluster itself are good examples. They feel very explosive and might not be what the user expects. They might not know that downgrading kubernetes version isn't possible on an existing cluster for example, or other fields like modifying the subnets a cluster uses. We do have the `apply --plan` to display that to them before the destruction occurs, is implementing more complex codepaths like this worth? I might be wrong, but it feels like a discussion worth having

michaelbeaumont: Yeah I was definitely assuming we have `apply --plan`, for sure. I don't think the codepath needs to be complex, I think we need only know, for each resource, on which other resources it depends.
With CAPA you can't make such changes, but it's running as a controller and there's no great way to handle the "approve" flow with custom resources, at the moment.

michaelbeaumont: How much we should consider the case of "`eksctl` as a controller" here I'm definitely not sure about and should be open to discussion IMO!

aclevername:
What are the req's for merging? I think this is a good baseline proposal, undoubtedly changes will be made over time but it feels like a good start to me

Callisto13:
> What are the req's for merging? I think this is a good baseline proposal, undoubtedly changes will be made over time but it feels like a good start to me

Yeh it does not need to be perfect now, we just need a baseline. Then we can open separate PRs to argue and embellish the finer points

Callisto13 (approved):
I am happy with this as a solid place to build on, nice work 🎉 .
Could you add a "Status: WIP" (or whatever) at the top to be clear on where we are at with it?
The next stage I guess would be to create individual tickets to address key points, questions, unknowns and extra sections, wdyt?

aclevername (approved):
LGTM

PR #3045, opened 2021-01-07
Add support for well known policies with IRSA

### LINKED ISSUE #2837: Addon Policy support for IRSA-enabled Service Accounts

I use [IAM Roles for Service Accounts](https://eksctl.io/usage/iamserviceaccounts/) rather than Instance Roles for better-constrained granting of permissions, and many of my IRSA `ServiceAccounts` are for features like Cluster Autoscaler, and eksctl already has an 'addon' describing the necessary policies for an _Instance Role_ to host that feature/application.

The use of IRSA is even more important for e.g., [Cluster Autoscaler on Fargate](https://github.com/kubernetes/autoscaler/issues/3008#issuecomment-717105714), as you are limited to one [Pod Execution Role](https://eksctl.io/usage/schema/#iam-fargatePodExecutionRoleARN) per fargate Profile, so you end up managing a bunch of Pod Execution Roles in eksctl, which _also_ don't support Addon Policies.

*Note*: I have not yet tested IRSA with Fargate myself, but running CA in Fargate is an improvement goal for my team's next EKS build, which is what reminded me of this issue. If IRSA doesn't work with Fargate, then I'd also want Addon Policy support for Fargate Profiles, although they don't even support `attachPolicy` so that is probably more work in eksctl and worth a separate FR.

A new key like [`withAddonPolicies`](https://eksctl.io/usage/schema/#nodeGroups-iam-withAddonPolicies) in the [`serviceAccounts` array](https://eksctl.io/usage/schema/#iam-serviceAccounts) is what I would expect to see.

I also suspect that `withAddonPolicy`, taking only a single policy, might be better, since using IRSA you probably only want one built-in policy per `ServiceAccount` anyway, as the whole point of IRSA is that you can run, e.g. Cluster Autoscaler and External DNS, each with their own `ServiceAccount` with only the IAM permissions _it_ needs.

```yaml
iam:
  serviceAccounts:
  - metadata:
      name: cluster-autoscaler
      namespace: kube-system
    withAddon: autoScaler
```

is much more succinct _and_ descriptive than the [documented example](https://eksctl.io/usage/iamserviceaccounts/):

```yaml
iam:
  serviceAccounts:
  - metadata:
      name: cluster-autoscaler
      namespace: kube-system
    attachPolicy:
      Version: "2012-10-17"
      Statement:
      - Effect: Allow
        Action:
        - "autoscaling:DescribeAutoScalingGroups"
        - "autoscaling:DescribeAutoScalingInstances"
        - "autoscaling:DescribeLaunchConfigurations"
        - "autoscaling:DescribeTags"
        - "autoscaling:SetDesiredCapacity"
        - "autoscaling:TerminateInstanceInAutoScalingGroup"
        Resource: '*'
```

and will also be covered when eksctl is updated to fix bugs or missing permissions in an addon.

As a specific example of such a benefit, the [`autoScaler` addon policy](https://github.com/weaveworks/eksctl/blob/b48a14044f599801e626d2781b635fa19b19088c/pkg/cfn/builder/iam_helper.go#L43-L55) includes `ec2:DescribeLaunchTemplateVersions` which is missing from the example in the documentation repeated above, and this difference has existed since 1cc92ee3fdeae5c9744ae667ad2e672202849b3e in mid 2019.

---

I'm surprised this isn't already a feature request. I had intended to try and implement it myself over the last few months, but time slipped away, and I guess I forgot to lodge an FR at the time when I was actively feeling the pain of copying large policy documents out of the addons code or service documentation into my eksctl config.

michaelbeaumont:
### Description
Fixes #2837 
I've only added these 5 for now. I want to confirm the rest of the existing `addonPolicies` make sense with IRSA

### Checklist
- [ ] Manually tested


aclevername:
can an example yaml be added please 😄

pkg/cfn/builder/iam_helper.go
```diff
@@ -18,6 +18,39 @@ type cfnTemplate interface {
...
+			customPolicy{Name: "PolicyExternalDNSChangeSet", Statements: changeSetStatements()},
+			customPolicy{Name: "PolicyExternalDNSHostedZones", Statements: externalDNSHostedZonesStatements()},
+		)
+	} else if wellKnownPolicies.CertManager {
+		policies = append(policies,
+			customPolicy{Name: "PolicyCertManagerChangeSet", Statements: changeSetStatements()},
+			customPolicy{Name: "PolicyCertManagerGetChange", Statements: certManagerGetChangeStatements()},
+		)
+		if wellKnownPolicies.ExternalDNS {
+			policies = append(policies, customPolicy{Name: "PolicyCertManagerHostedZones", Statements: certManagerHostedZonesStatements("route53:ListHostedZones", "route53:ListTagsForResource")})
```
TBBle, line 46: This can't be hit, because we're in an `else` after `if wellKnownPolicies.ExternalDNS` above.

pkg/cfn/template/iam.go
```diff
@@ -25,7 +25,7 @@ type IAMRole struct {
 	Path string `json:",omitempty"`
 
 	AssumeRolePolicyDocument MapOfInterfaces `json:",omitempty"`
-	ManagedPolicyArns        []string        `json:",omitempty"`
+	ManagedPolicyArns        []interface{}   `json:",omitempty"`
```
TBBle, line 28: I'm not clear why this is changed. Is it just because we're smuggling the return value from `makePolicyARN` out of `(rs *IAMServiceAccountResourceSet) AddAllResources()` inside this slice? Would it make sense to just stringify the value before adding it to `ManagedPolicyArns` instead?

(I'm not clear on what this `gfn` business is all about, so maybe this comment doesn't make sense.)

michaelbeaumont: Yes it's kind of a less than ideal situation because we do some cloud formation using our own types and some using goformation. It's here that they're being mixed. `makePolicyARN` uses Cloudformation intrinsics, which aren't serialized as strings. Although in this case it's probably not strictly necessary to use the intrinsic.

pkg/cfn/builder/iam_helper.go
michaelbeaumont: Yes, missed when going from handling elements of a list of policies to handling this struct.

examples/13-iamserviceaccounts.yaml
```diff
@@ -18,6 +18,11 @@ iam:
       labels: {aws-usage: "application"}
     attachPolicyARNs:
     - "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess"
+  - metadata:
+      name: aws-load-balancer-controller
+      namespace: kube-system
+    wellKnownPolicies:
```
aclevername, line 24: would `attachWellKnownPolicies` be better?

TBBle: I think it's a _little_ odd that this is would be a verb, when the general idea of declarative config suggests using nouns instead, to describe the resulting state, not how you get there.

`attachedWellKnownPolicies` perhaps, but that's pretty verbose, and only subtly different, so likely to lead to muscle-memory mistakes.

I think the various uses of `attach` in eksctl config stand out as verbs in a sea of nouns, but I guess using `attach` specifically is consistent here, e.g., we also have `securityGroups.attachIDs`, not `securityGroups.attachedIDs`.

michaelbeaumont: Yes, I was trying to avoid a verb but maybe `withWellKnownPolicies`?

pkg/cfn/builder/iam_helper.go
TBBle: I see this has been updated, but I think the logic has drifted slightly.

My understanding of the similar code in `createRole` is roughly:
* If CertManager or ExternalDNS, `changeSetStatements`
* If Certmanager, `certManagerGetChangeStatements`
* If CertManager and ExternalDNS, `certManagerHostedZonesStatements` + a couple of things ExternalDNS needs that aren't in that
* If CertManager but not ExternalDNS, `certManagerHostedZonesStatements`
* If ExternalDNS but not CertManager, `externalDNSHostedZonesStatements`.

This logic is currently
* If CertManager or ExternalDNS, `changeSetStatements`
* If CertManager, `certManagerGetChangeStatements`
* If CertManager and ExternalDNS, `certManagerHostedZonesStatements` + a couple of things ExternalDNS needs that aren't in that
* If CertManager and not ExternalDNS, `certManagerHostedZonesStatements`
* If ExternalDNS, `externalDNSHostedZonesStatements` _<==_ Applied even if CertManager.

`externalDNSHostedZonesStatements` and `certManagerHostedZonesStatements` both contain `route53:ListResourceRecordSets`, and the "couple of things" above are the other two statements from `externalDNSHostedZonesStatements`. So `externalDNSHostedZonesStatements` is a strict subset of "`certManagerHostedZonesStatements` + a couple of things ExternalDNS needs that aren't in that", so the former should only if the latter has not been applied.

To be fair, I don't know why the effort is made to avoid the overlap, unless there's an actual failure on the AWS API when a policy contains multiple statements for the same Effect/Resource/Action tuple? It just makes it harder to read and reason about.

michaelbeaumont: I've simplified that code now.

pkg/cfn/template/iam.go
michaelbeaumont: I'm going to leave the intrinsics because we use them for this all through out cloudformation

aclevername (approved):
LGTM

pkg/cfn/template/iam.go
michaelbeaumont: I'm going to leave the intrinsics because we use them elsewhere (with `makePolicyArns`) (and `interface{}` is sort of how things are done in `cfn/template` right now)

pkg/cfn/builder/iam_helper.go
michaelbeaumont: @TBBle please do take a look over the changes if you get a chance

examples/13-iamserviceaccounts.yaml
Callisto13: Would I be at risk of ruining everything by saying I am not a fan of `WellKnown`? `withCommonPolicies`?

pkg/cfn/builder/iam.go
```diff
@@ -224,10 +224,27 @@ func (rs *IAMServiceAccountResourceSet) AddAllResources() error {
...
+		}
+	}
 
 	roleRef := rs.template.NewResource("Role1", role)
 
+	for _, p := range wellKnownPolicies {
+		if p.Name != "" {
+			doc := cft.MakePolicyDocument(p.Statements...)
+			rs.template.AttachPolicy(p.Name, roleRef, doc)
+		}
```
Callisto13, line 245: Is it possible to move this up under line 236 in the other loop over `wellKnownPolicies`? Seems a shame to do it twice if we can get away with not.

michaelbeaumont: i've solved it by just returning a tuple of lists instead of a list of a sum type imitation.

examples/13-iamserviceaccounts.yaml
michaelbeaumont: the one problem with "common" is that it has two meanings, one "well known", the other "shared"...

Callisto13: yeh that's totally right. `withNamedPolicies`? although I suppose all policies have names.

pkg/cfn/builder/iam.go
Callisto13: looks good 👍

examples/13-iamserviceaccounts.yaml
aclevername: Struggling to find a better alternative to `WellKnownPolicies` so I'm happy to stick with that. I'm still unsure about `with` vs `attach`. I know that using the word `attach`  isn't technically correct, but IMO its worth using it to remain consistent with the other fields.

PR #3047, opened 2021-01-08
v1.0 proposal draft

aclevername:
### Description

Draft

[Formatted version](https://github.com/weaveworks/eksctl/blob/v1-proposal/docs/proposal-007-v1-changes.md)
### Checklist
- [ ] Manually tested

### DIFF
--- a/docs/proposal-007-v1-changes.md
+++ b/docs/proposal-007-v1-changes.md
@@ -25,6 +25,8 @@ This is a draft list of changes that could be made in a v1.0 release. Some of th
 
   - [`wait-nodes` command](https://github.com/weaveworks/eksctl/blob/2e3ce5b9d1fcc0bed960d88bfb4251da8479ab9f/pkg/ctl/utils/wait_nodes.go#L14-L16)
 
+  - [Fix typo in asset filename](https://github.com/weaveworks/eksctl/pull/3020)
+
 - New config spec. Needs its own proposal and will be heavily linked to [eksctl apply](https://github.com/weaveworks/eksctl/pull/3037)
 
 - Refactoring to consistently use or remove the `--approve` flag. This is used in some commands such as `eksctl upgrade cluster`
--- a/docs/proposal-007-v1-changes.md
+++ b/docs/proposal-007-v1-changes.md
@@ -1,0 +1,48 @@
+Authors: @weaveworks/eksctl-reviewers
+### Status: Design ongoing
+
+# v1.0 changes
+
+This proposal outlines a list of features and breaking changes to be included in the v1.0 release of eksctl. This
+proposal can be used as a place to reference other proposal for changes and should serve as an overarching document to
+see what will change in v1.0.
+
+### Draft list of changes
+This is a draft list of changes that could be made in a v1.0 release. Some of these may required their own proposals.
+
+- Introduction of [eksctl apply](https://github.com/weaveworks/eksctl/pull/3037)
+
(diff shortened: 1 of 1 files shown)


michaelbeaumont:
Should we tag the issues we think should be part of 1.0?

aclevername:
> Should we tag the issues we think should be part of 1.0?

sounds good! I've create the `area/v1.0` label

docs/proposal-007-v1-changes.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Goals
+- TODO
+
+## Non-Goals
+- TODO
+## Risks
+- Adoption of the new CLI version will be slow
+
+- Users will dislike migrating config file format
```
Callisto13, line 47: Probably a mitigation would be to do something fancy for them.  like we provide a translate subcommand

docs/proposal-007-v1-changes.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Community will ask us to maintain the older version
```
Callisto13, line 49: we'll need a nice clear timeline and be very strict on when we move from "actively developed" -> "maintenance" -> "EOL"

docs/proposal-007-v1-changes.md
```diff
@@ -0,0 +1,50 @@
...
+This proposal outlines a list of features and breaking changes to be included in the v1.0 release of eksctl. This
+proposal can be used as a place to reference other proposal for changes and should serve as an overarching document to
+see what will change in v1.0.
+
+### Draft list of changes
+This is a draft list of changes that could be made in a v1.0 release. Some of these may required their own proposals.
+
+- Introduction of [eksctl apply](https://github.com/weaveworks/eksctl/pull/3037)
+
+- Removal of deprecated functionality
```
Callisto13, line 15: with `apply` will all current subcommands and flags disappear?

michaelbeaumont:
Can't we just discuss everything on the issues themselves and tag them with `v1.0`? I.e. everything in this proposal is necessarily scoped to one issue.

Callisto13:
ah yeh you are right. we should use this proposal as just a grouping doc

aclevername:
> ah yeh you are right. we should use this proposal as just a grouping doc

I do propose that in the intro 😉  https://github.com/weaveworks/eksctl/pull/3047/files#diff-dc3c816fca42f7d3a40a3181f94cd419e321499ba38b0922b2b4be954046209dR6-R8

michaelbeaumont:
Do we need the proposal at all? What is there to propose about `v1.0` itself? If we assume it exists then it's defined by the issues that make it into `v1.0`.

Callisto13:
> I do propose that in the intro

Expecting me to actually read stuff was your first mistake

Callisto13:
> Do we need the proposal at all? What is there to propose about v1.0 itself? If we assume it exists then it's defined by the issues that make it into v1.0.

I like having a "thing" to group and link to for when people ask about v1, don't really mind what form it takes.

aclevername:
> > Do we need the proposal at all? What is there to propose about v1.0 itself? If we assume it exists then it's defined by the issues that make it into v1.0.
> 
> I like having a "thing" to group and link to for when people ask about v1, don't really mind what form it takes.

+ 1

I created it initially because we had a small mental list of things/break changing to go into `v1.0` and they were not being tracked. This proposal or another solution for tracking them and providing a space for community discussion is all we need really

PR #3057, opened 2021-01-11
Check if clusters are eksctl owned using stack name

aclevername:
### Description

Currently we look for the clusterName output on the CF stack, which doesn’t exist until the cluster is finished creating. Checking the stack name instead should ensure we always correctly detect if a stack belongs to a cluster or not.
### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -450,17 +450,21 @@ func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
 	return stacks, nil
 }
 
-func IsClusterStack(clusterName string, stacks []*Stack) bool {
+func (c *StackCollection) IsClusterStack() (bool, error) {
+	stacks, err := c.DescribeStacks()
+	if err != nil {
+		return false, err
+	}
 	for _, stack := range stacks {
-		if *stack.StackName == fmt.Sprintf("eksctl-%s-cluster", clusterName) {
+		if *stack.StackName == c.makeClusterStackName() {
 			for _, tag := range stack.Tags {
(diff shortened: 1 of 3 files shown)


pkg/cfn/manager/api.go
```diff
@@ -450,15 +450,21 @@ func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
...
+func (c *StackCollection) IsClusterStack() (bool, error) {
+	stacks, err := c.DescribeStacks()
+	if err != nil {
+		return false, err
+	}
 	for _, stack := range stacks {
-		for _, output := range stack.Outputs {
-			if *output.OutputKey == "ClusterStackName" {
-				return true
+		if *stack.StackName == c.makeClusterStackName() {
```
cPu1, line 459: Minor nit: `c.makeClusterStackName()` should be moved outside of the loop and assigned to a variable.

pkg/cfn/manager/api.go
```diff
@@ -450,15 +450,21 @@ func (c *StackCollection) DescribeStacks() ([]*Stack, error) {
...
+			for _, tag := range stack.Tags {
+				if *tag.Key == api.ClusterNameTag && *tag.Value == c.spec.Metadata.Name {
```
cPu1, line 461: `matchesClusterName` from this package also has this logic:

```suggestion
				if matchesClusterName(*tag.Key, *tag.Value, spec.Metadata.Name) {
```

cPu1 (approved):
LGTM

cPu1 (approved):

PR #3066, opened 2021-01-12
Flux v2 support

Callisto13:
**please nobody squash-merge this. i want small commits for this otherwise future us will be mad**

**also nobody click "Update branch". i hate that thing**

---

### New stuff:
- examples/12-gitops-toolkit.yaml
- pkg/actions/flux/enable.go
- pkg/actions/flux/enable_test.go
- pkg/apis/eksctl.io/v1alpha5/defaults.go
- pkg/apis/eksctl.io/v1alpha5/types.go
- pkg/ctl/cmdutils/configfile.go
- pkg/ctl/cmdutils/gitops.go
- pkg/ctl/enable/flux.go
- pkg/executor/executor.go
- pkg/flux/flux.go
- pkg/flux/flux_test.go
- pkg/gitops/gitops.go
- userdocs/src/usage/gitops.md
- pkg/ctl/enable/flux_test.go

### Reorganised/Refactored old stuff:
- pkg/actions/repo/enable.go
- pkg/actions/repo/enable_test.go
- pkg/actions/repo/wait.go
- pkg/ctl/enable/repo.go
- pkg/git/git_test.go

## Description

This PR lets users define Flux v2 (aka gitops toolkit) configuration in their cluster config. The config can be applied in one of two commands:

```sh
eksctl enable flux -f <file>  # install toolkit to existing cluster
eksctl create flux -f <file>  # flux2 will be installed in a new cluster after creation
```

As discussed in planning, I have introduced a new subcommand to `enable` because the flags for v2 in no way line up with those in v1, therefore simply adding a `--use-v2` style flag to the existing subcommand would have given a rubbish UX.

Edit: We landed on creating a new top-level config object since the old `git.` paths are soon to become redundant and have no relevance in the current plans for GitOps. `gitops` gives us room to add more toolkit options underneath later, and means we can cleanly build alongside `git` codepaths without interaction which will make removing things easier. There may be some confusion for users ("was the previous config _not_ flux/gitops?") but hopefully we can counter that will docs.

Implementation-wise, I went with simply shelling out to Flux. Ensuring that the binary is there is something we already insist on for the aws-cli and kubectl. Also, the Flux code doesn't wrap their bootstrap functionality in a library, and copy-pasting ~300 lines from there to here is worse.

---

### Flags:

There are **no additional flags** for this command (beyond `--config-file`). The reasons for this are:
1. Unlike Flux v1, bootstrapping Flux v2 components is very simple: just one command. The value in `eksctl` providing flags for v1 is that it turned a multi-step process into just 1 for users. There is no need for users to run one command with `eksctl`, with flags delegated to `flux` when they can just use `flux` directly. The value `eksctl` delivers here is in allowing users to specify a config file with the possibility of having it run automatically after a create, so that is what I have implemented. (If users want flags later we can do that for them maybe.)
2. We are moving towards a config-only world soon, and the project has been trying to avoid adding new flags for a while. With this command it seems to make sense to follow this.

---

### Config:

I have implemented the bare minimum of Flux2 config options and left many things to default in this first pass. We can expand and add more features as we need them. Whether we intend to 1:1 pass all config options through is a discussion for another time. 

The config file looks like this:

```yaml
gitops:
  flux:
    gitProvider: github             # required. options are github or gitlab
    owner: dr-who                   # required
    repository: my-cluster-gitops   # required
    personal: true                  # optional. if left false, assumes 'owner' is an org
    branch: main                    # optional
    namespace: "flux-system"        # optional
    path: "clusters/cluster-27"     # optional
```

---

### Refactoring:

I have moved some of the Flux v1 (aka `repo`) stuff around so that a) it fits with the `actions` organisation and b) the implementation sits alongside the new interface (I know it doesn't _need_ to but it is nice).

I have done some minimal refactoring and fixing of existing tests, but kept it to a minimum since the diff was already quite large (and that code will all be gone by the summer anyway).

---

### Still to come in other PRs:
- Integration tests
- Various missing improvements and configuration which has been deliberately left for another day

## Checklist
- [x] Manually tested
- [x] (Core team) Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)


Callisto13:
```
./eksctl enable toolkit -f test.yaml
[ℹ]  eksctl version 0.37.0-dev+b001ff82.2021-01-13T13:16:19Z
[ℹ]  ensuring v1 repo components not installed
[ℹ]  checking whether toolkit components already installed
[!]  helm-controller deployment was not found
[!]  kustomize-controller deployment was not found
[!]  notification-controller deployment was not found
[!]  source-controller deployment was not found
[ℹ]  running pre-flight checks
► checking prerequisites
✔ kubectl 1.19.4 >=1.18.0
✔ Kubernetes 1.18.9-eks-d1db3c >=1.16.0
✔ prerequisites checks passed
[ℹ]  bootstrapping gitops toolkit into cluster
► connecting to github.com
✔ repository created
✔ repository cloned
✚ generating manifests
✔ components manifests pushed
► installing components in flux-system namespace
namespace/flux-system created
networkpolicy.networking.k8s.io/allow-scraping created
networkpolicy.networking.k8s.io/allow-webhooks created
networkpolicy.networking.k8s.io/deny-ingress created
role.rbac.authorization.k8s.io/crd-controller-flux-system created
rolebinding.rbac.authorization.k8s.io/crd-controller-flux-system created
clusterrolebinding.rbac.authorization.k8s.io/cluster-reconciler-flux-system created
customresourcedefinition.apiextensions.k8s.io/buckets.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/gitrepositories.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmcharts.source.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/helmrepositories.source.toolkit.fluxcd.io configured
service/source-controller created
deployment.apps/source-controller created
customresourcedefinition.apiextensions.k8s.io/kustomizations.kustomize.toolkit.fluxcd.io configured
deployment.apps/kustomize-controller created
customresourcedefinition.apiextensions.k8s.io/helmreleases.helm.toolkit.fluxcd.io configured
deployment.apps/helm-controller created
customresourcedefinition.apiextensions.k8s.io/alerts.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/providers.notification.toolkit.fluxcd.io configured
customresourcedefinition.apiextensions.k8s.io/receivers.notification.toolkit.fluxcd.io configured
service/notification-controller created
service/webhook-receiver created
deployment.apps/notification-controller created
deployment "source-controller" successfully rolled out
deployment "kustomize-controller" successfully rolled out
deployment "helm-controller" successfully rolled out
deployment "notification-controller" successfully rolled out
✔ install completed
► configuring deploy key
✔ deploy key configured
► generating sync manifests
✔ sync manifests pushed
► applying sync manifests
◎ waiting for cluster sync
✔ bootstrap finished
[✔]  GitOps Toolkit installed successfully
[✔]  see https://toolkit.fluxcd.io/ for usage instructions
```

stefanprodan:
We do severals releases per week for flux2, unless eksctl plans to do the same to keep up with Flux changes I suggest we stick to using Flux CLI instead of a package. I don't expect users to just bootstrap a cluster, they will have to download Flux CLI anyway to build their GitOps pipelines.

Callisto13:
@stefanprodan We are using the CLI. We required users to have the CLI on the path, and then we shell out.

stefanprodan:
@Callisto13 maybe it's not clear from my comment, I did read the code changes in this PR and I'm in favour of using the Flux CLI. We could provide a bootstrap package but then eksctl would need to keep up with flux2 releases as well as Kubernetes.

Callisto13:
Sorry, morning brain!

So to clarify, you are saying to stick with the approach I have here because providing a library in Flux would in reality not be useful to us?

stefanprodan:
> I have here because providing a library in Flux would in reality not be useful to us?

This is not about usefulness but about velocity.

pkg/actions/repo/wait.go
```diff
@@ -0,0 +1,56 @@
...
+					return errors.Errorf("expected event type to be %T; got %T", &v1.Deployment{}, event.Object)
+				}
+				if deployment.Status.ReadyReplicas == deployment.Status.Replicas {
+					return nil
+				}
+			}
+		case <-timer.C:
+			return fmt.Errorf("timed out (after %v) waiting for deployment %q", timeout, name)
+		}
+	}
```
aclevername, line 55: can this be refactored to use [WaitForCondition?](https://github.com/weaveworks/eksctl/blob/f2b94cebfaa47126374ba01cd3ac686e03b035a5/pkg/utils/waiters/waiters.go#L15-L38). Example https://github.com/weaveworks/eksctl/blob/f2b94cebfaa47126374ba01cd3ac686e03b035a5/pkg/actions/cluster/unowned.go#L221-L237

pkg/actions/toolkit/enable.go
```diff
@@ -0,0 +1,123 @@
...
+
+//go:generate counterfeiter -o fakes/fake_flux_client.go . FluxClient
+type FluxClient interface {
+	PreFlight() error
+	Bootstrap(opts *api.Toolkit) error
+}
+
+type Manager struct {
+	Opts         *api.Toolkit
+	K8sClientSet kubeclient.Interface
```
aclevername, line 27: nit: I think we've just used `clientSet` elsewhere, is this just to differentiate `clientSet` and the client for flux? If so how about `KubeClient` to match `FluxClient`?

pkg/actions/toolkit/enable_test.go
```diff
@@ -0,0 +1,94 @@
...
+		fakeClientSet = fake.NewSimpleClientset()
+		fakeFluxClient = new(fakes.FakeFluxClient)
+		opts = &api.Git{Toolkit: &api.Toolkit{Namespace: "fluxv2"}}
+	})
+
+	JustBeforeEach(func() {
+		manager = &gitops.Manager{
+			K8sClientSet: fakeClientSet,
+			Opts:         opts.Toolkit,
+			FluxClient:   fakeFluxClient,
```
aclevername, line 36: big nit: rather than making the fields on the  struct public could you use an `export` pattern? Example: https://github.com/weaveworks/eksctl/blob/master/pkg/drain/export_test.go. This pattern is used a lot in the core library as a way of modify private fields for testing, e.g. https://golang.org/src/net/http/export_test.go

pkg/actions/toolkit/enable.go
```diff
@@ -0,0 +1,123 @@
...
+	m := &Manager{
+		Opts:         opts.Toolkit,
+		K8sClientSet: k8sClientSet,
+		FluxClient:   flux.NewClient(opts.Toolkit.AuthTokenPath, opts.Toolkit.GitProvider),
+	}
+
+	return m, nil
+}
+
+func (m *Manager) Run() error {
```
aclevername, line 45: `toolkitManager.Run()` doesn't really describe what its doing 😉

pkg/actions/toolkit/enable.go
```diff
@@ -0,0 +1,123 @@
...
+	logger.Info("ensuring v1 repo components not installed")
+	v1Namespace, err := m.checkV1()
+	if err != nil {
+		return errors.Wrap(err, "checking for flux v1 components")
+	}
+
+	if v1Namespace != "" {
+		logger.Warning("flux v1 components already installed in namespace %q. auto-migration not yet supported by eksctl. skipping installation", v1Namespace)
+		return nil
+	}
```
aclevername, line 57: could `checkV1` just return an error, wrap the logic for checking namespace value inside the check?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -803,10 +803,12 @@ func (n *NodeGroup) BaseNodeGroup() *NodeGroupBase {
 // cluster and linking it to a Git repository.
 // [Gitops Guide](/gitops-quickstart/)
 type Git struct {
-
 	// [Enable Repo](/usage/gitops/#installing-flux)
 	Repo *Repo `json:"repo,omitempty"`
 
+	// [Enable Toolkit](TODO docs link)
```
aclevername, line 809: TODO 👀

pkg/actions/repo/wait.go
Callisto13: see that thing i just said in slack

pkg/apis/eksctl.io/v1alpha5/types.go
Callisto13: that link wont exist until published

Callisto13: oh yeh

pkg/actions/repo/wait.go
aclevername: (she said no)

pkg/actions/toolkit/enable.go
Callisto13: oh lol i actually had it like that before. have no memory of how it got like this

pkg/actions/toolkit/enable.go
Callisto13: do you want a comment or a different func name?

pkg/gitops/gitops.go
```diff
@@ -10,43 +10,56 @@ import (
...
+		profilesSupported = false
+		logger.Info("gitops toolkit configuration detected, setting installer to Flux v2")
+	} else {
+		installer, err = repo.New(k8sRestConfig, k8sClientSet, cfg, timeout)
+		profilesSupported = true
+		logger.Info("gitops repo configuration detected, setting installer to Flux v1")
 	}
 
-	userInstructions, err := installer.Run(context.Background())
+	return installer, profilesSupported, err
```
aclevername, line 45: nit: private function below public

pkg/actions/toolkit/enable.go
aclevername: 🤕  I think it could be updated to be more descriptive, atm I don't really understanding what it does from the func name

pkg/actions/toolkit/enable.go
Callisto13: i swear i copied that from somewhere else 🤔 

but yeh i have no great attachment to it

pkg/actions/toolkit/enable.go
Callisto13: it makes slightly more sense in `installer.Run()`

Callisto13: `installer.Install()` gets me told off by the linter, as would `installer.InstallXComponents()` probably.

`installer.RunInstallation()`? I don't know if that is much better.

aclevername: `installer.Run` is good, before `manager.Run` sort of didn't really imply what was happening (what does telling the `manager` to `Run` do?), but I think telling the `installer` to `Run` shows its going to install 👌

pkg/actions/toolkit/enable.go
Callisto13: oh now i remember. we need to abort if something is found, but finding something is not an error so i don't want to just return the error (the logging is nicer), so it was this or use a bool.

pkg/actions/toolkit/enable.go
Callisto13: It was always meant to be `installer`, I just forgot to rename `manager` in that package 🙃

aclevername (approved):
LGTM 🎉

examples/27-gitops-toolkit.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+git:
+  toolkit:
+    gitProvider: github             # required. options are github or gitlab
+    owner: dr-who                   # required
+    repository: my-cluster-gitops   # required
+    personal: true                  # optional. if left false, assumes 'owner' is an org
+    branch: main                    # optional
+    namespace: "flux-system"        # optional
+    path: "flux/"                   # optional
```
stefanprodan, line 24: This needs to be `clusters/<cluster-name>/` since this path is used to define the cluster state not only Flux.

Callisto13: ah good to know, thanks 👍

aclevername:
@Callisto13 can you re-request a review when you've finished making changes 😄

stefanprodan:
There are are several options missing like private repos, ssh hostname and token-auth. You can bootstrap flux2 using token-based authentification, eksctl should support this option as many users have disabled SSH for GitLab on-prem. Others run Gitlab SSH server on a different hostname or on a different port.

Please check out `flux bootstrap github -h` and `flux bootstrap gitlab -h` to see all options.

Callisto13:
@stefanprodan for an MVP we have deliberately chosen to only enable the basic options. We are not sure if we want to have a 1:1 passthrough of all options, we may want to have an `additionalArgs` string as we did for the original enable repo flow.

pkg/ctl/enable/toolkit_test.go
```diff
@@ -0,0 +1,197 @@
...
+		It("should fail", func() {
+			cmd := mockEnableToolkitCmd("toolkit", "--cluster", "foo")
+			out, err := cmd.Execute()
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError("unknown flag: --cluster"))
+			Expect(out).To(ContainSubstring("Usage"))
+		})
+	})
+
+	When("--region is provided", func() {
```
michaelbeaumont, line 52: These feel like brittle/superfluous tests since we're explicitly declaring the incompatible flags in `cmd/cmdutils/gitops.go` and they should ideally be handled through a single interface. Do  these CLI option tests provide any value, what do we think?

pkg/actions/repo/enable.go
```diff
@@ -354,6 +374,96 @@ func mergeMaps(m1 map[string][]byte, m2 map[string][]byte) map[string][]byte {
...
+	}
+	err := portForward(namespace, "flux", 3030, "Flux", restConfig, cs, try)
+	return deployKey, err
+}
+
+type tryFunc func(rootURL string) error
+
+func portForward(namespace string, nameLabelValue string, port int, name string,
+	restConfig *rest.Config, cs kubeclient.Interface, try tryFunc) error {
+	fluxSelector := metav1.LabelSelector{
```
michaelbeaumont, line 421: very nitpicky but `gofmt` isn't opinionated enough here IMO, I usually do long argument list formatting like:

```
functionName(
    argument, argument, argument, argument, argument, argument,
    argument, argument, argument, argument, argument, argument, 
) error {
```
to separate the header from body but it's not very important either way

pkg/actions/toolkit/enable.go
```diff
@@ -0,0 +1,123 @@
...
+		if d.ObjectMeta.Name == "flux" {
+			return d.ObjectMeta.Namespace, nil
+		}
+	}
+
+	return "", nil
+}
+
+func (ti *Installer) checkInstallation() bool {
+	// TODO: this is the default set, we should maybe have an option to add whatever
```
michaelbeaumont, line 98: If we can punt on any the `TODO`s for this PR, should we leave them out of the code and make issues/track them in the main issue?

pkg/ctl/cmdutils/gitops.go
```diff
@@ -249,16 +250,114 @@ func (l *GitOpsConfigLoader) Load() error {
...
+	if l.cmd.ClusterConfig, err = eks.LoadConfigFromFile(l.cmd.ClusterConfigFile); err != nil {
+		return err
+	}
+
+	meta := l.cmd.ClusterConfig.Metadata
+	if meta == nil {
+		return ErrMustBeSet("metadata")
+	}
+
+	for f := range l.flagsIncompatibleWithConfigFile {
```
michaelbeaumont, line 266: granted, `cmdutils` is pretty sketchy, but can this be handled in one place in the code without too much refactoring?

pkg/executor/executor.go
```diff
@@ -22,14 +24,29 @@ func NewShellExecutor(envVars []string) Executor {
...
+	}
+
+	return cmd.Run()
+}
+
+func (e ShellExecutor) buildCmd(command string, args ...string) *exec.Cmd {
 	cmd := exec.Command(command, args...)
+
 	if len(e.envVars) > 0 {
 		cmd.Env = e.envVars
```
michaelbeaumont, line 46: unrelated refactoring but I feel like the `KEY=VALUE` string formatting should happen here and `envVars` should be `map[string]string`

pkg/actions/repo/enable.go
Callisto13: yep can change it. this was one of the things i moved from elsewhere

pkg/ctl/cmdutils/gitops.go
Callisto13: > `cmdutils` is pretty sketchy

biggest understatement ever 😂 

I think I intended to do something with several bits in that file, but then it fell off my radar (read: I got tired). Can definitely take another look at this bit tho.

pkg/actions/toolkit/enable.go
Callisto13: 👍  totally, is on my list to create issues for several "next-steps" and "missing pieces"

pkg/ctl/enable/toolkit_test.go
Callisto13: huh actually I don't even need to mark them as incompatible, since I have removed those flags altogether.

so yes, these tests are superfluous, but for a different reason 😄

pkg/executor/executor.go
Callisto13: I have no strong feelings about this, but have done it in the last commit. LMKWYT. (stretching the propriety of acronyms there)

pkg/actions/toolkit/enable_test.go
```diff
@@ -0,0 +1,95 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+var _ = Describe("Gitops", func() {
+	var (
+		fakeClientSet  *fake.Clientset
+		fakeFluxClient *fakes.FakeFluxClient
+		opts           *api.Git
+
```
aclevername, line 22: nit

Callisto13: the nit to end all nits

aclevername (approved):
LGTM 👍

pkg/actions/toolkit/enable_test.go
Callisto13: i am feeling generous so i'll allow it

stefanprodan:
Doesn't have to part of this PR, but we should add `fluxcd/tap/flux` to https://github.com/weaveworks/eksctl/blob/master/.goreleaser.brew.yml#L18

Callisto13:
@markramm are you happy with the subcommand on this?

Callisto13:
> Doesn't have to be part of this PR, but we should add fluxcd/tap/flux to https://github.com/weaveworks/eksctl/blob/master/.goreleaser.brew.yml#L18

oh yes good catch 👍

michaelbeaumont:
@Callisto13 With fluxv2 do any of the 3 existing fields under `git:` make any sense? We can probably just have `gitops` top level

Callisto13:
agreed 👍  as discussed in standup and with feedback from Mark Ramm; we are gonna abandon the `git` top level object and have a new `gitops` object.

So the config would look like:
```yaml
gitops:
  flux:
    gitProvider: github             # required. options are github or gitlab
    owner: dr-who                   # required
    repository: my-cluster-gitops   # required
    personal: true                  # optional. if left false, assumes 'owner' is an org
    branch: main                    # optional
    namespace: "flux-system"        # optional
    path: "clusters/cluster-27"     # optional
```

With the command being:
```sh
eksctl enable flux -f config.yaml
```

While `enable flux` may be confusing to users, it makes more sense in the context of a future Weave Gitops world. Eventually users would be able to call `enable` on other elements of WGO, or call `enable gitops` to install the entire configuration under that object.
The `git` object and everything below will be deprecated and removed along with the Repo/BootstrapProfiles implementation later in the year.

Gonna mark as a draft while I move over. I wont be changing any logic, just moving types around and renaming.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -604,6 +604,9 @@ type ClusterConfig struct {
 
 	// +optional
 	Git *Git `json:"git,omitempty"`
+
+	// +optional
+	GitOps *GitOps `json:"gitops,omitempty"`
```
aclevername, line 609: worth a comment to distinguish the two

aclevername (approved):
LGTM

PR #3093, opened 2021-01-18
add git token to stale github actions workflow

aclevername:
https://github.com/weaveworks/eksctl/actions/runs/493766615

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/stale@v3
       with:
-        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        repo-token: ${{ secrets.WEAVEWORKSBOT_TOKEN }}
         stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
         close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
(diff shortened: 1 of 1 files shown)


michaelbeaumont (approved):
Looks like it's optional in `master` (and in the docs at actions/stale) but they haven't made a release in 3 months :roll_eyes:

.github/workflows/stale.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     steps:
     - uses: actions/stale@v3
       with:
+        repo-token: ${{ secrets.WEAVEWORKSBOT_TOKEN }}
```
michaelbeaumont, line 13:
```suggestion
        repo-token: ${{ secrets.GITHUB_TOKEN }}
```

This is automatically included with all action runs and it should have write permission

michaelbeaumont (changes requested):

.github/workflows/stale.yml
aclevername: ah good to know

michaelbeaumont (approved):

.github/workflows/stale.yml
michaelbeaumont: the `WEAVEWORKSBOT_TOKEN` is necessary sometimes because actions performed by the actions `GITHUB_TOKEN` can't trigger other actions, like check runs for example.

aclevername: good to know! thanks

Issue #3102, opened 2021-01-18
panic: runtime error: invalid memory address with eksctl upgrade nodegroup 


jmolero-cortical:
**What were you trying to accomplish?**

Trying to change the launch template version of a node group in a EKS cluster NOT created with eksctl

**What happened?**

As reported at #2174 and similar to #3092. When using the command:

`eksctl upgrade nodegroup --name=NG  --cluster=testcluster  --kubernetes-version 1.18  --launch-template-version 2`

The output is the following:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2d8c2b3]

goroutine 1 [running]:
github.com/weaveworks/eksctl/pkg/actions/nodegroup.(*Manager).upgradeAndWait(0xc0006379e0, 0x7ffeefbffb60, 0x13, 0x7ffeefbffba5, 0x4, 0x7ffeefbffbc4, 0x1, 0x0, 0x0, 0x0)
	github.com/weaveworks/eksctl/pkg/actions/nodegroup/upgrade.go:60 +0x753
github.com/weaveworks/eksctl/pkg/actions/nodegroup.(*Manager).Upgrade(0xc0006379e0, 0x7ffeefbffb60, 0x13, 0x7ffeefbffba5, 0x4, 0x7ffeefbffbc4, 0x1, 0x0, 0xc0001ca8b0, 0xc000240d90)
	github.com/weaveworks/eksctl/pkg/actions/nodegroup/upgrade.go:38 +0x3e6
github.com/weaveworks/eksctl/pkg/ctl/upgrade.upgradeNodeGroup(0xc0006274a0, 0x7ffeefbffb60, 0x13, 0x7ffeefbffba5, 0x4, 0x7ffeefbffbc4, 0x1, 0x0, 0x1179565, 0xc0000d8400)
	github.com/weaveworks/eksctl/pkg/ctl/upgrade/nodegroup.go:86 +0x311
github.com/weaveworks/eksctl/pkg/ctl/upgrade.upgradeNodeGroupCmd.func1(0xc0000da000, 0xc00009cc60, 0x0, 0x6, 0x0, 0x0)
	github.com/weaveworks/eksctl/pkg/ctl/upgrade/nodegroup.go:27 +0xef
[...]

Issue #3092, opened 2021-01-18
Error: invalid Kubernetes version: strconv.ParseUint: parsing "": invalid syntax


cipriantarta:
**What were you trying to accomplish?**

I'm trying to upgrade one of the managed node groups with the following command
`eksctl -p *** upgrade nodegroup -f eksctl/dev.yaml --cluster *** system`
(Note, stars for actual names due to privacy reasons)

**What happened?**

Nothing happens, but the error in the title is returned.

**How to reproduce it?**

Just running the above command reproduces it 100% of the time.
**Logs**

**Anything else we need to know?**

**Versions**

```
❯ eksctl version
0.36.0

❯ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-15T11:35:50Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

My guess is that it has something to do with the minor version being `18+`?

aclevername:
@cipriantarta thanks for reporting this. It appears to be a bug introduced as part of recent work on the upgrade command. Previously the `--kubernetes-version` flag  was optional, this accidentally became mandatory as part of https://github.com/weaveworks/eksctl/pull/2953. I'll ensure this gets fixed ASAP.

I think a workaround for now will be to provide the `--kubernetes-version` flag with your current nodegroup version as the value. Can you try that and see if that works for the meantime? Thanks

cipriantarta:
@aclevername The workaround works. Thanks!

aclevername:
Fixed in https://github.com/weaveworks/eksctl/releases/tag/0.36.1

PR #3104, opened 2021-01-19
Make kubernetes-version optional and improve launch template validation for upgrade nodegroup

aclevername:
### Description
Improves the UX for this https://github.com/weaveworks/eksctl/issues/3102

Fixes this regression https://github.com/weaveworks/eksctl/issues/3092
- Previously it was always optional, see: https://github.com/weaveworks/eksctl/blob/6cd4d4270c344f6eb1aa2c2c37397f5d7e67d203/pkg/managed/service.go#L158-L162
### Checklist
- [ ] Manually tested
- [ ] (Core team) Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)

### DIFF
--- a/pkg/actions/nodegroup/upgrade.go
+++ b/pkg/actions/nodegroup/upgrade.go
@@ -3,6 +3,8 @@ package nodegroup
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
+
 	"github.com/pkg/errors"
 
 	"github.com/aws/aws-sdk-go/aws/request"
@@ -21,8 +23,10 @@ func (m *Manager) Upgrade(nodeGroupName, version, launchTemplateVersion string, 
 		return err
 	}
 
-	if _, err := semver.ParseTolerant(version); err != nil {
-		return errors.Wrap(err, "invalid Kubernetes version")
+	if version != "" {
+		if _, err := semver.ParseTolerant(version); err != nil {
+			return errors.Wrap(err, "invalid Kubernetes version")
+		}
 	}
 
 	if hasStacks {
@@ -46,19 +50,40 @@ func (m *Manager) upgradeAndWait(nodeGroupName, version, launchTemplateVersion s
 		Version:       &version,
 	}
 
+	describeNodegroupOutput, err := m.ctl.Provider.EKS().DescribeNodegroup(&eks.DescribeNodegroupInput{
+		ClusterName:   &m.cfg.Metadata.Name,
+		NodegroupName: &nodeGroupName,
+	})
+
+	if err != nil {
+		return err
+	}
+
 	if launchTemplateVersion != "" {
-		describeNodegroupOutput, err := m.ctl.Provider.EKS().DescribeNodegroup(&eks.DescribeNodegroupInput{
-			ClusterName:   &m.cfg.Metadata.Name,
-			NodegroupName: &nodeGroupName,
-		})
-		if err != nil {
-			return err
+		lt := describeNodegroupOutput.Nodegroup.LaunchTemplate
+		if lt == nil || (lt.Id == nil && lt.Name == nil) {
+			return errors.New("cannot update launch template version because the nodegroup is not configured to use one")
 		}
 
 		input.LaunchTemplate = &eks.LaunchTemplateSpecification{
 			Version: &launchTemplateVersion,
-			Id:      describeNodegroupOutput.Nodegroup.LaunchTemplate.Id,
 		}
+
+		if lt.Id != nil {
+			input.LaunchTemplate.Id = describeNodegroupOutput.Nodegroup.LaunchTemplate.Id
+		} else {
+			input.LaunchTemplate.Name = describeNodegroupOutput.Nodegroup.LaunchTemplate.Name
+
+		}
+	}
+
+	if version == "" {
+		// Use the current Kubernetes version
+		version, err := semver.ParseTolerant(*describeNodegroupOutput.Nodegroup.Version)
+		if err != nil {
+			return errors.Wrapf(err, "unexpected error parsing Kubernetes version %q", *describeNodegroupOutput.Nodegroup.Version)
+		}
+		input.Version = aws.String(fmt.Sprintf("%v.%v", version.Major, version.Minor))
 	}
 
 	upgradeResponse, err := m.ctl.Provider.EKS().UpdateNodegroupVersion(input)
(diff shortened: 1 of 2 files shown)


pkg/actions/nodegroup/upgrade.go
```diff
@@ -46,19 +50,39 @@ func (m *Manager) upgradeAndWait(nodeGroupName, version, launchTemplateVersion s
...
+				Id:      describeNodegroupOutput.Nodegroup.LaunchTemplate.Id,
+			}
+		} else if lt != nil && lt.Name != nil {
+			input.LaunchTemplate = &eks.LaunchTemplateSpecification{
+				Version: &launchTemplateVersion,
+				Name:    describeNodegroupOutput.Nodegroup.LaunchTemplate.Name,
+			}
+		} else {
+			return fmt.Errorf("cannot update launch template version because the nodegroup is not configured to use one")
 		}
```
cPu1, line 76: Very minor nit (just something I prefer) that can be ignored:

```suggestion
		if lt == nil || lt.Id == nil || lt.Name == nil {
			return errors.New("cannot update launch template version because the nodegroup is not configured to use one")
		}
		input.LaunchTemplate = &eks.LaunchTemplateSpecification{
			Version: &launchTemplateVersion,
			Id:      describeNodegroupOutput.Nodegroup.LaunchTemplate.Id,
			Name:    describeNodegroupOutput.Nodegroup.LaunchTemplate.Name,
		}
```

cPu1 (approved):
I have left a minor comment but otherwise LGTM.

pkg/actions/nodegroup/upgrade.go
aclevername: I like! 😄

aclevername: I like the refactor, but I think only one of ID or name need to be set?

cPu1: Yes, that makes sense.

cPu1 (approved):

pkg/actions/nodegroup/upgrade.go
```diff
@@ -46,19 +50,34 @@ func (m *Manager) upgradeAndWait(nodeGroupName, version, launchTemplateVersion s
...
 		input.LaunchTemplate = &eks.LaunchTemplateSpecification{
 			Version: &launchTemplateVersion,
 			Id:      describeNodegroupOutput.Nodegroup.LaunchTemplate.Id,
+			Name:    describeNodegroupOutput.Nodegroup.LaunchTemplate.Name,
+		}
+	}
+
+	if version == "" {
+		// Use the current Kubernetes version
+		version, err := semver.ParseTolerant(*describeNodegroupOutput.Nodegroup.Version)
```
michaelbeaumont, line 76: So in this case the version isn't changed? Should/can the behavior be the same as [`create nodegroup`](https://github.com/weaveworks/eksctl/blob/7d7a8150bed6ccb6f58f82f634cbf05c3d479b57/pkg/actions/nodegroup/create.go#L242)?

pkg/actions/nodegroup/upgrade.go
```diff
@@ -46,19 +50,34 @@ func (m *Manager) upgradeAndWait(nodeGroupName, version, launchTemplateVersion s
...
-			return err
+		lt := describeNodegroupOutput.Nodegroup.LaunchTemplate
+		if lt == nil || (lt.Id == nil && lt.Name == nil) {
+			return errors.New("cannot update launch template version because the nodegroup is not configured to use one")
 		}
-
 		input.LaunchTemplate = &eks.LaunchTemplateSpecification{
 			Version: &launchTemplateVersion,
 			Id:      describeNodegroupOutput.Nodegroup.LaunchTemplate.Id,
+			Name:    describeNodegroupOutput.Nodegroup.LaunchTemplate.Name,
```
michaelbeaumont, line 70: Just curious, why do we need to set `Name`?

aclevername: [one of the two need to be set](https://github.com/weaveworks/eksctl/pull/2953#discussion_r551967069), its unclear atm if this issue https://github.com/weaveworks/eksctl/issues/3102 is caused because the template only being referenced by the name in EKS, so I've added this as a precaution.

pkg/actions/nodegroup/upgrade.go
aclevername: yes the version doesn't get changed. I just copied this from how it used to be to preserve the UX. Whats the behaviour in `create nodegroup`?

michaelbeaumont: If the behavior before was to not update anything when `--version` is left out then it should stay that way I suppose.

I think the default with `create nodegroup` is to use the control plane version, so here I suppose the version would be updated to the control plane version? Like I imagine I should be able to write `upgrade nodegroup --cluster xx --name xx-ng` and it would pick the right version.

aclevername: Perhaps a good change we could make as part of `v1.0`? wdyt?

michaelbeaumont (approved):
@aclevername Does this definitely fix https://github.com/weaveworks/eksctl/issues/3102 then or are we not sure?

aclevername:
> @aclevername Does this definitely fix #3102 then or are we not sure?

I believe #3201 is just due to a user error, this PR will definitely improve the UX to prevent it from panicking. It does make me unsure if the `name` is needed, but I don't see any harm in keeping it in.

aclevername:
doing 1 more manual test before merging ⚠️

aclevername:
So it does appears to not like having both... very strange. I'm going to change it to use ID if its set, or name if its set.
```
./eksctl upgrade nodegroup --cluster jk-console --name ng-1 --launch-template-version 2
Error: InvalidParameterException: Either provide launch template ID or launch template name in the request.
{
  RespMetadata: {
    StatusCode: 400,
    RequestID: "fc59f526-90e0-4dc1-94c7-6aa3843f4ee2"
  },
  Message_: "Either provide launch template ID or launch template name in the request."
}
```

aclevername:
passed with changes to only set name or ID

michaelbeaumont (approved):

aclevername:
changed the PR to merge into `release-0.36` and force pushed over the branch to make it be based off `release-0.36` instead of master.

michaelbeaumont (approved):
LGTM for `0.36`

PR #3107, opened 2021-01-19
Add proposal for EC2 instance selector integration

cPu1:
### Checklist
- [ ] Manually tested
- [x] (Core team) Added labels for change area (e.g. `area/nodegroup`), target version (e.g. `version/0.12.0`) and kind (e.g. `kind/improvement`)


docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,277 @@
...
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: development
+  region: us-west-2
+
+managedNodeGroups:
+- name: <name>
+  instanceType: m5.large
```
aclevername, line 68: is this meant be set as well as `instanceTypes`?

cPu1: It's a typo. Thanks!

aclevername:
Lovely write up @cPu1.

Just to ensure if I'm understanding:
When the user runs `create cluster --dry-run` and the user has provided: 
```
managedNodeGroups:
- name: linux
  instanceSelector:
    vCPUs: 2
    memory: 4 
```

We compute `instanceTypes` using the `instanceSelector` values and present them, e.g.:
```
managedNodeGroups:
- name: linux
  instanceTypes: ["c3.large", "c4.large", "c5.large", "c5d.large", "c5n.large", "c5a.large"]
  instanceSelector:
    vCPUs: 2
    memory: 4 
```

Question:
Why do we need to show them instanceTypes? Is the intention that the users will see the list of `instanceTypes`, go to there config, remove the `instanceSelector` field and replace it with `instanceTypes` and set values that they like, e.g. lets say the user doesn't want any `c5*` instances, after seeing the above output from a `--dry-run` they would modify the config to be:

```
managedNodeGroups:
- name: linux
  instanceTypes: ["c3.large", "c4.large"]
```

If this is the case then why do we add the complexity of this being part of the `create cluster` flow? The user should either:
1. just set the `instanceSelector` fields and be happy with the decisions made by the EC2 Instance Selector
2. Provide a list of `instanceTypes` themselves manually, we could move there lives easier by having an `eksctl generate instance-types` command added


I don't quite see the value in having this as part of create cluster/nodegroup

Callisto13:
Really thorough @cPu1 🎉 

Furthering @aclevername's comment:

> 2. Provide a list of instanceTypes themselves manually, we could move there lives easier by having an eksctl generate instance-types command added

In this case, we would be better off not providing anything they can just run the instance selector themselves.

IIUC, the value to adding this to eksctl is so that users do not need to think about it. If they want to disagree with the Selector, it is probably the exact same effort to just run the selector themselves than to do a 2 step process through us.
That said we do still need to return the config to them anyway, since their record will no longer be current.

cPu1:
>1. just set the instanceSelector fields and be happy with the decisions made by the EC2 Instance Selector

Users can do this by omitting the `--dry-run` flag. This is specified in the proposal:
>If --dry-run is omitted, eksctl will proceed to creating the cluster and nodegroup with the instance types matching the instance selector resource criteria.




>2. Provide a list of instanceTypes themselves manually, we could make their lives easier by having an eksctl generate instance-types command added

Such a command would add little value because it'd be a mere wrapper around the  `ec2-instance-selector` tool, which users can run themselves. This proposal is more about integrating the EC2 Instance Selector with the `create cluster` and `create nodegroup` workflows.

aclevername:
> > 1. just set the instanceSelector fields and be happy with the decisions made by the EC2 Instance Selector
> 
> Users can do this by omitting the `--dry-run` flag. This is specified in the proposal:
> 
> > If --dry-run is omitted, eksctl will proceed to creating the cluster and nodegroup with the instance types matching the instance selector resource criteria.

👍 

> > 1. Provide a list of instanceTypes themselves manually, we could make their lives easier by having an eksctl generate instance-types command added
> 
> Such a command would add little value because it'd be a mere wrapper around the `ec2-instance-selector` tool, which users can run themselves. This proposal is more about integrating the EC2 Instance Selector with the `create cluster` and `create nodegroup` workflows.

I don't understand the value of it being in the `create cluster/nodegroup` workflow, having it printed in the output seems like a strange way to tell them. They could just run the binary themselves and provide the list themselves

cPu1:
>I don't understand the value it being printed out create cluster/nodegroup workflow. Having the feature is great, but if they don't like the values that get generated they could just run the binary themselves and provide the list themselves (this is why I suggested having an eksctl subcommand to wrap it to prevent them having to download it themselves)


The proposed `--dry-run` feature is a general-purpose option that is not specifically tied to `ec2-instance-selector`. The default behaviour of `eksctl create cluster` is to create a cluster with a 2-node nodegroup that uses `m5.large` instances. The dry-run feature lets users generate a base config that can be used as a starting point for making modifications. The purpose of dry-run is to output a ClusterConfig representing the CLI options passed. When this feature is used with the instance selector options, it also outputs the instance types matching the instance selector criteria.

aclevername:
> > I don't understand the value it being printed out create cluster/nodegroup workflow. Having the feature is great, but if they don't like the values that get generated they could just run the binary themselves and provide the list themselves (this is why I suggested having an eksctl subcommand to wrap it to prevent them having to download it themselves)
> 
> The proposed `--dry-run` feature is a general-purpose option that is not specifically tied to `ec2-instance-selector`. The default behaviour of `eksctl create cluster` is to create a cluster with a 2-node nodegroup that uses `m5.large` instances. The dry-run feature lets users generate a base config that can be used as a starting point for making modifications. The purpose of dry-run is to output a ClusterConfig representing the CLI options passed. When this feature is used with the instance selector options, for consistency, it also outputs the instance types matching the instance selector criteria.

I think as a feature aside from `ec2-instance-selector` it sounds like it has value, and might fit in well with https://github.com/weaveworks/eksctl/issues/2774, in particular it might have some overlap with generating manifests for existing imperatively created clusters. However just focusing on the EC2 selector support for the moment I still think that the user should either:

1. Just set the instanceSelector fields and be happy with the decisions made by the EC2 Instance Selector
2. Provide a list of instanceTypes themselves manually

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+<!-- /toc -->
+
+## Summary
+
+This proposal adds support for the EC2 instance selector for both managed and unmanaged nodegroups.
+
+
+## Motivation
+
+With support for multiple instance types and Spot instances for EKS Managed Nodegroups, users can create Spot instances where EKS will configure and launch an Autoscaling Group of Spot instances following  Spot best practices. Users can set `managedNodeGroup.spot` and supply a list of instance types in `managedNodeGroup.instanceTypes` to instruct eksctl to create a managed nodegroup using Spot instances.
```
tabern, line 23: This is very focused on spot. Can we change the motivation to focus on the general case (that of course includes spot) so that users are not confused?

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+It is a Spot best practice to use multiple instance types, but with over 270 EC2 instance types, users have to spend time figuring out which instance types would be well suited for their nodegroup.
```
tabern, line 25: Same call out here - this is an issue in general, not just for those using Spot.

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+The [EC2 Instance Selector](https://github.com/aws/amazon-ec2-instance-selector) tool addresses this problem by generating a list of instance types based on resource criteria such as vCPUs, memory, etc. This proposal aims to integrate the EC2 instance selector with eksctl to enable creating nodegroups with multiple instance types by passing the resource criteria.
+
+### Problem: a simple integration
+We could add the instance selector options—as CLI options (`--instance-selector-vcpus` and `--instance-selector-memory`) and in the config file—to `create cluster` and `create nodegroup` that would select the instance types matching the resource criteria for the user and proceed to creating the nodegroup. However, this gives users no control over selecting the instance types. Users may want to inspect and change the instances matching the instance selector criteria before proceeding to creating a nodegroup, and also to document the instance types being used for their nodegroups in the ClusterConfig file.
```
tabern, line 30: This is a strawman against a certain option. Can we be more clear about the problem we want to solve here?

In particular, is it critical that we have a "human in the loop" here? We could always modify the config file with the selections and output them. Is there benefit to having the manual selection process? Doesn't this require the user to have an understanding of the instance type options anyways?

tabern: Reading further - it looks like this is actually the proposal - but we negate it early to introduce the `--dry-run` functionality. Let's consider cutting or revising this section as its confusing.

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Proposal
+
+This design proposes adding a new general-purpose `--dry-run` option to `eksctl create cluster` and `eksctl create nodegroup` that skips cluster and nodegroup creation and instead outputs a ClusterConfig file representing the CLI options passed. This functionality extends to the instance selector integration where eksctl will set the instance types for a nodegroup when the instance selector options are supplied. This gives users the opportunity to inspect and, if required, change the instance types before proceeding with cluster and nodegroup creation. Users can omit the `--dry-run` option if they do not care about the instance types selected by eksctl.
```
tabern, line 34: `--dry-run` is one aspect of the feature. Shouldn't we lead with the new `--instance-selector-[ ]` functionality?

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+There are certain one-off options that cannot be represented in the ClusterConfig file, e.g., `--install-vpc-controllers`.
+Users would expect `eksctl create cluster --<options...> --dry-run > config.yaml` followed by `eksctl create cluster -f config.yaml` to be equivalent to running the first command without `--dry-run`. eksctl would therefore disallow passing options that cannot be represented in the config file when `--dry-run` is passed.
+
+## Design
+Along with the `--dry-run` option, two new options `--instance-selector-vcpus` and `--instance-selector-memory` will be added to `eksctl create cluster` and `eksctl create nodegroup`. These options will also be supported in the ClusterConfig file for both managed and self-managed nodegroups with the following schema:
+
+```yaml
+instanceSelector:
```
tabern, line 43: are vCPU and memory the only inputs we will accept?

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+  memory: <integer>
+```
+
+At launch only the `vCPUs` and `memory` options will be supported, additional options may be added later based on community feedback.
+
+**Note:** Because fields are case-sensitive, `vCPUs` may be difficult to get right, especially for users unfamiliar with Go naming conventions. However, `vCPUs` with the same case is standard enough that it's widely used by all major cloud providers. `vcpus` is easier to remember but it's inconsistent with the other field names.
+
+### eksctl create cluster
+
+When `eksctl create cluster` is called with the instance selector options and `--dry-run`, eksctl will output a ClusterConfig file containing a nodegroup representing the CLI options and the instance types set to the instances matched by the instance selector resource criteria.
```
tabern, line 54: we should ensure `--dry-run` outputs *all* options, not just node groups!

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+metadata:
+  name: development
+  region: us-west-2
+
+nodeGroups:
+- name: <name>
+  instancesDistribution:
+    instanceTypes: ["c5.large", "c5a.large", "c5ad.large", "c5d.large", "t2.medium", "t3.medium", "t3a.medium"]
+  instanceSelector:
```
tabern, line 93: should we automatically add a comment in the file "Ignored during execution" or something of the like?

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: cluster
+  region: us-west-2
+
+managedNodeGroups:
+- name: linux
+  instanceSelector:
```
tabern, line 119: is it confusing that this is only in the MNG section? How do we handle merging MNG and non-MNG groups?

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,269 @@
...
+
+metadata:
+  name: dev
+  region: us-west-2
+
+managedNodeGroups:
+- name: <name>
+  spot: true
+  instanceTypes: ["c3.large", "c4.large", "c5.large", "c5d.large", "c5n.large", "c5a.large"]
+  instanceSelector:
```
tabern, line 188: 👍

tabern (commented):
Generally looks good. Some comments on the overview for this feature as well as questions around how we represent. No major blockers in the current design.

docs/proposal-007-instance-selector.md
cPu1: We'd start off by adding vCPU and memory, and add more options based on user feedback. Otherwise we'd have to add a CLI option `instance-selector-*` for each instance selector option. Are there other options that you think we should add?

docs/proposal-007-instance-selector.md
cPu1: Good idea. We can have something like:

```yaml
  # Ignored during execution since `instanceTypes` is also set.
  instanceSelector:
```

docs/proposal-007-instance-selector.md
cPu1: I'm not sure I follow. This field is also present in the `nodeGroups` section below: https://github.com/weaveworks/eksctl/pull/3107/files#diff-af7ec3d70c271157e75ad67f9bb60c75c8596ace0bee45c3b6a83d7debd9408cR126

>How do we handle merging MNG and non-MNG groups?

Can you expand on this?

aclevername:
> I think as a feature aside from `ec2-instance-selector` it sounds like it has value, and might fit in well with #2774, in particular it might have some overlap with generating manifests for existing imperatively created clusters. However just focusing on the EC2 selector support for the moment I still think that the user should either:
> 
> 1. Just set the instanceSelector fields and be happy with the decisions made by the EC2 Instance Selector
> 2. Provide a list of instanceTypes themselves manually

@cPu1 I'm still slightly confused as to why we need the `--dry-run` functionality for this feature, I still stand by what I said above. Am I missing a different reason as to why we need it for this feature?

docs/proposal-007-instance-selector.md
tabern: you answered my question -must had missed it on the first pass.

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,360 @@
...
+
+There are certain one-off options that cannot be represented in the ClusterConfig file, e.g., `--install-vpc-controllers`.
+Users would expect `eksctl create cluster --<options...> --dry-run > config.yaml` followed by `eksctl create cluster -f config.yaml` to be equivalent to running the first command without `--dry-run`. eksctl would therefore disallow passing options that cannot be represented in the config file when `--dry-run` is passed.
```
tabern, line 80: :+1:

docs/proposal-007-instance-selector.md
tabern: GPU would be an important one

is `--instance-selector` too verbose?

docs/proposal-007-instance-selector.md
tabern: update looks good!

tabern (approved):
update LGTM :+1:

We may want to add GPU as an instance selector option. 
Also is `--instance-select-x` too verbose? May not be but wanted to see if you had thoughts on other options.

cPu1:
> We may want to add GPU as an instance selector option.

Noted, I'll add it to the proposal. To be clear, the `gpus` option from instance selector that specifies the total number of GPUs, right?

> Also is --instance-select-x too verbose? May not be but wanted to see if you had thoughts on other options.

It may be too verbose, but the instance selector option names by themselves are not very clear in the context of eksctl, e.g., just `--vCPUs` doesn't imply the use of instance selector. We need some prefix that indicates it's an instance selector option. Having the `--instance-selector-*` prefix also chimes in with the `instanceSelector` ClusterConfig field. It also helps users find all available instance selector options by typing `--instance-selector` and pressing <kbd>Tab</kbd>.

docs/proposal-007-instance-selector.md
```diff
@@ -0,0 +1,364 @@
+# EC2 Instance Selector Support for Nodegroups
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+  - [Dry Run](#dry-run)
```
aclevername, line 9: should these be in two separate proposal files? It feels a bit hard to distinguish them atm

cPu1: They could be, but I'll have to reword some of the bits because right now it's written within the context of instance selector. I was planning to rename the title and filename to include dry-run, so I'll do that.

aclevername (approved):

PR #3122, opened 2021-01-20
added RoleOnly option for iamserviceaccounts

### LINKED ISSUE #3042: Create an IAM Service Account role without creating/annotating the service account
**What happened?**
I used the IAM role name feature to give the iam role generated by eksctl a predictable name. Then I used the IAM role name in a GitOps setup and attached it with the `eks.amazonaws.com/role-arn` annotation to a service account which is provided by a helm chart.
The service account is generated be the helm chart and only the annotation and the IAM role is provided by me. I don't know the name of the service account generated by the helm chart.

I don't know if this is a bug or feature request but from the description of the PR (#2863) that introduced the feature:
>  For instance, when following a git-ops workflow it might be necessary to include the serviceaccount annotations for the IAM role ARNs in the manifest repo.

I thought that this is the intended use of the IAM role name feature.

I created a IAM role using the following cluster config.
The config generates two things: the IAM role and a Service Account.
I don't need the service account, I only want to use the IAM role and attach it to the service account generated by the helm chart.
```yaml
...
iam:
  withOIDC: true
  serviceAccounts:
    - metadata:
        name: some-name
        namespace: default
      roleName: role-for-helm-chart
      attachPolicy: ...
```

The helm chart is managed the `helm.fluxcd.io/v1` `HelmRelease` crd.
In the values section I added the following:
```yaml
    serviceAccount:
      annotations:
        eks.amazonaws.com/role-arn: arn:aws:iam::<redacted>:role/role-for-helm-chart
```

**How to reproduce it?**
1. Create a IAM role with a defined name using the eksctl
2. Attach the `eks.amazonaws.com/role-arn` annotation to a service account different than the one created by eksctl
3. see error `WebIdentityErr: failed to retrieve credentials caused by: AccessDenied: Not authorized to perform sts:AssumeRoleWithWebIdentity` in the application that uses the service account

https://github.com/weaveworks/eksctl/pull/2863#issuecomment-755526414

**Versions**
```
$ eksctl version
0.36.0-rc.0
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

Legion2:
### Description
Implementation of #3042
`RoleOnly` option is added to iamserviceaccounts, so it is possible to only create the IAM role but do not create or update any Service Account in the cluster.

The IAM role is still only valid for a specific namespace and name of a service account, but the service account can be managed by other tools such as helm.

Only the `create` command is affected by this change, the deletion command will also delete service accounts not created by eksctl. This is because it's not clear who created the service account, which should be implemented when the `apply` command is introduced #3037.

### Checklist
- [x] Manually tested

close #3042


Legion2:
@aclevername do I need to also commit the generated files, and if yes which ones?

aclevername:
> @aclevername do I need to also commit the generated files, and if yes which ones?

you need to include all auto-generated files 😄

Legion2:
I did not include files which I think are unrelated to my changes.

PS: I'm on windows and use dev container to run the make commands. Maybe there are line ending problems, because there is no `.gitattributes` files.

pkg/ctl/create/iamserviceaccount.go
```diff
@@ -46,6 +46,7 @@ func createIAMServiceAccountCmdWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *
 		fs.StringVar(&serviceAccount.Namespace, "namespace", "default", "namespace where to create the iamserviceaccount")
 		fs.StringSliceVar(&serviceAccount.AttachPolicyARNs, "attach-policy-arn", []string{}, "ARN of the policy where to create the iamserviceaccount")
 		fs.StringVar(&serviceAccount.RoleName, "role-name", "", "Set a custom name for the created role")
+		serviceAccount.RoleOnly = fs.Bool("role-only", false, "disable service account creation, only the role will be created")
```
michaelbeaumont, line 49: Can we use the same format as the above lines for adding args? Independent of which it's better, I think we should keep the same style.

userdocs/src/usage/iamserviceaccounts.md
```diff
@@ -69,8 +70,18 @@ CloudFormation will generate a role name that includes a random string. If you p
...
+The other toll is then responsible for adding the role ARN annotation.
+
+```console
+eksctl create iamserviceaccount --cluster=<clusterName> --name=<serviceAccountName> --role-only --role-name "custom-role-name"
+```
+
 Currently, to update a role you will need to re-create, run `eksctl delete iamserviceaccount` followed by `eksctl create iamserviceaccount` to achieve that.
 
+!!!note
+    `eksctl delete iamserviceaccount` deletes service accounts even if they were not created by `eksctl create iamserviceaccount`.
```
michaelbeaumont, line 83:
```suggestion
    `eksctl delete iamserviceaccount` deletes Kubernetes `ServiceAccounts` even if they were not created by `eksctl`.
```

userdocs/src/usage/iamserviceaccounts.md
```diff
@@ -69,8 +70,18 @@ CloudFormation will generate a role name that includes a random string. If you p
 eksctl create iamserviceaccount --cluster=<clusterName> --name=<serviceAccountName> --role-name "custom-role-name"
 ```
 
+When the service account is created and managed by some other tool, such as helm, use `--role-only` to prevent conflicts.
+The other toll is then responsible for adding the role ARN annotation.
```
michaelbeaumont, line 74:
```suggestion
The other tool is then responsible for maintaining the role ARN annotation.
```

pkg/ctl/cmdutils/filter/iamserviceaccount_filter_test.go
```diff
@@ -54,7 +60,72 @@ var _ = Describe("iamserviceaccount filter", func() {
 				"sa/test3",
 			)).To(BeTrue())
 		})
-		// TODO test SetExcludeExistingFilter() which requires mocking different kubernetes functions
+
+		It("exclude existing stacks works correctly", func() {
```
michaelbeaumont, line 64: Can we fix the `It` strings?

pkg/ctl/cmdutils/filter/iamserviceaccount_filter_test.go
```diff
@@ -54,7 +60,72 @@ var _ = Describe("iamserviceaccount filter", func() {
...
+			)).To(BeTrue())
+			Expect(excluded).To(HaveLen(3))
+			Expect(excluded.HasAll(
+				"sa/dev1",
+				"sa/dev2",
+				"sa/dev3",
+			)).To(BeTrue())
+		})
+
+		It("exclude existing service accounts works correctly", func() {
```
michaelbeaumont, line 90: also here

michaelbeaumont (changes requested):

pkg/ctl/create/iamserviceaccount.go
Legion2: `BoolVar` does not work, because it wants a bool pointer and not a pointer to a bool pointer, so I get a null pointer exception. That's because I use bool pointer in serviceAccount to model the absence of the RoleOnly option, maybe there is a better pattern.

pkg/ctl/cmdutils/filter/iamserviceaccount_filter_test.go
Legion2: sorry, I don't understand what you mean by "fix" here

pkg/ctl/create/iamserviceaccount.go
michaelbeaumont: I think you can just pass in `serviceAccount.RoleOnly` and add `RoleOnly: api.Disabled(),` to https://github.com/weaveworks/eksctl/blob/7901d1eebe27beb2205922d56be1b72b4d49fa72/pkg/ctl/create/iamserviceaccount.go#L28

pkg/ctl/cmdutils/filter/iamserviceaccount_filter_test.go
michaelbeaumont: Ah nevermind, I think I misread they were left copy pasted from a different `It`, sorry!

pkg/ctl/cmdutils/filter/iamserviceaccount_filter.go
```diff
@@ -57,6 +57,9 @@ func (f *IAMServiceAccountFilter) SetExcludeExistingFilter(stackManager serviceA
 
 	if !overrideExistingServiceAccounts {
 		err := f.ForEach(serviceAccounts, func(_ int, sa *api.ClusterIAMServiceAccount) error {
+			if api.IsEnabled(sa.RoleOnly) {
```
michaelbeaumont, line 60: This means with `roleOnly`, we'll always create the role, right? Can we also add that to the docs?

Legion2: The use case is that the service account already exists in the cluster or is created later by some other tool. But by default eksctl will not create iam roles for service accounts if the service account already exists. But with RoleOnly, the service account will never be created and overwritten by eksctl, so we should also not filter for them. I think this is the behavior a user expects when using the roleOnly option. Otherwise a user has to specify `--override-existing-serviceaccounts` and `--role-only` at the same time which is confusing, because the service accounts are not overwritten.

Legion2:
@michaelbeaumont what have to be done to merge this PR?

userdocs/src/usage/iamserviceaccounts.md
```diff
@@ -69,8 +70,18 @@ CloudFormation will generate a role name that includes a random string. If you p
 eksctl create iamserviceaccount --cluster=<clusterName> --name=<serviceAccountName> --role-name "custom-role-name"
 ```
 
+When the service account is created and managed by some other tool, such as helm, use `--role-only` to prevent conflicts.
+The other tool is then responsible for maintaining the role ARN annotation. Note that `--override-existing-serviceaccounts` has no effect on `roleOnly`/`--role-only` service accounts, the role will always be created.
+
+```console
+eksctl create iamserviceaccount --cluster=<clusterName> --name=<serviceAccountName> --role-only --role-name "custom-role-name"
```
cPu1, line 77: Minor nit, just to be consistent:

```suggestion
eksctl create iamserviceaccount --cluster=<clusterName> --name=<serviceAccountName> --role-only --role-name=<customRoleName>
```

cPu1:
Hi @Legion2. Thanks for the PR!

I have a minor improvement to suggest to this PR. When `roleOnly` is used, eksctl should log the role ARN annotation so that users could copy it and add it to their self-managed SA. 

> the deletion command will also delete service accounts not created by eksctl. This is because it's not clear who created the service account,

This is not the ideal behaviour IMO. eksctl should not delete resources it didn't create. When a SA is created as part of a Helm chart, outside of eksctl, the delete command can leave the Helm application in an inconsistent state. However, the current code also removes SAs it didn't create, so this case doesn't have to be addressed in this PR.

Legion2:
@cPu1 when using `roleOnly` you want a deterministic role ARN, so it is expected that it is specified using the `roleName` option. Also I have looked at the code and can not find how to get the generated role ARN when `roleName` is not given.

michaelbeaumont (approved):

Legion2:
Do I really need to commit generated files such as `pkg/addons/assets.go`, because I don't know why it changed and because I'm new to go and use windows with a docker setup to run the make commands, it could also be my setup and windows line endings, which cause the changes in the generated files.

michaelbeaumont:
There's a conflict right now because the schema needs to be regenerated with the newest changes from master

Legion2:
image
What about these generated files, I did not commit them, because these get also changed when running `make build` on the `main` branch.

michaelbeaumont:
You likely only need to commit `pkg/apis/eksctl.io/v1alpha5/schema.go` and `pkg/apis/eksctl.io/v1alpha5/assets/schema.json`

Legion2:
ok I had to run `make generate-all`, `make build` did not generate this file.

cPu1:
@Legion2 can you rebase and regenerate the assets by running `make generate-all`?

Legion2:
> Additionally, do you plan on addressing this in your PR?

@cPu1 Only if someone can help me, I don't know where to get this information from and where to print it. Else I don't think it is required, because if someone uses this option, he/she will also use the roleName option to generate a predictable name.

cPu1:
> @cPu1 Only if someone can help me, I don't know where to get this information from and where to print it. Else I don't think it is required, because if someone uses this option, he/she will also use the roleName option to generate a predictable name.

I suggested that as a nice-to-have feature and I think it can be left out. The only blocker for merging this PR is the conflict with the schema file.

Legion2:
@cPu1 I rebased it

cPu1 (approved):
LGTM. Thanks for the PR!

PR #3131, opened 2021-01-21
Update windows-nodes docs for `install-vpc-controller` in an existing cluster

dLobatog:
### Description

The docs here were providing the wrong `--cluster` argument, it should be trying to `install-vpc-controller` in the same cluster as the previous line.


dLobatog:
I can't set labels myself as a non-maintainer, which seems to be the thing that makes one of the PR checks fail

aclevername (approved):
Thanks @dLobatog

PR #3159, opened 2021-01-27
log undeleted stacks during cluster deletion

aclevername:
### Description

Example output:
```
./eksctl delete cluster --name jk-console
[ℹ]  eksctl version 0.38.0-dev+9f64d9de.2021-01-27T10:28:15Z
[ℹ]  using region us-west-2
[ℹ]  deleting EKS cluster "jk-console"
[!]  found the following undeleted stacks: eksctl-jk-console-addon-iamserviceaccount-baz-foo
[✔]  all cluster resources were deleted
```

I think printing `[✔]  all cluster resources were deleted` is misleading to the user as there are clearly some resources left, but eksctl doesn't know to clean them up so it shouldn't error either. Unsure what the UX should be, open to suggestions 
### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/actions/cluster/unowned.go
+++ b/pkg/actions/cluster/unowned.go
@@ -78,6 +78,10 @@ func (c *UnownedCluster) Delete(waitInterval time.Duration, wait bool) error {
 		return err
 	}
 
+	if err := logUndeletedStacks(c.ctl.NewStackManager(c.cfg)); err != nil {
+		return err
+	}
+
 	logger.Success("all cluster resources were deleted")
 	return nil
 }
--- a/pkg/actions/cluster/unowned.go
+++ b/pkg/actions/cluster/unowned.go
@@ -78,7 +78,7 @@ func (c *UnownedCluster) Delete(waitInterval time.Duration, wait bool) error {
 		return err
 	}
 
-	if err := logUndeletedStacks(c.ctl.NewStackManager(c.cfg)); err != nil {
+	if err := checkForUndeletedStacks(c.ctl.NewStackManager(c.cfg)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/actions/cluster/unowned.go
```diff
@@ -78,6 +78,10 @@ func (c *UnownedCluster) Delete(waitInterval time.Duration, wait bool) error {
 		return err
 	}
 
+	if err := logUndeletedStacks(c.ctl.NewStackManager(c.cfg)); err != nil {
+		return err
+	}
+
 	logger.Success("all cluster resources were deleted")
 	return nil
```
aclevername, line 86: I don't think this is great UX. We should either:
1. Print the warning and change the wording of the success message to "deletion complete"
2. Exit non-zero to indicate to the user that we didn't delete everything. Print a suggestion of manually deleting the cloudformation stacks
3. We should just delete the stacks if we find any rather than logging. However randomly deleting stacks that might have other dependencies that also need to be cleaned up could be problematic

michaelbeaumont: I'm for 2), if there are still stacks `eksctl` "knows" belongs to it but we didn't delete them, then it's definitely a failure to delete, IMO

aclevername:
```
./eksctl delete cluster --name jk-console
[ℹ]  eksctl version 0.38.0-dev+16e9c8ef.2021-01-27T14:45:31Z
[ℹ]  using region us-west-2
[ℹ]  deleting EKS cluster "jk-console"
[!]  found the following undeleted stacks: eksctl-jk-console-some-forgotten-stack
Error: failed to delete all resources
```

Callisto13 (approved):
LGTM 👍

Issue #3133, opened 2021-01-22
The Vpc Endpoint Service 'com.amazonaws.eu-south-1.ecr.dkr' does not exist (private cluster)


Callisto13:
Reported by @AndreaGal95 in https://github.com/weaveworks/eksctl/issues/3081

**What were you trying to accomplish?**
Trying to create a cluster in `eu-south-1` region

**What happened?**
Failed with:

```
services: InvalidServiceName: The Vpc Endpoint Service 'com.amazonaws.eu-south-1.ecr.dkr' does not exist
```

**How to reproduce it?**
Create a cluster in region `eu-south-1`

Config:

```yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: cloudbees-ci
  region: eu-south-1

vpc:
  id: "vpc-010798c958f4bfdff"
  subnets:
    private:
      eu-south-1a:
          id: "subnet-04ecf032c30889df0"
      eu-south-1b:
          id: "subnet-0da7f1efbf13add01"

privateCluster:
  enabled: true
  additionalEndpointServices:
  - "autoscaling"
  - "cloudformation"
  - "logs"

nodeGroups:
  - name: ng-1
    instanceType: m5.large
    desiredCapacity: 1
    privateNetworking: true


fargateProfiles:
  - name: fp-default
    selectors:
      - namespace: default
      - namespace: kube-system
  - name: fp-devops
    selectors:
      - namespace: devops

cloudWatch:
  clusterLogging:
    enableTypes: ["*"]
```

**Logs**

```
[ℹ]  eksctl version 0.36.1
[ℹ]  using region eu-south-1
[✔]  using existing VPC (vpc-010798c958f4bfdff) and subnets (private:[subnet-0da7f1efbf13add01 subnet-04ecf032c30889df0] public:[])
[!]  custom VPC/subnets will be used; if resulting cluster doesn't function as expected, make sure to review the configuration of VPC/subnets
[ℹ]  nodegroup "ng-1" will use "ami-066461a96ead1ce53" [AmazonLinux2/1.18]
[ℹ]  using Kubernetes version 1.18
[ℹ]  creating EKS cluster "cloudbees-ci" in "eu-south-1" region with Fargate profile and un-managed nodes
[ℹ]  1 nodegroup (ng-1) was included (based on the include/exclude rules)
[ℹ]  will create a CloudFormation stack for cluster itself and 1 nodegroup stack(s)
[...]

PR #3186, opened 2021-02-01
Detect and log unsupported region error for fully-private clusters

cPu1:
### Description

As discovered in #3133 , `eu-south-1` does not support fully-private clusters. This PR detects this error condition and logs the error.

### Checklist
- [x] Manually tested


pkg/cfn/builder/vpc_endpoint.go
```diff
@@ -166,10 +166,16 @@ func buildVPCEndpointServices(ec2API ec2iface.EC2API, region string, endpoints [
 		})
 
 		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "InvalidServiceName" {
+				return nil, &api.UnsupportedFeatureError{
```
michaelbeaumont, line 170: What's the advantage over `errors.Wrapf`?

cPu1: It's for the caller to identify errors of this type without relying on string comparison.

michaelbeaumont: In this case we only print the error anyway, right? I.e. why do we need the `errors.As` check?

michaelbeaumont: Not that it's an issue necessarily but is the handling of the error correct here?

cPu1: >In this case we only print the error anyway, right? I.e. why do we need the errors.As check?

We could just log the error here instead of wrapping it in a typed error, but I have avoided doing any I/O here because this code has no external side effects (if the provider implementation is mocked). We don't want logs for this case appearing in our unit tests. Additionally, the error type provides a standard format for unsupported features that the calling code can identify with a single check. The `errors.As` check is required for the caller to identify errors of this type because this check happens higher up in the chain where this error may already be wrapped by other errors.

michaelbeaumont: i don't mean logging it here but if you were to just return `errors.Wrap(err, fmt.Sprintf("fully-private clusters are not supported in region %q, please retry with a different region", region))` here or even `errors.New(fmt.Sprintf("fully-private clusters are not supported in region %q, please retry with a different region", region)`, wouldn't the user see the same thing they're seeing with this PR? We call `logger.Critical` in `pkg/ctl/create/cluster.go` anyway.
Like:
```
				ufe := &api.UnsupportedFeatureError{}
				if errors.As(err, &ufe) {
					logger.Critical(ufe.Message)
				}
				logger.Critical("%s\n", err.Error())
```
results in `logger.Critical` being called for this `UnsupportedFeatureError` twice basically

cPu1: With `errors.Wrap` or `errors.New`, the log message will not start with the "fully-private clusters are not supported" error,  the messages from the `errors.Wrap` call higher up in the chain will appear first, which can hide the user-friendly error message.

Having an error type with a `Message` field to represent this condition lets us log a more-specific, user-friendly message while still preserving the error context and logging it later.

michaelbeaumont: Ah the `UnsupportedFeatureError` itself is being wrapped again and isn't the highest level error? Now I get it.

cPu1: Yes, it isn't the highest-level error.

michaelbeaumont (approved):

PR #3194, opened 2021-02-02
master -> main

Callisto13:
Rename all the things.

Or some of them at least. GH will redirect all view links, so I have just updated the ones which are in "plain text".

### DIFF
--- a/build/scripts/tag-release.sh
+++ b/build/scripts/tag-release.sh
@@ -32,7 +32,7 @@ tag_version_and_latest "${m}" "${release_version}"
 git push --force origin "${release_branch}":docs
 
 git checkout "${default_branch}"
-git pull --ff-only origin main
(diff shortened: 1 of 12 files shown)


build/scripts/tag-release.sh
```diff
@@ -32,7 +32,7 @@ tag_version_and_latest "${m}" "${release_version}"
 git push --force origin "${release_branch}":docs
 
 git checkout "${default_branch}"
-git pull --ff-only origin master
+git pull --ff-only origin main
```
michaelbeaumont, line 35: oh no, i forgot one :facepalm:

michaelbeaumont (approved):

build/scripts/tag-release.sh
Callisto13: 🤦‍♀️

Issue #2931, opened 2020-12-11
Make the Provider less powerful


Callisto13:
Adding new things, specifically new things _with tests_ is quite challenging now due to the all powerful Provider.

The Provider contains all AWS interfaces, and is passed around everywhere, meaning that you can pretty much do anything, anywhere in the codebase.

There are few boundaries or tightly scoped components with clear responsibilities delineated by interfaces: nearly everything gets a Provider and needs little else to get things done.

This makes unit tests particularly hard, and not very "unity", as you often end up inadvertently testing things several layers down the stack in neighbouring packages. Unit tests being what they are: you should not need to think about stuff happening elsewhere.

This ticket is about fixing some of this, and namely removing some of the monopoly from the Provider. How that is done is player's choice 😉 .

aclevername:
## What is the provider?

The `ClusterProvider` is an [inteface](https://github.com/weaveworks/eksctl/blob/d64049cc2da36edd0a99099ad1065562fcdc38c2/pkg/apis/eksctl.io/v1alpha5/types.go#L517-L532)
```
// ClusterProvider is the interface to AWS APIs
type ClusterProvider interface {
	CloudFormation() cloudformationiface.CloudFormationAPI
	CloudFormationRoleARN() string
	CloudFormationDisableRollback() bool
	EKS() eksiface.EKSAPI
	EC2() ec2iface.EC2API
	ELB() elbiface.ELBAPI
	ELBV2() elbv2iface.ELBV2API
	STS() stsiface.STSAPI
	SSM() ssmiface.SSMAPI
	IAM() iamiface.IAMAPI
	CloudTrail() cloudtrailiface.CloudTrailAPI
	Region() string
	Profile() string
	WaitTimeout() time.Duration
}
```

The interface prescribes a powerful struct that provides the ability to interact with various AWS apis, e.g. `CloudFormation` and `EKS`. 

## Whats wrong with it?

### Used too widely 
We pass a struct implementation of this interface around everywhere- doing a code lookup finds its referenced 36 times. We pass this all-powerful struct into method/functions that only require a very slim subsection of its capabilitys, e.g. in the `Manager` package we only ever make `provider.CloudFormation` and `provider.EKS` calls. Giving the package the `provider` can lead to us to continue to over-use it, e.g. making `EC2` calls in the `manager` package when perhaps this logic should be abstracted out into a different location. **Having this powerful provider everywhere tempts us into bad design choices** (should the manager even be making EKS calls?)

[...]

PR #3216, opened 2021-02-05
use common stackManager interface and fakes across packages

aclevername:
### Description

First of many small PRs relating to #2931. This introduces a stackManager interface and fake that can be used to mock out the package directly, it should make testing this in the future easier.

This also highlights how INSANELY big the stack collection is, we should try to refactor this out into smaller sub packages in the future

### Checklist
- [ ] Manually tested


pkg/actions/addon/addon.go
```diff
@@ -22,19 +20,11 @@ type Manager struct {
 	clusterProvider *eks.ClusterProvider
 	withOIDC        bool
 	oidcManager     *iamoidc.OpenIDConnectManager
-	stackManager    StackManager
+	stackManager    manager.StackManager
 	clientSet       kubeclient.Interface
 }
 
-//go:generate counterfeiter -o fakes/fake_stack_manager.go . StackManager
-type StackManager interface {
```
michaelbeaumont, line 30: These custom interfaces are actually kind of useful for seeing which parts of the `StackManager` are used around the codebase, could be a guide to splitting it up

michaelbeaumont (approved):

pkg/actions/addon/addon.go
aclevername: Yeah I don't disagree, me and @Callisto13 had a chat about wether to keep these or not in the issue https://github.com/weaveworks/eksctl/issues/2931#issuecomment-748109410, I'm quite torn.

Callisto13: Mike does have a point that it could make things easier when we come to splitting that beast of a stack manager.

aclevername: It does, but like you pointed out in the issue it means loads of duplication in what is likely every package and if I had to guess would make people writing tests less likely 🤷  I'm leaning towards this at the moment as it enables everyone to very quick start adding tests

PR #3217, opened 2021-02-05
remove unnecessary ctl.Provider references

aclevername:
### Description

Part of #2931. Removes all of the simple scenarios where we pass in a `ctl.Provider` where a subset of the provider would be appropriate. Some of the more complicated scenarios are skipped and noted:

- `pkg/actions/addons/task.go` uses ctl extensively for things like kubeconfig + oidc creation, couldn't remove it cleanly yet.
- `pkg/cfn/manager/api.go` still uses ctl and I'm not sure if the partial refactor that I've done is actually worth it, could be reverted. 

Recommend reviewing by commit as it tackles sub-packages 1 by 1
### Checklist
- [ ] Manually tested


pkg/cfn/manager/api.go
```diff
@@ -79,17 +95,25 @@ func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) *
...
+		sharedTags:        tags,
+		cloudformationAPI: provider.CloudFormation(),
+		ec2API:            provider.EC2(),
+		eksAPI:            provider.EKS(),
+		iamAPI:            provider.IAM(),
+		cloudTrailAPI:     provider.CloudTrail(),
+		disableRollback:   provider.CloudFormationDisableRollback(),
+		roleARN:           provider.CloudFormationRoleARN(),
+		region:            provider.Region(),
+		waitTimeout:       provider.WaitTimeout(),
```
aclevername, line 108: I really don't think this is an improvement, I've made the change more to illustrate how crazy the provider usage is inside the manager package. 

I'm open for reverting it and trying to address the usage of ctl inside the manager package in a separate PR. LMK if anyone disagrees

pkg/managed/service.go
```diff
@@ -59,11 +66,13 @@ const (
 )
 
 // NewService creates a new Service
-func NewService(provider api.ClusterProvider, stackCollection *manager.StackCollection, clusterName string) *Service {
+func NewService(eksAPI eksiface.EKSAPI, ssmAPI ssmiface.SSMAPI, ec2API ec2iface.EC2API,
+	stackCollection *manager.StackCollection, clusterName string) *Service {
 	return &Service{
```
aclevername, line 71: I dislike 5 arguments as it feels like a lot, but at the same time having a new struct to pass the values in by wouldn't make sense as this is a constructor 😂

pkg/actions/addon/addon.go
```diff
@@ -34,18 +35,18 @@ type StackManager interface {
 	UpdateStack(stackName, changeSetName, description string, templateData manager.TemplateData, parameters map[string]string) error
 }
 
-func New(clusterConfig *api.ClusterConfig, clusterProvider *eks.ClusterProvider, stackManager StackManager, withOIDC bool, oidcManager *iamoidc.OpenIDConnectManager, clientSet kubeclient.Interface) (*Manager, error) {
+func New(clusterConfig *api.ClusterConfig, clusterProvider eksiface.EKSAPI, stackManager StackManager, withOIDC bool, oidcManager *iamoidc.OpenIDConnectManager, clientSet kubeclient.Interface) (*Manager, error) {
```
Callisto13, line 38: can we change the var name here as well?

aclevername: Oops! Yes

pkg/actions/label/label.go
```diff
@@ -22,13 +18,11 @@ type Manager struct {
 	clusterName string
 }
 
-func New(cfg *api.ClusterConfig, ctl *eks.ClusterProvider) *Manager {
-	sc := manager.NewStackCollection(ctl.Provider, cfg)
-	mc := managed.NewService(ctl.Provider, sc, cfg.Metadata.Name)
```
Callisto13, line 27: gah but now the same identical setup lines need to happen in 3 places. sacrificed for the greater good i suppose 😭

pkg/cfn/manager/api.go
Callisto13: i am really torn. on the one hand, this will make us do something about the massive manager package sooner rather than later. on the other, i hate it. and on a third hand, we _don't_ deal with it soon and someone can easily look at this and say "hmm we should have a thing which wraps all these APIs", which is probably how the provider began to begin with before people started throwing all the other junk in there.

so in conclusion: idk

Callisto13 (approved):

Issue #3214, opened 2021-02-05
VPCEndpointS3: CREATE_FAILED – Private DNS can't be enabled because the service com.amazonaws.us-east-1.s3 does not provide a private DNS name


m1hanick:
Hi,

Trying to create a new fully private EKS cluster with EKSCTL 0.32.0
Got the following error:
```
AWS::EC2::VPCEndpoint/VPCEndpointS3: CREATE_FAILED – "Private DNS can't be enabled because the service com.amazonaws.us-east-1.s3 does not provide a private DNS name. (Service: AmazonEC2; Status Code: 400; Error Code: InvalidParameter; Request ID: 1edf09f7-0572-46db-a6ba-9bac7dcabf18; Proxy: null)"
```

CloudFormation template contains this:
```    
"VPCEndpointS3": {
      "Type": "AWS::EC2::VPCEndpoint",
      "Properties": {
        "PrivateDnsEnabled": true,
        "SecurityGroupIds": [
          {
            "Ref": "ClusterSharedNodeSecurityGroup"
          }
        ],
        "ServiceName": "com.amazonaws.us-east-1.s3",
        "SubnetIds": [
          "subnet-xxxxxxxxxxxxx",
          "subnet-xxxxxxxxxxxxx",
          "subnet-xxxxxxxxxxxxx"
        ],
        "VpcEndpointType": "Interface",
        "VpcId": "vpc-xxxxxxxxxxxxxxx"
      }
    },
```

Firstly, it is surprising that it tries to create S3 endpoint with **VpcEndpointType** = '**Interface**', it was always  '**Gateway**' before.

As I've got, API call DescribeVpcEndpointServices returns two services for the S3

```{
    "ServiceDetails": [
        {
            "ServiceName": "com.amazonaws.us-east-1.s3",
            "ServiceId": "vpce-svc-0aa7c06513d4872d5",
            "ServiceType": [
                {
                    "ServiceType": "Gateway"
                }
            ],
            "AvailabilityZones": [
                "us-east-1a",
                "us-east-1b",
                "us-east-1c",
                "us-east-1d",
                "us-east-1e",
                "us-east-1f"
            ],
            "Owner": "amazon",
            "BaseEndpointDnsNames": [
                "s3.us-east-1.amazonaws.com"
            ],
            "VpcEndpointPolicySupported": true,
            "AcceptanceRequired": false,
            "ManagesVpcEndpoints": false,
            "Tags": []
        },
        {
[...]

PR #3228, opened 2021-02-08
Fix creating private cluster in some regions

cPu1:
### Description

AWS recently added support for interface endpoints for S3, which caused the `DescribeVPCEndpointServices` call to return an additional S3 endpoint. eksctl enables PrivateDNS for all interface endpoints but it is not supported for S3, so the endpoint creation was failing. This PR filters out the S3 interface endpoint and any other unexpected endpoint types before proceeding with creating VPC endpoints. 

Fixes https://github.com/weaveworks/eksctl/issues/3214

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/cfn/builder/vpc_endpoint.go
+++ b/pkg/cfn/builder/vpc_endpoint.go
@@ -189,7 +189,12 @@ func buildVPCEndpointServices(ec2API ec2iface.EC2API, region string, endpoints [
 		}
 
 		endpointType := *sd.ServiceType[0].ServiceType
-		if *sd.ServiceName == s3EndpointName && endpointType == ec2.VpcEndpointTypeInterface {
+
+		if *sd.ServiceName == s3EndpointName {
+			if endpointType != ec2.VpcEndpointTypeGateway {
+				continue
+			}
+		} else if endpointType != ec2.VpcEndpointTypeInterface {
 			continue
 		}
 
--- a/pkg/cfn/builder/vpc_endpoint.go
+++ b/pkg/cfn/builder/vpc_endpoint.go
@@ -139,15 +139,11 @@ func buildVPCEndpointServices(ec2API ec2iface.EC2API, region string, endpoints [
 	serviceNames := make([]string, len(endpoints))
 	serviceDomain := fmt.Sprintf("com.amazonaws.%s", region)
 	for i, endpoint := range endpoints {
-		name := fmt.Sprintf("%s.%s", serviceDomain, endpoint)
-		hasChinaPrefix, ok := chinaPartitionServiceHasChinaPrefix[endpoint]
-		if !ok {
-			return nil, errors.Errorf("couldn't determine endpoint domain for service %s", endpoint)
+		serviceName, err := makeServiceName(region, endpoint)
+		if err != nil {
+			return nil, err
 		}
-		if api.Partition(region) == api.PartitionChina && hasChinaPrefix {
-			name = "cn." + name
-		}
-		serviceNames[i] = name
+		serviceNames[i] = serviceName
 	}
 
 	var serviceDetails []*ec2.ServiceDetail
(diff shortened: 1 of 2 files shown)


pkg/cfn/builder/vpc_endpoint.go
```diff
@@ -181,27 +177,48 @@ func buildVPCEndpointServices(ec2API ec2iface.EC2API, region string, endpoints [
...
+	}
 
-	for i, sd := range serviceDetails {
+	for _, sd := range serviceDetails {
 		if len(sd.ServiceType) > 1 {
 			return nil, errors.Errorf("endpoint service %q with multiple service types isn't supported", *sd.ServiceName)
 		}
 
+		endpointType := *sd.ServiceType[0].ServiceType
+		if *sd.ServiceName == s3EndpointName && endpointType == ec2.VpcEndpointTypeInterface {
```
michaelbeaumont, line 192: Can we create an "expected endpoint type" map for all the endpoints so that it doesn't happen with any other types?

Callisto13:
Can we update the description to have a TLDR of that (quite long) issue thread to summarise the problem and the solution?

pkg/cfn/builder/vpc_endpoint.go
cPu1: It is unlikely that the other services will start offering gateway endpoints, but I agree that we can make this more robust. An "expected endpoint type" map, however, is not required to handle this. Except for S3, we need all of them to be interface endpoints.

michaelbeaumont: It's just a way of being more declarative. IMO it's easier to read something like:

```
endpointType := *sd.ServiceType[0].ServiceType

if endpointType != expectedEndpointTypes[*sd.ServiceName] { // or a func expectedEndpointTypes(string) string
	continue
}
```

cPu1: It comes at the cost of increased redundancy and it's not practical to create a map to represent every possible condition. I'd prefer a simple if-else statement over a map.

michaelbeaumont: As far as readability, a map is just one way, like I said a function is fine too. I just don't think it should be inline, that's the worst for readability.

But I don't really see why we should assume that every future endpoint is also supposed to be an interface endpoint? Feels like we can save time and a future headache if whoever adds a future endpoint is forced to be explicit about their assumptions. I'm not sure why it's impractical to add a line `newEndpoint: "interface"` or to have 10 lines listing expected endpoint types. I also don't see how redundancy comes into play or why it's negative here. Just because we _can_ shorten it to `if s3 && != gateway { continue}` and typing it is faster doesn't mean it's easier to read. In fact for reading it's much easier to just see a declared list of `endpoint: type`, IMO. Like it doesn't matter that only 1 endpoint should be a gateway. It could be 2, or 3.

What happens when the next endpoint comes and it too needs to be a gateway? As someone new to the code, there's nothing obviously special about why only `S3` needs a gateway and not an interface. This is just another hidden assumption that will bite future us. What happens to the `if` statement when it gets updated? Does it stay simple?

michaelbeaumont: I suppose that's just an opinion, the only thing I think is really important is for it to not be inline.

cPu1: I get your point, but there are several other things that can change. What if a future interface endpoint does not support `PrivateDNS`? What if a future gateway endpoint requires different logic around setting route table IDs? The code hasn't accounted for these cases either, and as is seen the S3 interface endpoint not supporting `PrivateDNS` is what caused the error. Do we add another map representing endpoints that support `PrivateDNS` because a future endpoint may not support it?

>As someone new to the code, there's nothing obviously special about why only S3 needs a gateway and not an interface.

IMO, understanding the difference between the two endpoint types is crucial to understanding this code. 

>the only thing I think is really important is for it to not be inline.

I never argued with the inline part, my comment was more about using a separate structure like a map representing all possible cases. I'm not against using a function to make the condition more readable.

michaelbeaumont: Once the difference between them is clear, there is still no inherent reason that only S3 needs a gateway, right? DynamoDB also has one. Here we are, sometimes enabling one, sometimes the other. Indeed as we saw, there are now both types supported for S3, so adding a new endpoint might very well require explicitly choosing which type to enable. A map would make it clearer that there's a more or less arbitrary, and from the point of `eksctl` irrelevant one way or the other, mapping between service name and endpoint type.

Maybe it's the style of programming I'm used to but in languages with a well-thought out type system you'd have a sum type `Service`. Good style would dictate making the mapping `Service -> EndpointType` explicit by listing every `Service` such that when a new `Service` is added, you'd get a compile time error. Not possible with go but can be done with map + runtime assertions `if _, ok := ....; !ok { panic() }`.

If there's some reason to think that `PrivateDNS` isn't inherently supported for every kind of interface then yes, I would also make that explicit. Same with setting route table IDs. All of these bugs are caused by implicit, hidden assumptions. There's zero cost to making them explicit that I can see.

michaelbeaumont (approved):

PR #3232, opened 2021-02-08
Support secondary CIDR blocks for VPCs

### LINKED ISSUE #2730: eksctl create not accepting vpc non-default cidr
**What happened?**
 eksctl -v4 create cluster -f eksctl/eks-course.yaml
2020-10-12T17:16:39Z [ℹ]  eksctl version 0.29.2
2020-10-12T17:16:39Z [ℹ]  using region us-east-1
Error: VPC CIDR block "10.205.117.0/24" is not the same as "10.205.251.0/25"

A new eks cluster would be created using the non-default vpc cidr

**How to reproduce it?**
create a vpc with a default cidr plus one additional one.  use the non-default on to specify the vpc cidr, including subnets under the none default cidr

```
kind: ClusterConfig

metadata:
  name: EKS-course-cluster
  region: us-east-1

vpc:
  id: "vpc-123"
  cidr: "<put in non-default cidr in vpc here>"
  subnets:
    private:
      subnet-us-east-1b:
        id: "subnet-123"
        cidr: "<put in non-default vpc subnet cidr here>"
      subnet-us-east-1a:
        id: "subnet-124"
        cidr: "<put in non-default vpc subnet cidr here>"

nodeGroups:
  - name: ng-1
    instanceType: t2.small
    desiredCapacity: 2
    volumeSize: 20
    ssh: # use existing EC2 key
      publicKeyName: 'yourkey'
    privateNetworking: true
```

**Versions**
```
eksctl version
0.29.2

kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

```

**Logs**
```
eksctl -v4 create cluster -f eksctl/eks-course.yaml 
2020-10-12T17:24:13Z [ℹ]  eksctl version 0.29.2
2020-10-12T17:24:13Z [ℹ]  using region us-east-1
2020-10-12T17:24:13Z [▶]  role ARN for the current session is "redacted"
Error: VPC CIDR block "redacted" is not the same as "redacted"

### PR DESCRIPTION
**Pull request by michaelbeaumont**
### Description
Fixes #2730 
VPCs can have secondary cidr ranges and the code unnecessarily enforced that only the primary was used.

### Checklist
- [x] Manually tested

### CODE REVIEW

File: pkg/vpc/vpc_test.go [hunk 1]
```diff
@@ -373,6 +376,52 @@ var _ = Describe("VPC - Import VPC", func() {
 			describeVPCError: nil,
 			error:            fmt.Errorf("invalid CIDR address: *"),
 		}),
+		Entry("VPC with secondary CIDR", importVPCCase{
```
Callisto13, line 379: Can we make the `Entry` strings a little more descriptive? They don't tell me anything about the expected behaviour in this situation.

Callisto13:
Is it possible to get a fuller description here explaining what the issue was and how checking the `CidrBlockAssociationSet` solves it?

pkg/vpc/vpc_test.go
michaelbeaumont: Should they? I think of the entry strings as just describing the case under test and then the `Expect` calls/the rest of the test case as specifying the expected behavior. Like if it passes I think "VPCs with secondary CIDR works as expected" if it fails then I see where it fails to meet expectations.

michaelbeaumont:
In the issue is a pretty good one I think, VPCs can have secondary cidr ranges and the code unnecessarily enforced that only the primary was used.

pkg/vpc/vpc_test.go
Callisto13: These strings in tests serve absolutely no purpose other than to describe what the code's behaviour, so it seems a shame not to make use of them. The perfect test file for me is one where I do not need to read the test code, or even the implementation code, to understand what is supposed to be happening. IMO when used properly they can make sharing context a dream and can really help people coming to a new area of code.

Callisto13:
Yeh, I get that. After reading the issue, then the code here, then googling `CidrBlockAssociationSet`. But why do I have to work so hard at 8am? Having clear TLDRs makes it easier on us for reviewing, builds a clearer historical record, and helps new people on the project. Right now figuring out why stuff went in is way more work than it should be.

pkg/vpc/vpc_test.go
michaelbeaumont: I think they just describe which case is being tested. What if there's 20 `Expect` statements? The `DescribeTable` already has `"can import VPC"`. I have no problem changing them but I'm not sure how else to word this one...

pkg/vpc/vpc_test.go
```diff
@@ -258,9 +258,12 @@ var _ = Describe("VPC - Import VPC", func() {
 				return input != nil
 			})).Return(mockResultFn, vpcCase.describeVPCError)
 
-			if err := importVPC(p, vpcCase.cfg, vpcCase.id); err != nil {
+			err := importVPC(p, vpcCase.cfg, vpcCase.id)
+			if vpcCase.error != nil {
+				Expect(err).To(HaveOccurred())
```
Callisto13, line 263: this check is redundant but nbd 🤷‍♀️

pkg/vpc/vpc_test.go
Callisto13: They don't need to be super verbose, like `"<in this case>, <does this thing>/<fails>/<does not fail>"`

michaelbeaumont: I get what you're saying but I think in these case we already have `DescribeTable("can import VPC", Entry("VPC with secondary CIDR", ...))`. I think the test case `struct` should tell me whether I'm expecting an error or it to work.

One problem is the `DescribeTable` description implies all the cases should work but some of them expect errors, I'm fixing the entire structure.

I've added a formatting function to improve the output.

pkg/vpc/vpc_test.go
michaelbeaumont: In the `if` statement I'm checking what's expected while the `Expect` asserts it to be the case

pkg/vpc/vpc_test.go
Callisto13: > I think the test case struct should tell me whether I'm expecting an error or it to work.

It _should_ but reading descriptions are faster and easier to line up outcome (pass/fail) with what was expected to happen.

> One problem is the DescribeTable description implies all the cases should work but some of them expect errors, I'm fixing the entire structure.

Yeh, then in this case it's fine if we expect everything in this bit to be "happy path". If there are some cases which error or in which the behaviour changes then it would be helpful in the description to point be clear on that.

pkg/vpc/vpc_test.go
Callisto13: On the next line you are checking the error message, so the has occurred bit is sort-of implied right?

Callisto13: but really doesn;t matter 🙃

michaelbeaumont: `.Error()` was being called so I need to check it's not null but apparently `MatchError` accepts strings so I've used that instead and also fixed the existing logic error everywhere

Callisto13 (approved):

PR #3289, opened 2021-02-16
Start integration tests from release workflow

michaelbeaumont:
### Description
It's not possible to depend on jobs from other workflows, so it needs to be done manually. The IT job doesn't really belong in the same workflow as the release jobs, in my opinion.
I realize the inline `github-script` JS is kind of heinous and it could be put into its own action... but there'd be so much boilerplate and probably we'd need to fork `actions/github-script` ... :weary:
WDYT?

~Just saw you an create composite run step actions :heart_eyes:, working on it~
Never mind, Github released a very limited version of composite actions [without `uses`](https://github.com/actions/runner/issues/646)

### Checklist
- [ ] Manually tested


Callisto13 (approved):

michaelbeaumont:
So yeah, this is one of the options, and it has to be duplicated to `release.yaml`. The other option would be to duplicate the integration test job so that it's in all 3 workflows. But I really don't like this because there's no one integration test job but rather 3 without connection to each other. This _could_ be put into an action but it'd be arguably even more annoying to maintain because I'd have to somehow fork `github-script` (otherwise check out this action for [how much javascript you have to write to create an action that makes a PR](https://github.com/peter-evans/create-pull-request/tree/master/src))

Callisto13:
> (otherwise, check out this action for how much javascript you have to write to create an action that makes a PR)

that is the most horrifying thing i have ever seen. surely there must be another way

michaelbeaumont:
I'm going to close this, I don't think it's convenient enough that it's worth committing so much inline JS

PR #3296, opened 2021-02-17
create clusters in before suites

aclevername:
### Description
When `create cluster` fails in the integration tests it ends up making all subsequent `Its` fail, making the output very hard to parse. This PR changes it so that the clusters are created in beforesuites. Its slightly annoying as a lot of the tests are actually testing the create cluster functionality, so having it in the before suite is slightly misleading. 

### Checklist
- [ ] Manually tested


michaelbeaumont:
Shouldn't the tests just be "fail fast" or something instead?

aclevername:
> Shouldn't the tests just be "fail fast" or something instead?

I think that would be best achieved by having all of the It's merged into one, but that has some downsides as not all of our tests required all the priors It's to pass for the later It's to pass. But pretty much every single test does require the cluster to be running, so its definitely a prerequisite for running the tests, so a before suite seems reasonable to me

michaelbeaumont (approved):

PR #3305, opened 2021-02-18
Fix bug in deleting clusters in unowned_cluster integration test

aclevername:
### Description
@Callisto13 :thinking: 

### Checklist
- [ ] Manually tested

### DIFF
--- a/integration/tests/unowned_cluster/unowned_cluster_test.go
+++ b/integration/tests/unowned_cluster/unowned_cluster_test.go
@@ -66,7 +66,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	AfterSuite(func() {
-		if !params.SkipDelete {
+		if !params.SkipCreate && !params.SkipDelete {
 			deleteStack(stackName, ctl)
 		}
 	})
@@ -287,7 +287,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	It("supports deleting clusters", func() {
-		if !params.SkipDelete {
+		if params.SkipDelete {
 			Skip("params.SkipDelete is true")
 		}
 		By("deleting the cluster")


Callisto13:
All part of the plan

Callisto13 (dismissed):

integration/tests/unowned_cluster/unowned_cluster_test.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	AfterSuite(func() {
-		if !params.SkipDelete {
```
michaelbeaumont, line 69: wait why is this correct?

integration/tests/unowned_cluster/unowned_cluster_test.go
```diff
@@ -287,7 +287,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	It("supports deleting clusters", func() {
-		if !params.SkipDelete {
```
michaelbeaumont, line 290: :100:

integration/tests/unowned_cluster/unowned_cluster_test.go
Callisto13: wait actually yeh. this is fine jake i am not completely rubbish

aclevername: The stack only gets created if `SkipCreate` is true, so we should only delete is if `SkipCreate` is true

Callisto13: but what if you wanted it created, but not deleted?

aclevername: then we should switch on both? :see_no_evil:

Callisto13: > The stack only gets created if SkipCreate is true
when it's `false` probably

i think skipping delete of things when `SkipDelete` is `true` is the way things should work

integration/tests/unowned_cluster/unowned_cluster_test.go
Callisto13: actually this is right too

@aclevername i suggest that this PR has no merit!

Callisto13: this should check for `true`, so it is doing the right thing

integration/tests/unowned_cluster/unowned_cluster_test.go
michaelbeaumont: `!Skip` is already way too many negations to keep together in my head

integration/tests/unowned_cluster/unowned_cluster_test.go
michaelbeaumont: Shouldn't it just be `SkipDelete || SkipCreate` like if I skip create then I probably want to skip delete? (I'm struggling to imagine when I want to `SkipCreate` in the first place tbh...)

Callisto13: oh yeh actually nvm

Callisto13: i read it wrong

aclevername: I don't get the use case for the various Skips. But it makes sense to only skip deleting the *cluster* if `SkipDelete` is true

integration/tests/unowned_cluster/unowned_cluster_test.go
aclevername: MY BRAIN IS MELTING:

We should only delete the the stack (*different to the cluster*) if:
1. It exists, so if SkipCreate was True it doesn't exist, so we shouldn't delete it
2. SkipDelete is true

integration/tests/unowned_cluster/unowned_cluster_test.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	AfterSuite(func() {
-		if !params.SkipDelete {
+		if !params.SkipCreate || !params.SkipDelete {
```
michaelbeaumont, line 69: | SkipCreate       | SkipDelete           | deleteStack  |
| ------------- |:-------------:| -----:|
| false      | false | true |
| false     | true      |   false |
| true |  false   |    false/true? | 
| true |  true   |    false |

aclevername: We don't have a stack to delete if SkipCreate is true, so the 3rd row outcome is `false`

michaelbeaumont: ok so then it should be `!params.SkipCreate && !params.SkipDelete` :dart:

integration/tests/unowned_cluster/unowned_cluster_test.go
```diff
@@ -287,7 +287,7 @@ var _ = Describe("(Integration) [non-eksctl cluster & nodegroup support]", func(
 	})
 
 	It("supports deleting clusters", func() {
-		if !params.SkipDelete {
+		if params.SkipDelete {
```
michaelbeaumont, line 290: | SkipCreate       | SkipDelete           | skip this test  |
| ------------- |:-------------:| -----:|
| false      | false | false |
| false     | true      |   true |
| true |  false   |  false/true?   | 
| true |  true   |  true  |

If I have `SkipCreate` do I have a cluster to delete?

integration/tests/unowned_cluster/unowned_cluster_test.go
aclevername: oops :+1:

integration/tests/unowned_cluster/unowned_cluster_test.go
aclevername: Yes because you provided a cluster

michaelbeaumont: i'd say we don't delete it then? since it's "managed" outside of the tests?

aclevername: but then `SkipCreate` is really `SkipCreateAndDelete`

michaelbeaumont: yeah basically or seen another way, the default of `SkipDelete` actually _depends_ on the value of `SkipCreate`

aclevername: the only other test using this functionality is `crud` and that follows the `SkipCreate` only means skip creating, not deleting too. I think this is correct for now.

Really this whole thing should get refactored, but thats a different PR

michaelbeaumont: ok!

michaelbeaumont (approved):
seems about right!

aclevername:
> seems about right!

I'll take it! Merging.

PR #3323, opened 2021-02-22
Create nodegroups/managed-nodegroups on unowned clusters

Callisto13:
### Description

Most of the mechanism for creating managed and unmanaged nodegroups is stays the same whether they are on an owned or an unowned cluster. This PR adds a VPC config "getter" which will fetch the required networking information from either the Cluster Stack (in the case of owned clusters), or the provided Cluster Config Spec.
To use this feature we expect users to provide their VPC spec, which I think is reasonable.

I updated some delete procedures to make sure we cover all types of nodegroups which need to die.

I also included an integration test here instead of in a separate PR.

Other random bits thrown in for free:
- Stop creating a new stack manager everywhere. Some places were creating a new stack manager, then calling a func which created another one.
- Shuffle some stuff around in ctl/create/cluster. Huge funcs inside other funcs really upsets me.
- New StackNotFoundErr type.
- Some reorganisation in actions/nodegroup create, and tidying up of how we init a new ng manager

I didn't go into refactoring here, since there are already 40+ files in this changeset. There is a lot I want to do around tests, which I will PR separately.

Still TODO:
- Tidy some unit tests (in another PR)
- Docs on running commands on non-eksctl created clusters (in another PR)

### Checklist
- [x] Manually tested


pkg/actions/nodegroup/create.go
```diff
@@ -86,20 +76,27 @@ func (m *Manager) Create(options CreateOpts, nodegroupFilter filter.NodeGroupFil
...
 
+	printer := printers.NewJSONPrinter()
 	if err := printer.LogObj(logger.Debug, "cfg.json = \\\n%s\n", cfg); err != nil {
 		return err
 	}
 
-	// TODO
-	if err := ctl.ValidateClusterForCompatibility(cfg, stackManager); err != nil {
-		return errors.Wrap(err, "cluster compatibility check failed")
+	if !isUnownedCluster {
```
aclevername, line 84: `isUnownedCluster` is only  used in an `if & else` statement or an `if !isUnownedCluster`, so could we called it `isOwnedCluster` instead so I never have to parse the ! in my head :smile: :sweat_smile:

aclevername (dismissed):
LGTM! 1 nit. Only thing I'm wondering if we have any edgecases for the different nodegroup types, but I can't think of anything obvious :shrug:

aclevername:
thought overnight, have you tested that `delete` works?

Callisto13:
Yeh I was doing a lot of tearing down stuff as I went, tho the integration test is now failing on that 😂

Callisto13:
Aaah. I had been deleting nodegroups, but not clusters. Deleting the whole cluster leave ng stacks behind, which does not happen when deleting them by themselves. Must be a discrepancy somewhere.

pkg/cfn/manager/delete_tasks.go
```diff
@@ -115,6 +115,63 @@ func (c *StackCollection) NewTasksToDeleteNodeGroups(shouldDelete func(string) b
...
+func (d *DeleteUnownedNodegroupTask) Describe() string {
+	return d.info
+}
+
+func (d *DeleteUnownedNodegroupTask) Do(errorchan chan error) error {
+	out, err := d.eksAPI.DeleteNodegroup(&eks.DeleteNodegroupInput{
+		ClusterName:   &d.cluster,
+		NodegroupName: &d.nodegroup,
+	})
+	go func() {
```
michaelbeaumont, line 141: I think you can use https://github.com/weaveworks/eksctl/blob/8b2a568175167a53bfe09a700b68ffa4f65a0162/pkg/eks/nodegroup.go#L361 here?

aclevername (approved):
LGTM

pkg/cfn/manager/interface.go
```diff
@@ -51,15 +52,17 @@ type StackManager interface {
...
 	NewClusterCompatTask() tasks.Task
 	NewTasksToCreateIAMServiceAccounts(serviceAccounts []*v1alpha5.ClusterIAMServiceAccount, oidc *iamoidc.OpenIDConnectManager, clientSetGetter kubernetes.ClientSetGetter) *tasks.TaskTree
 	DeleteTasksForDeprecatedStacks() (*tasks.TaskTree, error)
 	NewTasksToDeleteClusterWithNodeGroups(deleteOIDCProvider bool, oidc *iamoidc.OpenIDConnectManager, clientSetGetter kubernetes.ClientSetGetter, wait bool, cleanup func(chan error, string) error) (*tasks.TaskTree, error)
-	NewTasksToDeleteNodeGroups(shouldDelete func(string) bool, wait bool, cleanup func(chan error, string) error) (*tasks.TaskTree, error)
+	NewTasksToDeleteNodeGroups(wait bool, cleanup func(chan error, string) error) (*tasks.TaskTree, error)
+	NewTaskToDeleteUnownedNodeGroup(clusterName, nodegroup string, eksAPI eksiface.EKSAPI, waitCondition *DeleteWaitCondition) tasks.Task
 	NewTasksToDeleteOIDCProviderWithIAMServiceAccounts(oidc *iamoidc.OpenIDConnectManager, clientSetGetter kubernetes.ClientSetGetter) (*tasks.TaskTree, error)
 	NewTasksToDeleteIAMServiceAccounts(shouldDelete func(string) bool, clientSetGetter kubernetes.ClientSetGetter, wait bool) (*tasks.TaskTree, error)
```
Callisto13, line 67: @aclevername I am not sure where we are with removing this?

aclevername: what do you mean?

Issue #3031, opened 2021-01-05
Add support for Asia Pacific (Osaka) region


Callisto13:
Blocked on completion of https://github.com/aws/containers-roadmap/issues/1142. More details will be added later.

Callisto13:
We can go ahead and get this merged for the next release, ahead of the release from EKS.

Note that Fargate will not be available in that region, we used to have some code around this, we should bring it back.

PR #3347, opened 2021-02-24
Add support for Asia Pacific (Osaka) region

Callisto13:
### Description

Closes https://github.com/weaveworks/eksctl/issues/3031.

The `ap-northeast-3` region is not opt-in, so it can use the standard account ID, which means adding support for it is very simple (a little too simple, let me know if there is more to it 😅 ).

Fargate support in this region will lag behind for a couple of days, so there is now a check on create to log this. I have also added a warning in the docs, but since this will be removed in like a week, I wonder if it is necessary? Maybe simply pinning an issue in this repo will be enough to cover us for that time 🤷‍♀️ .

### Checklist
- [ ] Manually tested


pkg/eks/fargate.go
```diff
@@ -64,12 +65,15 @@ func DoCreateFargateProfiles(config *api.ClusterConfig, fargateClient FargateCli
...
-		switch errors.Cause(err).(type) {
-		case nil:
+		switch {
+		case err == nil:
 			logger.Info("created Fargate profile %q on EKS cluster %q", profile.Name, clusterName)
-		case *eks.ResourceInUseException:
+		case errors.As(err, &e):
 			logger.Info("Either Fargate profile %q already exists on EKS cluster %q or another profile is being created/deleted, no action taken", profile.Name, clusterName)
+		case fargate.IsUnauthorizedError(err):
+			return errors.Wrapf(err, "either account is not authorized to use Fargate or region %s is not supported", config.Metadata.Region)
```
aclevername, line 76: big nit: can we add a unit test :smile:

aclevername (approved):
LGTM

pkg/eks/fargate.go
Callisto13: yep. we do have a test which check for failing errors (unlike an other in this switch which would not fail only log), so it is kindof covered, but i can add something explicit for this error

Callisto13: the only difference here is that we alter the error message

Issue #3311, opened 2021-02-19
Need support for equivalent CLI "--fargate" option in a configuration file


ZillaG:
**What feature/behavior/change do you want?**
```
$ eksctl version
0.38.0
```
When creating a Fargate-only cluster, this is done on the command line
`$ eksctl create cluster ... --fargate`

However, I can't pass that option when using a config file
```
$ eksctl create cluster ... --fargate -f config.yaml
Error: cannot use --fargate when --config-file/-f is set
```
I don't see an equivalent option in [Config file schema](https://eksctl.io/usage/schema/#latest-schema) doc. Can you please add it?

**Why do you want this feature?**

I created my Fargate-only cluster using a config file, and chased a problem for 2 days on why `coredns` wasn't coming up. It's because, it was set up to `eks.amazonaws.com/compute-type: ec2`. I had to edit it to `fargate` to get it to come up, but this should really be in a configuration if it's not supported on a command line along with a config file option.

Callisto13:
Hey @ZillaG, thanks for raising this issue.

Unfortunately this is a case of crappy docs.

The equivalent of running `create cluster --fargate` is to add a `fargateProfiles` section in the config file (eg [this example](https://github.com/weaveworks/eksctl/blob/main/examples/16-fargate-profile.yaml)). The `fp-default` profile in that config is what the `--fargate` flag creates "behind the scenes".

I will review our docs and make sure that this is clear.

ZillaG:
Here's my config file, taken directly from the example you posted. As you can see I have fargate profiles. However I'm using existing VPC and only private subnets and only enable private endpoints. (I should've added this to my post, my bad.). I can also reproduce the problem even if I let EKS create it's own VPC, but only use private endpoints. If I let EKS create the cluster's VPC using both public and private endpoints, I don't get the issue.

```
kind: ClusterConfig
apiVersion: eksctl.io/v1alpha5

metadata:
  name: sandbox
  region: us-east-1
  version: "1.19"

# The VPC and subnets are for the data plane, where the pods will
# ultimately be deployed.
vpc:
  id: "vpc-xxxxx"
  clusterEndpoints:
    privateAccess: true
    publicAccess: false
  subnets:
    # us-east-1a is full
    private:
      us-east-1b:
        id: "subnet-xxxxx"
      us-east-1c:
        id: "subnet-yyyyy"

fargateProfiles:
  - name: fp-default
    selectors:
[...]

PR #3350, opened 2021-02-24
Clarify difference between fargate flag and config

Callisto13:
### Description

Closes https://github.com/weaveworks/eksctl/issues/3311.

We do have a UX discrepancy when it comes to enabling fargate. Users can currently run `create cluster --fargate` and have a default profile created for them without any customisation. There is no direct equivalence for this in the config file: users have to define `fargateProfiles` in the spec, and be explicit about what they want. The flag sets the `FargateProfile` object, which is what we use as the marker for deciding whether to prep the cluster for Fargate (and then create profiles).

I don't believe we should have an `enableFargate: bool` in the config. Having it on the command line is fine, since flags are more useful for people playing around with `eksctl`.  We tend to assume that folks use config to have finer control over their cluster, and to use more advanced options (there are several options we do not expose via flags at all). I believe `kubectl` has a similar principle. I also think that adding this field would lead to wider UX and internal changes.

### Checklist
- [ ] Manually tested

### DIFF
--- a/userdocs/src/usage/fargate-support.md
+++ b/userdocs/src/usage/fargate-support.md
@@ -47,7 +47,11 @@ $ eksctl create cluster --fargate
 [✔]  EKS cluster "ridiculous-painting-1574859263" in "ap-northeast-1" region is ready
 ```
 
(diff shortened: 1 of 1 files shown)


aclevername (approved):
LGTM

PR #3354, opened 2021-02-25
eksctl info command

### LINKED ISSUE #3029: eksctl info command
Currently we ask users to provide the output from `eksctl version` and `kubectl version` when opening a bug issue. I want users to be able to easily obtain all necessary information in one command.

A command `eksctl info` that provides the `eksctl version`, `kubectl version` and the OS type. This can be expanded to show more output that we deem useful. 

```bash
$ eksctl info
eksctl version: 0.35.0
kubectl version: 1.18.2
OS: MacOS
```

Thoughts @weaveworks/eksctl-reviewers ?

neha-viswanathan:
Closes #3029 

### Description

This MR contains an implementation of the `eksctl info` command

### Checklist
- [x] Manually tested

### DIFF
--- a/cmd/eksctl/info.go
+++ b/cmd/eksctl/info.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/info"
+)
+
+func infoCmd(cmd *cmdutils.Cmd) {
+	var output string
+
+	cmd.SetDescription("info", "Output the version of eksctl, kubectl and OS info", "")
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&output, "output", "o", "", "specifies the output format (valid option: json)")
+	})
+	cmd.CobraCommand.Args = cobra.NoArgs
+	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
+		switch output {
+		case "":
+			version := info.GetInfo()
+			fmt.Printf("eksctl version: %s\n", version.EksctlVersion)
+			fmt.Printf("kubectl version: %s\n", version.KubectlVersion)
+			fmt.Printf("OS: %s\n", version.OS)
+		case "json":
+			fmt.Printf("%s\n", info.String())
+		default:
+			return fmt.Errorf("unknown output: %s", output)
+		}
+		return nil
+	}
+}
--- a/pkg/info/info.go
+++ b/pkg/info/info.go
@@ -1,0 +1,62 @@
+package info
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+	"unicode"
+
+	"github.com/weaveworks/eksctl/pkg/version"
+)
+
+//go:generate go run ./release_generate.go
+
+// Info holds versions info
+type Info struct {
+	EksctlVersion  string
+	KubectlVersion string
+	OS             string
+}
+
+// GetInfo returns versions info
+func GetInfo() Info {
+	return Info{
+		EksctlVersion:  getEksctlVersion(),
+		KubectlVersion: getKubectlVersion(),
+		OS:             runtime.GOOS,
+	}
+}
+
+// getEksctlVersion returns the eksctl version
+func getEksctlVersion() string {
+	return version.GetVersion()
+}
+
+// getKubectlVersion returns the kubectl version
+func getKubectlVersion() string {
+	cmd := exec.Command("kubectl", "version", "--short", "--client")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Sprintf("error : %v", err)
+	}
+
+	values := strings.FieldsFunc(string(out), func(c rune) bool {
+		return unicode.IsSpace(c)
+	})
+	if len(values) == 0 {
+		return "unknown version"
+	}
+	return values[len(values)-1]
+}
+
+// String return info as JSON
+func String() string {
+	data, err := json.Marshal(GetInfo())
+	if err != nil {
+		return fmt.Sprintf("failed to marshal info into json: %q", err)
+	}
+
+	return string(data)
+}
--- a/pkg/info/info_test.go
+++ b/pkg/info/info_test.go
@@ -1,0 +1,46 @@
+package info
+
+import (
+	"encoding/json"
+	"strings"
+
+	semver "github.com/Masterminds/semver/v3"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Info", func() {
+	Context("GetInfo", func() {
+		It("returns the eksctl and kubectl versions, and the host os", func() {
+			result := GetInfo()
+
+			Expect(result.EksctlVersion).ToNot(Equal(""))
+			Expect(result.KubectlVersion).ToNot(Equal(""))
+			Expect(result.OS).ToNot(Equal(""))
+
+			_, err := semver.NewVersion(strings.TrimSpace(result.EksctlVersion))
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = semver.NewVersion(strings.TrimSpace(strings.TrimPrefix(result.KubectlVersion, "v")))
+			Expect(err).NotTo(HaveOccurred())
+
+			oses := []string{"aix", "android", "darwin", "dragonfly", "freebsd", "hurd", "illumos", "ios", "js", "linux", "nacl", "netbsd", "openbsd", "plan9", "solaris", "windows", "zos"}
+			Expect(result.OS).To(BeElementOf(oses))
+		})
+	})
+
+	Context("String", func() {
+		It("returns Info in json", func() {
+			result := String()
+
+			infos := Info{}
+			err := json.Unmarshal([]byte(result), &infos)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(infos.EksctlVersion).ToNot(Equal(""))
+			Expect(infos.KubectlVersion).ToNot(Equal(""))
+			Expect(infos.OS).ToNot(Equal(""))
+		})
+	})
+
+})
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -55,10 +55,9 @@ What type of AWS credentials are you using (i.e. default/named profile, MFA)? - 
 
 **Versions**
 <!--
-Please paste in the output of these commands:
+Please paste in the output of this command:
 -->
 ```
-$ eksctl version
-$ kubectl version
+$ eksctl info
 ```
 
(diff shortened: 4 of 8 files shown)


neha-viswanathan:
@aclevername 👋🏽 this is a rough first attempt of the `eksctl info` command. Please review and let me know your suggestions.

aclevername:
Thanks @neha-viswanathan ! Looks good, could we do some error handling? For example if we fail to discover the kubectl client we should just present `unknown` or something like that :smile:

neha-viswanathan:
@aclevername ready for round 2 :)

neha-viswanathan:
@Callisto13 Hi! I'll need your guidance here - I don't understand why the `make lint` is failing? Please help!

Callisto13:
No problem 😃 .

So the linter is reporting this error:
```
errors in package: [/__w/eksctl/eksctl/cmd/eksctl/main.go:45:49: undeclared name: infoCmd]"
```

Because the `infoCmd` function you are referencing on line 45 of `cmd/eksctl/main.go` does not exist. (This line here: `cmdutils.AddResourceCmd(flagGrouping, rootCmd, infoCmd)`.) You need to create that `infoCmd` function, which should call your new `GetVersion` function.

You can use the line below in `cmd/eksctl/main.go` which wires up the `versionCmd` as a guide. `versionCmd` is defined in `cmd/eksctl/version.go`, so following that pattern you can create a new `cmd/eksctl/info.go` which will define `infoCmd` and call your `GetVersion` function.

Hope this helps, let me know if you have any more questions or if anything is unclear 👍

neha-viswanathan:
@Callisto13 for some reason that file did not get committed earlier. 😕 Thanks for pointing it out. 😃

Callisto13:
> for some reason that file did not get committed earlier

Gah! that's annoying 😅

cmd/eksctl/info.go
```diff
@@ -0,0 +1,33 @@
...
+	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
+		fs.StringVarP(&output, "output", "o", "", "specifies the output format (valid option: json)")
+	})
+	cmd.CobraCommand.Args = cobra.NoArgs
+	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
+		switch output {
+		case "":
+			fmt.Printf("eksctl version: %s\n", info.GetVersion().EksctlVersion)
+			fmt.Printf("kubectl version: %s\n", info.GetVersion().KubectlVersion)
+			fmt.Printf("OS: %s\n", info.GetVersion().OS)
```
Callisto13, line 25: Do you think we could just call `GetVersion` just the once in this block? Would be good to avoid execing to kubectl 3 times in a row?

pkg/info/info.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Version holds versions info
+type Version struct {
+	EksctlVersion  string
+	KubectlVersion string
+	OS             string
+}
+
+// GetVersion returns versions info
+func GetVersion() Version {
```
Callisto13, line 24: Is is possible to have some tests for this file please?

pkg/info/info.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return values[len(values)-1]
+}
+
+// String returns eksctl info as a string
+func String() string {
+	if data, err := json.Marshal(GetVersion()); err == nil {
+		return string(data)
+	}
+	return ""
```
Callisto13, line 54: Can we do the inverse of this?
```suggestion
	if data, err := json.Marshal(GetVersion()); err != nil {
		return fmt.Sprintf("error : %v", err)
	}
	return string(data)
```

pkg/info/info.go
```diff
@@ -0,0 +1,55 @@
...
+type Version struct {
+	EksctlVersion  string
+	KubectlVersion string
+	OS             string
+}
+
+// GetVersion returns versions info
+func GetVersion() Version {
+	return Version{
+		EksctlVersion:  version.String(),
```
Callisto13, line 26: This returns quite a lot of stuff:
```
eksctl version: {"Version":"0.44.0","PreReleaseID":"dev","Metadata":{"BuildDate":"2021-03-26T15:34:25Z","GitCommit":"ee316d40"},"EKSServerSupportedVersions":["1.15","1.16","1.17","1.18","1.19"]}
```


I also think whatever is happening here is messing with the json (escaping everything):
```
$ ./eksctl info -o json
{"EksctlVersion":"{\"Version\":\"0.44.0\",\"PreReleaseID\":\"dev\",\"Metadata\":{\"BuildDate\":\"2021-03-26T15:34:25Z\",\"GitCommit\":\"ee316d40\"},\"EKSServerSupportedVersions\":[\"1.15\",\"1.16\",\"1.17\",\"1.18\",\"1.19\"]}","KubectlVersion":"v1.19.4","OS":"linux"}
```

I think some tests would help drive out desired UX in both these cases.

Callisto13 (changes requested):

pkg/info/info.go
neha-viswanathan: @Callisto13 instead of returning all this info, would it be enough to just display the version?

Callisto13: yes, ideally we would just show `eksctl version: 0.40.0` or similar

neha-viswanathan: @Callisto13 calling GetVersion() for retrieving eksctl's version returns a different version (version set in pkg/version/release.go). So I think we should go back to the previous implementation I had by executing the `eksctl version` command. Thoughts?
 ```
./eksctl info  
eksctl version: 0.44.0-dev
kubectl version: v1.19.7
OS: darwin

eksctl version
0.42.0
```

Comment (Callisto13) [reply pkg/info/info.go:26]: I would rather avoid shelling out to ourselves if we don't have to. Can we add a func somewhere to get the exact thing we need internally?

**Comment by Callisto13** [comment]
Hey @neha-viswanathan! Just checking in on the status of this, do you plan to finish or would you like it taken over? Thanks!

**Comment by neha-viswanathan** [comment]
> Hey @neha-viswanathan! Just checking in on the status of this, do you plan to finish or would you like it taken over? Thanks!

Hi @Callisto13 I'll try and wrap this up within the next week or so. I will keep you posted in any case.

**Comment by Callisto13** [comment]
Awesome 🎉

File: pkg/info/info.go (hunk 4 above)
Comment (neha-viswanathan) [reply pkg/info/info.go]: @Callisto13 🙋🏽‍♀️ Need some pointers on how I can extract the versions internally? Thanks!

Comment (Callisto13) [reply pkg/info/info.go]: Hey! Just looking back and I realised I made a mistake 🤦‍♀️ : 

Above when you built and ran the command locally you got:

```sh
./eksctl info  
eksctl version: 0.44.0-dev
kubectl version: v1.19.7
OS: darwin
```

and thought the version was wrong as it used the version set in pkg/version/release.go. This was actually correct, since the version is a local dev build built from main, which will not be the same as if you just run `eksctl version`. using your installed release version.

So TLDR: you _can_ use `version.GetVersion()`, it will return the correct thing.

pkg/info/info_test.go
```diff
@@ -0,0 +1,40 @@
...
+	It("wrong eksctl, kubectl versions and OS", func() {
+		info := GetVersion()
+
+		Expect(info).To(Equal(Version{
+			EksctlVersion:  "0.40.0",
+			KubectlVersion: "v1.10.1",
+			OS:             "amd64",
+		}))
+	})
+
```
Callisto13, line 39: I think this test can be simpler. The `kubectl` and `OS` are very unlikely to be ignored, although I am sure you can find a way to force them to disappear. Likewise there will never be a "wrong" version for any of them.

```suggestion
 var _ = Describe("Info", func() {
   Context("GetInfo", func() {
     It("returns the eksctl and kubectl versions, and the host os", func() {
       result := info.GetInfo()

       Expect(result.EksctlVersion).ToNot(Equal(""))
       Expect(result.KubectlVersion).ToNot(Equal(""))
       // we can't check for a specific version since it will change depending
       // on who is running the tests, but we can check that actual semvers are returned
       _, err := semver.NewVersion(strings.TrimRight(result.EksctlVersion, "\n"))
       Expect(err).NotTo(HaveOccurred())
       _, err = semver.NewVersion(strings.TrimRight(result.KubectlVersion, "\n"))
       Expect(err).NotTo(HaveOccurred())

       Expect(result.OS).ToNot(Equal(""))
       // technically a cheat, but
       // testing for the OS is harder, we could possibly have an array of OSes and see
       // if the one here  matches, if you feel like doing that
       Expect(result.OS).To(Equal(runtime.GOOS))
     })
   })

   Context("String", func() {
     It("returns Info in json", func() {
       result := info.String()

       // make processing easier, turn into a struct
       versions := info.Versions{}
       err := json.Unmarshal([]byte(result), &versions)
       Expect(err).NotTo(HaveOccurred())

       // same again basically, we can just check they are not nil
       Expect(versions.EksctlVersion).ToNot(Equal(""))
       Expect(versions.KubectlVersion).ToNot(Equal(""))
       Expect(versions.OS).ToNot(Equal(""))
     })
   })
   
   // There are is a potential failure case, but it is trickier
   // to test at this level. To test that the right thing happens when kubectl fails
   // to execute, you could remove the binary at the start of the test and put it back at the end
   // but if that goes wrong then you dont have a kubectl anymore :D 
   // Another option is to have the kubectl binary be an exported location which you can reset
   // in the test to a dummy bin. But this is a lot of work for a small functionality,
   // so if that sound like a pain I am happy to ignore it 
 })
```

neha-viswanathan: @Callisto13 thank you so much! Those tests definitely simplified everything.
1. regarding the OS test, it looks like these OSes are supported: https://github.com/golang/go/blob/master/src/go/build/syslist.go#L10. 
2. The kubectl test of moving the binary sounds a bit of an overkill. 

I feel the tests you have suggested look great. And thanks again for helping me out.

pkg/info/info.go
```diff
@@ -0,0 +1,60 @@
...
+	return values[len(values)-1]
+}
+
+// String return info as JSON
+func String() string {
+	if data, err := json.Marshal(GetInfo()); err == nil {
+		return string(data)
+	}
+	return ""
+}
```
Callisto13, line 60: let's print something here when the marshal fails, just so users don't get confused about why the command returned nothing:

```suggestion
func String() string {
	data, err := json.Marshal(GetInfo())
	if err != nil {
		return fmt.Sprintf("failed to marshal info into json: %q", err)
	}
	return string(data)
}
```

pkg/info/info_test.go
Callisto13: > regarding the OS test, it looks like these OSes are supported: https://github.com/golang/go/blob/master/src/go/build/syslist.go#L10.

cool, if you want to implement a match loop there feel free. 

> The kubectl test of moving the binary sounds a bit extra.

👍  no worries

pkg/info/info_test.go
```diff
@@ -0,0 +1,46 @@
...
+			Expect(result.OS).ToNot(Equal(""))
+
+			_, err := semver.NewVersion(strings.TrimSpace(result.EksctlVersion))
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = semver.NewVersion(strings.TrimSpace(strings.TrimPrefix(result.KubectlVersion, "v")))
+			Expect(err).NotTo(HaveOccurred())
+
+			oses := []string{"aix", "android", "darwin", "dragonfly", "freebsd", "hurd", "illumos", "ios", "js", "linux", "nacl", "netbsd", "openbsd", "plan9", "solaris", "windows", "zos"}
+			Expect(result.OS).To(BeElementOf(oses))
```
Callisto13, line 28: oooh i didn't know `BeElementOf` was a thing! nice!

Callisto13 (approved):
LGTM 🚀  thanks for working on this!

PR #3387, opened 2021-03-02
Merge 40 to main

Callisto13:
The auto job failed.

### DIFF
--- a/pkg/version/release.go
+++ b/pkg/version/release.go
@@ -6,7 +6,7 @@ package version
 var Version = "0.40.0"
 
 // PreReleaseID can be empty for releases, "rc.X" for release candidates and "dev" for snapshots
-var PreReleaseID = "dev"
+var PreReleaseID = "rc.0"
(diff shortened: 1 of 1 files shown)


pkg/version/release.go
```diff
@@ -6,7 +6,7 @@ package version
 var Version = "0.40.0"
 
 // PreReleaseID can be empty for releases, "rc.X" for release candidates and "dev" for snapshots
-var PreReleaseID = "dev"
+var PreReleaseID = "rc.0"
```
aclevername, line 9: we don't want this in main right?

Callisto13: honest to god no idea 🤷‍♀️

Callisto13: and some merge to main is failing below

aclevername: we definitely don't want this in main as we want to keep cutting dev releases by default

PR #3395, opened 2021-03-04
Cherry pick some things from main to 40

Callisto13:
Adds the following

- Fix 3393: fully-private clusters are not supported in region ""
- Create nodegroups/managed-nodegroups on unowned clusters
- Updated release notes


aclevername (approved):
LGTM

PR #3399, opened 2021-03-04
Set systemd cgroupdriver when HasMixedInstances 

Callisto13:
Adding one last thing to 40

- Set systemd cgroupdriver when HasMixedInstances (fix 3391)


aclevername (approved):
LGTM

Issue #3405, opened 2021-03-08
cli: Invalid images are accepted.


joshti:
**What were you trying to accomplish?**
Tried to deploy a cluster with Bottlerocket.

**What happened?**
AL2 deployed instead.

**How to reproduce it?**
`eksctl create nodegroup --cluster my-favourite-cluster --node-type t3.medium --node-ami Calcium --max-pods-per-node 100 --ssh-access`

**Anything else we need to know?**
There isn't a node-ami with the id of `Calcium`. `Bottlerocket` reproduces the same issue, but I'm assuming this could change with time. `node-ami-family` works as expected with `Bottlerocket`, but I believe doesn't error out either.

**Versions**

```
[cloudshell-user@ip-10-0-104-110 ~]$ eksctl version
0.37.0
```

cPu1:
Thanks for reporting this issue. This is indeed a bug in eksctl as it's not disallowing invalid AMI values, and a confusing behaviour because it defaults to using SSM parameters to resolve the AMI. A [PR](https://github.com/weaveworks/eksctl/pull/3412) has been opened to fix the issue, it will be out in the next release.

joshti:
Thanks much 😸

PR #3412, opened 2021-03-08
Disallow invalid AMI values

cPu1:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/3405

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/nodegroup_flags.go
+++ b/pkg/ctl/cmdutils/nodegroup_flags.go
@@ -39,7 +39,7 @@ func AddCommonCreateNodeGroupFlags(fs *pflag.FlagSet, cmd *Cmd, ng *api.NodeGrou
 	ng.SSH.PublicKeyPath = fs.String("ssh-public-key", "", "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
 	ng.SSH.EnableSSM = fs.Bool("enable-ssm", false, "Enable AWS Systems Manager (SSM)")
 
-	fs.StringVar(&ng.AMI, "node-ami", "", "'auto-ssm', 'auto', 'static' (deprecated) or an AMI id (advanced use)")
+	fs.StringVar(&ng.AMI, "node-ami", "", "'auto-ssm', 'auto' or an AMI ID (advanced use)")
 	fs.StringVar(&ng.AMIFamily, "node-ami-family", api.DefaultNodeImageFamily, "'AmazonLinux2' for the Amazon EKS optimized AMI, or use 'Ubuntu2004' or 'Ubuntu1804' for the official Canonical EKS AMIs")
 
 	fs.BoolVarP(&ng.PrivateNetworking, "node-private-networking", "P", false, "whether to make nodegroup networking private")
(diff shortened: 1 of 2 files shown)


Callisto13 (approved):
👍

pkg/ctl/cmdutils/nodegroup_flags.go
```diff
@@ -39,7 +39,7 @@ func AddCommonCreateNodeGroupFlags(fs *pflag.FlagSet, cmd *Cmd, ng *api.NodeGrou
 	ng.SSH.PublicKeyPath = fs.String("ssh-public-key", "", "SSH public key to use for nodes (import from local path, or use existing EC2 key pair)")
 	ng.SSH.EnableSSM = fs.Bool("enable-ssm", false, "Enable AWS Systems Manager (SSM)")
 
-	fs.StringVar(&ng.AMI, "node-ami", "", "'auto-ssm', 'auto', 'static' (deprecated) or an AMI id (advanced use)")
+	fs.StringVar(&ng.AMI, "node-ami", "", "'auto-ssm', 'auto' or an AMI ID (advanced use)")
```
aclevername, line 42: could we set a default string? Instead of relying on empty string meaning the user didn't set it

cPu1: >could we set a default string? Instead of relying on empty string meaning the user didn't set it


If we set a default string, in addition to coming up with a name for the default behaviour, we would also have to validate against the user setting that default value. This isn't a behaviour we want to expose to end users. 

>Instead of relying on empty string meaning the user didn't set it

The empty string (or the zero value) is often used to signal default behaviour, and is often preferred because it doesn't require a type to be initialised with certain values for it to be usable. This also means that code that calls `ResolveAMI` elsewhere, including tests, would have to ensure `ng.AMI` is initialised with a default value.

aclevername: What value does `""` represent though? What does 
```
		resolver = ami.NewMultiResolver(
			ami.NewSSMResolver(provider.SSM()),
			ami.NewAutoResolver(provider.EC2()),
		)
```

Mean? If we have a value for this the user can be specific in wanting this and note that we default to it

aclevername: > The AMI resolution is an implementation detail that users should not have to care about. 

But we do expose it for them to configure? This combination is exposed to the user, the way they can configure it so by omitting instead of specifying, which is what I think I find weird.

> This is also not a feature we want exposed to end users, because once SSM support for Ubuntu is out, the AutoResolver will be deprecated. We'll then have to handle deprecation and removal of the feature.

Ok I'm happy for it to stay as is then :+1:

aclevername (approved):
LGTM

pkg/ctl/cmdutils/nodegroup_flags.go
cPu1: >But we do expose it for them to configure? This combination is exposed to the user, the way they can configure it so by omitting instead of specifying, which is what I think I find weird.

Giving it a specific value that describes the behaviour and letting users configure it will expose it to them. Right now, the default behaviour is unknown to end users and can be changed later.

Issue #3403, opened 2021-03-05
`create cluster` makes too many CloudFormation API requests


cPu1:
**Problem**
The `create cluster` command uses CloudFormation to create the cluster stack, and polls the CloudFormation stack periodically with a consistent rate to enquire its status. During testing, it was observed that it can make up to 50 API requests just for the cluster's stack. This sometimes results in a rate-limiting error but the code continues to poll the stack at the same rate, exacerbating the problem.

For EKS 1.18 and below, it can take up to 15 minutes for the cluster to become active, for 1.19 (and above), it's under 9 minutes. After the CloudFormation stack has passed validation for most resources and has reached the `ControlPlane` CFN resource, there's no reason to keep polling every 15 to 20 seconds because it's known that it will take a few minutes for the cluster to become active. 

**Solution**
1. After the initial request, poll the CloudFormation stack every ~20 seconds to make sure any errors are surfaced immediately, but after the stack has reached the ControlPlane resource, increase the poll duration to a minute
1. Use an exponential backoff mechanism upon encountering a rate-limiting error
1. Use the result of the last DescribeStack call to avoid having to issue an additional request

cPu1:
Closed by #3421 and #3424

PR #3421, opened 2021-03-10
Minimize number of CFN calls for `create cluster`

cPu1:
### Description

This PR brings down the number of CloudFormation requests for `create cluster` from ~50 to ~15. After creating a stack, it waits 30 seconds for the first two requests to ensure any CloudFormation validation errors are surfaced immediately, and increases the wait duration to 1 minute for subsequent requests. This does mean that if an error is encountered later, e.g., while creating the EKS `ControlPlane` resource itself, it may take up to a minute for the error to surface. 

Addresses points (1) and (3) from https://github.com/weaveworks/eksctl/issues/3403

### Checklist
- [x] Manually tested


pkg/cfn/waiter/waiter.go
```diff
@@ -0,0 +1,40 @@
...
+	waitTimer := time.NewTimer(d)
+	select {
+	case <-waitTimer.C:
+		return p.Operation()
+
+	case <-ctx.Done():
+		waitTimer.Stop()
+		return false, ctx.Err()
+	}
+}
```
aclevername, line 40: could we not have used the existing waiter from the AWS package? It has the option to configure the timeout via an func https://github.com/aws/aws-sdk-go/blob/60fb75bcde84d6dc06ca60eef07626c542dedd3b/aws/request/waiter.go#L28-L34

cPu1: > It has the option to configure the timeout via an func

It does and our existing code relies heavily on that package, but I found a few issues with it:

1. It uses the concept of acceptors, matchers, PathAllWaiterMatch etc that is too complicated for our requirements. We just need to keep retrying an operation until it's successful. 
2. It's not type-safe. It extracts the value to match against using a JMESPath like `Stacks[].StackStatus`.
3. More importantly, that package was never intended to be used outside of the AWS SDK, as is also indicated by this comment https://github.com/aws/aws-sdk-go/blob/60fb75bcde84d6dc06ca60eef07626c542dedd3b/aws/request/waiter.go#L72-L73

aclevername: For 1 & 2 I haven't seen any problems with us using it, it seems to work well for the majority of our use cases. The comment for 3 is 4 years old and references a func that doesn't exist as far as I can see, so I'm not entirely sure thats true. I don't think we've had any problems using this waiter elsewhere in the codebase so I'm a bit against introducing a third waiter pattern. We already have a 2nd method [here](https://github.com/weaveworks/eksctl/blob/122dc492be096c30ddd4fe7b291838d16132678e/pkg/utils/waiters/waiters.go#L15-L38), could that be refactored to fit this use case if we won't use the AWS waiter? I'm just wary that this introduces a 3rd method for waiting in the codebase, which might be unnecessary

cPu1: >For 1 & 2 I haven't seen any problems with us using it, it seems to work well for the majority of our use cases. 

That's because it has been there forever and is rarely touched, but that doesn't mean it's not technical debt. Anyone new to that code would have to get familiar with acceptors, matchers etc and the fact that it's using JMESPath to extract values. 

>The comment for 3 is 4 years old and references a func that doesn't exist as far as I can see, so I'm not entirely sure thats true.

It's referring to the usage of the type itself and the `WaitUntil*` functions do exist in the service package. 

>We already have a 2nd method here, could that be refactored to fit this use case if we won't use the AWS waiter? I'm just wary that this introduces a 3rd method for waiting in the codebase, which might be unnecessary

I'd argue that the function can be removed in favour of the new one. That one doesn't take a context, accepts the error to return when a timeout error occurs, and there's no need for both a `time.Ticker` and `time.Timer`.

aclevername: > That's because it has been there forever and is rarely touched, but that doesn't mean it's not technical debt. Anyone new to that code would have to get familiar with acceptors, matchers etc and the fact that it's using JMESPath to extract values.

I added some usage for it for addons and it went reasonably smoothly, but I agree its not as intuitive
 
> I'd argue that the function can be removed in favour of the new one. That one doesn't take a context, accepts the error to return when a timeout error occurs, and there's no need for both a `time.Ticker` and `time.Timer`.

Yeah that sounds like a worthwhile refactor

aclevername (approved):
LGTM

Issue #3414, opened 2021-03-09
Required force option in eksctl manifest for addon


usman-github:
**What feature/behavior/change do you want?**

eksctl manifest file should have "force" property for addon like we have adhoc command for addon.

`eksctl update addon --name vpc-cni --version v1.7.9-eksbuild.1 --cluster <cluster-name> --force`

**Why do you want this feature?**

Recently EKS released the new version of CNI for addon. If we create the cluster using latest CNI version "1.7.9" then cluster get created successfully but addon fail to create with below error in health.

Error :

`Apply failed with 2 conflicts: conflicts with "kubectl-client-side-apply" using apps/v1: - .spec.template.spec.containers[name="aws-node"].image - .spec.template.spec.initContainers[name="aws-vpc-cni-init"].image`

Manifest :

```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: eks19
  region: eu-west-1
  version: "1.19"
  

addons:
  - name: vpc-cni
    version: v1.7.5-eksbuild.1
    attachPolicyARNs:
      - arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy


```

At the moment i am creating this feature request EKS cluster get created with [CNI version 1.7.5](https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html) by default and while creating the addon with version 1.7.9 using above manifest through the conflict error.

As we know a [conflict](https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts) is a special status error that occurs when an Apply operation tries to change a field, which another user also claims to manage. This prevents an applier from unintentionally overwriting the value set by another user.

Workaround :

As of now the workaround i am using is below.

1. creating the EKS cluster with CNI version 1.7.5 for addon.

Manifest :

```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: eks19
  region: eu-west-1
  version: "1.19"
  

addons:
  - name: vpc-cni
    version: v1.7.5-eksbuild.1
    attachPolicyARNs:
      - arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
```

2. updating the CNI version in addon to 1.7.9 using adhoc command.

`eksctl update addon --name vpc-cni --version v1.7.9-eksbuild.1 --cluster eks19 --force`

aclevername:
> **What feature/behavior/change do you want?**
> 
[...]

PR #3438, opened 2021-03-15
force addons creation during cluster creation

aclevername:
### Description

fixs https://github.com/weaveworks/eksctl/issues/3414 

The `force` argument tells the EKS addons API to override any existing resources during the addon creation, I think for any addon created during cluster creation it makes sense to always enforce this flag, as there shouldn't be any resources existing in the cluster that we don't want to have overridden as the user has not had a chance to create any of there own resources yet. In https://github.com/weaveworks/eksctl/issues/3414 they are hitting an error when creating a VPC-CNI addon against a fresh cluster (either by having the addon added during cluster creation or doing it manually afterwards). The error is `Apply failed with 2 conflicts: conflicts with "kubectl-client-side-apply" using apps/v1: - .spec.template.spec.containers[name="aws-node"].image - .spec.template.spec.initContainers[name="aws-vpc-cni-init"].image`, and the `force` argument fixes this.

This PR could be just to modify it to only run against the VPC-CNI, but I think to pre-empt any other addons that will face the same problem this makes sense.

### Checklist
- [ ] Manually tested


Callisto13:
Could you add some context here explaining what happened and why this is the fix?

pkg/actions/addon/tasks.go
```diff
@@ -58,6 +60,9 @@ func (t *createAddonTask) Do(errorCh chan error) error {
 	}
 
 	for _, a := range t.addons {
+		if t.forceAll {
+			a.Force = true
+		}
```
Callisto13, line 65: Is there an opportunity to set this to not be `true`? I guess I am wondering why we have it as a configurable var, when nothing seems to allow that value to be changed. The call in `pkg/ctl/create/cluster.go` simply sets it to `true`.

aclevername:
> Could you add some context here explaining what happened and why this is the fix?

my bad, done!

pkg/actions/addon/tasks.go
aclevername: yeah I did think about hard coding it straight in, but its not really clear that the tasks is only currently being used by the create cluster flow, it could be used elsewhere in the future (thinking apply as an option), so having it as something configured just makes it clear whats happening.

Callisto13 (approved):
👍

cPu1 (approved):

Issue #3426, opened 2021-03-11
Add support for --dry-run to `create cluster` and `create nodegroup`


cPu1:
This is outlined in this proposal here: https://github.com/weaveworks/eksctl/blob/main/docs/proposal-007-instance-selector-dry-run.md#dry-run

The gist of the proposal is that when `--dry-run` is supplied to `create cluster` or `create nodegroup`, eksctl will output a ClusterConfig containing the default values and representing the supplied CLI options.

An example invocation would look like this:

```shell
$ eksctl create cluster --dry-run --nodes=1 --nodes-max=2 --node-zones=us-west-2a,us-west-2d --alb-ingress-access --managed --spot --node-volume-size=50
```

```yaml
apiVersion: eksctl.io/v1alpha5
availabilityZones:
- us-west-2b
- us-west-2a
- us-west-2c
cloudWatch:
  clusterLogging: {}
iam:
  vpcResourceControllerPolicy: true
  withOIDC: false
kind: ClusterConfig
managedNodeGroups:
- amiFamily: AmazonLinux2
  availabilityZones:
  - us-west-2a
  - us-west-2d
  desiredCapacity: 1
  disableIMDSv1: false
  disablePodIMDS: false
  iam:
    withAddonPolicies:
      albIngress: true
      appMesh: false
      appMeshPreview: false
      autoScaler: false
      certManager: false
      cloudWatch: false
      ebs: false
      efs: false
      externalDNS: false
      fsx: false
      imageBuilder: false
      xRay: false
  instanceType: m5.large
  labels:
    alpha.eksctl.io/cluster-name: dry-run-cluster
    alpha.eksctl.io/nodegroup-name: ng
  maxSize: 2
  minSize: 1
  name: ng
  privateNetworking: false
  securityGroups:
    withLocal: null
    withShared: null
  spot: true
  ssh:
    allow: false
    enableSsm: false
    publicKeyPath: ""
  tags:
    alpha.eksctl.io/nodegroup-name: ng
    alpha.eksctl.io/nodegroup-type: managed
  volumeIOPS: 3000
  volumeSize: 50
  volumeThroughput: 125
  volumeType: gp3
metadata:
  name: dry-run-cluster
  region: us-west-2
  version: "1.18"
privateCluster:
  enabled: false
vpc:
  autoAllocateIPv6: false
  cidr: 192.168.0.0/16
  clusterEndpoints:
    privateAccess: false
    publicAccess: true
  nat:
    gateway: Single

[...]

PR #3451, opened 2021-03-18
Add support for dry-run

cPu1:
### Description

Adds dry-run support for `create cluster` and `create nodegroup`.

Closes https://github.com/weaveworks/eksctl/issues/3426

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -2,6 +2,8 @@ package create
 
 import (
 	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	"github.com/weaveworks/eksctl/pkg/kops"
@@ -67,6 +69,7 @@ func createClusterCmdWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *cmdutils.C
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
 		fs.BoolVarP(&params.InstallWindowsVPCController, "install-vpc-controllers", "", false, "Install VPC controller that's required for Windows workloads")
 		fs.BoolVarP(&params.Fargate, "fargate", "", false, "Create a Fargate profile scheduling pods in the default and kube-system namespaces onto Fargate")
+		fs.BoolVarP(&params.DryRun, "dry-run", "", false, "Dry-run mode that skips cluster creation and outputs a ClusterConfig")
 	})
 
 	cmd.FlagSetGroup.InFlagSet("Initial nodegroup", func(fs *pflag.FlagSet) {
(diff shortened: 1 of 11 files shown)


aclevername (changes requested):
Looks good! Two things:
- Just out of curiosity have you tried using the config you get as an output as input to eksctl? It sounds silly but with the way we mutate the config I wouldn't be surprised if some bug exists/gets introduced meaning that it doesn't work. Could potentially be worth an integration test?
- I can see the integration test failing a lot as whenever we add a new field to the cluster/nodegroup struct you need to remember to update this test, which won't get noticed in a github PR as we don't run the integration tests. Could we move this test/copy a portion of this test to live somewhere that ensures it gets run as part of `make test`/PR flow?

aclevername:
> It is! I was planning to address it in a separate PR, as part of adding an integration test for `create nodegroup`.

sweet!

> We don't run the integration tests for PRs, but we do build the integration tests, so any compilation errors will cause the PR checks to fail.

Its not compilation errors I'm worried about, its PRs like [this](https://github.com/weaveworks/eksctl/pull/2754) that are small so you wouldn't think to run the integration test, but would cause this test to fail as it introduces a new field. I think catching this before its merged would be really useful

cPu1:
The dry-run integration test is failing because it needs AWS credentials. I'm working on it.

aclevername:
> The dry-run integration test is failing because it needs AWS credentials. I'm working on it.

out of curiosity what sort of API calls do we make during dry-run that need creds? VPC validation?

cPu1:
>out of curiosity what sort of API calls do we make during dry-run that need creds? VPC validation?

One of the first checks performed by every command is `sts.GetCallerIdentity`, so it's failing on that.

cPu1:
Regarding running the dry-run integration test upon pushing to `main`, IMO, it's not worth the effort of introducing a new GH action that runs just the dry-run integration test to ensure any changes to the ClusterConfig schema are caught. I think we just have to accept that it's the same as other integration tests, and if the integration test suite didn't take too long to complete, we could've run it upon pushing to `main`. 

In the future, we can consider writing a custom check that reads the `defaultClusterConfig` variable and compares it with the output of the default ClusterConfig as part of the PR workflow.

pkg/ctl/create/cluster.go
```diff
@@ -107,6 +110,15 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
...
 		return err
 	}
+
+	if params.DryRun {
+		originalWriter := logger.Writer
+		logger.Writer = io.Discard
+		defer func() {
+			logger.Writer = originalWriter
+		}()
+	}
```
aclevername, line 120: does this stop `logger.Error` calls from output?

aclevername (approved):
LGTM, just need to update the integration test to create a cluster using the output of the dry run

PR #3456, opened 2021-03-19
Add release notes for 0.42

Callisto13:
These are hand-rolled so please make sure I got things under the right headings.

### DIFF
--- a/docs/release_notes/0.42.0.md
+++ b/docs/release_notes/0.42.0.md
@@ -1,0 +1,14 @@
+# Release 0.42.0
+
+## Features
+Add default security group rules to custom SGs (#3027)
+
+## Improvements
+Disallow setting internal fields (#3450)
+
+## Bug Fixes
+Force addons during cluster creation (#3438)
+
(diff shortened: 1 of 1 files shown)


docs/release_notes/0.42.0.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Features
+Add default security group rules to custom SGs (#3027)
+
+## Improvements
+Disallow setting internal fields (#3450)
+
+## Bug Fixes
+Force addons during cluster creation (#3438)
+Fix: nil pointer deref when vpc is empty (#3447)
```
cPu1, line 11: This bug didn't make it to any release because it was introduced by #3027. It may be confusing to users if we include it. We should still give attribution to the contributor (@liranp) though.

Callisto13: ah yes good point. i will remove it and leave his name in

aclevername (approved):

cPu1 (approved):

PR #3470, opened 2021-03-23
Add support for reconciling IAMServiceAccounts using  eksctl apply 

aclevername:
#wip

### Checklist
- [ ] Manually tested


aclevername:
I tried quite hard to refactor out any duplicate CF calls, so the stack gets passed around a lot.

pkg/actions/apply/apply.go
```diff
@@ -0,0 +1,63 @@
+package apply
+
+import (
+	"fmt"
+
+	"github.com/kris-nova/logger"
```
Skarlso, line 6: Any regard for ordering imports? 

I like the structure:
- stdlib
- third party
- under orgranization
- in the same repo

What do you think?

aclevername: great question! This is not something I've really ever "cared" too much for, I'd be happy for us to start taking an opinion. I wish the built in linter had the opinion for us :disappointed:

Skarlso: Yeah no linter check other then goimport -local I think. :/ But the IDE let's you set up import grouping format which is nice.

pkg/actions/apply/apply.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+	return nil
+}
+
+func logErrors(errs []error) {
+	for _, err := range errs {
+		logger.Info("%v", err)
+	}
+}
```
Skarlso, line 63: This could be something like this to simplify and minimise allocation:

```go
// I like formatting code the way that all path can be readable
// on the in one line. Meaning, if there is a decision, that should
// contain as less code as possible to not distract from the main flow.
if r.plan {
    return nil
}
errs := make(error, 0)
errs = append(errs, createTasks.DoAllSync()...)
errs = append(errs, updateTasks.DoAllSync()...)
errs = append(errs, deleteTasks.DoAllSync()...)

if len(errs) != 0 {
     // only need the for call if we know that there were errors anyways.
     for _, err := range errs {
         logger.Info("%v", err)
    }
    return fmt.Errorf("failed to reconcile cluster")
}
return nil
```

What do you think?

aclevername: sounds good!

Issue #2174, opened 2020-05-12
Use eksctl commands (e.g. update core-dns) on non-eksctl created clusters


kalbir:
**Why do you want this feature?**
`eksctl` has some commands that make it easy to do things on clusters. So regardless of whether you created the cluster with `eksctl` or some other method, it could save yout time if you could use one of those commands on your cluster.

Examples that come to mind are:
- `eksctl utils update-coredns`
- `eksctl create iamidentitymapping`
- `eksctl create fargateprofile` 
- `eksctl upgrade nodegroup` (https://github.com/weaveworks/eksctl/issues/2698)

**What feature/behavior/change do you want?**

My initial thought is that you could pass some identifier for the cluster to eksctl (e.g. name and region) that will identify a unique cluster. `eksctl` could then look up that cluster and apply the correct command to it. So on the surface of things, eksctl would not change from a user perspective. 

This would mean that `eksctl get clusters` could also show _all_ the clusters that you have in your account, and indicate whether they were created by `eksctl` or not.

Then if you tried a command on a non-eksctl cluster that we could not complete, you'd get a message telling you why you couldn't do that (e.g. `delete cluster`).

aclevername:
Thanks for opening the issue @kalbir. You raise some excellent points around the current user experience when using eksctl against clusters that weren't created with eksctl. I've opened two recent PRs to try and improve the current experience https://github.com/weaveworks/eksctl/pull/2779 and https://github.com/weaveworks/eksctl/pull/2775. However these PRs only adress `eksctl get clusters` and `eksctl create iamserviceaccount`. As you point out there are many commands where eksctl doesn't work. 

Your suggestion around changing it so that the UX seems reasonable to me. To recap the outcome of running a command against a non-eksctl created cluster would be either:
1. It works as expected
2. Where it doesn't work, it provides a good UX why, e.g. 
```bash
$ eksctl delete cluster --name cluster-created-in-console
[ℹ]  eksctl version 0.30.0
[ℹ]  Cannot delete cluster that was not created via eksctl (no associated CloudFormation stacks found)
```

It might be that over time we try to remove more occurrences of 2. and make eksctl smarter. For example at the moment it all of its interaction with creating clusters is through CloudFormation, meaning commands like `eksctl delete cluster` don't work as if you created the cluster manually it has no CloudFormation stack to delete. This could be overcome by making eksctl more aware of alternative methods of interacting with clusters.

aclevername:
[...]

PR #3471, opened 2021-03-23
Add docs for non eksctl-created clusters

Callisto13:
I know we are waiting for blogs and the like, but we should at least have _some_ docs, especially because people are asking about this.


aclevername (approved):
LGTM :tada:, feel free to close https://github.com/weaveworks/eksctl/issues/2174 when this is merged :smile:

Issue #2332, opened 2020-06-15
Allow modifying the minSize and maxSize value of a NodeGroup without changing desiredCapacity


TBBle:
**Why do you want this feature?**

I use the Cluster Autoscaler to manage the number of nodes in my nodegroups. However, I want to be able to change the limits (min and max) _without_ changing the number of currently active nodes.

This appears to be what was requested in #809 and #1893 as well. They were closed by #2022 which added `--nodes-max` and `--nodes-min`, but left `--nodes` as mandatory.

**What feature/behavior/change do you want?**

```
eksctl --cluster mycluster scale nodegroup --name ng-a --nodes-max 100
```
should change the maximum size to 100, but not affect current or minimum sizes.

As noted in #809's comments, this is equivalent to
```
aws autoscaling update-auto-scaling-group --auto-scaling-group-name $ASG_NAME --max-size=20
```
for `$ASG_NAME` of `eksctl-<mycluster>-nodegroup-<ng-a>-NodeGroup-<HASH>`.

TBBle:
I'm unaware of this being resolved at this time.

TBBle:
As a status update, #3511 has introduced the ability to change the minSize and maxSize of a NodeGroup without changing desiredCapacity, but has the side-effect that the ASG's size is reset to the desiredCapacity of the NodeGroup (by updating CloudFormation).

The Cluster Autoscaler interaction we need for the use case of this ticket is that somehow we need the minSize/maxSize to come from eksctl -> CloudFormation -> ASG -> Cluster Autoscaler, but desiredSize needs to come from Cluster AutoScaler -> ASG -> (?) -> Cloud Formation.

This is perhaps more complex to fix than I envisaged when I opened the ticket, because we can see in #3532 that updating the NodeGroup CF stack _at all_ will reset the ASG's desired size to the eksctl/CF desired size, losing the Cluster Autoscaler's size value.

So to make the UX work as I envisaged, eksctl would need to get the ASG's current desired size, and use _that_ as the "desired capacity" when updating the Nodegroup CF Stack, effectively applying Cluster Autoscaler's changes back from the ASG to the Nodegroup CF Stack.

#2331 is related to this approach.

----

It'd be nicer if AWS had a way to say "Cloud Formation stack doesn't manage 'desiredCapacity' of the ASG it templates.  That would also fix the fact that Cluster Autoscaler will trigger the Cloud Formation "drift check" because it updates the ASG.

[...]

PR #3515, opened 2021-04-03
Get nodegroup desired capacity from autoscaling instead of cloudformation

### LINKED ISSUE #2331: eksctl get nodegroups reports ASG Desired Size, not CF Desired Size

As a shorthand way of checking cluster scaling state, `eksctl get nodegroups --cluster X` is tempting.

However, it doesn't show the _current_ state, as updated by (for example) the Cluster Autoscaler.

This caught me out when I was trying to change the max size of a nodegroup with `eksctl scale nodegroup`, and you must provide the `--nodes` parameter to that, I used the value from `eksctl get nodegroups`, not realising this would scale the nodegroup back down to 1, until the Cluster Autoscaler was able to scale it back up.

This particular use-case was working around a different issue: #2332, but at the time I assumed CA had done its job between my two commands. It's only today, that when using `eksctl get nodegroups` and comparing it to `kubectl get nodes` that I realised the former never showed the effects of CA.

`eksctl get nodegroups --cluster X` should show the current capacity (the ASG Desired Size as updated by Cluster Autoscaler or other process), instead of the desired capacity (the CloudFormation value, set at original build-time to some arbitrary number, knowing that CA would fix it).

Perhaps you could argue that the fault lies in Cluster Autoscaler, as it changes the ASG and not the CloudFormation that created it, but I feel like that would be an uphill argument.

omnibrian:
Closes #2331 

### Description

Update nodegroup summaries to use desired capacity set at the autoscaling group level instead of in the cloudformation stack. This allows for getting more accurate desired capacity when using cluster-autoscaler since the autoscaling group desired capacity gets changed directly and isn't reflected in cloudformation.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/manager/api.go
+++ b/pkg/cfn/manager/api.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
 	"github.com/aws/aws-sdk-go/service/cloudtrail/cloudtrailiface"
 	"github.com/weaveworks/eksctl/pkg/cfn/waiter"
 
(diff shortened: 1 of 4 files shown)


aclevername:
Thanks for opening the PR @omnibrian. This sounds reasonable, it highlights how the relationship between the `desired` nodegroups in the cloudformation template differ from the actual state. This came up again recently here: https://github.com/weaveworks/eksctl/issues/2332#issuecomment-814350918

aclevername (approved):
lgtm

Issue #3563, opened 2021-04-12
(Future) Breaking: overrideBootstrapCommand soon to be required with Custom AMIs for AL2 and Ubuntu unmanaged nodegroups


Callisto13:
# Note 1: Future change is described below, no action required from users yet
# Note 2: This is an announcement thread, not a "problem with `overrideBootstrapCommand`" thread. Please do NOT spam this thread with problems you are having. Please open a new issue with your specific failure.

## Do I need to read this issue/Will this change affect me?

![Inception Template (1)](https://user-images.githubusercontent.com/8898786/127893505-d16658d3-54f7-4aa7-93b8-6d83c1029172.jpg)

## TLDR:
From `eksctl` version `0.47.0` onwards all AL2 and Ubuntu **unmanaged** nodes will be bootstrapped using the AMI provided bootstrapping script.

For most people, this change should not be noticeable: you don't need to take any action. For those using custom AL2/Ubuntu AMIs, you will soon (BUT NOT YET) need to provide some extra configuration. To give users more time to move over to this new requirement, the legacy codepath will still be present and followed when providing a custom AMI. When enough time has passed (release date TBD), this legacy path will be removed. When that happens, it means that when using a Custom AMI, it will be required to set the `overrideBootstrapCommand` in order for the nodes to successfully join the cluster.

In all other cases, this change should not have any effect.

## What's happening?

We  recently merged [some work](https://github.com/weaveworks/eksctl/pull/3564) to remove the custom bootstrapping scripts which `eksctl` drops on **AmazonLinux2** and **Ubuntu** machines so that they can join the k8s cluster. From version `0.47.0`, all `eksctl` nodes will be bootstrapped using the `bootstrap.sh` script provided by standard (non-custom) EKS AMIs (mostly, there are still some odd "wrapper" scripts there to tidy things up).

A side effect of this improvement is that (in the near future) an `overrideBootstrapCommand` will be required when creating **un**managed nodegroups with **custom AL2/Ubuntu amis**. In previous `eksctl` versions this was not necessary, since `eksctl` would simply send everything it needed to the node, bootstrapping nodes with custom and non-custom AMIs alike. Now that `eksctl` delegates to the built-in `/etc/eks/bootstrap.sh` on standard public images, any node which does not have that file on disk will fail to join the cluster.

[...]

PR #3492, opened 2021-03-30
Use built-in AMI bootstrapping scripts for unmanaged nodes

Callisto13:
### Description

Closes https://github.com/weaveworks/eksctl/issues/3464.

This PR removes all custom nodebootstrapping for Ubuntu and AL2 unmanaged nodes. We will now use the built-in `/etc/eks/bootstrap.sh` for all our bootstrapping needs. The new implementation _should_ have complete parity with the old one; users should not notice that anything has changed.

Also: 
- I have tidied-up the Windows and Bottlerocket bootstrap code and tests so that they fit the new interface. Windows was already using the built-in bootstrapper. Bottlerocket is doing something else which I don't want to get into.
- Speaking of interfaces, the NodeGroupResourceSet now contains a "bootstrapper". This decouples the cfn builder from nodebootstrapper, and means cfn no longer unit-tests userdata generation. (That test file is still a disaster, it is on my list)
- I have not touched Managed nodegroup bootstrapping as that is using the built-in bootstrapper, and this diff was already large enough.
- Deleted so much stuff!

See https://github.com/weaveworks/eksctl/issues/3464#issuecomment-810135117 for more info on decisions.

### Checklist
- [x] Added tests that cover your change (if possible)
- [x] Added/modified documentation as required (such as the `README.md`, or the `userdocs` directory)
- [x] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes
- [x] (Core team) Added labels for change area (e.g. `area/nodegroup`) and kind (e.g. `kind/improvement`)

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [x] Backfilled missing tests for code in same general area :tada:
- [x] Refactored something and made the world a better place :star2:

PR #3564, opened 2021-04-12
Use built-in AMI bootstrapping scripts for AL2/Ubuntu unmanaged nodes

Callisto13:
### Description

See original PR here: https://github.com/weaveworks/eksctl/pull/3492

This PR introduces the same change as in the above, with the re-addition of the previously removed "legacy" bootstrappers.

As discussed, while we want to move over to use the built-in bootstrapping scripts, we want to ensure that users have time to prepare for the new requirements for nodes created with custom AL2 and Ubuntu AMIs. (See notice here https://github.com/weaveworks/eksctl/issues/3563)

The diff between this PR and the previous is the addition of `pkg/nodebootstrap/legacy` which is where I have unceremoniously dumped our previous bootstrapping logic for AL2 and Ubuntu nodes. I have **not** done anything more than ensuring the new/legacy switch works, I have **not** tidied anything up in that dir, and I wont. Do not `@` me 🙃 . In order to determine whether the AMI is custom, I am checking for the presence of `ng.AMI` _before_ the AMI resolver sets it (because of this we cannot rely on that field any later) and setting a hacky internal Boolean field which we can then use to select the correct bootstrapper later. The field + checks will obviously disappear when the legacy code does.

### Checklist
- [x] Manually tested


aclevername (approved):
changes to `legacy` LGTM

Issue #3569, opened 2021-04-13
Skip add-on compatibility checks during creation of ARM nodegroups


ggml1:
**Why do you want this feature?**

When trying to create nodegroups with ARM instances, we faced an error message which did not allow us to continue.

> 2021-04-13 12:32:02 [✖]  to create an ARM nodegroup kube-proxy, coredns and aws-node addons should be up to date. Please use `eksctl utils update-coredns`, `eksctl utils update-kube-proxy` and `eksctl utils update-aws-node` before proceeding.

But our addons were already updated and would support ARM instances with no problems (we were already running them with ARM nodes in other cluster). Still, we had to run the update utils in order to proceed. Running them is not ideal in our case because it would override specific configurations we had on the deployment resources of some addons, which would cause issues in our cluster. 

**What feature/behavior/change do you want?**

Due to this, it would be cool if there was a way of skipping this add-on compatibility check when creating nodegroups, and therefore leaving this responsibility to the user. I think this could be done by passing a flag to the `create nodegroup` command.

aclevername:
Thanks for opening the issue @ggml1, sounds reasonable to me.

PR #3570, opened 2021-04-13
Add flag that allows skipping add-on compatibility checks during creation of ARM nodegroups

ggml1:
### Description

Implements a flag for #3569.

### Checklist
- [x] Manually tested
- [i think so] Made sure the title of the PR is a good description that can go into the release notes

### DIFF
--- a/pkg/actions/nodegroup/create.go
+++ b/pkg/actions/nodegroup/create.go
@@ -30,6 +30,7 @@ type CreateOpts struct {
 	InstallNeuronDevicePlugin bool
 	InstallNvidiaDevicePlugin bool
 	DryRun                    bool
+	SkipOutdatedAddonsCheck   bool
 	ConfigFileProvided        bool
 }
 
@@ -45,7 +46,7 @@ func (m *Manager) Create(options CreateOpts, nodegroupFilter filter.NodeGroupFil
 		return err
 	}
 
-	if err := checkARMSupport(ctl, m.clientSet, cfg); err != nil {
+	if err := checkARMSupport(ctl, m.clientSet, cfg, options.SkipOutdatedAddonsCheck); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


aclevername:
Thanks for opening the PR @ggml1, sounds reasonable to me.

pkg/actions/nodegroup/create.go
```diff
@@ -30,6 +30,7 @@ type CreateOpts struct {
 	InstallNeuronDevicePlugin bool
 	InstallNvidiaDevicePlugin bool
 	DryRun                    bool
+	CheckOutdatedAddons       bool
```
aclevername, line 33: since we want the default behaviour to be checking the addons, shall we change this to `SkipOutdatedAddonsCheck` or something similar, so we can rely on the defaulting to `false` to preserve the existing behaviour :smile:

pkg/actions/nodegroup/create.go
```diff
@@ -289,7 +290,7 @@ func checkARMSupport(ctl *eks.ClusterProvider, clientSet kubernetes.Interface, c
 		if err != nil {
 			return err
 		}
-		if !upToDate {
+		if checkOutdatedAddons && !upToDate {
 			logger.Critical("to create an ARM nodegroup kube-proxy, coredns and aws-node addons should be up to date. " +
 				"Please use `eksctl utils update-coredns`, `eksctl utils update-kube-proxy` and `eksctl utils update-aws-node` before proceeding.")
 			return errors.New("expected default addons up to date")
```
aclevername, line 296: worth keeping some sort of log, just not erroring when it occurs. WDYT?

ggml1: Hmmmm.. what information do you think we should log and in which case? Do you mean leaving the current log in both cases, just changing the return behavior?

aclevername: yeah leave the current log, change the return behaviour. Although thinking over it again if your providing this flag you probably already know and don't care... so perhaps its not worth it

Its probably worth updating the log line to explain that you can get around this check by providing the skip flag though :smile:

ggml1: Good idea! I have updated the log =D

aclevername (approved):
LGTM

PR #3583, opened 2021-04-15
Allow adding Docker registry mirrors to nodes via config file option dockerRegistryMirrors

jcttrll:
### Description

For organizations who run their own mirror of Docker Hub (for security, for
efficiency, or to avoid Docker Hub throttling), it is necessary to customize
/etc/docker/daemon.json to supply an array of "registry-mirrors".

This change allows passing Docker registry mirrors to eksctl via a new
dockerRegistryMirrors config file option.

UserData is affected as follows:
```
- content: |-
    {
      "bridge": "none",
      "exec-opts": [
        "native.cgroupdriver=systemd"
      ],
      "live-restore": true,
      "log-driver": "json-file",
      "log-opts": {
        "max-file": "10",
        "max-size": "10m"
      },
      "max-concurrent-downloads": 10,
      "registry-mirrors": [
        "https://docker.example.com"
      ]
    }
```

### Checklist
- [X] Manually tested


Callisto13:
Thanks for opening this PR @jcttrll 👍 

A couple of things to note:
- Most of the way we currently do bootstrapping is going to change (https://github.com/weaveworks/eksctl/pull/3564), which will affect this addition. In that PR we are removing as much responsibility for bootstrapping from `eksctl` as possible.
- This is the first bit of docker config we are exposing, so we should think about the pattern this starts. I am not sure whether we want to get into the habit of exposing 1:1 mappings of docker config in our api as it can get messy. For setting kubelet config we have a `KubeletExtraConfig` object which allows users to provide a map of options which we can just write in without caring about. What do folks think about potentially having a similar thing here?

Callisto13:
Hey @jcttrll! Just checking in on the status of this, do you plan to continue development or would you like it taken over? Thanks!

Skarlso:
Hello.

I'm closing this due to inactivity, and many things changed in bootstrapping land after we deprecated things. :)

Please feel free to reopen if you would like to continue working on this! 🤗 

Thanks!

PR #3637, opened 2021-04-27
Reset cgroup drivers back to cgroupfs

Callisto13:
**Description**

Since we have stopped providing custom bootstrapping scripts and are now
relying on what comes with the AMI, finding the right places to change
the cgroup driver setting for both kubelet and docker is trickier. GPU nodes now fail
to come up. The GPU AMIs are going through some changes, so those
locations will be even harder to pin down for a while.
I am undoing this for now, and plan to look into it later when I have
more time.

**More background**
Integration tests were failing to create a GPU node.

Creating a GPU node is failing for me locally, honestly I don't know how people have not been complaining about this?
I initially thought that the nvidia plugin was unhappy, but none of the pods which should be there are coming up. Coredns is failing because there aren't any nodes available for scheduling. Which is correct, since the nodes are not coming up. So why are they not?

...tbc

Config I am using:
```yaml
---
kind: ClusterConfig

metadata:
  name: cb-test-gpu
  region: us-west-2

nodeGroups:
  - name: ng-2
    desiredCapacity: 1
    instanceType: p2.xlarge
    # privateNetworking: true <- integration tests have this, but this is not the problem so i am not setting it
```

found it:
```
$ journalctl -u kubelet.service
...
failed to run Kubelet: misconfiguration: kubelet cgroup driver: "systemd" is different from docker cgroup driver: "cgroupfs"
...
```

I TESTED THIS FFS

Okay I have [undone](https://github.com/weaveworks/eksctl/pull/3637/commits/ac3e862d0285ee0340ade13109843850d9e7b8ff) some work from a while back and we seem to be fine now. I will re-run again and merge if okay.


aclevername (approved):

aclevername:
@Callisto13 whats the latest on this? any luck?

Callisto13:
fixing now

Callisto13:
fixed i think: https://github.com/weaveworks/eksctl/pull/3637/commits/ac3e862d0285ee0340ade13109843850d9e7b8ff

aclevername:
Awesome!!!

aclevername:
@cpu1 green run last night https://github.com/weaveworks/eksctl/actions/runs/831342180

integration/tests/crud/creategetdelete_test.go
```diff
@@ -200,9 +200,10 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 		})
 
 		Context("and add a second (GPU) nodegroup", func() {
-			PIt("should not return an error", func() {
+			It("should not return an error", func() {
 				cmd := params.EksctlCreateCmd.WithArgs(
 					"nodegroup",
+					"--timeout=45m0s",
```
cPu1, line 206:
```suggestion
					"--timeout=45m",
```

cPu1 (approved):

PR #3646, opened 2021-04-29
Set the version of an addon with a numeric version that has no prefix or suffix

### LINKED ISSUE #3574: eksctl addons supports pattern matching the version
Currently when creating an addon you have to specify a version like `v1.7.5-eksbuild.1`. As a user of addons I would like just to be able to specify `v1.7.5` or `1.7.5` and have eksctl figure out the suffix & `v` prefix
To make it easier to manage my addons

nikimanoledaki:
### Description

Closes #3574.

The current behaviour for setting the version of an addon when creating or updating it is to specify the full version, such as `v1.7.5-eksbuild.2` or `v1.7.5-eksbuild.1`, or set to `latest`, which will list and select the latest version available for this addon.

This PR lets the user specify a shortened form of the version such as `v1.7.5` or `1.7.5` that does not have an `eksbuild-1` or `eksbuild-2` suffix, whether it does or does not have a `v` prefix. It then finds the latest build for this numeric version using the same logic as when `latest` lists and selects the latest version for an addon.

Demo
1) Create an addon with version `1.7.5`
```
$ ./eksctl create addon --name vpc-cni --version 1.7.5 --cluster "ridiculous-party-1619777406"

2021-04-30 11:31:07 [!]  no IAM OIDC provider associated with cluster, try 'eksctl utils associate-iam-oidc-provider --region=us-west-2 --cluster=ridiculous-party-1619777406'
2021-04-30 11:31:07 [ℹ]  Kubernetes version "1.19" in use by cluster "ridiculous-party-1619777406"
2021-04-30 11:31:07 [ℹ]  when creating an addon to replace an existing application, e.g. CoreDNS, kube-proxy & VPC-CNI the --force flag will ensure the currently deployed configuration is replaced
2021-04-30 11:31:07 [!]  OIDC is disabled but policies are required/specified for this addon. Users are responsible for attaching the policies to all nodegroup roles
2021-04-30 11:31:08 [ℹ]  creating addon
2021-04-30 11:31:09 [ℹ]  successfully created addon
```

2) Get addon to check that it was set to latest build
```
./eksctl get addon --name vpc-cni --cluster "ridiculous-party-1619777406"

2021-04-30 11:32:56 [ℹ]  eksctl version 0.48.0-dev+625febb8.2021-04-30T10:49:48Z
2021-04-30 11:32:56 [ℹ]  using region us-west-2
2021-04-30 11:32:57 [ℹ]  Kubernetes version "1.19" in use by cluster "ridiculous-party-1619777406"
2021-04-30 11:32:58 [ℹ]  to see issues for an addon run `eksctl get addon --name <addon-name> --cluster <cluster-name>`
NAME    VERSION                 STATUS  ISSUES  IAMROLE UPDATE AVAILABLE
vpc-cni v1.7.5-eksbuild.2       ACTIVE  0               v1.7.10-eksbuild.1,v1.7.9-eksbuild.2,v1.7.9-eksbuild.1,v1.7.6-eksbuild.2,v1.7.6-eksbuild.1
```

3) Update addon to `1.7.7`
```
$ ./eksctl update addon --name vpc-cni --version 1.7.9 --cluster "ridiculous-party-1619777406"

2021-04-30 12:09:31 [!]  no IAM OIDC provider associated with cluster, try 'eksctl utils associate-iam-oidc-provider --region=us-west-2 --cluster=ridiculous-party-1619777406'
2021-04-30 12:09:31 [ℹ]  Kubernetes version "1.19" in use by cluster "ridiculous-party-1619777406"
2021-04-30 12:09:32 [ℹ]  new version provided v1.7.9-eksbuild.2
2021-04-30 12:09:32 [ℹ]  updating addon
```

4) Case where version was not found
```
$ ./eksctl update addon --name vpc-cni --version 1.7.8 --cluster "ridiculous-party-1619777406"

2021-04-30 12:08:39 [!]  no IAM OIDC provider associated with cluster, try 'eksctl utils associate-iam-oidc-provider --region=us-west-2 --cluster=ridiculous-party-1619777406'
2021-04-30 12:08:39 [ℹ]  Kubernetes version "1.19" in use by cluster "ridiculous-party-1619777406"
Error: failed to fetch addon version: version 1.7.8 not available for "vpc-cni"
```

Also: 
- Updated user docs on the available options when setting addon versions.
- Added a test for the edge case where the set version was not found when creating or updating the addon.
- Updated the unit tests for creating and updating an addon to use the full version of the addon, unless of course the purpose of the test is to check the logic of setting a short form of the version. This is to minimize what the other tests are testing (as in, the default API functionality rather than `eksctl` logic).
- Filled in some missing unit tests for updating an addon in `pkg/actions/addon/update_test.go`, such as for when the version is not set and `eksctl` preserves the version (but may suggest a more updated version). 
- Refactored the `update` unit tests to remove repeating code.

### Checklist
- [x] Manually tested


pkg/actions/addon/update.go
```diff
@@ -41,7 +42,7 @@ func (a *Manager) Update(addon *api.Addon, wait bool) error {
 		updateAddonInput.AddonVersion = &summary.Version
 	} else {
 		version := addon.Version
-		if version == "latest" {
+		if version == "latest" || !strings.Contains(version, "eksbuild") {
```
Callisto13, line 45: this will match on when `version == ""`, is that a problem?

userdocs/src/usage/addons.md
```diff
@@ -74,6 +74,14 @@ You can see what addons are enabled in your cluster by running:
 eksctl get addons --cluster <cluster-name>
 ```
 
+## Setting the addon's version
+
+Setting the version of the addon is optional. If it is not set, `eks` will set the default version for this addon. Note that this might not necessarily be the latest version available. 
+
+The addon version can be set to `latest`. Alternatively, the version can be set with the EKS built specified, such as `v1.7.5-eksbuild.1` or `v1.7.5-eksbuild.2`. It can also be set to the release version of the addon, such as `v1.7.5` or `1.7.5`, and the `eksbuild` will be discovered and set for you.
```
Callisto13, line 81:
```suggestion
The addon version can be set to `latest`. Alternatively, the version can be set with the EKS build tag specified, such as `v1.7.5-eksbuild.1` or `v1.7.5-eksbuild.2`. It can also be set to the release version of the addon, such as `v1.7.5` or `1.7.5`, and the `eksbuild` suffix tag will be discovered and set for you.
```

userdocs/src/usage/addons.md
```diff
@@ -74,6 +74,14 @@ You can see what addons are enabled in your cluster by running:
 eksctl get addons --cluster <cluster-name>
 ```
 
+## Setting the addon's version
+
+Setting the version of the addon is optional. If it is not set, `eks` will set the default version for this addon. Note that this might not necessarily be the latest version available.
```
Callisto13, line 79: > `eks` will set the default version

eks or eksctl?

pkg/actions/addon/update.go
Callisto13: actually nvm if i expand the lines above i can see we do something else when it is `""`

Callisto13 (commented):
very tidy! just a couple of questions right now

userdocs/src/usage/addons.md
nikimanoledaki: eks! I'll rephrase this so that it's clearer :)

Callisto13: 👍  cool, was not sure. have not looked at the addon stuff before :)

Callisto13 (commented):
LGTM, but leaving to @aclevername to approve since he knows the area better 🎉

pkg/actions/addon/addon.go
```diff
@@ -94,13 +95,26 @@ func (a *Manager) getLatestVersion(addon *api.Addon) (string, error) {
...
+		switch addonVersion {
+		case "latest":
+			versions = append(versions, v)
+		default:
+			if strings.Contains(*addonVersionInfo.AddonVersion, addonVersion) {
+				versions = append(versions, v)
+			}
+		}
+	}
+
```
aclevername, line 115: does this mean we allow any version value that is a substring of an actual version value? e.g `eks` would be a valid version value now, or just `1.` ?

nikimanoledaki: `eks` should not be a valid version value and I just realised that this edge case is not handled (I thought it was), thank you! if the given version is not `latest`, the input version should be parsed with something like `version.NewVersion(version)` to make sure that it is a valid version. 

just `1` would find the latest version that starts with that, yes!

aclevername: > `eks` should not be a valid version value and I just realised that this edge case is not handled (I thought it was), thank you! if the given version is not `latest`, the input version should be parsed with something like `version.NewVersion(version)` to make sure that it is a valid version.

:+1: 

> just `1` would be a valid version value, the latest version that starts with `1` would be returned! is that ok?

mhmm good question, I think the UX I've seen before is being able to set `latest` or atleast `major.minor`, so `1.2` would be valid, but `1` wouldn't be. However I don't think its really worth having too much of a fuss about it. If the `NewVersion` can parse the inputs, thats probably good enough. I would of *guessed* that `1` isn't a valid input to it, but `1.0` is but I could be wrong :smile:

nikimanoledaki: just added a small thing so that anything that can't be parsed returns an error `malformed version: <version>`! the versioning library is so handy! 😄 

just checked what happens when the value input is `1` and it doesn't return any error, it just discovers the latest `1.x` version. I don't mind the UX - I guess it would be unusual to ask for a major version `1` or `2`, it wouldn't necessarily be wrong to do that, right?

pkg/actions/addon/create_test.go
```diff
@@ -201,6 +201,18 @@ var _ = Describe("Create", func() {
...
+			When("the version is set to an invalid value", func() {
+				It("returns an error", func() {
+					err := manager.Create(&api.Addon{
+						Name:             "my-addon",
+						Version:          "eks",
+						AttachPolicyARNs: []string{"arn-1"},
+					}, false)
+					Expect(err).To(HaveOccurred())
+					Expect(err).To(MatchError(ContainSubstring("Malformed version: eks")))
+				})
```
aclevername, line 214: nice 😄

pkg/actions/addon/addon.go
```diff
@@ -86,6 +87,14 @@ func (a *Manager) waitForAddonToBeActive(addon *api.Addon) error {
 }
 
 func (a *Manager) getLatestVersion(addon *api.Addon) (string, error) {
```
aclevername, line 89: do we wanna rename this now? Its sort of "get the latest version matching my string" 😂 , maybe "getLatestMatchingVersion" or something to the effect, wdyt?

pkg/actions/addon/addon.go
```diff
@@ -96,11 +105,23 @@ func (a *Manager) getLatestVersion(addon *api.Addon) (string, error) {
...
+		}
+
+		switch addonVersion {
+		case "latest":
+			versions = append(versions, v)
+		default:
+			if strings.Contains(*addonVersionInfo.AddonVersion, addonVersion) {
+				versions = append(versions, v)
+			}
 		}
```
aclevername, line 120:
```suggestion
		if addonVersion == "latest" || strings.Contains(*addonVersionInfo.AddonVersion, addonVersion) {
		  versions = append(versions, v)
		}
```

pkg/actions/addon/create.go
```diff
@@ -27,11 +27,11 @@ const (
 
 func (a *Manager) Create(addon *api.Addon, wait bool) error {
 	version := addon.Version
-	if addon.Version == "latest" {
+	if version == "latest" || !strings.Contains(version, "eksbuild") && version != "" {
```
aclevername, line 30: what if they just specify `1.7.5.eksbuild` and want to let us figure out if its eksbuild1 or eksbuild2?

pkg/actions/addon/create_test.go
```diff
@@ -204,7 +250,7 @@ var _ = Describe("Create", func() {
 					Version:          "latest",
 					AttachPolicyARNs: []string{"arn-1"},
 				}, false)
-				Expect(err).To(MatchError(ContainSubstring("failed to parse version \"tottaly not semver\":")))
+				Expect(err).To(MatchError(ContainSubstring("failed to parse version \"totally not semver\":")))
```
aclevername, line 253: 😱

pkg/actions/addon/delete.go
```diff
@@ -24,7 +22,7 @@ func (a *Manager) Delete(addon *api.Addon) error {
 	})
 
 	if err != nil {
-		if awsError, ok := err.(awserr.Error); ok && awsError.Code() == awseks.ErrCodeResourceNotFoundException {
```
aclevername, line 27: nit: we tend to use `awseks` as a way of avoiding the fact `eks` clashes with the local package we have `pkg/eks`, here they don't use both so its not a big deal, but thats why you will see `awseks` all over the codebase 😄

pkg/actions/addon/create_test.go
nikimanoledaki: thanks! it's from the versioning package! :D

pkg/actions/addon/delete.go
nikimanoledaki: good to know! thanks 😊

pkg/actions/addon/create.go
nikimanoledaki: yep, good question.. I can change the highlighted line^ to check if the string contains `eksbuild-1` or `eksbuild-2`! I was hoping to tackle both of these by checking for `eksbuild` but you're right, and it's an easy fix!

pkg/actions/addon/addon.go
aclevername: yeah I think its OK, but if people actually do this I will judge them :joy:

pkg/actions/addon/create.go
nikimanoledaki: or we can just exclude `ekdsbuild-`, assuming there is something after the `-`. does that sound like a reasonable catch-all?

aclevername: Mhmmm I can see the value in not specifying the eksbuild number though, it's a reasonable use case

nikimanoledaki: Agreed, in that case I will just remove the bit of code that parses the version and throws the `malformed version` error. It will just return an error if no matching version was found! :)

aclevername: sounds good!

pkg/actions/addon/addon.go
```diff
@@ -94,13 +95,21 @@ func (a *Manager) getLatestVersion(addon *api.Addon) (string, error) {
...
-		versions = append(versions, v)
+
+		if addonVersion == "latest" || strings.Contains(*addonVersionInfo.AddonVersion, addonVersion) {
+			versions = append(versions, v)
+		}
+	}
+
+	if len(versions) == 0 {
+		return "", fmt.Errorf("version %s not available for %q", addonVersion, addon.Name)
```
aclevername, line 112: shall we change this to something like "no versions found matching %q"

pkg/actions/addon/update.go
```diff
@@ -41,9 +42,9 @@ func (a *Manager) Update(addon *api.Addon, wait bool) error {
...
```
aclevername, line 45: we don't want this anymore right?

nikimanoledaki: that's true! thank you 😊

pkg/actions/addon/addon.go
nikimanoledaki: sure, that's better wording!

pkg/actions/addon/create.go
aclevername, line 30: same here!

aclevername: perhaps worth adding a test for `1.2.3.eksbuild`?

nikimanoledaki: yup! awesome, thanks I'll put all this feedback in a commit!

aclevername (approved):
:tada:

PR #3651, opened 2021-04-30
Add plan for Flux2 expansion

Callisto13:
I wrote something of a proposal so that:
- We are all happy with this UX
- The plan is mapped out before I go away for 3 weeks
- Work is broken down into small chunks

[rendered](https://github.com/Callisto13/eksctl/blob/more-flux/docs/proposal-008-flux2.md)

### DIFF
--- a/docs/proposal-008-flux2.md
+++ b/docs/proposal-008-flux2.md
@@ -1,0 +1,370 @@
+# Flux v2 integration plan: next steps
+
+## Authors
+
+Claudia Beresford (@Callisto13)
+
+## Status
+
+Approved.
+
+## Table of Contents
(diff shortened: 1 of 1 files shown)


aclevername (approved):
LGTM. do we want to wait/have the first step be PR into flux the changes from array to slice? then we can document/test in the code that they are using a minimum supported version of flux

Callisto13:
yep 👍  this PR is already open https://github.com/fluxcd/flux2/pull/1390, and I will add a note to our plan to verify the version is what we want (we probably don't _need_ to fail because it only affects one flag for the bootstrap 🤔  so maybe we can log)

PR #3702, opened 2021-05-14
Fix taints support in unmanaged nodegroups

### LINKED ISSUE #3694: Taints in unmanaged nodegroups do not support duplicate keys
**Problem**
Kubernetes supports specifying taints with the same key but a different taint `effect` value, e.g.,

`$ kubectl taint node <node> key1=value1:NoSchedule key1=value2:NoExecute`

This construct cannot be replicated in eksctl because the taints structure in unmanaged nodegroups is a `map[string]string`, so only one of the two keys will be preserved in the `taints` map. 

Managed nodegroups do not suffer from this issue as it uses a slice of `NodeGroupTaint` values to represent taints:

```golang
type NodeGroupTaint struct {
	Key    string             `json:"key,omitempty"`
	Value  string             `json:"value,omitempty"`
	Effect corev1.TaintEffect `json:"effect,omitempty"`
}
```

**Solution**
Support specifying taints as `[]NodeGroupTaint` for unmanaged nodegroups, while still supporting the old structure to avoid a breaking change. Additionally, log a deprecation message when the old structure is used, suggesting users to migrate to the new structure.

cPu1:
### Description

Fixes #3694 

### Checklist
- [x] Manually tested


aclevername (approved):
LGTM, do we need to update the example config file, or is the user input unchanged?

cPu1:
> LGTM, do we need to update the example config file, or is the user input unchanged?

The old format (`map[string]string`) will continue to work, but I'll update the example with the new format.

PR #3728, opened 2021-05-18
Refactor and increase test coverage for nodegroup creation 

nikimanoledaki:
### Checklist
- [ ] Manually tested


pkg/actions/nodegroup/create.go
```diff
@@ -97,100 +94,102 @@ func (m *Manager) Create(options CreateOpts, nodegroupFilter filter.NodeGroupFil
...
-		}
+	if err := nodegroupFilter.SetOnlyLocal(m.ctl.Provider.EKS(), m.stackManager, cfg); err != nil {
+		return err
+	}
 
-		awsNodeUsesIRSA, err := eks.DoesAWSNodeUseIRSA(ctl.Provider, m.clientSet)
-		if err != nil {
-			return errors.Wrap(err, "couldn't check aws-node for annotation")
-		}
+	logFiltered := cmdutils.ApplyFilter(cfg, filter.NewNodeGroupFilter())
```
cPu1, line 130: This should be using the existing `NodeGroupFilter` that has been constructed with user-supplied options, rather than creating a new one, otherwise it'll attempt creation of nodegroups that already exist in the cluster. The newly introduced `NodegroupFilter` interface type, however, doesn't have the methods that `ApplyFilter` needs.

nikimanoledaki: Excellent feedback, thank you for finding this issue! :) Added a commit to fix this by using the user-supplied `NodegroupFilter`, then adding the missing functions to the interface. This also solved the issue of not having the full `cfg` for the nodegroup! 😄

pkg/actions/nodegroup/create.go
```diff
@@ -34,24 +32,22 @@ type CreateOpts struct {
 	ConfigFileProvided        bool
 }
 
-func (m *Manager) Create(options CreateOpts, nodegroupFilter filter.NodeGroupFilter) error {
+func (m *Manager) Create(options CreateOpts, nodegroupFilter filter.NodegroupFilter) error {
```
aclevername, line 35: nit: we tend to do `NodeGroup` instead of `Nodegroup` in the code, although we definitely aren't consistent :smile:

pkg/actions/nodegroup/nodegroup.go
```diff
@@ -29,9 +31,28 @@ func New(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, clientSet *kubernetes
...
+// MockNodeGroupService can be used for passing a mock of the nodegroup initialiser.
+func (m *Manager) MockNodeGroupService(ngSvc eks.NodeGroupInitialiser) {
+	m.init = ngSvc
+}
+
+// MockStackManager can be used for passing a mock of the stack manager.
+func (m *Manager) MockStackManager(sm manager.StackManager) {
+	m.stackManager = sm
+}
+
```
aclevername, line 55: should these be in an `export_test.go` file, to "hide" them from consumers of Manager? Example:
https://github.com/weaveworks/eksctl/blob/main/pkg/drain/export_test.go

the `export_test.go` is part of the package, so you can call it from within your tests. But because its in an `_test.go` file it doesn't get compiled by go when somebody uses this package externally

pkg/actions/nodegroup/create.go
nikimanoledaki: I completely agree with you! The reason for it here is that the struct is `NodeGroupFilter` and the interface is `NodegroupFilter` (not very creative..)

pkg/cfn/manager/create_tasks.go
```diff
@@ -81,6 +81,23 @@ func (c *StackCollection) NewUnmanagedNodeGroupTask(nodeGroups []*api.NodeGroup,
...
+		for _, err := range errs {
+			if err != nil {
+				logger.Critical("%s\n", err.Error())
+			}
+		}
+		return fmt.Errorf("failed to create nodegroups for cluster %q", name)
+	}
+	return nil
+}
+
```
aclevername, line 100: does this need to live in the `manager` package? I think it could live in `pkg/actions/nodegroup`, since it just executes tasks it doesn't really belong to the stack manager

pkg/actions/nodegroup/nodegroup.go
nikimanoledaki: very good idea, I'll move it!

pkg/cfn/manager/create_tasks.go
nikimanoledaki: sure! I don't see any issues with moving it, sounds good

pkg/actions/nodegroup/nodegroup.go
nikimanoledaki: Moved it :)

pkg/cfn/manager/create_tasks.go
nikimanoledaki: Moved it too!

nikimanoledaki:
@aclevername I implemented your review feedback and then manually created a nodegroup to test the changes - no regression!

pkg/actions/nodegroup/nodegroup.go
```diff
@@ -18,6 +18,8 @@ type Manager struct {
 	cfg          *api.ClusterConfig
 	clientSet    *kubernetes.Clientset
 	wait         WaitFunc
+	init         eks.NodeGroupInitialiser
```
aclevername, line 21: I feel like `init` is a bit vague, but not sure on a better name :man_shrugging:

aclevername (approved):
LGTM! Big PR so worth waiting for 2 accepting reviews before merging IMO :smile:

pkg/actions/nodegroup/nodegroup.go
Callisto13: i think `initialiser`, while verbose, would be clearer. `init` as we know it in Go _does_ shorthand initialise, but it may be helpful to clarify that we don't mean _that kind_ of `init`.

pkg/actions/nodegroup/create_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}),
+
+	Entry("fails when it does not support ARM", ngEntry{
+		mockCalls: func(k *fakes.FakeKubeProvider, init *fakes.FakeNodeGroupInitialiser, f *utilFakes.FakeNodegroupFilter) {
+			k.NewRawClientReturns(nil, fmt.Errorf("err"))
+		},
+		expErr: fmt.Errorf("err"),
+	}),
+
+	Entry("fails to load VPC from config", ngEntry{
```
Callisto13, line 78: can we make what this tests expects to happen a bit clearer?
```suggestion
	Entry("when cluster is unowned, fails to load VPC from config if config is not supplied", ngEntry{
```

pkg/actions/nodegroup/create_test.go
```diff
@@ -0,0 +1,290 @@
...
+		mockCalls: func(k *fakes.FakeKubeProvider, init *fakes.FakeNodeGroupInitialiser, f *utilFakes.FakeNodegroupFilter) {
+			k.NewRawClientReturns(&kubernetes.RawClient{}, nil)
+			k.ServerVersionReturns("1.17", nil)
+			k.LoadClusterIntoSpecFromStackReturns(nil)
+			k.SupportsManagedNodesReturns(false, errors.New("err"))
+		},
+		expErr: errors.New("err"),
+	}),
+
+	Entry("fails when NodeGroupService fails to match instances", ngEntry{
```
Callisto13, line 97: what does this mean? match instances to what?

pkg/ctl/cmdutils/filter/nodegroup_filter.go
```diff
@@ -23,7 +30,7 @@ type NodeGroupFilter struct {
 	remoteNodegroups sets.String
 }
 
-// NewNodeGroupFilter create new NodeGroupFilter instance
+// New create new NodeGroupFilter instance
```
Callisto13, line 33: this comment seems inaccurate

pkg/actions/nodegroup/create.go
Callisto13: Maybe `NGFilter` for the struct name? Or just `Filter` for the interface?

pkg/ctl/cmdutils/filter/nodegroup_filter.go
```diff
@@ -59,16 +66,16 @@ func (f *NodeGroupFilter) AppendIncludeNames(names ...string) {
 	f.delegate.AppendIncludeNames(names...)
 }
 
-// A stackLister lists nodegroup stacks
-type stackLister interface {
+// StackLister lists nodegroup stacks
+type StackLister interface {
 	ListNodeGroupStacks() ([]manager.NodeGroupStack, error)
 }
```
Callisto13, line 72: Can we move this to the top of the file, have all definitions together?

Callisto13 (commented):
Great work 🎉 

I just have a couple of comments on really superficial things.

pkg/ctl/cmdutils/filter/nodegroup_filter.go
nikimanoledaki: Yup, done!

pkg/ctl/cmdutils/filter/nodegroup_filter.go
nikimanoledaki: Good spot, updated it!

pkg/actions/nodegroup/create_test.go
nikimanoledaki: Updated it to make it more descriptive :)

pkg/ctl/cmdutils/filter/nodegroup_filter.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/weaveworks/eksctl/pkg/cfn/manager"
 )
 
+//go:generate counterfeiter -o fakes/fake_nodegroup_filter.go . NodegroupFilter
```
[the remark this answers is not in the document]
nikimanoledaki: Good shout, updated all the gen comments!

Callisto13 (approved):
:shipit:

PR #3753, opened 2021-05-24
Configure parallel upgrades with UpdateConfig and MaxUnavailable (or MaxUnavailablePercentage)

### LINKED ISSUE #3613: Parallel managed nodegroup upgrades
EKS is introducing new fields to configure how a nodegroup handles parallel upgrades.

- [x] **Create Nodegroup**

Let's add this to the config and set these fields during `create nodegroup`.

- [x] **Update Nodegroup**

These values can also be updated after the nodegroup is created via the `update-nodegroup-config` API call. We don't have an obvious eksctl command that this change could be made in. 

We can either:
~1. Add new flags to `eksctl upgrade nodegroup`, so you can change the configuration via that command. Slightly confusing since `updateConfig` is actually "updating" the config, not upgrading it.~

**2. Introduce a new `eksctl update nodegroup` command, slightly confusing as its only applicable for this small spec of information.**

~3. Intro a new `eksctl utils update-nodegroup-config`, a better name but some `utils` command.~

nikimanoledaki:
### Description

Addresses the `Create` part of #3613

Multiple managed nodes can be upgraded simultaneously. `updateConfig` can be used to configure how to handle parallel upgrades. 

To avoid any downtime due to upgrading multiple nodes at once, the number of nodes that can become unavailable during an upgrade can be set with `maxUnavailable` like this:
```
updateConfig:
  maxUnavailable: 5
```

`maxUnavailablePercentage` defines the maximum number of unavailable nodes as a percentage of the total number of nodes. 

The default number of `maxUnavailable` is 1 but it's set by EKS, not eksctl.

This feature is only available for managed nodes.

**Implementation details**
* Validate that `updateConfig` has either `maxUnavailable` or `maxUnavailablePercentage` set, but not both at the same time.
* Validate that `maxUnavailable` is not greater than `maxSize`
* Add `updateConfig` to config file schema and CF template.
* Updated user docs about how to add `updateConfig` while creating managed nodegroups

### Checklist
- [x] Manually tested


examples/15-managed-nodes.yaml
```diff
@@ -15,6 +15,8 @@ managedNodeGroups:
     maxSize: 4
     availabilityZones: ["us-west-2a", "us-west-2b"]
     volumeSize: 20
+    updateConfig:
+      maxUnavailable: 20 # or maxUnavailablePercentage if specified in percentage
```
aclevername, line 19: is it `20%` or `20` when setting percentages? Might be worth updating the comment to be clear :smile:

go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80 // indirect
 	github.com/vektra/mockery v1.1.2
 	github.com/voxelbrain/goptions v0.0.0-20180630082107-58cddc247ea2 // indirect
-	github.com/weaveworks/goformation/v4 v4.10.2-0.20210512074147-4fd9b878d08c
+	github.com/weaveworks/goformation/v4 v4.10.2-0.20210512152024-6bd11e940025
```
aclevername, line 56: is there a goformation PR we need to merge before this PR?

examples/15-managed-nodes.yaml
nikimanoledaki: true! thanks I'll reword :)

go.mod
nikimanoledaki: yes! https://github.com/weaveworks/goformation/pull/28

examples/15-managed-nodes.yaml
ravisinha0506: a note: maxUnavailable can't be greater than maxSize of the node group

nikimanoledaki: thanks for reviewing @ravisinha0506, this has been updated (this code snippet is outdated)!

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -844,6 +844,10 @@ type NodeGroup struct {
 	// +optional
 	Taints taintsWrapper `json:"taints,omitempty"`
 
+	// UpdateConfig configures how to update NodeGroups.
+	// +optional
+	UpdateConfig *NodeGroupUpdateConfig `json:"updateConfig,omitempty"`
```
Callisto13, line 849: Would this be better named `UpgradeConfig` given the update vs upgrade mess we have had in the past?

nikimanoledaki: Very valid point, that logic/naming would definitely make more sense for `eksctl`, however the AWS docs will be calling this new field `UpdateConfig`, so it might be better to keep it like this for consistency despite the slight cognitive dissonance?

Callisto13: omigod they suck. uuuugggh really hate calling this update

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -585,6 +586,15 @@ func ValidateManagedNodeGroup(ng *ManagedNodeGroup, index int) error {
 		ng.DesiredCapacity = ng.MinSize
 	}
 
+	if ng.UpdateConfig != nil {
+		if ng.UpdateConfig.MaxUnavailable != nil && ng.UpdateConfig.MaxUnavailablePercentage != nil {
+			return fmt.Errorf("cannot use maxUnavailable and maxUnavailablePercentage at the same time")
+		}
+		if aws.IntValue(ng.UpdateConfig.MaxUnavailable) > aws.IntValue(ng.MaxSize) || aws.IntValue(ng.UpdateConfig.MaxUnavailablePercentage) > aws.IntValue(ng.MaxSize) {
```
ravisinha0506, line 593: maxUnavailablePercentage can be greater than maxSize of the node group. For ex: a node group with 20 nodes should be able to upgrade 5 nodes at a time which makes it 25% ( greater than 20max size )

pkg/apis/eksctl.io/v1alpha5/managed_nodegroup_test.go
```diff
@@ -170,4 +170,57 @@ var _ = Describe("Managed Nodegroup Validation", func() {
...
+	},
+		Entry("max unavailable set", updateConfigEntry{
+			unavailable: aws.Int(1),
+			valid:       true,
+		}),
+		Entry("max unavailable specified in percentage", updateConfigEntry{
+			unavailablePercentage: aws.Int(1),
+			valid:                 true,
+		}),
+		Entry("returns an error if both are set", updateConfigEntry{
```
ravisinha0506, line 210: Should there be a test to validate that either unavailable or unavailablePercentage is defined?

pkg/apis/eksctl.io/v1alpha5/validation.go
nikimanoledaki: of course. 🤦 changed it, thanks for spotting this!

userdocs/src/usage/eks-managed-nodes.md
```diff
@@ -184,6 +185,11 @@ To upgrade to a specific AMI release version instead of the latest version, pass
 eksctl upgrade nodegroup --name=managed-ng-1 --cluster=managed-cluster --release-version=1.19.6-20210310
 ```
 
+## Handling parallel upgrades for nodes
+
+Multiple managed nodes can be upgraded simultaneously. `upgradeConfig` can be used to handle parallel upgrades. To avoid any downtime to your workloads due to upgrading multiple nodes at once, limit the number of nodes that can become unavailable during an upgrade as `maxnavailable`. Alternatively, use `maxUnavailablePercentage`, which defines the maximum number of unavailable nodes as a percentage of the total number of nodes. Note that this feature is only available for managed nodes.
```
Callisto13, line 190: should this be `updateConfig`?

(I mean it _should_ be upgrade, but I think we/they have called it update like savages 😉 )

nikimanoledaki: good spot! and yep ++ agreed 👀 

the user docs needed some extra TLC: https://github.com/weaveworks/eksctl/pull/3753/commits/a58cf6ff7ea98c39cef56f094a3ecf1744a0f771

aclevername (approved):

PR #3760, opened 2021-05-25
Revert "separate integration tests into separate github workflow jobs"

aclevername:
Reverts weaveworks/eksctl#3513


cPu1 (approved):

Callisto13:
why are we reverting this?

PR #3762, opened 2021-05-25
Support Bottlerocket as a node AMI family for managed nodegroups

### LINKED ISSUE #3709: Support Bottlerocket as a node AMI family for managed nodegroups
Managed nodegroups using Bottlerocket can be created by supplying a custom AMI and a bootstrap script, e.g.,:

```yaml
managedNodeGroups:
  - name: x86-bottlerocket-mng
    ami: ami-0acfc3fffaaa4e015
    instanceType: m5.large
    minSize: 1
    desiredCapacity: 2
    maxSize: 3
    privateNetworking: true
    disableIMDSv1: true
    volumeSize: 50
    labels: { role: x86-bottlerocket-mng }
    iam:
      attachPolicyARNs:
    overrideBootstrapCommand: |
      [settings.kubernetes]
      api-server = "https://redacted.gr7.us-east-1.eks.amazonaws.com"
      cluster-certificate = "redacted"
      cluster-name = "testing-cluster"
```

Unmanaged nodegroups, however, support this via the `amiFamily` field. 

Add native support for Bottlerocket for managed nodegroups so that they can be created by setting `amiFamily` to `Bottlerocket`.

cPu1:
### Description

Adds support for Bottlerocket as a node AMI family. 
Allows creation of Bottlerocket nodes in managed nodegroups by setting `amiFamily` to `Bottlerocket`. 

Closes #3709 

### Checklist
- [x] Manually tested


pkg/cfn/builder/managed_launch_template_test.go
```diff
@@ -44,8 +45,9 @@ var _ = Describe("ManagedNodeGroup builder", func() {
 		fakeVPCImporter.SecurityGroupsReturns(gfnt.Slice{gfnt.MakeFnImportValueString("eksctl-lt::ClusterSecurityGroupId")})
 		fakeVPCImporter.SubnetsPublicReturns(gfnt.MakeFnSplit(",", gfnt.MakeFnImportValueString("eksctl-lt::SubnetsPublic")))
 
-		stack := NewManagedNodeGroup(provider.MockEC2(), clusterConfig, m.ng, NewLaunchTemplateFetcher(provider.MockEC2()), false, fakeVPCImporter)
-		stack.UserDataMimeBoundary = "//"
+		bootstrapper := nodebootstrap.NewManagedAL2Bootstrapper(m.ng)
```
Callisto13, line 48: I think this should be a fake bootstrapper here

pkg/cfn/builder/managed_nodegroup_test.go
```diff
@@ -77,12 +78,14 @@ func TestManagedPolicyResources(t *testing.T) {
...
 
 			ng := api.NewManagedNodeGroup()
+			api.SetManagedNodeGroupDefaults(ng, clusterConfig.Metadata)
 			ng.IAM.WithAddonPolicies = tt.addons
 			ng.IAM.AttachPolicyARNs = prefixPolicies(tt.attachPolicyARNs...)
 
 			p := mockprovider.NewMockProvider()
 			fakeVPCImporter := new(vpcfakes.FakeImporter)
-			stack := NewManagedNodeGroup(p.EC2(), clusterConfig, ng, nil, false, fakeVPCImporter)
+			bootstrapper := nodebootstrap.NewManagedBootstrapper(clusterConfig, ng)
```
Callisto13, line 87: same here these should probably be fakes

Callisto13 (commented):
This would also be a good time to add `pkg/nodebootstrap/managed_al2_test.go` and bulk up the `pkg/nodebootstrap/bottlerocket_test.go`.

pkg/nodebootstrap/managed_al2_test.go
```diff
@@ -0,0 +1,84 @@
...
+				Name: "ssm-enabled",
+				SSH: &api.NodeGroupSSH{
+					Allow:         api.Enabled(),
+					PublicKeyName: aws.String("test-keypair"),
+					EnableSSM:     api.Enabled(),
+				},
+			},
+		},
+
+		expectedUserData: `TUlNRS1WZXJzaW9uOiAxLjANCkNvbnRlbnQtVHlwZTogbXVsdGlwYXJ0L21peGVkOyBib3VuZGFyeT0vLw0KDQotLS8vDQpDb250ZW50LVR5cGU6IHRleHQveC1zaGVsbHNjcmlwdA0KQ29udGVudC1UeXBlOiBjaGFyc2V0PSJ1cy1hc2NpaSINCg0KIyEvYmluL2Jhc2gKCnNldCAtbyBlcnJleGl0CnNldCAtbyBwaXBlZmFpbApzZXQgLW8gbm91bnNldAoKeXVtIGluc3RhbGwgLXkgYW1hem9uLXNzbS1hZ2VudApzeXN0ZW1jdGwgZW5hYmxlIGFtYXpvbi1zc20tYWdlbnQKc3lzdGVtY3RsIHN0YXJ0IGFtYXpvbi1zc20tYWdlbnQKDQotLS8vLS0NCg==`,
```
Callisto13, line 38: Can we decode the userdata and check against human readable fields?

cPu1: Yeah, that's how I wrote it initially but it involved replacing CRLF sequences. I agree that it's much more readable than encoded values. I'll see how we can best express it.

pkg/cfn/builder/managed_launch_template_test.go
cPu1: A fake bootstrapper would make the test less useful. This is more of an E2E unit test that asserts that for a given nodegroup, the correct CloudFormation template (which includes user data) is generated.

pkg/cfn/builder/managed_nodegroup_test.go
cPu1: A fake doesn't achieve much here and using a real bootstrapper also ensures the generated template more closely resembles a working nodegroup template.

pkg/cfn/builder/managed_launch_template_test.go
Callisto13: we should be able to test that in a the nodebootstrap unit, having all these things coupled unnecessarily is causing problems when we try to refactor/add new things

pkg/cfn/builder/managed_nodegroup_test.go
Callisto13: the bootstrapper interface exists to decouple it from the builder, so if i change something in the bootstrapper it doesn;t break things in the builder

pkg/cfn/builder/managed_launch_template_test.go
cPu1: This test is for testing the launch template generation, an important part of which is the user data, for a given nodegroup. It predates the bootstrapper interface. I would have liked to have a test that matches the real world, but seeing as I have added tests for the managed bootstrapper unit separately, I think I agree that this can be replaced with a fake.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -516,9 +516,13 @@ func validateNodeGroupIAM(iam *NodeGroupIAM, value, fieldName, path string) erro
...
 // ValidateManagedNodeGroup validates a ManagedNodeGroup and sets some defaults
 func ValidateManagedNodeGroup(ng *ManagedNodeGroup, index int) error {
-	if ng.AMIFamily != NodeImageFamilyAmazonLinux2 {
-		return fmt.Errorf("only %s is supported for Managed Nodegroups", NodeImageFamilyAmazonLinux2)
+	switch ng.AMIFamily {
+	case NodeImageFamilyAmazonLinux2, NodeImageFamilyBottlerocket:
+
+	default:
+		return errors.Errorf("%q is not supported for managed nodegroups", ng.AMIFamily)
 	}
```
aclevername, line 524: does this need to be a switch statement? looks weird seeing an empty `case`

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -619,6 +623,9 @@ func ValidateManagedNodeGroup(ng *ManagedNodeGroup, index int) error {
 		if !IsAMI(ng.AMI) {
 			return errors.Errorf("invalid AMI %q (%s.%s)", ng.AMI, path, "ami")
 		}
+		if ng.AMIFamily != NodeImageFamilyAmazonLinux2 {
+			return errors.Errorf("cannot set amiFamily to %s when using a custom AMI", ng.AMIFamily)
+		}
```
aclevername, line 628: does this need to include bottlerocket? slightly confused

pkg/apis/eksctl.io/v1alpha5/validation.go
cPu1: It can be an `if` statement but a `switch` is more concise when comparing against the same value. Ubuntu support will add another value to the `case` clause.

pkg/cfn/manager/nodegroup.go
```diff
@@ -76,7 +77,8 @@ func (c *StackCollection) createManagedNodeGroupTask(errorCh chan error, ng *api
 	name := c.makeNodeGroupStackName(ng.Name)
 
 	logger.Info("building managed nodegroup stack %q", name)
-	stack := builder.NewManagedNodeGroup(c.ec2API, c.spec, ng, builder.NewLaunchTemplateFetcher(c.ec2API), forceAddCNIPolicy, vpcImporter)
+	bootstrapper := nodebootstrap.NewManagedBootstrapper(c.spec, ng)
+	stack := builder.NewManagedNodeGroup(c.ec2API, c.spec, ng, builder.NewLaunchTemplateFetcher(c.ec2API), bootstrapper, forceAddCNIPolicy, vpcImporter)
```
aclevername, line 81: big nit: can this function get moved into a `managed_nodegroup.go` file? I find it weird that it lives in the `unmanaged` `nodegroup.go` file.

pkg/eks/nodegroup_service.go
```diff
@@ -65,8 +65,14 @@ func (m *NodeGroupService) NewAWSSelectorSession(provider api.ClusterProvider) *
 func (m *NodeGroupService) Normalize(nodePools []api.NodePool, clusterMeta *api.ClusterMeta) error {
 	for _, np := range nodePools {
 		switch ng := np.(type) {
+		case *api.ManagedNodeGroup:
+			if ng.AMIFamily != api.NodeImageFamilyAmazonLinux2 && !api.IsAMI(ng.AMI) {
+				if err := ResolveAMI(m.Provider, clusterMeta.Version, np); err != nil {
+					return err
+				}
+			}
```
aclevername, line 73: again slightly confused on why we don't have to `if` on bottlerocket?

aclevername (approved):
1 nit & 1 comment open but otherwise LGTM

Pair reviewed with @Callisto13

pkg/eks/nodegroup_service.go
cPu1: Only Bottlerocket requires AMI resolution in managed nodegroups. We plan on adding support for Ubuntu and Windows later, so instead of explicitly stating all of them, I chose to compare it with the one AMI family that doesn't require AMI resolution. I agree that this condition can be made clearer, so I'll work on something.

cPu1: I have pushed a change that should make the condition more clear.

pkg/apis/eksctl.io/v1alpha5/validation.go
cPu1: The alternative would be to include all the unsupported AMI families. I'll try to make this more explicit in the PR for Ubuntu support.

pkg/cfn/manager/nodegroup.go
cPu1: This file contains a lot of stuff related to managed nodegroups. I think at this point `nodegroup` refers to both managed and unmanaged nodegroups.

pkg/apis/eksctl.io/v1alpha5/validation.go
aclevername:
```
if ng.AMIFamily != NodeImageFamilyAmazonLinux2 {			
  return errors.Errorf("cannot set amiFamily to %s when using a custom AMI", ng.AMIFamily)		
}
```

wait but doesn't this mean that if you provide bottlerocket, it will fail?

pkg/cfn/manager/nodegroup.go
aclevername: thats extremely confusing! We should definitely refactor it out

pkg/apis/eksctl.io/v1alpha5/validation.go
cPu1: Yes, that's the intent. I meant the alternative would to be do a `if ng.AMIFamily != unsupported AMI families`.

rverma-nsl:
@cPu1 observed one bug with this. When we have a policy for strict ebs encryption, we observe that nodes are not coming up and its fail to launch the instance. On further analysis it was observed that, since bottlerocket uses 2 voluments , 2GB + root, the default 2GB volume is no longer encrypted i.e `/dev/xvda` & not included in launch template too.  This causes failure.

PR #3770, opened 2021-05-26
Add network-interface to TagSpecification

Aym3nTN:
### Description

I noticed that AWS added the ability to tag the network interfaces created via Launch Template, so I made a small research on which [attribute](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TagSpecification.html) to modify in order to make this happen automatically. 

image

### Checklist
- [ ] This is really a minor change and won't disrupt anything.

EDIT:
I updated the cfn testdata JSON files and nodegroup_test.go file


Callisto13 (approved):
LGTM, thanks 🎉

PR #3778, opened 2021-05-27
Write kubeconfig to temporary location before `enable flux` runs

Callisto13:
### Description

This PR only affects the `enable flux` command. If you are bootstrapping flux as part of a cluster create, the config and context are already set as part of the post create steps.

Before running any Flux commands, we write the cluster kubeconfig to a temporary location. We then set the `kubeconfig` flag on the Flux opts (to pass into those commands). A `defer` will clean up the created temp file.

I am not thrilled about where I have implemented this; it is in the `ctl` package which is basically untestable. Testing is weird in itself, since the outcome is that a file _does not_ exist. The only checkable thing is that a flag has been set, but I could not find a nice place to do that. The change is covered by the integration test, which would fail if this didn't work.

Do we want users to ever set their own kubeconfig/context? It doesn't make a lot of sense in the context of eksctl, but I have put a guard in there anyway. If we don't think they should ever be setting this, then we can have a thing in the flux package to discount those flags.

Integration test passing.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/enable/flux.go
+++ b/pkg/ctl/enable/flux.go
@@ -1,12 +1,16 @@
 package enable
 
 import (
+	"io/ioutil"
+	"os"
+
 	"github.com/kris-nova/logger"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/weaveworks/eksctl/pkg/actions/flux"
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 	"github.com/weaveworks/eksctl/pkg/ctl/cmdutils"
+	"github.com/weaveworks/eksctl/pkg/utils/kubeconfig"
 	"github.com/weaveworks/eksctl/pkg/version"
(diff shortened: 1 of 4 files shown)


pkg/ctl/enable/flux.go
```diff
@@ -46,6 +50,32 @@ func flux2Install(cmd *cmdutils.Cmd) error {
...
+		}
+
+		kubeCfgPath, err := ioutil.TempFile("", cmd.ClusterConfig.Metadata.Name)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := os.Remove(kubeCfgPath.Name()); err != nil {
+				logger.Debug("failed to remove temporary kubeconfig", kubeCfgPath)
+			}
```
aclevername, line 69: should this be `logger.Critical`?

Callisto13: meh probably

aclevername (approved):

PR #3810, opened 2021-06-04
Return k8s version when listing nodegroup(s)

### LINKED ISSUE #3809: Return k8s version when listing nodegroup(s)
Currently, the k8s version of a nodegroup is not returned when doing `get nodegroup`. 

The nodegroup's version can only be retrieved by checking through the AWS console.

nikimanoledaki:
### Description

Closes #3809

This PR adds the k8s version as an output field for better UX, parity with the EKS console, and easier debugging.

It also adds a bit more info in the user docs about listing nodegroups with the output logged in json or yaml format, which returns more info than with the default table logger.

### Managed nodegroups
The implementation is pretty straightforward for managed nodegroups because the EKS API for getting summaries returns the `version` anyway.

Output:
```bash
./eksctl get nodegroup --cluster test-2 --output yaml 
- AutoScalingGroupName: eks-10bceb40-fe58-1dca-387b-e5adf8e4a11a
  Cluster: test-2
  CreationTime: "2021-06-04T10:11:56.439Z"
  DesiredCapacity: 2
  ImageID: AL2_x86_64
  InstanceType: m5.large
  MaxSize: 2
  MinSize: 2
  Name: test-ng
  NodeInstanceRoleARN: arn:aws:iam::083751696308:role/eksctl-test-2-nodegroup-test-ng-NodeInstanceRole-1HEKJUXC8492Z
  StackName: eksctl-test-2-nodegroup-test-ng
  Status: ACTIVE
  Version: "1.18"
- AutoScalingGroupName: eks-8abceb4b-162d-7999-421d-1d7279a5afa4
  Cluster: test-2
  CreationTime: "2021-06-04T10:33:58.902Z"
  DesiredCapacity: 2
  ImageID: AL2_x86_64
  InstanceType: m5.large
  MaxSize: 2
  MinSize: 2
  Name: test-ng-2
  NodeInstanceRoleARN: arn:aws:iam::083751696308:role/eksctl-test-2-nodegroup-test-ng-2-NodeInstanceRole-148FQ6YOES9ZJ
  StackName: eksctl-test-2-nodegroup-test-ng-2
  Status: ACTIVE
  Version: "1.18"
```

### Unmanaged nodegroups

The stack and template do not contain the k8s version, so the only way to get the version of the nodegroup is by doing `k get nodes` with the kube client. In the loop that goes through the stacks and prepares the nodegroup summaries, it queries k8s to list the nodes that have the label of the current Nodegroup. Since all nodes in a nodegroup are the same k8s version, only 1 node is required to do this.

Output:
```bash
./eksctl get nodegroup --cluster n-test --name ng-11e8fae8 --output yaml
ngName: ng-11e8fae8- AutoScalingGroupName: eksctl-n-test-nodegroup-ng-11e8fae8-NodeGroup-OUE252BHQMNG
  Cluster: n-test
  CreationTime: "2021-06-07T07:50:44.495Z"
  DesiredCapacity: 2
  ImageID: ami-038d10226c688c779
  InstanceType: m5.large
  MaxSize: 2
  MinSize: 2
  Name: ng-11e8fae8
  NodeInstanceRoleARN: arn:aws:iam::083751696308:role/eksctl-n-test-nodegroup-ng-11e8fa-NodeInstanceRole-10Y91XCXIKMP8
  StackName: eksctl-n-test-nodegroup-ng-11e8fae8
  Status: CREATE_COMPLETE
  Version: 1.19.6
```
### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/nodegroup/get.go
+++ b/pkg/actions/nodegroup/get.go
@@ -61,6 +61,7 @@ func (m *Manager) GetAll() ([]*manager.NodeGroupSummary, error) {
 			CreationTime:         describeOutput.Nodegroup.CreatedAt,
 			NodeInstanceRoleARN:  *describeOutput.Nodegroup.NodeRole,
 			AutoScalingGroupName: strings.Join(asgs, ","),
+			Version:              *describeOutput.Nodegroup.Version,
 		})
 	}
 
--- a/pkg/actions/nodegroup/get.go
+++ b/pkg/actions/nodegroup/get.go
@@ -13,7 +13,7 @@ import (
 )
 
 func (m *Manager) GetAll() ([]*manager.NodeGroupSummary, error) {
-	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("")
+	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("", m.clientSet.CoreV1().Nodes())
 	if err != nil {
 		return nil, errors.Wrap(err, "getting nodegroup stack summaries")
 	}
@@ -77,7 +77,7 @@ func getInstanceTypes(ng *awseks.Nodegroup) string {
 }
 
 func (m *Manager) Get(name string) (*manager.NodeGroupSummary, error) {
-	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries(name)
+	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries(name, m.clientSet.CoreV1().Nodes())
 	if err != nil {
 		return nil, errors.Wrap(err, "getting nodegroup stack summaries")
 	}
@@ -114,5 +114,6 @@ func (m *Manager) Get(name string) (*manager.NodeGroupSummary, error) {
 		CreationTime:         describeOutput.Nodegroup.CreatedAt,
 		NodeInstanceRoleARN:  *describeOutput.Nodegroup.NodeRole,
 		AutoScalingGroupName: asg,
+		Version:              *describeOutput.Nodegroup.Version,
 	}, nil
 }
--- a/pkg/actions/nodegroup/get.go
+++ b/pkg/actions/nodegroup/get.go
@@ -10,12 +10,20 @@ import (
 	"github.com/kris-nova/logger"
 	"github.com/pkg/errors"
 	"github.com/weaveworks/eksctl/pkg/cfn/manager"
+	kubewrapper "github.com/weaveworks/eksctl/pkg/kubernetes"
 )
 
 func (m *Manager) GetAll() ([]*manager.NodeGroupSummary, error) {
-	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("", m.clientSet.CoreV1().Nodes())
+	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("")
 	if err != nil {
 		return nil, errors.Wrap(err, "getting nodegroup stack summaries")
+	}
+
+	for _, summary := range summaries {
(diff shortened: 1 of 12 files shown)


Callisto13 (approved):

aclevername:
@nikimanoledaki does this only work for `managed` nodegroups? What happens in `unmanaged`?

nikimanoledaki:
Thanks for the approval + feedback @Callisto13 @aclevername I've just added stuff to also return the kube version of unmanaged nodegroups! :)

pkg/actions/nodegroup/get.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func (m *Manager) GetAll() ([]*manager.NodeGroupSummary, error) {
-	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("")
+	summaries, err := m.stackManager.GetUnmanagedNodeGroupSummaries("", m.clientSet.CoreV1().Nodes())
```
aclevername, line 16: I don't feel overly strongly about this but:

It would be nice if the `stackManager` package just did `cloudformation` calls, how would you feel about moving the kubernetes API calls into this func? I just think it reduces the scope of what the `stackManager` package does, which is already very large :smile: 

We definitely do kuberntes calls in the `stackManager` package elsewhere, but I think aiming for "this package just makes cloudformation calls" would be nice.

WDYT?

aclevername (commented):
LGTM- 1 possible refactor comment. Two questions:

- For a private cluster, does this break `eksctl get nodegroups`? Did that work before but no we querying k8s, so it would fail? I don't know, so maybe worth trying :smile: 
- Is there an integration test where we do `get nodegroups` that needs to get updated for the new format?

Other than that awesome PR :smile:

pkg/actions/nodegroup/get.go
nikimanoledaki: I agree - I'll refactor this to move the entire `GetUnmanagedNodeGroupSummaries` function to this package as a helper func, agree that we should only use `stackManager` for the cloudformation calls!

aclevername: :ok_hand:

nikimanoledaki: I've been making some changes to move `GetUnmanagedNodeGroupSummaries` out of the `stackManager` but it's quite tightly coupled with the other `stackManager` methods. In `GetUnmanagedNodeGroupSummaries`, the Kube API call is part of a loop that iterates over all unmanaged nodegroups, so it would require another loop if it were to be moved by itself, which wouldn't be very efficient. So I think it's worth creating an issue to refactor `GetUnmanagedNodeGroupSummaries` and its `stackManager` methods. What do you think?

nikimanoledaki: Moved `kubewrapper.GetNodegroupKubernetesVersion()` to this function in https://github.com/weaveworks/eksctl/pull/3810/commits/47d6e735a51613a5519eb2702f1cc331de0896c5!

nikimanoledaki:
> LGTM- 1 possible refactor comment. Two questions:
> 
> * For a private cluster, does this break `eksctl get nodegroups`? Did that work before but now we querying k8s, so it would fail? I don't know, so maybe worth trying 😄
> * Is there an integration test where we do `get nodegroups` that needs to get updated for the new format?
> 
> Other than that awesome PR 😄

Thanks for reviewing again 😄 
* did some digging and found that private clusters aren't accessible by eksctl
* good shout! I added an assertion on the CRUD test to check that the k8s version is returned and ran it locally, all good!

aclevername (approved):

PR #3828, opened 2021-06-10
Bootstrap `update nodegroup` command

### LINKED ISSUE #3827: Bootstrap new `update nodegroup` command
## New Feature: `eksctl update nodegroup`

The new `update nodegroup` command will be used for updating a nodegroup's configuration. 

`update nodegroup` is not to be confused with `upgrade nodegroup`, which is used for upgrading a nodegroup and its resources, for example to use a different AMI with a newer k8s version.

Instead, `update nodegroup` can be used to update configuration of nodegroups. This is important especially if the config needs to be changed prior to running an upgrade. 

Note that this is only available for managed nodegroups.

This PR relates to #3613 - it will be used to updating the new `UpdateConfig` field that will be used for handling parallel upgrades.

nikimanoledaki:
### Description

Closes #3827 

This PR bootstraps the `update nodegroup` command. 

Validates that:
- Nodegroup must be a managed nodegroup
- When used with a config file, a warning will log any unsupported fields that were present in the config file but could not be updated

```bash
./eksctl update nodegroup --config-file examples/15-managed-nodes.yaml
2021-06-16 15:49:20 [!]  unchanged fields: the following fields could not be updated because they are not supported by `eksctl update nodegroup`: InstanceType, AvailabilityZones, ScalingConfig, VolumeSize, SSH, Labels, Tags, IAM, SecurityGroups, Taints, UpdateConfig
2021-06-16 15:49:22 [ℹ]  nodegroup successfully updated
```

We don't want users to confused `upgrade` and `update` so every effort has been made to explain this in the `ctl` and user docs. We also don't want users to try to update fields in the config that cannot be updated through `update nodegroup` so a lot of validation was added to make sure only very specific fields in the config can be used with this command.

### Checklist
- [x] Manually tested


userdocs/src/usage/eks-managed-nodes.md
```diff
@@ -184,6 +184,11 @@ To upgrade to a specific AMI release version instead of the latest version, pass
 eksctl upgrade nodegroup --name=managed-ng-1 --cluster=managed-cluster --release-version=1.19.6-20210310
 ```
 
+## Updating managed nodegroups
+It is also possible to update specific fields of a managed nodegroup with the command `eksctl update nodegroup` without using `upgrade`. This new command currently only supports a few fields and is intended to be used as a way to modify the configuration of a nodegroup without triggering an entire upgrade.
```
aclevername, line 188: shall we add a list of the fields? or I guess in this PR there isn't any yet?

aclevername:
@nikimanoledaki I think we don't want to merge this unless we have the PR that builds on top of this ready to get, otherwise we are shipping a new command that has no functionality :thinking:

userdocs/src/usage/eks-managed-nodes.md
nikimanoledaki: I had added a link to the example config but linter didn't like it because the link will appear broken until this PR is merged! https://github.com/weaveworks/eksctl/pull/3828/commits/4d888bc40944257b1a8e9bbbdb7a02779f262292

nikimanoledaki:
@aclevername no problem! today I'll be making a branch with this one as its base to add `updateConfig` so that we can ~yeet~ merge both at the same time!

userdocs/src/usage/eks-managed-nodes.md
nikimanoledaki: I've made a note to make a one-liner PR after this one is merged to add the link to the example again, so that the linter doesn't complain :)

examples/29-update-nodegroup.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: cluster-29
+  region: us-west-2
+
+managedNodeGroups:
+- name: mng
```
Callisto13, line 10: what's this example config demonstrating?

pkg/ctl/cmdutils/configfile.go
```diff
@@ -761,3 +762,84 @@ func NewDeleteIAMServiceAccountLoader(cmd *Cmd, sa *api.ClusterIAMServiceAccount
...
+	l.validateWithConfigFile = func() error {
+		var clusterConfig *api.ClusterConfig
+		var err error
+		if clusterConfig, err = eks.LoadConfigFromFile(l.ClusterConfigFile); err != nil {
+			return err
+		}
+
+		l := len(clusterConfig.ManagedNodeGroups)
+		if l > 1 {
+			return errors.New("please update one NodeGroup at a time")
```
Callisto13, line 779: so just to clarify on the UX...

say I had a config file like
```yaml
managedNodeGroups:
- name: one
  someOtherField: foobar
- name: two
  updateConfig:
    shrugEmoji: true
- name: three
```
in order to run this command i would have to edit my config file to remove those nodegroups?

pkg/ctl/update/nodegroup_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+		}
+		config := ctltest.CreateConfigFile(cfg)
+		cmd := newMockCmd("nodegroup", "--config-file", config)
+		_, err := cmd.execute()
+		Expect(err).To(HaveOccurred())
+		Expect(err).To(MatchError(ContainSubstring("unsupported field: MaxPodsPerNode cannot be used with `eksctl update nodegroup`")))
+	})
+
+	It("returns error if cfg contains unsupported field in NodeGroup fields", func() {
```
Callisto13, line 56: I may have missed something so this question may be irrelevant, but are we validating other fields for any other reason apart from clarity? Presumably we are only reading the `updateConfig` field for this command?

userdocs/src/usage/eks-managed-nodes.md
Callisto13: Can we add more to make it very clear that this will only ever update upgrade configuration?

I mean, who knows what we may use it for later, but let's not lead anyone on.

pkg/ctl/cmdutils/configfile.go
nikimanoledaki: right now yes but I'm not dead set on this UX, I meant to ask about this and I'm happy to change it. do we want `update nodegroup` to be able to update multiple nodegroups at once?

pkg/ctl/update/nodegroup_test.go
nikimanoledaki: it's just for clarity :)

pkg/ctl/cmdutils/configfile.go
Callisto13: I think so yes. People are likely to have their cluster config stored somewhere so it feels weird that we would require them to remove bits to run this command. With the config above I would have to comment-run-comment 3 times which does not seem fun.

pkg/ctl/update/nodegroup_test.go
Callisto13: cool cool

examples/29-update-nodegroup.yaml
nikimanoledaki: nothing right now, this PR is just for bootstrapping `eksctl update nodegroup`! I'm creating another PR to use `update nodegroup` to update `updateConfig!

nikimanoledaki: see: https://github.com/weaveworks/eksctl/pull/3855/commits/2f353953cddf57eb99fdc641a4941aacc2e7279b

pkg/ctl/cmdutils/configfile.go
nikimanoledaki: good shout! opened a new issue to add this afterwards: https://github.com/weaveworks/eksctl/issues/3857

userdocs/src/usage/eks-managed-nodes.md
nikimanoledaki: definitely, more info will be added in the PR that joins `update nodegroup` and `updateConfig`

nikimanoledaki:
@Callisto13 I removed the validation that returned an error if there were `> 1` ng in the cfg file, so now the config file that's used for creating the cluster can be used! :) Will do the `multiple nodegroup` bit in a separate PR to keep things tidy. 

Also, it doesn't return errors anymore for unsupported fields and instead it just logs the fields that could not be updated (see description) ☺️ I like it! What do you think?

integration/tests/managed/managed_nodegroup_test.go
```diff
@@ -425,37 +423,59 @@ var _ = Describe("(Integration) Create Managed Nodegroups", func() {
...
+						"nodegroup",
+						"--config-file", "-",
+						"--verbose", "4",
+					).
+					WithoutArg("--region", params.Region).
+					WithStdin(testutils.ClusterConfigReader(clusterConfig))
+
+				Expect(cmd).To(RunSuccessfullyWithOutputStringLines(
+					ContainElement(ContainSubstring("unchanged fields")),
+					ContainElement(ContainSubstring("Spot")),
```
aclevername, line 474: whats this testing?

pkg/actions/nodegroup/update.go
```diff
@@ -0,0 +1,28 @@
...
+func (m *Manager) Update() error {
+	ngName := m.cfg.ManagedNodeGroups[0].Name
+	_, err := m.ctl.Provider.EKS().DescribeNodegroup(&eks.DescribeNodegroupInput{
+		ClusterName:   &m.cfg.Metadata.Name,
+		NodegroupName: &ngName,
+	})
+
+	if err != nil {
+		if managed.IsNotFound(err) {
+			return fmt.Errorf("update is only supported for managed nodegroups; could not find one with name %q", ngName)
```
aclevername, line 21: slightly confusing because it could be either:

1. it doesn't exist
2. it exists as a `unmanaged` nodegroup

I think it might be worth distinguishing the two error cases, since this command might add a lot of confusion for users who try to update `unmanaged` nodegroups

userdocs/src/usage/eks-managed-nodes.md
```diff
@@ -195,6 +194,14 @@ Note that `maxUnavailable` cannot be higher than `maxSize`. Also, `maxUnavailabl
 
 This feature is only available for managed nodes.
 
+## Updating managed nodegroups
+It is also possible to update specific fields of a managed nodegroup with the command `eksctl update nodegroup` without using `upgrade`. 
+
+This new command currently only supports a few fields. It is intended to be used as a way to modify the configuration of a nodegroup without triggering an entire upgrade.
```
aclevername, line 200: its none in this PR right? and then the first field is introduced in the follow PR? Might be worth moving the docs to the follow up PR

Callisto13:
> I removed the validation that returned an error if there were > 1 ng in the cfg file, so now the config file that's used for creating the cluster can be used! 

So does this mean it will silently just act on the first in the array? Do we log that this is happening?

integration/tests/managed/managed_nodegroup_test.go
Callisto13: I believe it is testing that other config is ignored.

I know we don't normally do this and each command only looks at what is cares about, but since this is `update` and we often get asked things like "Can i update xyz on my ng without recreating?" it may make sense to be very clear and log the things in the config which are not going to be changed.

pkg/actions/nodegroup/update.go
```diff
@@ -0,0 +1,28 @@
...
+		}
+		return err
+	}
+
+	logger.Info("nodegroup successfully updated")
```
Callisto13, line 26: Where is the update call happening in this func? Am I being blind? Is github diff messing with me? 😂

Callisto13: o wait is it happening in https://github.com/weaveworks/eksctl/pull/3855?

pkg/ctl/cmdutils/configfile.go
```diff
@@ -761,3 +763,84 @@ func NewDeleteIAMServiceAccountLoader(cmd *Cmd, sa *api.ClusterIAMServiceAccount
...
+		}
+		return nil
+	}
+	return l
+}
+
+// validateSupportedConfigFields parses a config file's fields, evaluates if non-empty fields are supported,
+// and returns an error if a field is not supported.
+func validateSupportedConfigFields(i interface{}, supportedFields []string, unsupportedFields []string) ([]string, error) {
+	v := reflect.ValueOf(i)
```
Callisto13, line 809: can we rename this `i`? bit of shadowing.

pkg/ctl/cmdutils/configfile.go
```diff
@@ -761,3 +763,84 @@ func NewDeleteIAMServiceAccountLoader(cmd *Cmd, sa *api.ClusterIAMServiceAccount
...
+		ng := clusterConfig.ManagedNodeGroups[0]
+		if unsupportedFields, err = validateSupportedConfigFields(*ng.NodeGroupBase, []string{"Name"}, unsupportedFields); err != nil {
+			return err
+		}
+
+		if unsupportedFields, err = validateSupportedConfigFields(*ng, []string{"NodeGroupBase"}, unsupportedFields); err != nil {
+			return err
+		}
+
+		logger.Warning("unchanged fields: the following fields remain unchanged; they are not supported by `eksctl update nodegroup`: %s", strings.Join(unsupportedFields[:], ", "))
```
Callisto13, line 793: will this log even if that `unsupportedFields` array is empty?

integration/tests/managed/managed_nodegroup_test.go
nikimanoledaki: yes^ 👏 

I'd say informing the user of which fields were _not_ affected is so important for clarity of use that it's almost a feature and should be tested as such... it's part of making this command extra crystal clear to compensate for having a potentially confusing UX (`update` vs `upgrade`, `updateConfig` for upgrades)

nikimanoledaki: I'd (probably) be confused 😛

pkg/actions/nodegroup/update.go
nikimanoledaki: that's true, thank you, I'll find another way :)

pkg/actions/nodegroup/update.go
nikimanoledaki: yeah there's nothing happening here 🙈 all the fun happens over at #3855

pkg/ctl/cmdutils/configfile.go
nikimanoledaki: good shout, changing it to check `len` of `unsupportedFields` first!

pkg/actions/nodegroup/update.go
nikimanoledaki: would it be less confusing and capture both error cases if the error message was simply reworded to `could not find managed nodegroup with name %q`?

userdocs/src/usage/eks-managed-nodes.md
nikimanoledaki: do you mean that we should move all user docs about `update` to the follow-up PR? the other PR is pretty much ready to go right after this one and they're meant to be released together so I don't think it makes much difference tbh, what do you think?

pkg/actions/nodegroup/update.go
aclevername: > would it be less confusing and capture both error cases if the error message was reworded to could not find managed nodegroup with name %q?

sounds good!

userdocs/src/usage/eks-managed-nodes.md
aclevername: I think in that case lets leave it, but for the future If I was to go to this doc for a git blame, I wouldn't find the PR that actually introduced this functionality which would be slightly confusing, although we link to that PR in this PR so its not that bad :smile:

pkg/ctl/cmdutils/configfile.go
```diff
@@ -761,3 +763,84 @@ func NewDeleteIAMServiceAccountLoader(cmd *Cmd, sa *api.ClusterIAMServiceAccount
...
+			return err
+		}
+
+		l := len(clusterConfig.ManagedNodeGroups)
+		if l < 1 {
+			return ErrMustBeSet("managedNodeGroups field")
+		}
+
+		var unsupportedFields []string
+		ng := clusterConfig.ManagedNodeGroups[0]
```
aclevername, line 784: do we not need to validate all managed nodegroups, since they are all getting updated?

aclevername (approved):
left another comment, but otherwise LGTM

pkg/ctl/cmdutils/configfile.go
nikimanoledaki: for now it's just 1 nodegroup but I'll open another PR right afterwards to iterate over all nodegroups! https://github.com/weaveworks/eksctl/pull/3828#discussion_r651658891

userdocs/src/usage/eks-managed-nodes.md
nikimanoledaki: ok I'll point to the other PR in this PR's merge commit!

nikimanoledaki:
> So does this mean it will silently just act on the first in the array? Do we log that this is happening?

there's more logging in the other PR since not much happens in this one! sorry, it's a bit confusing!

PR #3829, opened 2021-06-11
Create nodegroups in integration tests with SSM enabled

cPu1:
### Description

In a future release, EKS AMIs might include the SSM agent by default. In preparation for that, this changelist adds an env var to enable SSM for all nodegroups, and sets it for integration tests.

### Checklist
- [ ] Manually tested


Callisto13 (changes requested):
I don't think this is something we should give to customers, even if it is not documented. It may be more work, but let's update the integration tests

cPu1:
It's a lot more manual work to update the integration tests because both `create cluster` and `create nodegroup` need to be updated to pass `--enable-ssm` and include `ssh.enableSsm` for every nodegroup if a config file is being used. It's possible to miss some tests this way. 

I'm thinking a better alternative is to use an environment variable that clearly signifies it's internal and may be removed in a future release, like `_____INTERNAL_NODEGROUP_ENABLE_SSM`.

Callisto13:
I am really against releasing anything which has nothing to do with customers or even eksctl.

There are like 13 test files, and they are not that long. It will take like less than 10 mins to update them all, and if we miss one 🤷‍♀️  we'll find out soon enough. This really isn't something we should be spend more than 30 seconds thinking about

Callisto13 (approved):

PR #3845, opened 2021-06-14
Make addon.Force unconfigurable

cPu1:
### Description

`addon.Force` is an internal field but it's still configurable in ClusterConfig. 

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/addon.go
+++ b/pkg/apis/eksctl.io/v1alpha5/addon.go
@@ -27,7 +27,7 @@ type Addon struct {
 	// +optional
 	Tags map[string]string `json:"tags,omitempty"`
 	// Force applies the add-on to overwrite an existing add-on
-	Force bool
+	Force bool `json:"-"`
 }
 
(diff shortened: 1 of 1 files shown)


aclevername:
Is there a reason why we wouldn't want this to be configurable from inside the config file?

cPu1:
> Is there a reason why we wouldn't want this to be configurable from inside the config file?

I believe this field was added for supporting the `--force` flag. If we fully support this field, the UX for `--force` along with `addon.force: false` can be tricky to get right (does `--force` override `addon.force`?). Furthermore, `addon.force` in itself does not convey whether it's for forcing the creation, update or deletion, so if we do want to support this feature, I think it should be named appropriately (`forceCreate` or `forceUpdate`?).

aclevername:
does the current `--force` flag work when a config file is provided?

cPu1:
> does the current `--force` flag work when a config file is provided?

I think so, which is why the current UX can be confusing.

aclevername (approved):

PR #3885, opened 2021-06-22
Add default tags to OIDC provider resources

cPu1:
### Description

Adds default eksctl tags to OIDC provider resources.

```
alpha.eksctl.io/cluster-name: <cluster>
alpha.eksctl.io/eksctl-version: <version>
```

Unlike many other resources, eksctl does not tag OIDC provider resources. I have run into this issue while trying to debug orphan OIDC provider resources which had no information about the cluster they were created for. At the very least, eksctl should tag OIDC provider resources with the cluster information, like it does for other resources. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/eks.go
+++ b/pkg/eks/eks.go
@@ -8,9 +8,9 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/weaveworks/eksctl/pkg/cfn/waiter"
-
 	"github.com/weaveworks/eksctl/pkg/cfn/manager"
+	"github.com/weaveworks/eksctl/pkg/cfn/waiter"
+	"github.com/weaveworks/eksctl/pkg/version"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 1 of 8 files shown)


pkg/iam/oidc/api_test.go
```diff
@@ -60,26 +60,26 @@ var _ = Describe("EKS/IAM API wrapper", func() {
 		})
```
Callisto13, line 60: can we add tests for the tags?

pkg/eks/eks.go
```diff
@@ -274,7 +274,16 @@ func (c *ClusterProvider) NewOpenIDConnectManager(spec *api.ClusterConfig) (*iam
...
-	return iamoidc.NewOpenIDConnectManager(c.Provider.IAM(), parsedARN.AccountID, *c.Status.ClusterInfo.Cluster.Identity.Oidc.Issuer, parsedARN.Partition)
+	return iamoidc.NewOpenIDConnectManager(c.Provider.IAM(), parsedARN.AccountID,
+		*c.Status.ClusterInfo.Cluster.Identity.Oidc.Issuer, parsedARN.Partition, sharedTags(c.Status.ClusterInfo.Cluster))
+}
+
+func sharedTags(cluster *awseks.Cluster) map[string]string {
+	return map[string]string{
+		api.ClusterNameTag:   *cluster.Name,
+		api.EksctlVersionTag: version.GetVersion(),
+	}
```
Callisto13, line 285: since users are not providing these, can we just set an array of `*awsiam.Tag` here and pass it through, rather than creating a `map[string]string` and then looping to translate in `pkg/iam/oidc/api.go`?

cPu1: I wouldn't be too keen to do that, `tags` are best represented as a `map[string]string` rather than a `[]*awsiam.Tag`, and that `Tag` type is specifically made for the AWS IAM APIs. It also breaks the abstraction.

pkg/iam/oidc/api_test.go
cPu1: Yup, I added one.

Callisto13 (approved):

PR #3914, opened 2021-06-29
`Update` command can update multiple nodegroups

### LINKED ISSUE #3857: Update nodegroup command accepts config file with multiple nodegroups
Currently the `update nodegroup` command has validation for accepting only 1 nodegroup in a config file at a time. However, it would be nice to be able to iterate over multiple nodegroups, since a user might just pass the same config file that was used to create a cluster, which may have multiple nodegroups.

For more info, see discussion in #3828 with this comment:

> I think so yes. People are likely to have their cluster config stored somewhere so it feels weird that we would require them to remove bits to run this command. With the config above I would have to comment-run-comment 3 times which does not seem fun.

_Originally posted by @Callisto13 in https://github.com/weaveworks/eksctl/pull/3828#discussion_r651672224_

Creating a separate issue for this here so that the other work in #3828 can be merged as soon as possible.

nikimanoledaki:
### Description

Closes #3857 

A config file containing multiple managed nodegroups will iterate over them and update all of them.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/nodegroup/update.go
+++ b/pkg/actions/nodegroup/update.go
@@ -6,7 +6,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/kris-nova/logger"
-	"github.com/pkg/errors"
 
 	"github.com/weaveworks/eksctl/pkg/managed"
 )
@@ -30,10 +29,6 @@ func (m *Manager) Update() error {
 
 		if ng.UpdateConfig == nil {
 			return fmt.Errorf("the submitted config didn't contain changes for nodegroup %s", ng.Name)
-		}
-
-		if describeNodegroupOutput.Nodegroup.UpdateConfig == nil {
-			return errors.New("cannot update updateConfig because the nodegroup is not configured to use one")
 		}
 
 		logger.Info("updating nodegroup %s's UpdateConfig", ng.Name)
--- a/pkg/actions/nodegroup/update.go
+++ b/pkg/actions/nodegroup/update.go
@@ -12,47 +12,51 @@ import (
 )
 
 func (m *Manager) Update() error {
-	ng := m.cfg.ManagedNodeGroups[0]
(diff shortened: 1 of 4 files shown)


nikimanoledaki:
Waiting for #3914 to merge to test it manually! :)

Callisto13:
Does it make sense to have the integration test update 2 ngs?

nikimanoledaki:
> Does it make sense to have the integration test update 2 ngs?

@Callisto13 I don't know 🤔 I was looking at other commands that take multiple nodegroups such as `create nodegroups`  and I didn't find an integration test for that. I'm inclined to say no because it would make the tests longer for a very small piece of functionality (essentially a couple of for loops) but also it would be easy to make this change. wdyt?

Callisto13:
yeh I think you are right, no need 👍

pkg/actions/nodegroup/update.go
```diff
@@ -12,47 +12,51 @@ import (
...
-	ng := m.cfg.ManagedNodeGroups[0]
-	describeNodegroupOutput, err := m.ctl.Provider.EKS().DescribeNodegroup(&eks.DescribeNodegroupInput{
-		ClusterName:   &m.cfg.Metadata.Name,
-		NodegroupName: &ng.Name,
-	})
-
-	if err != nil {
-		if managed.IsNotFound(err) {
-			return fmt.Errorf("could not find managed nodegroup with name %q", ng.Name)
+	for _, ng := range m.cfg.ManagedNodeGroups {
```
Callisto13, line 15: LGTM, just one small nit which is more of a personal preference:

can we move the contents of this `for` into another private func? simply because the process is quite long so it is easy to lose sight that this is all in a loop 🤷‍♀️

nikimanoledaki: Good idea! I split the contents into `updateNodegroup` and the strictly `updateConfig` bit into `updateUpdateConfig` (I know, cringe)

Callisto13: 😂  excellent

Callisto13 (approved):

PR #3927, opened 2021-07-02
Prepare for next development iteration

weaveworksbot:


Callisto13:
don't merge this yet, i might need it for a thing

aclevername:
> don't merge this yet, i might need it for a thing

this shouldn't prevent you from doing an `rc.1`? Or at least I don't think it should...

Callisto13:
yeh i don't think it should either... but i am not 100% sure so there is no harm leaving it 🤷‍♀️

Callisto13 (approved):

Issue #3929, opened 2021-07-04
Regression in 0.55.0: `--spot is only valid with managed nodegroups`


nbusseneau:
Hi,

We use `eksctl` on [`cilium/cilium-cli`](https://github.com/cilium/cilium-cli) for deploying an EKS cluster and running tests on it via GitHub Actions ([link](https://github.com/cilium/cilium-cli/blob/8d0139bb8af6b885da71c4c1d0e14af41f901274/.github/workflows/eks.yaml)).

First, we download the `latest` release:
```yaml
      - name: Install eksctl CLI
        run: |
          curl -LO "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_$(uname -s)_amd64.tar.gz"
          sudo tar xzvfC eksctl_$(uname -s)_amd64.tar.gz /usr/bin
          rm eksctl_$(uname -s)_amd64.tar.gz
```

Then, we deploy a cluster without a nodegroup:
```yaml
      - name: Create EKS cluster without nodegroup
        run: |
          eksctl create cluster \
            --name ${{ env.clusterName }} \
            --tags "usage=${{ github.repository_owner }}-${{ github.event.repository.name }},owner=${{ steps.vars.outputs.owner }}" \
            --without-nodegroup
```

And finally, we setup a spot managed nodegroup:
```yaml
      - name: Add managed spot nodegroup
        run: |
          eksctl create nodegroup \
            --cluster ${{ env.clusterName }} \
            --nodes 2 \
            --instance-types "t3.medium,t3a.medium" \
            --node-volume-type gp3 \
            --node-volume-size 10 \
            --managed \
            --spot \
            --node-private-networking
```

Looking at [the list of scheduled runs](https://github.com/cilium/cilium-cli/actions/workflows/eks.yaml?query=event%3Aschedule) (running every 6 hours) we can see that the [last working run was on Jul 2 around 9 CEST](https://github.com/cilium/cilium-cli/actions/runs/992727965), with [the next run around 15 CEST failing](https://github.com/cilium/cilium-cli/actions/runs/993695438) at the managed spot nodegroup creation step, along with all subsequent runs:
```
Error: couldn't create node group filter from command line options: --spot is only valid with managed nodegroups (--managed)
```

Nothing changed in how we setup EKS clusters between the last working and first failing runs mentioned above.

[...]

PR #3930, opened 2021-07-05
Do not fail on valid flag combinations for managed ngs

Callisto13:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/3929

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/cmdutils/configfile.go
+++ b/pkg/ctl/cmdutils/configfile.go
@@ -440,8 +440,11 @@ func makeManagedNodegroup(nodeGroup *api.NodeGroup, options CreateManagedNGOptio
 }
 
 func validateManagedNGFlags(cmd *cobra.Command, managed bool) error {
-	flagsValidOnlyWithMNG := []string{"spot", "instance-types"}
+	if managed {
+		return nil
+	}
 
+	flagsValidOnlyWithMNG := []string{"spot", "instance-types"}
 	if flagName, found := findChangedFlag(cmd, flagsValidOnlyWithMNG); found {
 		return errors.Errorf("--%s is only valid with managed nodegroups (--managed)", flagName)
 	}
--- a/pkg/ctl/create/nodegroup_test.go
+++ b/pkg/ctl/create/nodegroup_test.go
@@ -57,23 +57,31 @@ var _ = Describe("create nodegroup", func() {
 
 		DescribeTable("invalid flags or arguments",
 			func(c invalidParamsCase) {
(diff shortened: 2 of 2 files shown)


pkg/ctl/cmdutils/configfile.go
```diff
@@ -440,8 +440,11 @@ func makeManagedNodegroup(nodeGroup *api.NodeGroup, options CreateManagedNGOptio
 }
 
 func validateManagedNGFlags(cmd *cobra.Command, managed bool) error {
-	flagsValidOnlyWithMNG := []string{"spot", "instance-types"}
+	if managed {
+		return nil
+	}
```
aclevername, line 445: why is the func call `validateManagedFlags` and a bool is passed in to indicate its managed? Is it not called because its validating flags for managed nodes? :see_no_evil:

Callisto13: i think it was named to validate the use of managed flags (not managed _nodegroup_ flags), now it only checks for the use of managed flags on unmanaged ngs

it is confusing because technically it is named accurately, but we use "managed" as a shorthand for "managed nodegroups" so our minds fill in a gap which is not there

aclevername: ohhhh. Gotcha. Since for managed its always `if managed; return nil` would it make sense to move that logic outside of the func? Might make it less confusing

> i think it was named to validate the use of managed flags (not managed _nodegroup_ flags), now it only checks for the use of managed flags on unmanaged ngs

it does have NG in the func name though :wink:

aclevername (approved):
LGTM, 1 nit

aclevername:
oh lol automerge was enabled. smh

pkg/ctl/cmdutils/configfile.go
Callisto13: > would it make sense to move that logic outside of the func? Might make it less confusing

i did think about that, and decided to do the `if thing` in one place rather than more than 1, don't really care tho 🤷‍♀️ 

> it does have NG in the func name though

oh shit yeh, what was i reading 🤣

Callisto13:
you only have yourself to blame tbh

aclevername:
totally not the fault of the person who enabled it :shrug:

Callisto13:
stop approving unapprovable crap!

aclevername:
it was approvalable, my comments were just a nit

Callisto13:
hmmm i wouldn't have approved it :trollface:

PR #3951, opened 2021-07-06
Fix cfssl build issue after go mod tidy

Skarlso:
### Description

The fix is basically pinning to versions we already install. This is only working since go 1.16. Pinning to a module version makes the install module aware and ignore go mod files otherwise located in the project. So if that is not okay, the "easier" fix is changing install depts such as it does this:

```bash
# Install all other Go build requirements
while IFS= read -r req; do
+  go get "${req}"
  go install "${req}"
done < ${REQUIREMENTS_FILE}
```

Since go install does not install the missing things for the right version.

### TODO:
- [x] Do a `make -f Makefile.docker push-build-image` (can now just `make push-build-image`)
- [x] fix `go list` or even remove it. I have no idea what it is for

### Checklist
- [ ] Manually tested


aclevername (approved):
MY HERO :rocket:

PR #3956, opened 2021-07-07
Fix --wait issue for eksctl upgrade nodegroup command

### LINKED ISSUE #3889: --wait=false does not work when upgrading nodegroups
When using eksctl 0.54.0 to update a managed node group from 1.19 to 1.20, the --wait=false flag still causes the command to wait for CloudFormation.

When using the --timeout flag that will make it only wait for the time specified, but the exit code is 1, causing it to fail any scripts I use the command in.

This is the command I used:
`eksctl upgrade nodegroup --cluster cluster-name --name=nodegroup-name --wait=false --kubernetes-version=1.20`

vchetu:
Fixes #3889 

### Description

Currently --wait flag is ignored while performing nodegroup upgrade action. This PR checks for wait flag and accordingly performs the action.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/actions/nodegroup/upgrade.go
+++ b/pkg/actions/nodegroup/upgrade.go
@@ -134,4 +134,3 @@ func (m *Manager) waitForUpgrade(options managed.UpgradeOptions) error {
 	logger.Info("nodegroup successfully upgraded")
 	return nil
 }
-
--- a/pkg/actions/nodegroup/upgrade.go
+++ b/pkg/actions/nodegroup/upgrade.go
@@ -42,6 +42,8 @@ func (m *Manager) Upgrade(options managed.UpgradeOptions, wait bool) error {
 		return m.waitForUpgrade(options)
 	}
 
+	logger.Info("nodegroup upgrade request submitted successfully")
+
 	return nil
 
 }
--- a/pkg/actions/nodegroup/upgrade.go
+++ b/pkg/actions/nodegroup/upgrade.go
@@ -16,7 +16,7 @@ import (
 	"github.com/weaveworks/eksctl/pkg/utils/waiters"
 )
 
-func (m *Manager) Upgrade(options managed.UpgradeOptions) error {
+func (m *Manager) Upgrade(options managed.UpgradeOptions, wait bool) error {
 	stackCollection := manager.NewStackCollection(m.ctl.Provider, m.cfg)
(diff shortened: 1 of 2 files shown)


nikimanoledaki:
Thank you for opening a PR @vchetu, a member of the team will review shortly!

pkg/actions/nodegroup/upgrade.go
```diff
@@ -34,10 +34,10 @@ func (m *Manager) Upgrade(options managed.UpgradeOptions) error {
 		return managedService.UpgradeNodeGroup(options)
 	}
 
-	return m.upgradeAndWait(options)
+	return m.upgradeAndWait(options, wait)
```
cPu1, line 37: A cleaner way would be to split `upgradeAndWait` into two functions `upgrade` and `waitForUpgrade`, and call the latter one only if `wait` is set:

```suggestion
	if err := m.upgrade(options); err != nil {
		return err
	}
	if wait {
		return m.waitForUpgrade(options)
	}
	return nil
```

cPu1 (changes requested):

pkg/actions/nodegroup/upgrade.go
vchetu: Thanks @cPu1. Modified the code accordingly

pkg/actions/nodegroup/upgrade.go
```diff
@@ -34,10 +34,19 @@ func (m *Manager) Upgrade(options managed.UpgradeOptions) error {
...
-	return m.upgradeAndWait(options)
+	if err := m.upgrade(options); err != nil {
+		return err
+	}
+
+	if wait {
+		return m.waitForUpgrade(options)
+	}
+
+	return nil
```
cPu1, line 45: We should also log that the request has been submitted if `wait` is not set:

```suggestion
	logger.Info("nodegroup upgrade request submitted successfully")
	return nil
```

cPu1:
Thanks for the PR, @vchetu. It needs just one more change and it's good to merge.

pkg/actions/nodegroup/upgrade.go
vchetu: @cPu1 Added the log info.

cPu1 (approved):

Issue #2971, opened 2020-12-19
Could you please expose EksAllAccess and IamLimitedAccess policies as individually accessible source files


amcginlay:
I would like this feature so we can curl the latest version of these documents, apply the <account_id> replacements and call `aws iam create-policy` action from a simple script.

Currently it would appear that these documents are only available as sub-portions of the markdown here:
https://raw.githubusercontent.com/weaveworks/eksctl/master/userdocs/src/usage/minimum-iam-policies.md

Thanks in advance
Al

Callisto13:
Thanks for this request @amcginlay 👍 

We are happy to accept PRs if you would like to work on this 😄  otherwise the team will get to this after the holidays.

orsenthil:
Hi @amcginlay , @Callisto13 - This PR https://github.com/weaveworks/eksctl/pull/4001 might help address this issue.

Callisto13:
Closing this as investigation shows it is not feasible in a useful/clean way: https://github.com/weaveworks/eksctl/pull/4001#issuecomment-884977352

Issue #4000, opened 2021-07-21
Document Minimal required IAM Policies for EKS Cluster Creation Upfront


orsenthil:
**What were you trying to accomplish?**

I tried to follow the README.md to create a EKS cluster and I was constantly running into lack of permissions starting with 

> arn:aws:iam::account-id:user/senthil is not authorized to perform: ssm:GetParameter on resource:

And more such permissions issues.  

I realized that [Minimum IAM Policies](https://eksctl.io/usage/minimum-iam-policies/) were documented recently https://github.com/weaveworks/eksctl/pull/2267 but it should be mentioned upfront in the README.md document, so that instructions can be followed without any assumptions or troubleshooting.

**What happened?**

Unable to create a EKS Cluster, ran into permission issues, like

> arn:aws:iam::account-id:user/senthil is not authorized to perform: ssm:GetParameter on resource:

**How to reproduce it?**

1. Create a new Amazon AWS Account.
2. Create a IAM User.
3. Try to follow README.md instructions to create the EKS Cluster.
4. 
**Logs**

**Anything else we need to know?**

**Versions**

```
[~/git/eksctl]$ eksctl info
eksctl version: 0.56.0
kubectl version: v1.21.2
OS: darwin
```

nikimanoledaki:
Hi @orsenthil, closing this issue since it has been addressed by #4001. Thank you for contributing! 🙌

PR #4001, opened 2021-07-21
Document Minimal required IAM Policies for EKS Cluster Creation Upfront

orsenthil:
Fixes

* https://github.com/weaveworks/eksctl/issues/4000
* https://github.com/weaveworks/eksctl/issues/2971

### Checklist
- [x] Manually tested


Callisto13:
Thanks @orsenthil !

I don't think it quite fixes #2971, since the OP there would like to be able to `curl` a file with the complete policies, but the additional documentation is very welcome 😄

Callisto13 (approved):

orsenthil:
Thanks for merging my contribution, @Callisto13 . 🎉 

> since the OP there would like to be able to curl a file with the complete policies

I notice two details with this.

1. A person will have to change the account-id in that json document.
2. The size of the json file may be large for IAM Inline json, and AWS will either ask to break those into multiple jsons or request size limit with the json size. I tried and encountered this.

Both seem not helpful to me. What are your thoughts?

Callisto13:
Thanks @orsenthil 👍 I had not yet looked into the implementation of that request, and what you have said makes sense. I think we can therefore close that issue as something we cannot/would not be sensible to do.

Thanks for your help on this 🎉

Issue #3979, opened 2021-07-15
Expose support for containerd


Callisto13:
**Description**

From \<date tbc\>, AL2 AMIs will support bootstrapping nodes with containerd as the runtime. See: https://github.com/awslabs/amazon-eks-ami/pull/698.

Dockerd will still be the default. Users will not notice the change on our side if they wish to leave things as there are.
I have verified that the new behaviour does not break either the legacy or the new bootstrapping codepath in eksctl.

K8s versions `1.15` onwards will be supported.

AFAIK, this change will only be in effect in AL2 AMIs. So this will not include:
- Ubuntu AMIs (these are built by Canonical, so no timeline here)
- Any GPU AMIs (I might be wrong about this, so test it out)

**Implementation notes**
This will require a new field. Something along the lines of:

```yaml
nodeGroups:
- name: foo
  containerRuntime: containerd # or cri, runtime, etc
```

It should default to `dockerd`. Alternatively we could have a `bool` `useContainerd` field, but I like that less.

_(Sidebar: Ideally we would have a higher-level `bootstrap` field or something, to which things like `runtime` are subfields. We could move existing things under there, but that would ofc be breaking. Our config was not built up with much forethought, so now there are lots of unfortunately named fields everywhere (see the `volumeX` lot) mixed in which some which had a little more planning. Something to think about when we reshuffle for v1? Unless anyone thinks we can go for something tidier with this now 🤷‍♀️ . Actually maybe I don't care because this flag will be deprecated not long after we add it.)_

Validation will need to ensure that the flag is only set (or read) if the AMI family is AL2. We do not need to validate against the version (yet, see below).

The flag needs to be passed to the bootstrap script like so:

```bash
/etc/eks/bootstrap.sh <cluster-name> --other-flags etc --container-runtime containerd
```

Dockerd will be no longer be an option from 1.22 (current target, may change); from then containerd will become the default. This means that:
- we will then have to deprecate this option once the last min supported version which has this flag disappears (in this case, 1.21). It may be a good idea to mark this as deprecated right away, so that users are aware it is a temporary measure
- once 1.22 (or whatever it ends up being) is release we _will_ need to validate against version, to ensure the flag is only passed to the supported AMIs.

We should not add this for the legacy bootstrap path.

PR #4051, opened 2021-08-04
Expose nodegroup container runtime configuration

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/3979

### Checklist
- [x] Manually tested


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -1349,13 +1361,25 @@ type NodeGroupBase struct {
 	// This is a hack, will be removed shortly. When this is true for Ubuntu and
 	// AL2 images a legacy bootstrapper will be used.
 	CustomAMI bool `json:"-"`
+
+	// ContainerRuntime defines the runtime to use for the container.
```
Callisto13, line 1365: just the one container? 🙃 

```suggestion
	// ContainerRuntime defines the runtime (CRI) to use for containers on the node
```

Skarlso: Fair point

integration/tests/managed/managed_nodegroup_test.go
```diff
@@ -549,6 +550,31 @@ var _ = Describe("(Integration) Create Managed Nodegroups", func() {
 			})
 		})
 
+		Context("and creating a nodegroup with containerd runtime", func() {
```
Callisto13, line 553: the code you have added is for unmanaged nodegroups, this is a managed test.

we also need to do the opposite for each:
- pass the flag to the managed bootstrap code
- have integration tests for unmanaged nodegroups

Skarlso: Ah, thanks

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -268,6 +269,18 @@ func validateNodeGroupBase(ng *NodeGroupBase, path string) error {
 		return fmt.Errorf("AMI Family %s is not supported - use one of: %s", ng.AMIFamily, strings.Join(supportedAMIFamilies(), ", "))
 	}
 
+	if ng.ContainerRuntime != nil {
+		if *ng.ContainerRuntime == ContainerRuntimeContainerD && ng.AMIFamily != NodeImageFamilyAmazonLinux2 {
+			// check if it's dockerd or containerd
+			return fmt.Errorf("%s as runtime is only support for AL2 ami family", ContainerRuntimeContainerD)
+		}
+		if *ng.ContainerRuntime != ContainerRuntimeDockerD {
+			if *ng.ContainerRuntime != ContainerRuntimeContainerD {
```
Callisto13, line 278: possible to one-line this? or did you not like the length?
```suggestion
		if *ng.ContainerRuntime != ContainerRuntimeDockerD && *ng.ContainerRuntime != ContainerRuntimeContainerD {
```

Skarlso: I wanted to do something else in there. But I think I eventually did it somewhere else. :D thanks for the spot! :D

Callisto13:
And don't forget the docs 😜  In which we should highlight that we are not exposing a flag for this thing, just config

pkg/nodebootstrap/ubuntu_test.go
```diff
@@ -49,7 +49,8 @@ var _ = Describe("Ubuntu User Data", func() {
 API_SERVER_URL=
 B64_CLUSTER_CA=
 NODE_LABELS=
-NODE_TAINTS=`, "\n")))
+NODE_TAINTS=
+CONTAINER_RUNTIME=`, "\n")))
```
Callisto13, line 53: this flag isn't for ubuntu, granted the var wont be used, but could it be confusing if a user comes across it? maybe have `makeBootstrapEnv` only add the var if al2? wdyt?

Skarlso: Ah, agreed. It is confusing.

pkg/nodebootstrap/al2_test.go
```diff
@@ -149,7 +167,8 @@ API_SERVER_URL=
 B64_CLUSTER_CA=
 NODE_LABELS=
 NODE_TAINTS=
-CLUSTER_DNS=172.16.0.10`,
+CLUSTER_DNS=172.16.0.10
+CONTAINER_RUNTIME=`,
```
Callisto13, line 171: i am surprised that this var is empty if we always have that config set on the node (containerd if asked for, dockerd the rest of the time) 🤔  am i reading wrong? did you do a test ng create without that conf option?

Skarlso: This test doesn't call defaults it sets it manually so defaults aren't applied to it. I think normally, the set defaults is always called, so I thought it's fine in this test. Do you think it would be better to ~hardcode it~ set it as default somewhere else?

Callisto13: aaaah yeh that makes all the sense. nope what you have is fine i think.

integration/tests/crud/creategetdelete_test.go
```diff
@@ -138,14 +145,14 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 			Expect(config.CurrentContext).To(ContainSubstring(params.Region))
 		})
 
-		Context("and listing clusters", func() {
+		XContext("and listing clusters", func() {
```
Callisto13, line 148: this looks promising 😂

Skarlso: Argh :D

Skarlso: I already started a full run, just came back to check it but committed the previous run's edit. :D

Skarlso: @Callisto13 

Whole suite (41 tests) passed except for one which failed because it timed out.

```
2021-08-06 07:13:11 [!]  retryable error (RequestError: send request failed
caused by: Post "https://cloudformation.us-west-2.amazonaws.com/": net/http: TLS handshake timeout) from cloudformation/ListStacks - will retry after delay of 43.44878ms

• Failure [60.009 seconds]
(Integration) Create, Get, Scale & Delete
/Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/crud/creategetdelete_test.go:59
  cluster with 1 node
  /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/crud/creategetdelete_test.go:126
    and listing clusters
    /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/crud/creategetdelete_test.go:148
      should return the previously created cluster [It]
      /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/crud/creategetdelete_test.go:149

      Timed out after 60.000s.
      Expected process to exit.  It did not.

      /Users/skarlso/goprojects/weaveworks/eksctl/integration/runner/runner.go:125

```

I wonder if we should do something about that. The current implementation doesn't really wait that long. Maybe we should increase the wait time on that if a retry-able error happens?

cw-sakamoto:
Thanks for the great PR!!

If setting `containerRuntime` to `containerd`, I think you don't need the following command, is that correct?
https://github.com/weaveworks/eksctl/blob/50659eb7c041f12e5a481225db40b7ae11662be7/pkg/nodebootstrap/bindata/assets/bootstrap.al2.sh#L24

Skarlso:
> Thanks for the great PR!!
> 
> If setting `containerRuntime` to `containerd`, I think you don't need the following command, is that correct?
> https://github.com/weaveworks/eksctl/blob/50659eb7c041f12e5a481225db40b7ae11662be7/pkg/nodebootstrap/bindata/assets/bootstrap.al2.sh#L24

Hi! :) 

Yes, normally, I would agree, however, there is an issue here https://github.com/awslabs/amazon-eks-ami/pull/698#discussion_r677378723 that some people reported that this is breaking CI and their flow in some way. I thought about leaving this in there in that case. But I'm not a 100% sure about that. @Callisto13 any thoughts on the comment in the issue?

cw-sakamoto:
@Skarlso 

Thanks for the quick reply!

If keep it, we need to change the path of `containerd`'s sock path below in docker service file.

```
$ cat /usr/lib/systemd/system/docker.service | grep ExecStart
ExecStartPre=/bin/mkdir -p /run/docker
ExecStartPre=/usr/libexec/docker/docker-setup-runtimes.sh
ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock $OPTIONS $DOCKER_STORAGE_OPTIONS $DOCKER_ADD_RUNTIMES
```

FYI: https://github.com/awslabs/amazon-eks-ami/blob/master/files/containerd-config.toml#L6

Callisto13:
@Skarlso I think I referred to this line in a comment somewhere (maybe in slack?). This restart was added when we would add/edit the docker daemon config, which I believe we no longer do. So it is likely some legacy stuff which I forgot to remove when I removed some bits which changed the cgroup driver.

So we can:
1. Try removing it completely, see if it is needed when `--container-runtime=dockerd`. (I feel it really should not be needed but good to check)
2. If I am wrong and it _is_ needed, then it will only been required if using `dockerd` in which can we can throw an `if` around it.

make sense?

pkg/nodebootstrap/bindata/assets/bootstrap.al2.sh
```diff
@@ -11,16 +11,15 @@ echo "eksctl: running /etc/eks/bootstrap"
...
+  --kubelet-extra-args "${KUBELET_EXTRA_ARGS}" \
+  --container-runtime "${CONTAINER_RUNTIME}"
 
 echo "eksctl: merging user options into kubelet-config.json"
 trap 'rm -f ${TMP_KUBE_CONF}' EXIT
 jq -s '.[0] * .[1]' "${KUBELET_CONFIG}" "${KUBELET_EXTRA_CONFIG}" > "${TMP_KUBE_CONF}"
 mv "${TMP_KUBE_CONF}" "${KUBELET_CONFIG}"
 
 systemctl daemon-reload
-echo "eksctl: restarting docker daemon"
```
Skarlso, line 22: Running a test with this to see if it's needed at all.

Skarlso: Container runtime is up:

```
Container runtime
docker://19.3.13
```

pkg/ctl/cmdutils/test_data/managed-nodegroup-with-container-runtime.yaml
```diff
@@ -0,0 +1,14 @@
+# A simple example of ClusterConfig object:
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: test-cluster-1
+  region: eu-north-1
+
+managedNodeGroups:
```
Callisto13, line 10: unmanaged right?

Skarlso: Nope, this is supposed to fail. It's testing that you *cannot* do this.

userdocs/src/usage/container-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+    instanceType: m5.xlarge
+    desiredCapacity: 10
+    privateNetworking: true
+    amiFamily: AmazonLinux2
+    containerRuntime: containerd
+  - name: ng-2-api
+    labels: { role: api }
+    instanceType: m5.2xlarge
+    desiredCapacity: 2
+    privateNetworking: true
```
Callisto13, line 29: we can keep keep inline examples as simple as possible

```suggestion
  - name: ng-1
    instanceType: m5.xlarge
    desiredCapacity: 2
    amiFamily: AmazonLinux2
    containerRuntime: containerd
```

userdocs/src/usage/container-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+This value is set to `dockerd` by default to preserve backwards compatibility, but will soon be
+deprecated.
+
+The container runtime value is then passed to the bootstrap.sh script by `eksctl` as follows:
+
+```bash
+/etc/eks/bootstrap.sh "${CLUSTER_NAME}" \
+  ...
+  --container-runtime "${CONTAINER_RUNTIME}"
+```
```
Callisto13, line 41: We don't need to share this with them 

```suggestion
```

userdocs/src/usage/container-runtime.md
```diff
@@ -0,0 +1,68 @@
...
+  - name: ng2
+    instanceType: m5.large
+    ami: ami-custom1234
+managedNodeGroups:
+  - name: m-ng-2
+    ami: ami-custom1234
+    instanceType: m5.large
+    overrideBootstrapCommand: |
+      #!/bin/bash
+      /etc/eks/bootstrap.sh <cluster-name> --container-runtime containerd
```
Callisto13, line 67: again only what they need to know for this one thing
```suggestion
managedNodeGroups:
  - name: m-ng-1
    instanceType: m5.large
    overrideBootstrapCommand: |
      #!/bin/bash
      /etc/eks/bootstrap.sh <cluster-name> <other flags> --container-runtime containerd
```

Skarlso: Kind of wanted to show that you can have these together as well. ¯\_(ツ)_/¯

pkg/ctl/cmdutils/test_data/managed-nodegroup-with-container-runtime.yaml
Callisto13: aaaah i thought this file was an example.yaml 🤦‍♀️  ignore me

userdocs/src/usage/container-runtime.md
Callisto13: at this point they already know how nodegroups + config work. we don't pull full examples for each little feature, otherwise the docs would be literally nothing but yaml. if you want to have a full example, create a new file in the `examples/` dir. but this is such a small feature it probably is not necessary

Skarlso: oh yeah, I agree, I'm just saying that's why I put it in there. :D

userdocs/src/usage/container-runtime.md
```diff
@@ -0,0 +1,46 @@
...
+  - name: ng-1
+    instanceType: m5.xlarge
+    desiredCapacity: 2
+    amiFamily: AmazonLinux2
+    containerRuntime: containerd
+```
+
+This value is set to `dockerd` by default to preserve backwards compatibility, but will soon be
+deprecated.
+
```
Callisto13, line 27:
```suggestion

_Note that there is no equivalent flag for setting the container runtime, this can only be done via a config file._
```

Callisto13 (approved):
LGTM!

PR #4065, opened 2021-08-09
Add preBootstrapCommands to userdata when using customAMI

### LINKED ISSUE #4032: Possibility to set preBootstrapCommands when using custom ami with managed nodegroups

When pinning AMI for a managed nodegroup, currently eksctl ignores preBootstrapCommands. It would nice to have preBoostrapCommands.

AWS [blog](https://aws.amazon.com/blogs/containers/introducing-launch-template-and-custom-ami-support-in-amazon-eks-managed-node-groups/) talks about the workflow with eksctl, but eksctl does not respect preBootstrapCommands when using custom AMI.

image

This will allow us to do additional configuration before bootstrapping.

kishoregv:
### Description

Will fix #4032 

### Checklist
- [x] Manually tested


Callisto13:
👍  thanks @kishoregv.

Can you make this a draft while it is still WIP? That way it will be easy for us to spot when it is ready for review 🙏

kishoregv:
> Can you make this a draft while it is still WIP? That way it will be easy for us to spot when it is ready for review 🙏
Thanks for the advice. Will do that in future.

The PR is now ready for review.

Thanks

Callisto13:
👍  could you add some testing to `pkg/nodebootstrap/managed_al2_test.go` please?

pkg/nodebootstrap/managed_al2.go
```diff
@@ -82,12 +81,29 @@ func (m *ManagedAL2) UserData() (string, error) {
...
-		return base64.StdEncoding.EncodeToString([]byte(*ng.OverrideBootstrapCommand)), nil
+		scripts = append(scripts, *ng.OverrideBootstrapCommand)
 	}
 
-	return "", nil
+	if len(scripts) == 0 {
+		return "", nil
+	}
+
+	if err := createMimeMessage(&buf, scripts, nil, mimeBoundary); err != nil {
```
kishoregv, line 102: @Callisto13 or Any one who is going to review this
Would it be better to use cloudConfig instead of MIME format?

cloudConfig would allow all the preBootstrapCommands to be run as single script file. Where as with MIME format each string in preBootstrapCommands is it's script file.

The benefit of allowing all the preBootstrapCommands to run as one script file is, any environment variables initialized earlier will be applicable to later commands.

Other alternative to make preBootstrapCommands as one script while using MIME format is, I can join the preBootstrapCommands with a newline char, like below

```

	if len(ng.PreBootstrapCommands) > 0 {
		scripts = append(scripts, ng.PreBootstrapCommands...)
	}

```

please share your thoughts.

Callisto13: That makes sense to me @kishoregv; we don't appear to add the Override commands in the MIME format when using custom AMIs, so if adding the Pre- commands as a single cloudconfig script works, then that should be fine.

I may be missing some context here, however; @cPu1 do you know if there is a good reason to use MIME in this case?

kishoregv: @Callisto13 If the PR looks good you folks, I am ok to merge the PR as is without any future changes that I was proposing above.

Thanks

Callisto13: 👍  sounds good

If you were able to manually verify that the current changes fix the bug, then I am happy with it; the priority is to get the bug fixed.

If you'd like to submit a follow-up pr to do the suggested change when we have more time to look into it, please do.

Himangini:
@kishoregv can you resolve the conflicts, please. We can then proceed to review and merge this PR :)

pkg/nodebootstrap/managed_al2.go
kishoregv: I have manually verified that the current changes fix the bug.

kishoregv:
> @kishoregv can you resolve the conflicts, please. We can then proceed to review and merge this PR :)

@himangini Done.👍

Callisto13 (approved):
LGTM!

PR #4072, opened 2021-08-10
Add to 61 rc

Callisto13:
Adding more to this branch to fix a bug introduced by the last rc. Bug fix not added to release notes since the initial implementation is in the notes, so that would be weird 🤷‍♀️ 

Merged head of `main` into release 0.61.

Decided to simply pull in main as the only new commits are related to docs, internal changes or fixing the bug that this rc is being created for.

When this is merged I will trigger an rc.1


pkg/version/release.go
```diff
@@ -3,7 +3,7 @@ package version
 // This file was generated by release_generate.go; DO NOT EDIT.
 
 // Version is the version number in semver format X.Y.Z
-var Version = "0.61.0"
+var Version = "0.62.0"
```
aclevername, line 6:
```suggestion
var Version = "0.61.0"
```

aclevername (changes requested):

aclevername (approved):

PR #4098, opened 2021-08-16
Fetch instance type from launch template in `eksctl get nodegroups`

aclevername:
### Description

Previously we showed `"-"` when a nodegroup has no instances types. This happens when the nodegroup is configured to use a launch template. Changes the code so now we query the launch template to fetch the instance type

### Checklist
- [ ] Manually tested


pkg/actions/nodegroup/get.go
```diff
@@ -78,11 +80,25 @@ func (m *Manager) GetAll() ([]*manager.NodeGroupSummary, error) {
...
 }
 
-func getInstanceTypes(ng *awseks.Nodegroup) string {
+func (m *Manager) getInstanceTypes(ng *awseks.Nodegroup) string {
 	if len(ng.InstanceTypes) > 0 {
 		return strings.Join(aws.StringValueSlice(ng.InstanceTypes), ",")
 	}
-	logger.Info("no instance types reported by EKS for nodegroup %q", *ng.NodegroupName)
+
+	if ng.LaunchTemplate != nil {
```
nikimanoledaki, line 88: this could be refactored to keep the happy path on the left and exit early, e.g.:
```
if ng.LaunchTemplate == nil {
  logger.Info("no instance type found for nodegroup %q", *ng.NodegroupName)
  return "-"
}
etc
```
this works too but it'll be clearer that way :)
wdyt?

pkg/actions/nodegroup/get_test.go
```diff
@@ -140,10 +179,123 @@ var _ = Describe("Get", func() {
 					Expect(ngSummary[0].Version).To(Equal("1.18"))
 				})
 			})
+
+			When("a nodegroup is is associated with a launch template", func() {
```
nikimanoledaki, line 183:
```suggestion
			When("a nodegroup is associated with a launch template", func() {
```

pkg/actions/nodegroup/get.go
aclevername: done!

nikimanoledaki (approved):
LGTM!

pkg/actions/nodegroup/get.go
```diff
@@ -151,6 +129,34 @@ func (m *Manager) Get(name string) (*manager.NodeGroupSummary, error) {
...
+		LaunchTemplateId: ng.LaunchTemplate.Id,
+	})
+
+	if err != nil {
+		return "-"
+	}
+
+	for _, template := range resp.LaunchTemplateVersions {
+		if strconv.Itoa(int(*template.VersionNumber)) == *ng.LaunchTemplate.Version {
+			return *template.LaunchTemplateData.InstanceType
```
nikimanoledaki, line 152: this would only return the first `InstanceType` when looping over `resp.LaunchTemplateVersions`, what if there are more? are we returning too early?

aclevername: Its not that its only the first return, but rather the lauch template version that matches the version is use.  Alaunch template has multiple versions. In this case we want to get the instance type from a given launch template & version. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-launch-templates.html#manage-launch-template-versions

pkg/actions/nodegroup/get.go
```diff
@@ -135,6 +129,34 @@ func (m *Manager) Get(name string) (*manager.NodeGroupSummary, error) {
...
+
+	if ng.LaunchTemplate == nil {
+		logger.Info("no instance type found for nodegroup %q", *ng.NodegroupName)
+		return "-"
+	}
+
+	resp, err := m.ctl.Provider.EC2().DescribeLaunchTemplateVersions(&ec2.DescribeLaunchTemplateVersionsInput{
+		LaunchTemplateId: ng.LaunchTemplate.Id,
+	})
+
```
Callisto13, line 145:
```suggestion
```
this blank line ruined my day but otherwise lgtm

Callisto13 (approved):

pkg/actions/nodegroup/get.go
nikimanoledaki: Right! That makes sense ☺️

PR #4105, opened 2021-08-18
Fix dry-run integration tests

cPu1:
### Checklist
- [ ] Manually tested

### DIFF
--- a/integration/tests/dry_run/dry_run_test.go
+++ b/integration/tests/dry_run/dry_run_test.go
@@ -86,7 +86,6 @@ nodeGroups:
     withShared: true
   ssh:
     allow: false
-    enableSsm: false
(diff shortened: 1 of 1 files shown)


aclevername:
These tests are a constant pain :confused: is there anything we can do to detect failures quicker? It feels like this is the set of tests that fail the most often

aclevername (approved):
see above comment

Himangini (approved):

cPu1:
> These tests are a constant pain  is there anything we can do to detect failures quicker? It feels like this is the set of tests that fail the most often

That's true, we can't always remember to update this when the schema changes. But we cannot run this as part of the unit tests  because it needs credentials. We'll have to think of a better way to catch these schema changes early.

PR #4124, opened 2021-08-23
Fix markdown issue in "Managing nodegroups" user docs

EHJ-52n:
I just fixed a markdown issue on the "Managing nodegroups" topic.

See https://eksctl.io/usage/managing-nodegroups/#creating-a-nodegroup-from-a-config-file


nikimanoledaki (approved):
Thanks @EHJ-52n, good catch! 🎉

Issue #4154, opened 2021-08-31
kubernetes.NewList fails if an entire item is a comment or whitespace in a YAML file


cPu1:
Given this example [pkg/kubernetes/testdata/misc-sample-multidoc-empty-lists-1.yaml](https://github.com/weaveworks/eksctl/blob/main/pkg/kubernetes/testdata/misc-sample-multidoc-empty-lists-1.yaml), the tests fail if a comment or whitespace is added at the top before the next item starts. `kubectl` handles these cases well. This isn't an issue if the YAML is controlled by eksctl but it might be worth handling this for YAMLs obtained from external sources that may have comments or whitespaces as an entire item.

PR #4159, opened 2021-09-01
Add func for lax parsing of YAML data

cPu1:
Closes https://github.com/weaveworks/eksctl/issues/4154

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/kubernetes/manifests.go
+++ b/pkg/kubernetes/manifests.go
@@ -35,17 +35,8 @@ func NewRawExtensions(manifest []byte) ([]runtime.RawExtension, error) {
 	return objects, nil
 }
 
-// NewList decoded data into a list of Kubernetes resources
+// NewList decodes data into a list of Kubernetes resources, ignoring any whitespace and comment nodes
 func NewList(data []byte) (*metav1.List, error) {
-	return parseList(data, false)
-}
-
-// ParseListLax is like NewList but ignores comment and whitespace nodes
-func ParseListLax(data []byte) (*metav1.List, error) {
-	return parseList(data, true)
-}
-
-func parseList(data []byte, lax bool) (*metav1.List, error) {
 	var list metav1.List
 	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewBuffer(data), 4096)
 
@@ -58,11 +49,9 @@ func parseList(data []byte, lax bool) (*metav1.List, error) {
 			return nil, err
 		}
 
-		if lax {
-			r.Raw = bytes.TrimSpace(r.Raw)
-			if len(r.Raw) == 0 || bytes.Equal(r.Raw, []byte("null")) {
-				continue
-			}
+		r.Raw = bytes.TrimSpace(r.Raw)
+		if len(r.Raw) == 0 || bytes.Equal(r.Raw, []byte("null")) {
+			continue
 		}
 
 		if err := AppendFlattened(&list, r); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/manifests.go
```diff
@@ -37,19 +37,35 @@ func NewRawExtensions(manifest []byte) ([]runtime.RawExtension, error) {
...
 // NewList decoded data into a list of Kubernetes resources
 func NewList(data []byte) (*metav1.List, error) {
-	list := metav1.List{}
+	return parseList(data, false)
+}
+
+// ParseListLax is like NewList but ignores comment and whitespace nodes
+func ParseListLax(data []byte) (*metav1.List, error) {
+	return parseList(data, true)
+}
```
aclevername, line 46: how do we know when to call this vs `NewList`? Shouldn't we just always ignore comments in every yaml resource we parse?

cPu1: We'd want to use `ParseListLax` for YAMLs obtained from external sources. 

>Shouldn't we just always ignore comments in every yaml resource we parse?

Probably. I kept the implementation of NewList unchanged because I didn't want to change the existing behaviour for its callers.

aclevername: > We'd want to use `ParseListLax` for YAMLs obtained from external sources.

shall we change the name of the function to represent that then? Wouldn't really of known otherwise :see_no_evil: 

> > Shouldn't we just always ignore comments in every yaml resource we parse?
> 
> Probably. I kept the implementation of NewList unchanged because I didn't want to change the existing behaviour for its callers.

We could just make it the default behaviour :shrug: don't see any harm

cPu1: >shall we change the name of the function to represent that then? Wouldn't really of known otherwise see_no_evil

What are you thinking of?

>We could just make it the default behaviour shrug don't see any harm

I was worried about breaking existing code as NewList is being used in a few places, but I'll go ahead with making it the default behaviour and run the integration tests on this branch before merging.

aclevername: > > shall we change the name of the function to represent that then? Wouldn't really of known otherwise see_no_evil
> 
> What are you thinking of?
> > We could just make it the default behaviour shrug don't see any harm
> 
> I was worried about breaking existing code as NewList is being used in a few places, but I'll go ahead with making it the default behaviour and run the integration tests on this branch before merging.

If its going to be the default behaviour we don't need to rename it :smile: but if we do end up keeping them separate somethings more descriptive e.g. `ParseKubernetesResourcesWithComments`. I just find `ParseListLax` means nothing to me :joy:

cPu1: >If its going to be the default behaviour we don't need to rename it

Yup, we don't. 

>somethings more descriptive e.g. ParseKubernetesResourcesWithComments

How about `ParseResourcesIgnoreComments`?

aclevername: sounds good, although hopefully it will just be the default behaviour :crossed_fingers:

cPu1:
I have made lax YAML parsing the default behaviour and triggered the integration tests.

cPu1:
Tests for `integration/scripts/download-previous-release.sh` and EKS Anywhere failed but they are not related to this change.

aclevername:
> Tests for `integration/scripts/download-previous-release.sh` and EKS Anywhere failed but they are not related to this change.

fixed on main

aclevername (approved):

Issue #4158, opened 2021-09-01
Add support for wellKnownPolicies when defining addons


aclevername:
**What feature/behavior/change do you want?**
When creating an IAMSerivceAccount you can reference `wellKnownPolicies`, e.g.:
```yaml
iam:
  withOIDC: true
  serviceAccounts:
  - metadata:
      name: aws-load-balancer-controller
      namespace: kube-system
    wellKnownPolicies:
      awsLoadBalancerController: true
```

but you can't do this for addons, depsite it creating an equiavlent IAMServiceAccount under the hood.

## Example

The following config file:
```yaml
addons:
- name: some-new-addon
  wellKnownPolicies:
    awsLoadBalancerController: true
```

Should result in the policies for `awsLoadBalancerController` getting added to the addons IAMServiceAccount. It should be equivalent to:

```yaml
addons:
- name: some-new-addon
  attachPolicy: # inline policy can be defined along with `attachPolicyARNs`
    Version: "2012-10-17"
    Statement:
    - Effect: Allow
      Action:
      - "awsLoadBalancerController-actions"
      ...
      Resource: '*'
      ....
```

**Why do you want this feature?**
As more addons get introduced some of the existing wellKnownPolicies will be applicable, being able to use this is convienant

Callisto13:
Just to clarify, these `wellKnownPolicies` will be extra things that the user wants the addon to have access to, not policies that the addon actually needs to be able to work?

aclevername:
> Just to clarify, these `wellKnownPolicies` will be extra things that the user wants the addon to have access to, not policies that the addon actually needs to be able to work?

Well it can be either. For example when EBS CSI Driver addon gets introduced the user can select the well known policy for that #4188

Callisto13:
why would we not just automatically attach the required policies? i know things like that are coming up, but why expose this?

Skarlso:
The name, `wellKnown` was misleading. I literally thought it's an AWS policy feature I somehow missed. :)

Turns out it's just common policies people attach to clusters. I really would prefer this to be called `commonPolicies` or `commonClusterPolicies` or even `mostUsedPolicies` instead. :) But it's not a blocker.

Why? Because `wellKnown` is a connotation that brings baggage so to speak. It's a loaded phrase with a different historical meaning to it that made me think it's an embedded feature.

aclevername:
> why would we not just automatically attach the required policies? i know things like that are coming up, but why expose this?

I think theres a few reasons:
[...]

PR #4161, opened 2021-09-02
Add support for specifying wellKnownPolicies when using addons

aclevername:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4158

### Checklist
- [ ] Manually tested


pkg/actions/addon/create.go
```diff
@@ -221,18 +218,19 @@ func (a *Manager) createRoleUsingAttachPolicyARNs(addon *api.Addon, namespace, s
...
+		resourceSet = builder.NewIAMRoleResourceSetWithAttachPolicyARNs(addon.Name, namespace, serviceAccount, addon.PermissionsBoundary, addon.AttachPolicyARNs, a.oidcManager)
+	} else if addon.WellKnownPolicies.HasPolicy() {
+		logger.Info("creating role using provided well known policies")
+		resourceSet = builder.NewIAMRoleResourceSetWithWellKnownPolicies(addon.Name, namespace, serviceAccount, addon.PermissionsBoundary, addon.WellKnownPolicies, a.oidcManager)
+	} else {
+		logger.Info("creating role using provided policies")
+		resourceSet = builder.NewIAMRoleResourceSetWithAttachPolicy(addon.Name, namespace, serviceAccount, addon.PermissionsBoundary, addon.AttachPolicy, a.oidcManager)
 	}
-	return resourceSet.OutputRole, nil
+	return resourceSet, resourceSet.AddAllResources()
```
Skarlso, line 233: This is so weird. It returns the resource, but then also returns the output of calling AddAllResources on the created resource. The cognitive load of this "simple" function is pretty high. I would prefer this to be separated and the caller has the responsibility to further add all resources if necessary. The `createroleResourceSet` doesn't suggest that this will also add all of them. It suggest construction.

Skarlso (approved):
One question / note. Otherwise the generated function and logic look okay to me. I'm not sure though if anything might be missing from somewhere. In my limited understanding of the current code structure and focal points of `eksctl` this looks okay. :)

pkg/actions/addon/create.go
aclevername: I think whats weird is how the resourceSets work. You use a constructor to "build" it, but its only actually built/useful after you've called `AddAllResources`. I could move this call further up, but its just moving the error handling around even more :shrug:

Skarlso: Ah, well that's annoying. :D Fine, leave it as is :)

Callisto13 (approved):

Issue #2510, opened 2020-08-06
Cluster creation fails if VPC Endpoint Route already exists in the RouteTable that is associated with a Subnet


michaelbeaumont:
> Notice you're getting a completely different error with 0.24 as in your first comment.
> First error on master:
>
> AWS::EC2::VPCEndpoint/VPCEndpointS3: CREATE_FAILED – "route table rtb-09ec7cd1e8effc6cf already has a route with destination-prefix-list-id pl-68a54001 (Service: AmazonEC2; Status Code: 400; Error Code: RouteAlreadyExists; Request ID: 9815caa0-c444-46fc-9116-624b749e477a)"

I could replicate the above error, too. I validated it on master.

This error occurs when the S3 VPC Endpoint Route already exists in the RouteTable that is associated with the Subnet.
My fix was not for this error.

_Originally posted by @hiraken-w in https://github.com/weaveworks/eksctl/issues/2473#issuecomment-669691737_

lkr2des:
Hi

For me it still continues to give error about s3 endpoint being route table creation failure.
Our VPC already has all the necessary vpc endpoints created using separate template and not expecting eksctl to create it.

Here's my cluster yaml
I tried with eksctl version 26, 27, 28 same error.

apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: ml-dev-eks-cluster
  region: us-west-2
  version: '1.17'

privateCluster:
 enabled: true

vpc:
  id: "vpc-38c13640"
  cidr: "172.20.0.0/18"
  subnets:
    private:
      us-west-2a:
          id: "subnet-0d7b1446"
          cidr: "172.20.2.0/24"
      us-west-2b:
          id: "subnet-b4cf8dcd"
          cidr: "172.20.18.0/24"

[ℹ]  eksctl version 0.28.0-rc.0
[ℹ]  using region us-west-2
[✔]  using existing VPC (vpc-38c13640) and subnets (private:[subnet-0d7b1446 subnet-b4cf8dcd] public:[])
[!]  custom VPC/subnets will be used; if resulting cluster doesn't function as expected, make sure to review the configuration of VPC/subnets
[ℹ]  nodegroup "m5a-xlarge-ng-1" will use "ami-064270ad2910f51a2" [AmazonLinux2/1.17]
[ℹ]  using EC2 key pair "dev-keypair"
[ℹ]  using Kubernetes version 1.17
[...]

PR #4164, opened 2021-09-02
Fix duplicate destination prefix list id in routing

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/2510

### Checklist
- [ ] Manually tested


pkg/cfn/builder/vpc_endpoint.go
```diff
@@ -97,18 +101,43 @@ func (e *VPCEndpointResourceSet) subnetsForAZs(azs []string) []*gfnt.Value {
...
-			m[routeTableStr] = true
+		if !routeTables[routeTableStr] {
+			routeTables[routeTableStr] = true
+		} else {
+			continue
+		}
+
+		destinationExists := false
+		output, err := e.ec2API.DescribeRouteTables(&ec2.DescribeRouteTablesInput{
+			RouteTableIds: aws.StringSlice([]string{subnet.RouteTable.String()}),
```
Skarlso, line 119: Uh, this has one weird case where this is 👍 

```
RouteTableIds: ["{map[Ref:PrivateRouteTableUSWEST2A]}"]
```

That will def not work. :/

Skarlso: I probably will have to change all of this to properly pass in the routing table IDs because that's not the routing table ID. :sadpanda:

Skarlso: Hmm, these should be IDs though.

```go
subnetRoutes[*rta.SubnetId] = *rt.RouteTableId
```

So the test must be wrong somewhere.

pkg/cfn/builder/vpc_endpoint_test.go
```diff
@@ -29,6 +31,107 @@ type vpcResourceSetCase struct {
...
+				resourceJSON, err := rs.template.JSON()
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedJSON, err := ioutil.ReadFile("testdata/custom_vpc_private_endpoint_same_destination_prefix_list.json")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(resourceJSON).To(MatchJSON(expectedJSON))
+			})
+		})
+
+		When("same destinations prefix list ids are in the routing tables", func() {
```
Skarlso, line 128: Continue this on Monday. Going to extract the setup into a BeforeEach.

Issue #4052, opened 2021-08-04
When using MFA enabled IAM users, eksctl is not caching MFA token and is requesting for a new token for every command execution.


skkumsi:
**What happened?**

When AWS CLI is configured with MFA enabled IAM users, for every eksctl command execution a prompt for MFA token is being given out. The 'eksctl' client is not caching the token unlike AWS CLI where the MFA token credentials are cached under ~/.aws/cli/cache directory.

**How to reproduce it?**

file: ~/.aws/config

>[profile tester]
output = json
region = us-east-1
mfa_serial = arn:aws:iam::36xxxxxxx:mfa/tester
role_arn = arn:aws:iam::36xxxxxxx:role/awsqs-eks-cluster
source_profile = tester

In the above CLI config the MFA enabled IAM user 'tester' is assuming an IAM Role 'awsqs-eks-cluster'

when 'aws sts get-caller-identity' is executed for the first time, the MFA token prompt is being requested. Upon providing a valid MFA token, the command returns the following output.
```
$ aws sts get-caller-identity
Enter MFA code for arn:aws:iam::36xxxxxxx:mfa/tester:
{
    "UserId": "AROAVKP2SENF2BUEXAMPLE:botocore-session-1628114328",
    "Account": "36xxxxxxx",
    "Arn": "arn:aws:sts::36xxxxxxx:assumed-role/awsqs-eks-cluster/botocore-session-1628114328"
}
```
```
$ ls -l ~/.aws/cli/cache/
total 8
-rw-------  1 admin  staff  1410 Aug  4 17:59 6b8b3ff506784382ca184e327173d67cdb3bf574.json
```
For all consecutive AWS CLI commands, the cached credentials are used and users needs not enter MFA token again. If the above mentioned credentials.json file is deleted, AWS CLI also prompts for a MFA token again.

But in case of 'eksctl' client for every execution MFA token prompt is being given out. The 'eksctl' client is not using the cached credentials in ~/.aws/cli/cache/ , neither storing them after first execution elsewhere. 
```
$ eksctl get clusters --name eks-cluster-name
Assume Role MFA token code: 525923
2021-08-04 18:07:38 [i]  eksctl version 0.59.0
2021-08-04 18:07:38 [i]  using region us-east-1
NAME			VERSION	STATUS	CREATED			VPC			SUBNETS							SECURITYGROUPS
[...]

PR #4194, opened 2021-09-08
Add token cache for MFA users and credentials

Skarlso:
### Description

_NOTE_: Going to re-write this whole thing. We don't need a lock that is this complex. We don't handle multiple creds, only the current one for example. And the mocking can be done a lot cleaner I think.

_NOTE2_: Significantly simplified the caching. We don't need locks as eksctl is not expected to be called concurrently. And we don't need role / cluster caching. We only cache per profile. I removed all the mocks, because I want to test with temporary filesystem rather than os mocks.

Closes https://github.com/weaveworks/eksctl/issues/4052

### TODO:

- [x] Switch to turn caching on. This shouldn't be the default so existing functionality doesn't change.
- [x] Test the new cacher.

### Checklist
- [x] Manually tested


pkg/eks/filecache/filecache_test.go
```diff
@@ -0,0 +1,509 @@
...
+}
+
+func TestNewFileCacheProvider_Unlockable(t *testing.T) {
+	c := credentials.NewCredentials(&stubProvider{})
+
+	_, _, testFlock := getMocks()
+
+	// unable to lock
+	testFlock.success = false
+	testFlock.err = errors.New("lock stuck, needs wd-40")
```
Callisto13, line 242: lol

pkg/eks/api.go
```diff
@@ -161,6 +161,15 @@ func New(spec *api.ProviderConfig, clusterSpec *api.ClusterConfig) (*ClusterProv
 	// later re-use if overriding sessions due to custom URL
 	s := c.newSession(spec)
 
+	cache := os.Getenv(EksctlGlobalEnableCachingEnvName)
```
Skarlso, line 164: hmm, should add an api test here. will be though to see the change

Skarlso: this part will have to be covered by an integration test

Himangini (approved):
LGTM, although have a second pair of eyes for review 👀  👍🏻

integration/tests/caching/credential_caching_test.go
```diff
@@ -0,0 +1,102 @@
...
+				Expect(cmd).Should(RunSuccessfully())
+
+				_, err := os.Stat(filepath.Join(tmp, "credentials.yaml"))
+				Expect(os.IsNotExist(err)).To(BeTrue())
+			})
+		})
+		// Note: This proves to be a challenge since normal providers like, static and file providers, do not support
+		// expiry, therefore, they cannot be cached. This requires an AssumeRole or an EC2 role provider.
+		// For now, this part is tested via unit tests.
+		XWhen("credential caching is enabled", func() {
```
aclevername, line 78: remember to remove :smile:

integration/tests/caching/credential_caching_test.go
```diff
@@ -0,0 +1,102 @@
...
+				cmd := params.EksctlGetCmd.WithArgs(
+					"cluster",
+					"--name", clusterName,
+				).WithoutArg("--region", params.Region)
+				Expect(cmd).Should(RunSuccessfully())
+
+				content, err := ioutil.ReadFile(filepath.Join(tmp, "credentials.yaml"))
+				Expect(err).NotTo(HaveOccurred())
+				Expect(content).NotTo(BeEmpty())
+			})
```
aclevername, line 99: can we test if the caching actually works, instead just if the file exists or not? or does the `get` command existing successfully mean the caching is working (it runs successfully in both scenarios so I'm guessing no)

integration/tests/caching/credential_caching_test.go
```diff
@@ -0,0 +1,102 @@
...
+		clusterName string
+	)
+	BeforeSuite(func() {
+		clusterName = params.NewClusterName("cache")
+		cmd := params.EksctlCreateCmd.WithArgs(
+			"cluster",
+			"--name", clusterName,
+			"--without-nodegroup",
+		).WithoutArg("--region", params.Region)
+		Expect(cmd).Should(RunSuccessfully())
```
aclevername, line 45: do we need a cluster? Could just running `eksctl get clusters` be enough?

pkg/eks/filecache.go
```diff
@@ -0,0 +1,227 @@
...
+	Now() time.Time
+}
+
+// RealClock defines a clock using time.Now()
+type RealClock struct{}
+
+// Now returns the current time.
+func (r *RealClock) Now() time.Time {
+	return time.Now()
+}
```
aclevername, line 37: ahhh, a classic :smile:

pkg/eks/filecache.go
```diff
@@ -0,0 +1,227 @@
...
+
+// Put puts the given cachedCredential with a given key into the map. It will overwrite
+// if the key already exists.
+func (c *cacheFile) Put(key string, credential cachedCredential) {
+	c.ProfileMap[key] = credential
+}
+
+// Get returns cachedCredential if it exists in the cred store.
+func (c *cacheFile) Get(key string) (credential cachedCredential) {
```
aclevername, line 64:
```suggestion
func (c *cacheFile) Get(key string) (cachedCredential) {
```

image

Skarlso: Pfff :D

Skarlso: That's the reason for named return types. :D

pkg/eks/filecache.go
Skarlso: Whaaat, you don't like my Clock? :D :D :D

integration/tests/caching/credential_caching_test.go
Skarlso: Maaaaybe. 🤔

integration/tests/caching/credential_caching_test.go
Skarlso: This test isn't working and needs another ticket to investigate how it could be made working. I need an AssumeRole provider and a user with creds that can be cached. I'm not sure how to achieve that from an integration test. Feel free to share if you have any ideas?

integration/tests/caching/credential_caching_test.go
Skarlso: Yeah, you're probably right. For the sake of the test a simple call to get should be enough, doesn't need a cluster for that. :)

pkg/eks/filecache.go
Skarlso: I can replace it with a simple `func() time.Time` if you'd prefer that. :P

aclevername: no I love it! Its a classic

integration/tests/caching/credential_caching_test.go
aclevername: no idea :/

Skarlso: YourMyOnlyHope-vblog

pkg/eks/filecache.go
Callisto13: to irritate people with better taste? :trollface:

Skarlso: Pfpfpfpfff, you people are just jealous.

pkg/eks/filecache.go
```diff
@@ -0,0 +1,228 @@
+package eks
```
Callisto13, line 1: why is this in the `eks` package? does it call the eks api?

pkg/eks/filecache.go
```diff
@@ -0,0 +1,228 @@
...
+// and the default credential provider is used. Caches are per profile.
+func NewFileCacheProvider(profile string, creds *credentials.Credentials, clock Clock) (FileCacheProvider, error) {
+	if creds == nil {
+		return FileCacheProvider{}, errors.New("no underlying Credentials object provided")
+	}
+	filename, err := cacheFilename()
+	if err != nil {
+		return FileCacheProvider{}, fmt.Errorf("failed to get cache file: %w", err)
+	}
+	_ = os.MkdirAll(filepath.Dir(filename), 0700)
```
Callisto13, line 84: do we not want to check the err?

pkg/eks/filecache.go
Skarlso: Why wouldn't it be? It's the other way around. eks is using this thing. I mean it can live in a separate package, sure if you'd prefer.

Callisto13: doesn't bother me really if there is a logical reason for it to be there. we just tend to overload that package a bit and oftentimes i end up there like "how did i even get here?"

pkg/eks/filecache_test.go
```diff
@@ -0,0 +1,177 @@
...
+	stubProvider
+	expiration time.Time
+}
+
+func (s *stubProviderExpirer) ExpiresAt() time.Time {
+	return s.expiration
+}
+
+var _ = Describe("filecache", func() {
+	Context("a cached based provider is request", func() {
```
Callisto13, line 44: what is this meant to say?

pkg/eks/filecache_test.go
```diff
@@ -0,0 +1,177 @@
...
+			Expect(string(content)).To(Equal(`profiles:
+  profile:
+    credential:
+      accesskeyid: id
+      secretaccesskey: secret
+      sessiontoken: token
+      providername: stubProvider
+    expiration: 0001-01-01T00:00:00Z
+`))
+			Expect(p.IsExpired()).NotTo(BeTrue())
```
Callisto13, line 89: is the same thing (ie the content of the file) being checked twice here?

pkg/eks/filecache.go
Skarlso: :D That's a reasonable thought to me! :) To understand why I put it there.. It's dealing with the cred coming back from the API, right? So I thought it's reasonable to put it next to it.

But I also thought of just putting it into a cache package, or creds package, or providers package... What do you think would be a nice package name for it then? apicache, credcache, provider, cache...?

pkg/eks/filecache.go
Skarlso: Ops, thanks for the spot.

pkg/eks/filecache_test.go
```diff
@@ -0,0 +1,177 @@
...
+				p, err := NewFileCacheProvider("profile", credentials.NewStaticCredentials("id", "secret", "token"), fakeClock)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = p.Retrieve()
+				Expect(err).ToNot(HaveOccurred())
+				_, err = os.Stat(filepath.Join(tmp, "credentials.yaml"))
+				Expect(os.IsNotExist(err)).To(BeTrue())
+			})
+
+		})
+		When("the cache file is not private", func() {
```
Callisto13, line 154: instead of private (which is hard to tell what it means) can we say what is expected in octal?

pkg/eks/filecache_test.go
Skarlso: Ops, I missed that I'm already checking this above a couple lines. NICE spot Claudia. I'm impressed. :D

pkg/eks/filecache_test.go
Skarlso: I probably had a stroke writing that.

pkg/eks/filecache.go
Callisto13: > It's dealing with the cred coming back from the API, right? So I thought it's reasonable to put it next to it.

that makes sense.

and yeh the trouble of moving to its own package is that we are leaning towards a two-word name 😂 . I like `credcache` best simply because it is most descriptive 🤷‍♀️  or maybe just `credential`

pkg/eks/filecache_test.go
Callisto13: yes learning to read was a great decision

aclevername: image

pkg/eks/filecache.go
Skarlso: Done :)

pkg/credentials/filecache.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return credential, err
+	}
+	f.cachedCredential = cachedCredential{
+		Credential: credential,
+		Expiration: expiration,
+	}
+	// overwrite whatever was there before. we don't care about multiple creds for various clusters.
+	// if user switches to another role and another profile they have to re-authenticate.
+	cache, _ := readCacheFile(filename)
```
Callisto13, line 199: err?

pkg/credentials/filecache.go
```diff
@@ -0,0 +1,230 @@
...
+		clock:            clock,
+	}, nil
+}
+
+// readCacheFile reads the contents of the credential cache and returns the
+// parsed yaml as a cachedCredential object.
+func readCacheFile(filename string) (cacheFile, error) {
+	lock := flock.New(filename)
+	defer func() {
+		_ = lock.Unlock()
```
Callisto13, line 121: do we want to log or smth if unlocking fails? 🤔 i know we will find out that the lock is still claimed the next time it is called, but could be helpful 🤷‍♀️

pkg/credentials/filecache.go
Skarlso: Doesn't matter at this point. It will just be overwritten anyways. If there was any kind of error, it will create a new file, create a new cache, create a new thing all together.

pkg/credentials/filecache.go
Skarlso: Sure, I can add some log lines

Skarlso: Added :)

Callisto13: is that message correct?

pkg/eks/filecache_test.go
Callisto13: did you change this?

pkg/credentials/filecache.go
Skarlso: Now it is. :D :D

Skarlso: You know, reading really helped further your carrier.

Skarlso:
@Callisto13 Are you okay now with merging this?

Callisto13:
The test still has that duplication in it, should that be there?

otherwise yeh all good

Callisto13 (approved):

Skarlso:
> The test still has that duplication in it, should that be there?
> 
> otherwise yeh all good

@Callisto13 Wait, what duplication? I thought you mean the IsExpire check which was duplicated. :D The file content checking is not. It's checking that the file contains whatever the `value` object returned plus the expiry date. Or I really can't read. D:

pkg/eks/filecache_test.go
Callisto13: > @Callisto13 Wait, what duplication? I thought you mean the IsExpire check which was duplicated. :D The file content checking is not. It's checking that the file contains whatever the value object returned plus the expiry date and the profile at which it was cached. Or I really can't read. :D

is `p.Retrieve` not just pulling out the content of that file? (i can;t remember and cnba to check the code 😁 )

ckdarby:
```bash
# This works
aws s3 ls
....
....bunch of info...

# Right after try to create cluster prompted for MFA
eksctl create cluster
Assume Role MFA token code:
```

```
eksctl version                                                          
0.77.0
```

This is no longer working by the way. @Skarlso

Skarlso:
You have to explicitly enable caching like this:

```
export EKSCTL_ENABLE_CREDENTIAL_CACHE=1
```

If you do that, does it still ask?

Skarlso:
@ckdarby ^^

ckdarby:
@Skarlso It doesn't look like any of the AWS sts is carrying forward anymore.

```
export EKSCTL_ENABLE_CREDENTIAL_CACHE=1

## Valid sts
aws sts get-caller-identity
{...}

##Valid output
aws s3 ls
...


╰─ ~/bin ❯ eksctl create cluster                                                   
2022-01-10 14:57:13 [!]  Cache file /home/ckdarby/.eksctl/cache/credentials.yaml does not exist.
2022-01-10 14:57:13 [ℹ]  No cached credential available.  Refreshing...
Assume Role MFA token code: ******
Assume Role MFA token code: ******
Error: checking AWS STS access – cannot get role ARN for current session: AccessDenied: MultiFactorAuthentication failed, unable to validate MFA code.  Please verify your MFA serial number is valid and associated with this user.
	status code: 403

**Comment by Skarlso** [comment]
That is super strange. Can you check if the aws cli is working? It looks like it can't authenticate in the first place so it's unable to cache the credentials.

**Comment by ckdarby** [comment]
@Skarlso Yes, I am able to do the normal `aws` calls without issue

**Comment by Skarlso** [comment]
No, I mean, can you try the aws cli WITH mfa?

**Comment by Skarlso** [comment]
Oh, right, you are saying that you are able to use `aws` normally. Ummm... can you try deleting the aws cred cache and try your MFA again please?

**Comment by ckdarby** [comment]
```bash

rm -fr ~/.aws/cli/cache/
export AWS_PROFILE=st4

aws sts get-caller-identity
Enter MFA code for arn:aws:iam::**********:mfa/cdarby: 
## successful sts output {userId, Account, Arn}

export EKSCTL_ENABLE_CREDENTIAL_CACHE=1
eksctl create cluster
2022-01-10 15:39:26 [!]  Cache file /home/ckdarby/.eksctl/cache/credentials.yaml does not exist.
2022-01-10 15:39:26 [ℹ]  No cached credential available.  Refreshing...
Assume Role MFA token code:
```

:smiling_face_with_tear:

Skarlso:
And if you try again now, it's still not working?

ckdarby:
@Skarlso Correct.

Skarlso:
Okay, we updated the aws go-sdk pretty recently. :/ They might have messed something up. I will have to take a look at that. In the mean time, maybe pop over and see in recent issue if there is something about this.

It's rather late in my TZ so I'm gonna drop off and take a look this tomorrow. Thanks for telling me.

PR #4201, opened 2021-09-09
Delete IAM role if RegisterCluster fails

cPu1:
### Description

`RegisterCluster` fails if the SLR is not present. This changelist deletes the IAM role created for the connector if the API call fails. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/connector/connector.go
+++ b/pkg/connector/connector.go
@@ -28,9 +28,9 @@ const (
 )
 
 type ExternalCluster struct {
-	Name          string
-	Provider      string
-	ConnectorRole string
+	Name             string
+	Provider         string
+	ConnectorRoleARN string
 }
 
 type EKSConnector struct {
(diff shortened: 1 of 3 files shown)


pkg/connector/connector.go
```diff
@@ -75,26 +75,43 @@ func (c *EKSConnector) RegisterCluster(cluster ExternalCluster) (*ManifestList,
...
+			connectorRoleName, err := roleNameFromARN(connectorRoleARN)
+			if err != nil {
+				return errors.Wrap(err, "error parsing connector role ARN")
+			}
+			return c.deleteRole(connectorRoleName)
+		}
+
+		if deleteErr := deleteRoleIfOwned(); deleteErr != nil {
+			err = errors.Wrap(err, deleteErr.Error())
+		}
```
aclevername, line 102: why is this a func inside a func, that only gets called once? Seems a bit odd :thinking: (compared to just even not having the func at all, or having this func defined outside of this func)

cPu1: This is mainly to avoid having to check the `error` value multiple times in order to `Wrap` it below with the error from `registerCluster`. 

>or having this func defined outside of this func

Having it inside lets it access the closed-over variables and this function is not relevant outside.

aclevername: > Having it inside lets it access the closed-over variables and this function is not relevant outside.

I don't really think thats a reason to not put it outside, we do this literally all the time. This is also happening inside this file already, `ownsIAMRole()` is defined outside of the function its called in, and is only called inside that function, same with `createConnectorRole()` :shrug: having it separated just makes it easier to read and is the normal convention

cPu1: What I meant by relevant is not that it's used only once, but rather that the logic it contains is very specific to the closed-over variables and it'd be difficult to come up with a good name for a function defined outside and its parameters. If we remove the first `if`, we can make a func called `deleteRoleByARN` but the caller will have to guard it with an `if`. We don't gain anything from this. 

>why is this a func inside a func, that only gets called once?

This is no reason for not having an anonymous function. There should be examples of Go code on popular GitHub repos where an anonymous function is defined and called only once.

aclevername: > What I meant by relevant is not that it's used only once, but rather that the logic it contains is very specific to the closed-over variables and it'd be difficult to come up with a good name for a function defined outside and its parameters. If we remove the first `if`, we can make a func called `deleteRoleByARN` but the caller will have to guard it with an `if`. We don't gain anything from this.

The function name you've provided it is already pretty descriptive and understandable though `deleteRoleIfOwned`? You could expand it in the function outside with something like `deleteOwnedRoleIfExists` or something like that.

Callisto13: > This is mainly to avoid having to check the error value multiple times in order to Wrap it below with the error from registerCluster.

would it even be that many times just to call each bit in sequence? i don't see what value we get from this tbh, just an extra layer of "stuff" to parse 🤷‍♀️

cPu1: >The function name you've provided it is already pretty descriptive and understandable though deleteRoleIfOwned?

It is, but like I pointed not much is gained from this, and it's not uncommon to define logic like this in an anonymous function. I will put that block of code outside in a function though. 

>You could expand it in the function outside with something like deleteOwnedRoleIfExists or something like that.

`ifExists` is not required because it's known here that the role exists.

cPu1: >would it even be that many times just to call each bit in sequence? i don't see what value we get from this tbh, just an extra layer of "stuff" to parse woman_shrugging

Two times, but it results in code duplication and I found it unreadable.

aclevername (approved):

Issue #4259, opened 2021-10-01
Add ipFamily field to config


Callisto13:
Part of https://github.com/weaveworks/eksctl/issues/4255

**Description**
Task here is to add a new field in VPC: `ipFamily`. Type is `string`.

```yaml
vpc:
   ipFamily: ipv6 # or ipv4
```

Value can be `ipv6` or `ipv4`, default is `ipv4`.

Add validation and defaulting here for the type.

**Scope**
- Literally just adding the field in Types. This will unblock the rest of the work. Do not be a hero.
- Bootstrap docs (`examples/` and `userdocs/`)

**Acceptance Criteria**
- As an `eksctl` user
- When I set `vpc.ipFamily` in my cluster configuation to `ipv6`
- The setting is validated
- If I don't specify the value, it's defaulted to `ipv4`
- I can see an example ipv6 config in `examples/`
- I can find some eksctl docs about ipv6

PR #4292, opened 2021-10-04
Add ipFamily to config

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4259

### Checklist
- [x] Manually tested


pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -523,6 +523,67 @@ var _ = Describe("ClusterConfig validation", func() {
...
+			cfg = api.NewClusterConfig()
+			vpc = api.NewClusterVPC()
+			cfg.VPC = vpc
+		})
+
+		Context("ipFamily", func() {
+			It("should not error default ipFamily setting", func() {
+				err = cfg.ValidateVPCConfig()
+				Expect(err).ToNot(HaveOccurred())
+			})
```
aclevername, line 543: do we wanna check the default is ipv4?

Skarlso: ops, sure

examples/29-vpc-with-ip-family.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ClusterConfig
+
+metadata:
+  name: cluster-2
+  region: eu-north-1
+
+vpc:
+  ipFamily: ipv6
+
+nodeGroups: []
```
nikimanoledaki, line 14: unmanaged nodegroups won't be supported so this line should be `managedNodeGroups`

nikimanoledaki (approved):
LGTM! one small thing only :)

examples/29-vpc-with-ip-family.yaml
Skarlso: Ah, thanks! :D But does it even matter, if it's empty? 🤔 Can I even just delete it? :D

aclevername (approved):

examples/29-vpc-with-ip-family.yaml
nikimanoledaki: it doesn't since it's empty! you could delete it :)

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -182,6 +183,8 @@ type (
 		ID string `json:"id,omitempty"`
 		// +optional
 		CIDR *ipnet.IPNet `json:"cidr,omitempty"`
+		// +optional
```
cPu1, line 186: The generated schema will list the supported values if you add this comment:

```suggestion
		// Valid variants are `IPFamily` constants
		// +optional
```

I think you might also need to add a comment for the constants.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -315,6 +315,20 @@ const (
 	DefaultNodeVolumeGP3IOPS = 3000
 )
 
+type IPFamily string
+
+const (
```
cPu1, line 320: This will let the schema generation code list supported values for the IPFamily field:

```suggestion

// Values for `IPFamily`
const (
```

cPu1 (commented):
I know I am late to review this, but I had already typed my comments so I'll submit it.

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -315,6 +315,20 @@ const (
...
+	// IPV4Family defines an ip family of v4 to be used when creating a new VPC.
+	IPV4Family IPFamily = "ipv4"
+	// IPV6Family defines an ip family of v6 to be used when creating a new VPC.
+	IPV6Family IPFamily = "ipv6"
```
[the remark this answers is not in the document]
nikimanoledaki: I agree that it would be more readable, however, CF expects lowercase input values - see https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-eks/pull/56/files#diff-4ba214c0c8b1426b24d21b0ed5bb2beb7fdc52f5e6899c6e77b2fd1e04e731e0R45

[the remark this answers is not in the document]
cPu1: CF is an implementation detail while this is the UX for this feature. We don't need to follow the API schema when designing the UX.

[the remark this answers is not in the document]
nikimanoledaki: That's true, I think we were taking `ipv4` for granted until now but I agree with you that capitalising it would make a cleaner UX. 

Your other comments are also good catches! Would you like to open a quick PR to add your suggestions?

Skarlso:
I can add a new PR with the suggested changes by @cPu1.

pkg/apis/eksctl.io/v1alpha5/types.go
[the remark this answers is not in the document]
Skarlso: Why would we allow multiple values for this setting @cPu1?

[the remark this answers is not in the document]
cPu1: Ah, we won't. That example `ipFamilies` snippet was taken from the Kubernetes `Service` resource. It was only to show that it uses the uppercase version.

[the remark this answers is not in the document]
Skarlso: Oh, okay. :D Thanks. :DD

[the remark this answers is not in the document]
cPu1: >Would you like to open a quick PR to add your suggestions?

Looks like Gergely is working on it.

pkg/apis/eksctl.io/v1alpha5/types.go
Skarlso: Oh cool, I didn't know that.

PR #4296, opened 2021-10-05
Change ip to IP and update schema description

Skarlso:
### Description

Additional comments update for #4292.

### Checklist
- [ ] Manually tested


cPu1 (approved):

cPu1:
That was quick!

nikimanoledaki (approved):

Skarlso:
@cPu1 Thanks for the additional cool improvement review! :)

pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -183,6 +183,7 @@ type (
 		ID string `json:"id,omitempty"`
 		// +optional
 		CIDR *ipnet.IPNet `json:"cidr,omitempty"`
+		// Valid variants are `IPFamily` constants
 		// +optional
 		IPFamily *string `json:"ipFamily,omitempty"`
```
nikimanoledaki, line 188: how come this is a pointer rather than a string? I would expect a string since this is the config file input

Skarlso: Because we need to be able to differentiate between default has to be set because it hasn't been defined, or the user specifically set it to ipv4.

aclevername: :thinking: but how is it different to `metadata.Version` (string), that can be empty or set by the user

Skarlso: Isn't metadata.Version always set though? For VPC, we don't always set this value, we set it only when it's being created. Also, if we set this to empty string, it will show up as `ipFamily = ""` if vpc is set but this is not in the config files. Which is rather confusing.

Right now, we aren't setting this setting if a user defines a vpc for example. But if you leave this empty, it will show up as empty in yaml marshalling which the code can ignore of course.

metadata.Version is always set to a version as far as I can tell. But please correct me if I'm wrong.

Skarlso: @aclevername ^^ :D

Issue #4273, opened 2021-10-01
User can create a new VPC with a custom IPv6 CIDR


nikimanoledaki:
Part of https://github.com/weaveworks/eksctl/issues/4255

**Description**

Users should be able to import their own v6 IPs into AWS and have them used as part of their VPC creation.

This is a slightly different UX from the current IPv4 workflow in which users can set a `vpc.cidr` with an actual CIDR and don't need to import anything beforehand.

_Note: setting CIDR for IPv6 is a slightly different experience than for IPv4. For IPv4 users can provide whatever numbers (within reason) and no other work is required on their part. For IPv6, users have to do the work to import their CIDR into their account before they use eksctl. So there is an extra step that they will have to be made aware of in  Docs, as the behaviour of this config is not quite the same as the equivalent IPv4 flag._

In order to support this for IPv6 we would have to expose 2 fields which would map to EC2::IPv6CidrBlock and EC2::IPv6Pool.

```yaml
vpc:
  cidr: 192.0.2.0/24 <- users should still be able to set a v4 custom cidr
  ipv6Cidr: 2001:db8::/32 <- for the custom v6 CIDR
  ipv6Pool: <- must be set alongside ipv6Cidr
```

The rest of the new VPC setup is the same as detailed in the above section, the requested CIDR(s) will be set in the generated Cloud Formation template.

**Scope**
- The current `vpc.cidr` workflow should work with IPv6 as well. Users should be able to set both.
- If we validate v4 CIDRS then we need to validate v4 CIDRs as well. 

**Acceptance Criteria**
Story 1
- As an eksctl user who has imported their own IPv6 IPs,
- I can set my custom IPv6 CIDR in my config file
- And when I create my cluster
- I see that my cluster and VPC have been created using that CIDR.

Story 2
- As an eksctl user who has imported their own IPv4 and IPv6 IPs,
- When I set IPv4 CIDR and IPv6 CIDR in my config file,
- And when I create my cluster
- I see that my cluster and VPC have been created using both given IPv4 and IPv6 CIDRs.

PR #4378, opened 2021-10-28
User can create a new VPC with a custom IPv6 CIDR

aclevername:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4273

TODO:
-  ~~Add validation `vpc.ipv6Cidr` and `vpc.ipv6CidrPool` are always both set together~~
- ~~Test out it works with a real cluster~~
  - This look way too expensive to test https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-byoip.html
- ~~Add support for the current `vpc.cidr` workflow~~

### Checklist
- [ ] Manually tested
- [X] (Core team) Added labels for change area (e.g. `area/nodegroup`) and kind (e.g. `kind/improvement`)

- [X] Refactored something and made the world a better place :star2:


pkg/apis/eksctl.io/v1alpha5/assets/schema.json
```diff
@@ -577,6 +577,12 @@
             "IPv6"
           ]
         },
+        "ipv6Cidr": {
+          "type": "string"
+        },
+        "ipv6CidrPool": {
```
nikimanoledaki, line 583: Looks good :) I realise how late it is to make this change, but could we rename the new `ipv6CidrPool` field to `ipv6Pool`? We should have caught this way earlier (we also probably didn't know as much as we do now) but I don't think there's such a thing as a `CIDR pool`: it's really just an `address pool` and the CIDR range gets derived from it.

aclevername: yeah sure!

nikimanoledaki: yep, `ipv6Pool ` checks out: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Ipv6Pool.html not sure how this slipped through!

aclevername: done!

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -199,6 +208,22 @@ func (c *ClusterConfig) ValidateVPCConfig() error {
...
+	}
+
+	if c.VPC.IPv6Pool != "" {
+		count++
+	}
+
+	if count == 0 || count == 2 {
+		return nil
+	}
+	return fmt.Errorf("Ipv6Cidr and Ipv6CidrPool must both be configured to use custom ipv6 CIDR pool")
```
nikimanoledaki, line 224:
```suggestion
	return fmt.Errorf("Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool")
```

pkg/cfn/builder/vpc_ipv4_test.go
```diff
@@ -557,3 +534,13 @@ func makeRTOutput(subnetIds []string, main bool) *ec2.DescribeRouteTablesOutput
 		}},
 	}
 }
+
+func assertIpv6CidrBlockCreatedWithSelect(cidrBlock interface{}, expectedFnCIDR string) {
```
nikimanoledaki, line 538: nice :)

Himangini:
OMG, you ticked the checkboxes in PR description 🎉 💟

Himangini:
```
 Expected
            <*errors.errorString | 0xc000156840>: {
                s: "Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool",
            }
        to match error
            <string>: Ipv6Cidr and Ipv6CidrPool must both be configured to use custom ipv6 CIDR pool

        /home/runner/work/eksctl/eksctl/pkg/apis/eksctl.io/v1alpha5/validation_test.go:734
 ```       😅

**Review by nikimanoledaki — APPROVED** [review]
lgtm, couple of notes:
- unit test needs updating as mentioned by @Himangini :)
- let's discuss the expensive manual test with the team during standup

**Comment by aclevername** [comment]
> ```
>  Expected
>             <*errors.errorString | 0xc000156840>: {
>                 s: "Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool",
>             }
>         to match error
>             <string>: Ipv6Cidr and Ipv6CidrPool must both be configured to use custom ipv6 CIDR pool
> 
>         /home/runner/work/eksctl/eksctl/pkg/apis/eksctl.io/v1alpha5/validation_test.go:734
>  ```       😅 
>         
> ```

fixing! Do you have any other PR feedback?

File: pkg/apis/eksctl.io/v1alpha5/validation.go [hunk 4]
```diff
@@ -164,8 +164,13 @@ func (c *ClusterConfig) ValidateVPCConfig() error {
 	if c.VPC.IPFamily != IPV4Family && c.VPC.IPFamily != IPV6Family {
 		return fmt.Errorf("invalid value %s for ipFamily; allowed are %s and %s", c.VPC.IPFamily, IPV4Family, IPV6Family)
 	}
+
 	// This is the new vpc check, I need this check when the user sets it.
 	if c.VPC.IPFamily == IPV6Family {
+		if err := c.ipv6CidrsValid(); err != nil {
+			return err
```
Skarlso, line 171: Proper fmt.Errorf here too please.

Skarlso: ah, never mind, I guess it's fine because you do wrap in the calling function.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -190,6 +195,10 @@ func (c *ClusterConfig) ValidateVPCConfig() error {
 		if IsEnabled(c.VPC.AutoAllocateIPv6) {
 			return fmt.Errorf("auto allocate ipv6 is not supported with IPv6")
 		}
+	} else {
```
Skarlso, line 198: This else hanging here after like 40-50 lines of code is super confusing and reads a bit uneasy.

Would you mind just doing a 

```go
if c.VPC.IPFamily == IPV4Family {
```

which reads a lot better and is more easy to follow along please? :)

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -199,6 +208,22 @@ func (c *ClusterConfig) ValidateVPCConfig() error {
...
+	if c.VPC.IPv6Pool != "" {
+		count++
+	}
+
+	if count == 0 || count == 2 {
+		return nil
+	}
+	return fmt.Errorf("Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool")
+}
+
```
Skarlso, line 226: I don't get this function. :D You aren't saying `"Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool` but you are allowing for them to be not configured at all on that count == 0 or.

So is it now that they need to be configured if either is defined or not? Because if they need to be configured than the count is not needed at all.

Also, I find that reading this with the counting increases the cognitive load of this function significantly mostly because it's unexpected format. I would rather see a logical match like:

```go
func (c *ClusterConfig) ipv6CidrsValid() error {
	if (c.VPC.IPv6Cidr == "" && c.VPC.IPv6Pool == "") || (c.VPC.IPv6Cidr != "" && c.VPC.IPv6Pool != "") {
		return nil
	}
	return fmt.Errorf("Ipv6Cidr and Ipv6Pool must both be configured to use a custom IPv6 CIDR and address pool")
}
```

which just reads a lot smoother. WDYT?

pkg/apis/eksctl.io/v1alpha5/validation.go
aclevername: image

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -708,6 +710,38 @@ var _ = Describe("ClusterConfig validation", func() {
...
+
+					cfg.VPC.IPFamily = api.IPV4Family
+					cfg.VPC.IPv6Cidr = ""
+					cfg.VPC.IPv6Pool = "bar"
+					err = cfg.ValidateVPCConfig()
+					Expect(err).To(MatchError("Ipv6Cidr and Ipv6CidrPool is only supportd when IPFamily is set to IPv6"))
+				})
+			})
+
+			When("only one of IPv6Cidr or IPv6CidrPool is provided and ipv6 is set", func() {
```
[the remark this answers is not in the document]
aclevername: thats just the default behaviour? e.g. every test that does ipv6 will assert this. Not sure if it needs a new test?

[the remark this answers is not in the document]
Skarlso: oh that's right

Skarlso (approved):

Himangini:
>let's discuss the expensive manual test with the team during standup
Don't think we have discussed this in today's sup, lets chat tomorrow 👍🏻

Skarlso:
@Himangini Oh Gods you are absolutely right 🤦

PR #4464, opened 2021-11-16
Support scaling all nodegroups found in a config file

### LINKED ISSUE #2655: eksctl scale nodegroup --config-file --name <nodegroup>: node group <nodegroup> not found
**What happened?**
`eksctl scale nodegroup --config-file config.yaml --name <nodegroup>` results in `node group <nodegroup> not found`

1. `eksctl scale nodegroup --config-file config.yaml --name <nodegroup>` should succeed in applying the current desiredCapacity, minSize, maxSize properties of the nodeGroups/managedNodeGroups instance(s) from the config.yaml file.

2. Also would be great if `--name` could be omitted (not required) in which case ALL nodegroups would be scaled to current supplied config spec.

**How to reproduce it?**
1. Launch a cluster using a valid config.yaml with at least one nodegroup and then attempt: `eksctl scale nodegroup --config-file config.yaml --name <nodegroup>` where `config.yaml` is your cluster configuration file and `<nodegroup>` is the name of one of your node groups defined in that cluster config file.

Also note that `eksctl scale nodegroup --cluster <cluster> --name <nodegroup> --minSize 0` works fine even when using the same values that should be read from `config.yaml` if `--config-file config.yaml` were to be used instead. In other words, the bug appears specific to the `--config-file config.yaml` option.

macOS Homebrew eksctl binary

**Versions**
```
$ eksctl 0.27.0
$ kubectl 1.16.9
```

nikimanoledaki:
### Description

Closes #2655

This PR addresses the second part of that issue e.g. `Also would be great if --name could be omitted (not required) in which case ALL nodegroups would be scaled to current supplied config spec.`

### Checklist
- [x] Manually tested


integration/tests/crud/creategetdelete_test.go
```diff
@@ -428,6 +415,50 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
...
+					"nodegroup",
+					"--cluster", params.ClusterName,
+					"--name", "n1",
+					"-o", "yaml",
+				)
+				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MaxSize: 5")))
+				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MinSize: 5")))
+				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("DesiredCapacity: 5")))
+			})
+		})
```
aclevername, line 460: can we do a check on the before vs after? Hard to tell if the existing size is already 5 :smile: perhaps its obvious from the above parts in the test. wdyt?

integration/tests/crud/creategetdelete_test.go
```diff
@@ -428,6 +415,50 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
...
+				Expect(cmd).To(RunSuccessfully())
+			})
+
+			It("should scale all nodegroups", func() {
+				By("scaling all nodegroups in the config file o the desired capacity, max size, and min size")
+				cmd := params.EksctlScaleNodeGroupCmd.WithArgs(
+					"--config-file", "-",
+				).
+					WithoutArg("--region", params.Region).
+					WithStdin(clusterutils.ReaderFromFile(params.ClusterName, params.Region, "testdata/scale-nodegroups.yaml"))
```
aclevername, line 437: I have mixed feelings about this :/ I really dislike that I have to go look elsewhere to see whats in this file, vs in-lining the config like using the struct and then writing it too a file

integration/tests/crud/testdata/scale-nodegroups.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: <generated>
+
+nodeGroups:
+  - name: n1
+    desiredCapacity: 5
+    maxSize: 5
+    minSize: 5
+
+managedNodeGroups:
+  - name: ng-0
```
aclevername, line 15: `ng-0`, `n1` :ahhhhh:

Himangini:
⭐  for ticking all the boxes in description 🎉

pkg/ctl/cmdutils/scale_test.go
```diff
@@ -33,171 +33,208 @@ var _ = Describe("scale node group config file loader", func() {
...
+			It("fails", func() {
+				cmd := &Cmd{
+					CobraCommand: newCmd(),
+				}
+
+				err := NewScaleAllNodeGroupLoader(cmd).Load()
+				Expect(err).To(HaveOccurred())
+				Expect(err).To(MatchError("a config file is required when scaling multiple nodegroups"))
 			})
 		})
```
aclevername, line 239: what is this scenario :thinking: ? `eksctl scale nodegroup --name foo`? The error message implies your using a config file, but your not passing one in so perhaps a more generic error is better?

pkg/ctl/scale/nodegroup.go
```diff
@@ -57,10 +56,38 @@ func scaleNodeGroupWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *cmdutils.Cmd
...
+		}
+	} else {
+		if err := cmdutils.NewScaleNodeGroupLoader(cmd, ng).Load(); err != nil {
+			return err
+		}
+		if err := scaleNodegroup(cmd, ng); err != nil {
+			return err
+		}
 	}
+	return nil
```
aclevername, line 74: nice. Maybe rather than `if else` do a `if` with a return at the end? Less brain power required :smile: 

```suggestion
	if ng.Name == "" && cmd.NameArg == "" {
		if err := cmdutils.NewScaleAllNodeGroupLoader(cmd).Load(); err != nil {
			return err
		}
		return scaleAllNodegroups(cmd);
	}
	
        if err := cmdutils.NewScaleNodeGroupLoader(cmd, ng).Load(); err != nil {
        	return err
        }
        return scaleNodegroup(cmd, ng)

```

integration/tests/crud/testdata/scale-nodegroups.yaml
nikimanoledaki: the names of the nodegroups created in the `crud` suite predate this PR but I'm happy to change them for consistency 👍

integration/tests/crud/creategetdelete_test.go
nikimanoledaki: image

nikimanoledaki: hehehehe

pkg/ctl/cmdutils/scale.go
```diff
@@ -63,14 +85,11 @@ func validateNameArgument(cmd *Cmd, ng *api.NodeGroupBase) error {
 		ng.Name = cmd.NameArg
 	}
 
-	if ng.Name == "" {
```
Skarlso, line 66: Huh?

pkg/ctl/scale/nodegroup.go
```diff
@@ -1,11 +1,10 @@
 package scale
 
 import (
-	"github.com/weaveworks/eksctl/pkg/actions/nodegroup"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/weaveworks/eksctl/pkg/actions/nodegroup"
```
Skarlso, line 7: :shef'skiss:

integration/tests/crud/creategetdelete_test.go
aclevername: image

nikimanoledaki: AMAZING

pkg/ctl/scale/nodegroup.go
nikimanoledaki: I started noticing this after you mentioned it a couple of weeks ago! :D

integration/tests/crud/testdata/scale-nodegroups.yaml
nikimanoledaki: you're right, I'll change the ng names in the `crud` package to something more consistent - what do you think of the following:
```
	const (
		initNG = "ng-1"
		testNG = "ng-2"

		unmNG1 = "unm-ng-1"
		unmNG2 = "unm-ng-2"
	)
```

aclevername: I think we just do `mng-1` and `ng-1` in other tests for managed vs unmanaged, but its still not great either

pkg/ctl/cmdutils/scale.go
Skarlso: Oh, of course, because now we can do this. Right?

nikimanoledaki: right! if `scale` is called without `--name`, it scales all nodegroups in the config file, provided there is a config file. 

the name will never be `""` at this point. we only get to this point if either `ng.Name` or `cmd.NameArg` are set, which is checked before the config file loader, [here](https://github.com/weaveworks/eksctl/pull/4464/files/c0fb213ac7f81ae11a0db4acda1bc9f23a6161d1#diff-e9fe30ff1a878b24cb1d115bad2744799bdd46cccbeaee1cdb9c66360262f3a8R59)

pkg/ctl/cmdutils/scale_test.go
nikimanoledaki: I updated this file, hopefully the scenarios are a bit clearer now! Also added a couple of missing unit tests :)

Himangini (approved):
LGTM 👍🏻

PR #4465, opened 2021-11-16
Add config-file option to get nodegroup

Skarlso:
### Description

Does the `nodegroup` part of https://github.com/weaveworks/eksctl/issues/1126.

```
eksctl get nodegroup -f cluster-managed.yaml
2021-11-16 16:03:36 [ℹ]  eksctl version 0.75.0-dev+c5f6d0ef.2021-11-16T15:10:41Z
2021-11-16 16:03:36 [ℹ]  using region us-west-2
CLUSTER                         NODEGROUP       STATUS  CREATED                 MIN SIZE        MAX SIZE        DESIRED CAPACITY     INSTANCE TYPE    IMAGE ID        ASG NAME
test-nodegroup-cluster-config   managed-ng-1    ACTIVE  2021-11-16T12:02:30Z    1               2               1                    m5.large AL2_x86_64      eks-52be9450-42a1-0ff9-b901-3b51d3bda8d3
test-nodegroup-cluster-config   managed-ng-2    ACTIVE  2021-11-16T14:56:25Z    1               2               1                    m5.large AL2_x86_64      eks-f8be949f-df08-fa4e-4e46-2590f597f474
```

It's not allowed to define `--name` with `--config-file`. I'm on the fence with that. Like, it would be nice to define which nodegroup you are interested in... But then again, if you use the config-file you want to see all of the things the config-file provides and this behaviour is mimicked with others. 

This was easier since nodegroup already was using the config thing to load in data, so it was easy to just overwrite it if the setting is set. I tried without, with and only name and only name argument.

### Testing

- [x] with `--config-file`
- [x] with named argument
- [x] with `--cluster` + `--name`
- [x] with `--config-file` + `--name`
- [x] without anything

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/get/nodegroup.go
+++ b/pkg/ctl/get/nodegroup.go
@@ -38,15 +38,18 @@ func getNodeGroupCmd(cmd *cmdutils.Cmd) {
 		cmdutils.AddRegionFlag(fs, &cmd.ProviderConfig)
 		cmdutils.AddCommonFlagsForGetCmd(fs, &params.chunkSize, &params.output)
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
 func doGetNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, params *getCmdParams) error {
+	if err := cmdutils.NewMetadataLoader(cmd).Load(); err != nil {
+		return err
+	}
 	cfg := cmd.ClusterConfig
 
-	// TODO: move this into a loader when --config-file gets added to this command
 	if cfg.Metadata.Name == "" {
 		return cmdutils.ErrMustBeSet(cmdutils.ClusterNameFlag(cmd))
 	}


pkg/ctl/get/nodegroup.go
```diff
@@ -38,15 +38,18 @@ func getNodeGroupCmd(cmd *cmdutils.Cmd) {
...
 		cmdutils.AddCommonFlagsForGetCmd(fs, &params.chunkSize, &params.output)
 		cmdutils.AddTimeoutFlag(fs, &cmd.ProviderConfig.WaitTimeout)
+		cmdutils.AddConfigFileFlag(fs, &cmd.ClusterConfigFile)
 	})
 
 	cmdutils.AddCommonFlagsForAWS(cmd.FlagSetGroup, &cmd.ProviderConfig, false)
 }
 
 func doGetNodeGroup(cmd *cmdutils.Cmd, ng *api.NodeGroup, params *getCmdParams) error {
+	if err := cmdutils.NewMetadataLoader(cmd).Load(); err != nil {
```
Skarlso, line 48: Oh, I failed to write some unit tests for this 🤦

pkg/ctl/get/nodegroup.go
```diff
@@ -38,15 +38,18 @@ func getNodeGroupCmd(cmd *cmdutils.Cmd) {
...
+		return err
+	}
 	cfg := cmd.ClusterConfig
 
-	// TODO: move this into a loader when --config-file gets added to this command
```
nikimanoledaki, line 49: I think this TODO is saying to move all the validation below into the loader to match the pattern in other `ctl`s 🤔

Skarlso: Oh... you might be right. :D

pkg/ctl/get/nodegroup_test.go
```diff
@@ -15,12 +17,41 @@ var _ = Describe("get", func() {
...
+
+		It("setting --name and --config-file at the same time", func() {
+			f, err := os.CreateTemp("", "configfile")
+			Expect(err).NotTo(HaveOccurred())
+			_, err = f.WriteString(configFile)
+			Expect(err).NotTo(HaveOccurred())
+			cmd := newMockCmd("nodegroup", "--name", "name", "--config-file", f.Name())
+			_, err = cmd.execute()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("Error: cannot use --name when --config-file/-f is set"))
```
cPu1, line 41: There's also `MatchError`:

```suggestion
			Expect(err).To(MatchError(ContainSubstring("Error: cannot use --name when --config-file/-f is set")))
```

pkg/ctl/get/nodegroup_test.go
```diff
@@ -15,12 +17,41 @@ var _ = Describe("get", func() {
...
+
+		It("setting --cluster and --config-file at the same time", func() {
+			f, err := os.CreateTemp("", "configfile")
+			Expect(err).NotTo(HaveOccurred())
+			_, err = f.WriteString(configFile)
+			Expect(err).NotTo(HaveOccurred())
+			cmd := newMockCmd("nodegroup", "--cluster", "name", "--config-file", f.Name())
+			_, err = cmd.execute()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("Error: cannot use --cluster when --config-file/-f is set"))
```
cPu1, line 52: Same as the previous comment about using `MatchError`.

pkg/ctl/get/nodegroup_test.go
```diff
@@ -15,12 +17,41 @@ var _ = Describe("get", func() {
 		})
 
 		It("setting --name and argument at the same time", func() {
-			cmd := newMockCmd("nodegroup", "ng", "--cluster", "dummy", "--name", "ng")
```
cPu1, line 18: Any reason why the `--cluster` argument was removed from this test?

Skarlso: This one is a bit of a interesting one. I think it was testing the wrong thing. However, with the introduction of the new `--config-file` option the error message is not correctly returning that `--name and nameArg can't be together, where as previously, it was doing the wrong thing with the `--cluster` param. Very weird. :D

cPu1 (approved):

pkg/ctl/get/nodegroup.go
nikimanoledaki: Ah I meant that this could be done as part of this PR! WDYT?

nikimanoledaki:
`It's not allowed to define --name with --config-file`
This reminds me of https://github.com/weaveworks/eksctl/pull/4383 where, if `--config-file` and `--name` are both set, `FindNodegroup` discovers the nodegroup in question in the config file. WDYT? Maybe worth creating a new issue for this feature though!

pkg/ctl/get/nodegroup.go
Skarlso: definitely not :DD

Skarlso:
@nikimanoledaki yeah sure. The problem is that it's actually in the main thing and it would change behaviour for all commands. Which might be fine of course. Just need to document it ( if we do ) and be aware.

pkg/ctl/get/nodegroup.go
aclevername: why not? Seems related to this PR :thinking:

nikimanoledaki:
That's true, it would be a big change in behaviour that would be hard to justify if there hasn't been any demand from users. The difference with the PR I linked is that there had been user demand! So I guess let's leave it as is and let user demand drive this, like all other things? :)

Skarlso:
Yep, sounds good to me.

pkg/ctl/get/nodegroup_test.go
Skarlso: Uh, I just understood what's happening here and will provide a fix in an upcoming PR.

Issue #3718, opened 2021-05-17
instanceSelector support for allow/deny list


kr3cj:
We are starting to use the (`instanceSelector`)[https://github.com/weaveworks/eksctl/pull/3498] feature in some our EKS configs. It greatly simplifies our configuration and is easier to understand. Thank you for this feature.

However, in order to use it everywhere, we would need to be able to specify some kind of allow/deny list for specific instanceTypes.

The reason we need this is because we've performed profiling of different instance classes with our workloads and found that certain ones aren't something we want to adopt in production (t3 specifically). They're good for staging/dev environments but for production, we try to stick with other instance classes.

We noticed that the underlying (`ec2-instance-selector`)[https://github.com/aws/amazon-ec2-instance-selector] already supports the following arguments:

```
      --allow-list string                 List of allowed instance types to select from w/ regex syntax (Example: m[3-5]\.*)
      --deny-list string                 List of instance types which should be excluded w/ regex syntax (Example: m[1-2]\.*)
```

It would be great if those could be exposed in the eksctl config schema. Something like:

```
managedNodeGroups:
  - name: my-node-group
    instanceSelector:
      vCPUs: 4
      memory: "16"
      allow: "[cm]5\.*"
```

Or

```
managedNodeGroups:
  - name: my-node-group
    instanceSelector:
      vCPUs: 4
      memory: "16"
      deny: "t\.*"
```

This came out of the discussion on https://github.com/weaveworks/eksctl/issues/3531#issuecomment-814225408

cPu1:
This is a useful feature, but we're wary of adding more instance selector options to eksctl to avoid bloating it with features only few are using. We're going to discuss this with the EKS team to check if other users would benefit from this feature or not.

bhops:
> This is a useful feature, but we're wary of adding more instance selector options to eksctl to avoid bloating it with features only few are using. We're going to discuss this with the EKS team to check if other users would benefit from this feature or not.

I would argue that part of the reason it may mot be very widely used could be because of how limited the included options are. For my use case, the inability to filter the list of returned instances is a non-starter for using the instanceSelector option entirely as it brings in very old generation as well as burstable instances which we do not want running in production.

akestner:
[...]

PR #4466, opened 2021-11-17
Adds instanceSelector support for allow and deny rules

bhops:
### Description

Added the ability to specify `allow` and `deny` attributes as part of the instanceSelector
configuration as well as the command line arguments which allow instance types to be
filtered based on regexes. These fields behave like the `--allow-list` and `--deny-list`
arguments from https://github.com/aws/amazon-ec2-instance-selector.

Issue #3718 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/assets/schema.json
+++ b/pkg/apis/eksctl.io/v1alpha5/assets/schema.json
@@ -816,6 +816,11 @@
     },
     "InstanceSelector": {
       "properties": {
+        "allow": {
+          "type": "string",
+          "description": "Allow is a regex that matching instances must meet to be selected",
+          "x-intellij-html-description": "Allow is a regex that matching instances must meet to be selected"
+        },
         "cpuArchitecture": {
           "type": "string",
           "description": "CPU Architecture of the EC2 instance type. Valid variants are: `\"x86_64\"` `\"amd64\"` `\"arm64\"`",
@@ -825,6 +830,11 @@
             "amd64",
             "arm64"
           ]
+        },
+        "deny": {
+          "type": "string",
+          "description": "Deny is a regex that prevents matching instances from being selected",
+          "x-intellij-html-description": "Deny is a regex that prevents matching instances from being selected"
         },
         "gpus": {
           "type": "integer",
@@ -846,7 +856,9 @@
         "vCPUs",
         "memory",
         "gpus",
-        "cpuArchitecture"
+        "cpuArchitecture",
+        "allow",
+        "deny"
       ],
       "additionalProperties": false,
       "description": "holds EC2 instance selector options",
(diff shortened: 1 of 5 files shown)


bhops:
I am not sure where to write good unit tests for this, as most of the tests around the instanceSelector logic appears to be mocked, and aside from validating that the fields exist on the structures they don't seem to be very effective tests for the new logic.

I did however test this manually by building and running with the new flags.

cPu1:
Thanks for the PR, @bhops. I skimmed through it and it LGTM, but we need to discuss this with the EKS team before merging. I have left a [comment](https://github.com/weaveworks/eksctl/issues/3718#issuecomment-971254071) on the issue.

bhops:
Thanks @cPu1. I’ve also left a comment on the issue but understand the hesitation as well. This can happily sit tight while the team decides what to do.

rothgar:
This works a bit different than `ec2-instance-selector` behavior and might be surprising to users. With ec2-instance-selector you can provide a comma separated list or a regex. In this case only regex or a single instance is supported.

Is there a reason for these advanced queries someone can't use a config file or a subshell?

For example, this works without needing to change `eksctl`
```
eksctl create cluster --instance-types $(ec2-instance-selector --service eks --deny-list 'c5.18xlarge,c5.24xlarge' -o one-line)
```

bhops:
> This works a bit different than `ec2-instance-selector` behavior and might be surprising to users. With ec2-instance-selector you can provide a comma separated list or a regex. In this case only regex or a single instance is supported.

Noted. The descriptions and help text specify it must be a regex but I can see why the inability to pass a comma separated list could be unexpected to users.

> Is there a reason for these advanced queries someone can't use a config file or a subshell?

Subshell would work. Not sure what you mean by a config file though? The use case that I was trying to support was actually to be able to pass these options via the cluster YAML file like so:

```yaml
         instanceSelector:
            cpuArchitecture: amd64
            gpus: 0
            memory: "32"
            vCPUs: 8
            deny: "t.*"
```

bhops:
Just to confirm, would this be considered an acceptable change if the CLI options were dropped, and the deny (and presumably allow) configuration could be passed as a list rather than an individual regex via the YAML configuration?

I believe this would result in more changes to the logic of eksctl since eksctl would have to parse the instance types returned by `ec2-instance-selector` rather than simply "passing through" the allow/deny option to `ec2-instance-selector`. Is that desired? The original reason this was flagged was that there was some hesitation behind adding even more functionality into `eksctl` and that feels like moving the selection logic into eksctl rather than letting `ec2-instance-selector` make the decision.

rothgar:
Sorry if I wasn't clear. I'm not a maintainer on eksctl so I have no say if the PR will be accepted.

My opinion was that we should try to keep the CLI interface simpler and any time someone wants to do more complex things they should use a config file. It'll help with long term maintenance and encourage better practices.

Either way I think ec2-instance-selector should be the library making the selection decision and if possible that complexity be left out of eksctl.

Skarlso:
`eksctl` agrees on the CLI part. We are trying to move away from too many CLI options as they are pretty confusing and there is a LOT of them now.

I also suggest to move this into a config file for better handling, sensible defaults and living documentation.

PR #4480, opened 2021-11-19
Karpenter install

Skarlso:
### Description

First half of https://github.com/weaveworks/eksctl/issues/4430. This PR adds the following: 

- Ability to install Helm Charts
- Ability to install Helm Repos
- The beginnings of managing the new CloudFormation stack that will be used to install Karpenter
- Unit testing around Karpenter ( I'm writing the integration test ( although it won't work at first ) )
- Documentation ( in the process of writing that too :D )

**Reviewers**: This is a draft PR, but I wanted to share the skeleton with you, because it's getting rather large as I write the integration test and the documentation and everything else... So I wanted to you to have an early bird access to it. ;)

### Checklist
- [ ] Added tests that cover your change (if possible)
- [ ] Added/modified documentation as required (such as the `README.md`, or the `userdocs` directory)
- [ ] Manually tested
- [ ] Made sure the title of the PR is a good description that can go into the release notes
- [ ] (Core team) Added labels for change area (e.g. `area/nodegroup`) and kind (e.g. `kind/improvement`)

### BONUS POINTS checklist: complete for good vibes and maybe prizes?! :exploding_head:
- [ ] Backfilled missing tests for code in same general area :tada:
- [ ] Refactored something and made the world a better place :star2:

Issue #4432, opened 2021-11-08
As a user of karpenter I want to be able to update the AWS resources associated to it


aclevername:
Follow up from https://github.com/weaveworks/eksctl/issues/4430

As a user of karpenter
I want to be able to update the AWS resources eksctl created for it
So that I can successfully upgrade the karpenter installation myself manually

## Background
As part of https://github.com/weaveworks/eksctl/issues/4430 we added support for:
- creating the various AWS resources required for karpenter
- installing the helm chart

As new versions of karpenter gets released it will require new AWS resources- e.g. new additions to the policies attached to the IAM role. I should be able to update the AWS resources eksctl created via a command, so that I can update my karpenter installation.

## Proposal
a new command:
```
eksctl utils update-karpenter-resources
```

*Naming open to suggestion*. It needs to convey that *it only updates the AWS resources, not karpenter itself!*. We don't want eksctl becoming a helm-wrapper package manager. This command should have all the default flags:
```
General flags:
  -c, --cluster string       EKS cluster name
  -r, --region string        AWS region
  -f, --config-file string   load configuration from a file (or stdin if set to '-')
      --approve              Apply the changes

AWS client flags:
  -p, --profile string   AWS credentials profile to use (overrides the AWS_PROFILE environment variable)

Common flags:
  -C, --color string   toggle colorized logs (valid options: true, false, fabulous) (default "true")
  -h, --help           help for this command
  -v, --verbose int    set log level, use 0 to silence, 4 for debugging and 5 for debugging with AWS debug logging (default 3)
```

As part of https://github.com/weaveworks/eksctl/issues/4430 we created the AWS resources in its separate stack. The implementation should just be to update the stack via a changset.

## Acceptance Criteria
Happy path :partying_face: 

- As an eksctl user
- Who has a pre-existing eksctl cluster with karpenter installed
- When I run `eksctl utils update-karpenter-resources -f config.yaml`
- I can inspect the cloudformation stack to see that AWS related karpenter resource are updated to the latest requirements
- I should find relevant docs for this

Himangini:
During PR review decide a proper name for this command

Skarlso:
Thinking about this, it might be a bit tricky. Which resources do we want to update? Right now, the Helm Chart and helm installation are not tracked as part of the stack or something like that. So even if you find the stack ( which is tagged ) those are only the IAM resources. The Helm install, of course, can't be tracked.

[...]

PR #4488, opened 2021-11-24
Create required AWS resources during karpenter installation

### LINKED ISSUE #4430: Add support for installing karpenter as part of eksctl create cluster
As a user of eksctl
I want to be able to have karpenter pre-installed on my cluster
So that I don't have to go through a complicated manual install (which involves creating various AWS resources and modifying resources created by eksctl)

## Background
[Karpenter](https://karpenter.sh/docs/getting-started/) is a just-in-time node auto-scalling tool from AWS. `Karpenter automatically launches just the right compute resources to handle your cluster's applications. It is designed to let you take full advantage of the cloud with fast and simple compute provisioning for Kubernetes clusters.`. 

Currently if a user wants to install Karpenter onto an eksctl cluster they have to (paraphrasing from https://karpenter.sh/docs/getting-started/):
1. Tag subnets in the created cluster with tags that Karpenter can use to auto-
discover them.
1. Create necessary IAM resources (instance profile, node role, and controller
policy) to allow Karpenter and the nodes it provisions to have permissions they
need.
1. Create IAMIdentityMapping within the cluster created to allow nodes
provisioned to connect to the cluster.
1.  Create an IAM Role for Service Accounts (IRSA) to allow the Karpenter
controller to provision nodes for the cluster.
1. Install the Karpenter Helm chart.

The user would then create a set of `Provisioner` CRDs to setup karpenter, e.g:
```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default
spec:
  requirements:
    - key: node.k8s.aws/capacity-type
      operator: In
      values: ["spot"]
  provider:
    instanceProfile: KarpenterNodeInstanceProfile-${CLUSTER_NAME}
    cluster:
      name: ${CLUSTER_NAME}
      endpoint: $(aws eks describe-cluster --name ${CLUSTER_NAME} --query "cluster.endpoint" --output json)
  ttlSecondsAfterEmpty: 30
```

Of these 5 steps, 4 of them already well-within scope for what eksctl can do (creating AWS resources and updating the iam-identitymapping config). Helm installation is a new mechanism- however pre-installing some resources is not new to eksctl, for example we support pre-installing [flux](https://eksctl.io/usage/gitops-v2/) and various addons (coredns etc).

## Proposal
A user should be able to provide within their cluster config:
```yaml
---
kind: ClusterConfig

metadata:
  name: cluster-1
  region: eu-north-1
  
karpernter:
  version: 0.1.0 # example semver version
```

This will indicate to eksctl that it needs to create all the AWS resources above, and then install the helm chart at the specific version.

Q: Why is the user providing the version? Shouldn't it just be `karpenter: true/false` and then we install the latest version?
A: Two reasons:
  - Karpenter is a fast-moving-project. They do regular breaking changes between minor versions, so a user will care about a particular Karpenter version they want to install as their `Provisioner` CRDs will be tied to them.
  - Karpenter has also suggested to us that they might at a reasonable frequency change the set of IAM policies required. This means that eksctl will need to be kept up-to-date with the set of policies required for a given version. E.g.:
     - Karpenter `0.1.0` requires policy `x`
     - Karpenter `0.2.0` requires policy `x & y`

## All resources must be created in a separate stack (*IMPORTANT*)
Like with IAMServiceAccounts, and Fargate, lets create any karpenter-related resources in a new cloudformation stack. Not as part of the cluster stack. This is so it can be updated separately from the cluster stack :smile: 

### Version pinning suggestion (just an idea)
We need to pin eksctl to a "max version" of Karpenter, as we know that the set of AWS resources we create will work for version X, but might not work for a newer version. E.g.:
- `eksctl 0.74.0` was created/tested against `karpenter 0.3.0`. Then a new karpernter helm chart `0.4.0` comes along, that requires some new policies. Rather than letting an eksctl user of `0.74.0` install `0.4.0`, we should instead error and suggest the user update to the latest eksctl version.

Karpenter have indicated to us they would be willing to (try as an experiment) open PRs against eksctl when AWS resources change/new version of karpenter are released. This is shifting the cost of knowing what AWS resources we need to create, along with what the max karpenter version is to them. If we go down this route we should document how they can update this, and make the code/tests/integration-tests as easy as possible to contribute to.

## Acceptance Criteria
Happy path :partying_face: 

- As an eksctl user
- Who configured `karpenter.version` to a supported karpenter version
- When I do `eksctl create cluster -f config.yaml`
- I can see that karpenter is installed (pods running successfully) and
- I should be able to successfully deploy a [Provisioner](https://karpenter.sh/docs/provisioner-crd/) (to sanity-check the installation works)
- I should be able to find relevant docs

Sad path :cry: : 

- As an eksctl user
- Who configured `karpenter.version` to an un-supported karpenter version
- When I do `eksctl create cluster -f config.yaml`
- I should get an error telling me that the version it not supported and
- A suggestion of what version I can use instead
- I should be able to find relevant docs

Goes without saying the above criteria should be supported with well-tested performant code.

Skarlso:
### Description

Installs the AWS resources for Karpenter.

Closes #4430

### Checklist
- [x] Manually tested


pkg/actions/karpenter/create_test.go
```diff
@@ -0,0 +1,219 @@
...
+				CTL:                ctl,
+				Config:             cfg,
+				KarpenterInstaller: fakeKarpenterInstaller,
+			}
+			Expect(install.Create()).To(Succeed())
+			Expect(fakeKarpenterInstaller.InstallCallCount()).To(Equal(1))
+		})
+		FWhen("DescribeSubnets fails", func() {
+			BeforeEach(func() {
+				p.MockEC2().On("DescribeSubnets", mock.Anything).Return(nil, errors.New("nope"))
```
Skarlso, line 120: @aclevername HALP :D This mock is not working for some reason, I don't understand. :D See anything wrong with it?

Skarlso:
Notes on testing. Using `defaultProvisioner: true` resulted in a nil error: https://github.com/aws/karpenter/issues/854

pkg/ctl/create/cluster.go
```diff
@@ -424,6 +417,75 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
...
+		return fmt.Errorf("failed to create client for auth config: %w", err)
+	}
+	identityArn := fmt.Sprintf("arn:aws:iam::%s:role/%s-%s", parsedARN.AccountID, builder.KarpenterNodeRoleName, cfg.Metadata.Name)
+	id, err := iam.NewIdentity(identityArn, authconfigmap.RoleNodeGroupUsername, authconfigmap.RoleNodeGroupGroups)
+	if err != nil {
+		return fmt.Errorf("failed to create new identity: %w", err)
+	}
+	if err := acm.AddIdentity(id); err != nil {
+		return fmt.Errorf("failed to add new identity: %w", err)
+	}
```
nikimanoledaki, line 475: are there unit tests for this? maybe these 2 sections could be refactored to test them more easily?

pkg/karpenter/karpenter.go
```diff
@@ -12,28 +12,37 @@ import (
 const (
 	// DefaultKarpenterNamespace default namespace for Karpenter
 	DefaultKarpenterNamespace = "karpenter"
+	// DefaultKarpenterServiceAccountName is the name of the service account which is needed for Karpenter
+	DefaultKarpenterServiceAccountName = "karpenter"
```
nikimanoledaki, line 16: these 2 could be renamed to `karpenter.DefaultNamespace` & `karpenter.DefaultServiceAccountName`

pkg/karpenter/karpenter.go
```diff
@@ -12,28 +12,37 @@ import (
...
-	karpenterHelmRepo         = "https://charts.karpenter.sh"
-	karpenterHelmChartName    = "karpenter/karpenter"
-	karpenterReleaseName      = "karpenter"
-	controllerClusterName     = "controller.clusterName"
-	controllerClusterEndpoint = "controller.clusterEndpoint"
-	createServiceAccount      = "serviceAccount.create"
-	addDefaultProvisioner     = "defaultProvisioner.create"
+	karpenterHelmRepo      = "https://charts.karpenter.sh"
+	karpenterHelmChartName = "karpenter/karpenter"
+	karpenterReleaseName   = "karpenter"
```
nikimanoledaki, line 20: same here - it seems redundant to include `karpenter` in the name of these vars since they are in pkg `karpenter`

pkg/karpenter/karpenter.go
```diff
@@ -12,28 +12,37 @@ import (
...
+	ClusterName          string
+	CreateServiceAccount bool
+	ClusterEndpoint      string
+	Version              string
 }
 
 // InstallKarpenter defines a functionality to install Karpenter.
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+//counterfeiter:generate -o fakes/fake_karpenter_installer.go . InstallKarpenter
 type InstallKarpenter interface {
```
nikimanoledaki, line 46: there's some repetition here too with `karpenter.InstallKarpeneter`, which could be omitted

integration/tests/karpenter/karpenter_test.go
```diff
@@ -0,0 +1,64 @@
...
+			cmd := params.EksctlCreateCmd.
+				WithArgs(
+					"cluster",
+					"--config-file=-",
+					"--verbose=4",
+				).
+				WithoutArg("--region", params.Region).
+				WithStdin(clusterutils.ReaderFromFile(clusterName, params.Region, "testdata/cluster-config.yaml"))
+
+			Expect(cmd).To(RunSuccessfully())
```
nikimanoledaki, line 61: is there any other assertion that could be added here to check that karpenter was installed successfully on the cluster?

pkg/actions/karpenter/tasks.go
```diff
@@ -0,0 +1,142 @@
...
+// if not, it will create them.
+func (k *karpenterTask) maybeUpdateSubnetTags() error {
+	var ids []string
+	for _, subnet := range k.cfg.VPC.Subnets.Private {
+		ids = append(ids, subnet.ID)
+	}
+	for _, subnet := range k.cfg.VPC.Subnets.Public {
+		ids = append(ids, subnet.ID)
+	}
+	sort.Strings(ids)
```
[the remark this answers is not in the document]
Skarlso: Because they are coming from a map.

pkg/cfn/builder/karpenter_test.go
```diff
@@ -0,0 +1,164 @@
...
+			CreateServiceAccount: api.Disabled(),
+		}
+	})
+
+	Context("AddAllResources", func() {
+		It("generates the correct CloudFormation template", func() {
+			krs := builder.NewKarpenterResourceSet(cfg)
+			Expect(krs.AddAllResources()).To(Succeed())
+			result, err := krs.RenderJSON()
+			fmt.Println(string(result))
```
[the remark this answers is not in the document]
Skarlso: Uuuu yes. :D nice catch. :D

pkg/karpenter/karpenter.go
Skarlso: This one has a history from the previous PR. Naming this wasn't super easy. :D So we decided to keep this one around. :) Check it out here: https://github.com/weaveworks/eksctl/pull/4480

Skarlso: However, if you have a better name PLEASE SHARE!! :D

integration/tests/karpenter/karpenter_test.go
Skarlso: Hmmm, hmmm. Check pod status could be added.

pkg/ctl/create/cluster.go
Skarlso: Nope, because the cluster wires real implementation. The only way without re-writing the whole thing to test was the integration test.

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
nikimanoledaki: right but why do they need to be ordered? 🤔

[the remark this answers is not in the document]
nikimanoledaki: not sure what I'm not seeing :')

pkg/actions/karpenter/karpenter.go
```diff
@@ -1,40 +1,65 @@
...
-	cfg                *api.ClusterConfig
-	clientSet          kubernetes.Interface
-	wait               WaitFunc
-	kubeProvider       eks.KubeProvider
-	karpenterInstaller karpenter.InstallKarpenter
+	StackManager       manager.StackManager
+	CTL                *eks.ClusterProvider
+	Config             *api.ClusterConfig
+	Wait               WaitFunc
+	KarpenterInstaller karpenter.InstallKarpenter
```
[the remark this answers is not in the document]
Skarlso: The whole naming around these is confusing. It could be better. Feel free to spew ideas at me. :D

pkg/ctl/create/cluster.go
Skarlso: Ah wait, that's not true. Maybe I can do something here.

Himangini:
Can you update the description/title for this PR please? i genuinely thought this PR is for #4432 from reading the title 😅

integration/tests/karpenter/karpenter_test.go
Skarlso: Added this one. I'm checking for Pods being healthy in `karpenter` namespace. 👍

Skarlso:
Go home link checker, you're drunk.

pkg/ctl/create/cluster.go
```diff
@@ -424,6 +417,75 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 	return nil
 }
 
+// installKarpenter prepares the environment for Karpenter, by creating the following resources:
```
Skarlso, line 420: Trying to add unit tests fro this next.

integration/tests/karpenter/karpenter_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+var params *tests.Params
+
+func init() {
+	testing.Init()
+	if err := api.Register(); err != nil {
+		panic(fmt.Errorf("unexpected error registering API scheme: %w", err))
+	}
+	params = tests.NewParams("")
```
cPu1, line 28: This needs to be given a prefix if you're planning to use the cluster name generated by `tests.NewParams`.

integration/tests/karpenter/karpenter_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+var _ = Describe("(Integration) Karpenter", func() {
+
+	var (
+		clusterName string
+	)
+
+	BeforeEach(func() {
+		clusterName = fmt.Sprintf("it-karpenter-%d", time.Now().Unix())
```
cPu1, line 42: Any reason why you are not using the generated `params.ClusterName`? Are you planning to create multiple clusters in this test suite?

Skarlso: Too long. :/ Errors out on the Role name creation scheme.

integration/tests/karpenter/karpenter_test.go
Skarlso: I'm not planning on that.

integration/tests/karpenter/karpenter_test.go
cPu1: It might be worth adding a comment here because most of the other tests rely on the cluster name generated by `tests.NewParams` :slightly_smiling_face:

pkg/ctl/create/cluster.go
Skarlso: @nikimanoledaki So I looked at this now for a long time, but in order to unit test this, I have to refactor a LOT of things.

While I think that has a lot of merit, doing that in this Pr would be a nightmare to review. Maybe this could be added to the testing refactor? Or I could create a separate ticket to refactor the `ctl` package to be unit testable.

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
Skarlso: @nikimanoledaki Can I resolve this? :)

pkg/actions/karpenter/karpenter.go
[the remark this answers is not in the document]
Himangini: how about `Install karpenter.InstallKarpenter` ? just suggesting tho, naming is hard :D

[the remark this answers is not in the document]
aclevername: do we need an `installer` struct here? If it only has the 1 func to create the task can just be a public func with a big `InstallOptions` struct for args

pkg/actions/karpenter/tasks.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+	tags := map[string]string{
+		api.KarpenterNameTag: name,
+	}
+	if err := k.stackManager.CreateStack(name, stack, tags, nil, errs); err != nil {
+		return fmt.Errorf("failed to create stack: %w", err)
+	}
+
+	if err := k.maybeUpdateSubnetTags(); err != nil {
```
aclevername, line 67:
```suggestion
	if err := k.ensureSubnetsHaveTags(); err != nil {
```

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
aclevername: >  Because the unit test wouldn't be able to match the calling order

annoying! You can get around this by doing something like:

```
                     p.MockEC2().On("DescribeSubnets", mock.Anything)..Run(func(args mock.Arguments) {
                          Expect(args).To(HaveLen(1))
                          Expect(args[0]).To(BeAssignableToTypeOf(&awseks.DescribeSubnetsInput{}))
                          Expect(args[0].(*awseks.DescribeSubnetsInput).Subnets).To(ConsistOf(.....) //unordered matcher
                      }.Return(&ec2.DescribeSubnetsOutput{
				Subnets: []*ec2.Subnet{
					{
						SubnetId: aws.String(privateSubnet1),
						VpcId:    aws.String(cfg.VPC.ID),
					},
					{
						SubnetId: aws.String(privateSubnet2),
						VpcId:    aws.String(cfg.VPC.ID),
						Tags: []*ec2.Tag{
							{
								Key:   aws.String("kubernetes.io/cluster/" + clusterName),
								Value: aws.String("shared"),
							},
						},
					},
					{
						SubnetId: aws.String(publicSubnet1),
						VpcId:    aws.String(cfg.VPC.ID),
					},
					{
						SubnetId: aws.String(publicSubnet2),
						VpcId:    aws.String(cfg.VPC.ID),
					},
				},
			}, nil)
```

>

pkg/actions/karpenter/tasks.go
```diff
@@ -0,0 +1,142 @@
...
+			Tags: []*ec2.Tag{
+				{
+					Key:   aws.String(clusterTag),
+					Value: aws.String(""),
+				},
+			},
+		}); err != nil {
+			return fmt.Errorf("failed to add tags for subnets: %w", err)
+		}
+	}
```
aclevername, line 125: nah fam thats a no from me

either 
```suggestion
                createTagsInput := &ec2.CreateTagsInput{
			Resources: aws.StringSlice(updateSubnets),
			Tags: []*ec2.Tag{
				{
					Key:   aws.String(clusterTag),
					Value: aws.String(""),
				},
			},
		}
		if _, err := k.ec2API.CreateTags(input); err != nil {
			return fmt.Errorf("failed to add tags for subnets: %w", err)
		}
	}
```

or 

```suggestion
		_, err := k.ec2API.CreateTags(&ec2.CreateTagsInput{
			Resources: aws.StringSlice(updateSubnets),
			Tags: []*ec2.Tag{
				{
					Key:   aws.String(clusterTag),
					Value: aws.String(""),
				},
			},
		})
		
		if err != nil {
			return fmt.Errorf("failed to add tags for subnets: %w", err)
		}
	}
```

aclevername, line 125: nah fam thats a no from me

either 
```suggestion
                createTagsInput := &ec2.CreateTagsInput{
			Resources: aws.StringSlice(updateSubnets),
			Tags: []*ec2.Tag{
				{
					Key:   aws.String(clusterTag),
					Value: aws.String(""),
				},
			},
		}
		if _, err := k.ec2API.CreateTags(input); err != nil {
			return fmt.Errorf("failed to add tags for subnets: %w", err)
		}
	}
```

or 

```suggestion
		_, err := k.ec2API.CreateTags(&ec2.CreateTagsInput{
			Resources: aws.StringSlice(updateSubnets),
			Tags: []*ec2.Tag{
				{
					Key:   aws.String(clusterTag),
					Value: aws.String(""),
				},
			},
		})
		
		if err != nil {
			return fmt.Errorf("failed to add tags for subnets: %w", err)
		}
	}
```

pkg/actions/karpenter/create.go
```diff
@@ -0,0 +1,7 @@
+package karpenter
+
+// Create creates a Karpenter installer task and waits for it to finish.
+func (i *Installer) Create() error {
+	taskTree := NewTasksToInstallKarpenter(i.Config, i.StackManager, i.CTL.Provider.EC2(), i.KarpenterInstaller)
+	return doTasks(taskTree)
+}
```
aclevername, line 7: could the tasks live in this file? I know we use the `tasks.go` conventioin in `pkg/action/addons` but thats because your either using tasks or the func directly. Here we only ever use the tasks, so I feel the two can go together. wdyt?

pkg/actions/karpenter/tasks.go
```diff
@@ -0,0 +1,142 @@
...
+	name := k.makeKarpenterStackName()
+
+	logger.Info("building nodegroup stack %q", name)
+	stack := builder.NewKarpenterResourceSet(k.cfg)
+	if err := stack.AddAllResources(); err != nil {
+		return err
+	}
+	tags := map[string]string{
+		api.KarpenterNameTag: name,
+	}
```
aclevername, line 62: might be worth having the karpenter version tag too? I think it will help debugging down the line

pkg/actions/karpenter/tasks.go
```diff
@@ -0,0 +1,142 @@
...
+		for _, tag := range subnet.Tags {
+			if aws.StringValue(tag.Key) == clusterTag {
+				hasTag = true
+				break
+			}
+		}
+		if !hasTag {
+			updateSubnets = append(updateSubnets, *subnet.SubnetId)
+		}
+	}
```
aclevername, line 111:
```suggestion
	var updateSubnets []string
	for _, subnet := range output.Subnets {
		for _, tag := range subnet.Tags {
			if aws.StringValue(tag.Key) == clusterTag {
				updateSubnets = append(updateSubnets, *subnet.SubnetId)
			}
		}
	}
```

keys are unique afaik so this should be okay

pkg/ctl/create/cluster.go
```diff
@@ -424,6 +417,75 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
...
+	if err != nil {
+		return fmt.Errorf("failed to create installer: %w", err)
+	}
+	if err := installer.Create(); err != nil {
+		return fmt.Errorf("failed to install Karpenter: %w", err)
+	}
+
+	return nil
+}
+
```
aclevername, line 488: can some of this just not live in `ctl/create/`? Might make unit testing easier

pkg/karpenter/karpenter.go
aclevername: `karpenter.ChartInstaller`?

aclevername: or `helm.KarpenterInstaller`

pkg/actions/karpenter/karpenter.go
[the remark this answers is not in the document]
Skarlso: I hate that option :(

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
Skarlso: Hmmm, hmmm. I like the self-verifying call order. But I'll consider this. :D

pkg/karpenter/karpenter.go
Skarlso: ChartInstaller! I like that one! I already have a package called `helm` so I can't use that. ChartInstaller isn't bad.

pkg/actions/karpenter/create.go
Skarlso: I'm thinking about when we add the update feature. But I'll have a thought about that.

pkg/actions/karpenter/tasks.go
Skarlso: The point here is to only update if the there is no such tag. We stop searching the tags once we find that it has the tag.

pkg/ctl/create/cluster.go
Skarlso: Pfff, maybe. There is a lot of context which would have to be passed around. But I can try.

pkg/actions/karpenter/karpenter.go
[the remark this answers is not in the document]
Skarlso: Let me elaborate. It's a lot harder to work around a public function and makes a for a rigid structure which can be a pain than to move away from. Especially since then I will have to have a local function as well actually which holds the correct one. And just adds complexity to things. I would rather have a slight naming problem then a bigger problem with a public function. :)

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
aclevername: I prefer this because I'd rather have a bit more complexity in the test than in the code. Also easy for someone to see the code and go "this is overly complicated, let me remove the sorting logic"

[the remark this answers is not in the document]
Skarlso: Okay, fine. :D I'll add it.

pkg/karpenter/karpenter.go
aclevername: it fits! go for it

pkg/actions/karpenter/create.go
aclevername: oh fair point. idm either way then

pkg/actions/karpenter/tasks.go
aclevername: oh sorry completely misread!

pkg/actions/karpenter/tasks.go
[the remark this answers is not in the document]
Skarlso: This is just horrible, to be honest. :D 

```
                          Expect(args[0].(*awseks.DescribeSubnetsInput).Subnets).To(ConsistOf(.....) //unordered matcher

``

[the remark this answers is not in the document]
Comment (Skarlso) [reply pkg/actions/karpenter/tasks.go]: Yeah, I hate this. I also have to change all of the CreateTag calls too because they contain the updateSubnetIds which can be in any order. Instead of doing all that in the tests I can simplify the whole thing with a nice and easy sort. Which is what I'm going to do. :) If you truly hate it, you can change it in an append PR. :D I will merge it. :D

File: pkg/ctl/create/cluster.go (hunk 18 above)
Comment (Skarlso) [reply pkg/ctl/create/cluster.go]: yay, this turned out nicely. Also, kudos to @nikimanoledaki for first flagging it as an issue and booo to me for not extracting it. :D Thank you both. :)

File: pkg/ctl/create/cluster.go [hunk 19]
```diff
@@ -413,6 +401,27 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
...
 
+// installKarpenter prepares the environment for Karpenter, by creating the following resources:
+// - iam roles and profiles
+// - service account
+// - identity mapping
+// then proceeds with installing Karpenter using Helm.
+func installKarpenter(ctl *eks.ClusterProvider, cfg *api.ClusterConfig, stackManager manager.StackManager, clientSet *kubeclient.Clientset) error {
+	if cfg.Karpenter == nil {
+		return nil
+	}
```
aclevername, line 412: do we validate elsewhere `.version` is set?

aclevername (approved):
1 comment- otherwise LGTM

pkg/ctl/create/cluster.go
Skarlso: Like, that it's a required field?

Skarlso: Huh, guess I don't check it. :O

aclevername: :smile:

Skarlso: Added :D

rothgar:
Should I be able to build the karpenter branch to test this?

Looking at the example config
```
kind: ClusterConfig
metadata:
  name: cluster-with-karpenter
  region: us-west-2
  version: '1.21'
karpenter:
  version: '0.5.1'
  createServiceAccount: true
```
Will that also enable oidc in the cluster to be used with the service account?

Should I be able to run karpenter on fargate by adding this?

```
fargateProfiles:
  - name: karpenter
    selectors:
    - namespace: karpenter
```

rothgar:
I'm assuming the fargate profile causes an error with the helm chart because it creates the namespace in the cluster

```
2021-12-08 15:22:15 [ℹ]  1 error(s) occurred while installing Karpenter, you may wish to check your Cluster for further information
2021-12-08 15:22:15 [✖]  failed to install Karpenter chart: failed to install chart: rendered manifests contain a resource that already exists. Unable to continue with install: ServiceAccount "karpenter" in namespace "karpenter" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: key "app.kubernetes.io/managed-by" must equal "Helm": current value is "eksctl"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "karpenter"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "karpenter"
```

rothgar:
I tried with this config

```
kind: ClusterConfig
metadata:
  name: karpenter
  region: us-east-2
  version: '1.21'
karpenter:
  version: '0.5.1'
  createServiceAccount: true
```
and got this error
```
2021-12-08 15:44:11 [ℹ]  1 error(s) occurred while installing Karpenter, you may wish to check your Cluster for further information
2021-12-08 15:44:11 [✖]  failed to install Karpenter chart: failed to install chart: rendered manifests contain a resource that already exists. Unable to continue with install: ServiceAccount "karpenter" in namespace "karpenter" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: key "app.kubernetes.io/managed-by" must equal "Helm": current value is "eksctl"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "karpenter"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "karpenter"
Error: failed to install Karpenter: failed to install Karpenter on cluster
```
We'll need to disable the service role in the helm chart if it's declared in the cluster config

Skarlso:
Hey @rothgar!

Thanks for chiming in! Wow that is something else entirely. :D

I didn't try Karpenter with Fargate. I hope it works without that. I'll try running it with your config. Did you try without `createServiceAccount: true`? 

Hm. I think there is a bug in which when this is set to true, I should skip creating the service account because Karpenter would like to do that. I think that's the bug you discovered. I'll check it out. Well done! :)

Skarlso:
Also, that's interesting. You don't have nodegroups defined. I'm not a 100% that will work since it won't be able to create the webhook and the controller.

Skarlso:
I believe I found a fix. The service account was created even if the createServiceAccount was enabled. I have a fix up in a sec.

Issue #4252, opened 2021-09-27
Replace golint with revive


cPu1:
[golint](https://github.com/golang/lint) is archived and no longer maintained. [revive](https://github.com/mgechev/revive) is a drop-in replacement for golint and is supported by golangci-lint.

PR #4489, opened 2021-11-26
Deprecate golint and apply some changes from revive

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4252

### Checklist
- [x] Manually tested


.golangci.yml
```diff
@@ -77,9 +77,30 @@ linters-settings:
...
-  golint:
-    # minimal confidence for issues, default is 0.8
-    min-confidence: 0.8
+  revive:
+    confidence: 0.8
+    severity: warning
+    errorCode: 0
+    warningCode: 0
+    rules:
+      - name: blank-imports
```
Skarlso, line 86: These are the golint defaults. They are defined one-by-one for clarity on what's being checked.
I removed `dot-import` rule. The rest I fixed. If we decide, we can enable some additional rules.

pkg/actions/identityproviders/wait.go
```diff
@@ -38,8 +39,5 @@ func (m *Manager) waitForUpdate(
 		clusterName,
 	)
 
-	if err := waiters.Wait(clusterName, msg, acceptors, newRequest, timeout, nil); err != nil {
```
Skarlso, line 41: For these and the rest. These are empty ifs. I know that some prefer to return explicitly, but these are empty statements. If they would add logging or at least wrap the error, I would also leave them be. As such, they add fluff and extra code lines for no reason. :)

aclevername (commented):
do we need to update the readme/dependencies file? Also do we need to update the build-image it/github actions so it has revive installed?

Skarlso:
>do we need to update the readme/dependencies file? Also do we need to update the build-image it/github actions so it has revive installed?

Good question... Docs: no. Image: I honestly don't know.... Do we?

aclevername:
> > do we need to update the readme/dependencies file? Also do we need to update the build-image it/github actions so it has revive installed?
> 
> Good question... Docs: no. Image: I honestly don't know.... Do we?

well presumably `make lint` on this branch will fail the github workflows? since revive isn't installed?

Are you sure we don't have to update docs/depencies? Isn't having this installed basiclly a requirement now for developing on eksctl? Check out the contributing doc

Skarlso:
I did checkout all docs. Nothing says you need golint installed. I guess it's an implicit requirement or naturally comes along.

Also, that's a good point on CI. We don't install golint anyways. I think that's done by goreleaser or the image comes installed with it, no idea. Besides... If that would be the case, my workflow would fail too, right? Because goreleaser right now on this branch IS running revive..

Skarlso:
Yep. It's installed by golangci-lint:

```
go: downloading github.com/mgechev/revive v1.1.0
```

aclevername:
> I did checkout all docs. Nothing says you need golint installed. I guess it's an implicit requirement or naturally comes along.

https://github.com/weaveworks/eksctl/blob/32cdc50072b8dd5dc22987f89bcdf8fa178504c0/.requirements#L4

This gets used in `make install-build-deps` which is referenced in the contributing doc :smile: 


> Also, that's a good point on CI. We don't install golint anyways. I think that's done by goreleaser or the image comes installed with it, no idea. Besides... If that would be the case, my workflow would fail too, right? Because goreleaser right now on this branch IS running revive..

Dockerimage installs golang-ci through the same install-build-dep above https://github.com/weaveworks/eksctl/blob/32cdc50072b8dd5dc22987f89bcdf8fa178504c0/build/docker/Dockerfile#L60

aclevername:
> > This gets used in make install-build-deps which is referenced in the contributing doc
> 
> ... and? Sorry, I'm not following. How is that relevant?

Docker file runs `make install-build-deps`,  `make install-build-deps` looks at the `.requirments` file and `runs go install <x>` for each line. Thats how `golangci-lint` gets installed in the dockerimage

Skarlso:
Discussed over slack. `golangci-lint` installs its own dependencies if defined.

```
/Users/skarlso/go/bin/golangci-lint run /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/backwards_compat --allow-parallel-runners --fast
integration/tests/backwards_compat/backwards_compatibility_test.go:4:1  revive  var-naming: don't use an underscore in package name
```
Locally:

```
➜  which revive
revive not found
```

aclevername (approved):
LGTM!

PR #4531, opened 2021-12-09
Do not create service account if Karpenter is enabled to do so

### LINKED ISSUE #4528: Karpenter fails installation with Fargate and/or with `createServiceAccount: true`
For more detail check the discussion here: https://github.com/weaveworks/eksctl/pull/4488#issuecomment-989288498

Skarlso:
### Description

Closes #4528 

Testing:

```
2021-12-09 13:13:58 [ℹ]  kubectl command should work with "/Users/skarlso/.kube/config", try 'kubectl get nodes'
2021-12-09 13:13:58 [✔]  EKS cluster "test-karpenter-sa-3" in "us-west-2" region is ready
➜  karpenter-stack
```

### TODO

- Add link to karpenter documentation for the default provisioner setup

### Checklist
- [ ] Manually tested


pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -325,3 +325,11 @@ func (c *ClusterConfig) SetDefaultFargateProfile() {
 		},
 	}
 }
+
+// SetDefaultKarpenterCreateServiceAccount disabled the create service account option for Karpenter if it's not
```
cPu1, line 329:
```suggestion
// SetDefaultKarpenterCreateServiceAccount disables the create service account option for Karpenter if it's not
```

pkg/actions/karpenter/create.go
```diff
@@ -27,18 +27,21 @@ func (i *Installer) Create() error {
...
-				Namespace: karpenter.DefaultNamespace,
+	i.Config.SetDefaultKarpenterCreateServiceAccount()
+	if api.IsDisabled(i.Config.Karpenter.CreateServiceAccount) {
+		karpenterServiceAccountTaskTree := i.StackManager.NewTasksToCreateIAMServiceAccounts([]*api.ClusterIAMServiceAccount{
+			{
+				ClusterIAMMeta: api.ClusterIAMMeta{
+					Name:      karpenter.DefaultServiceAccountName,
+					Namespace: karpenter.DefaultNamespace,
+				},
+				AttachRoleARN: roleArn,
```
aclevername, line 38: wait so if `IsDisabled`, we create it? I'm confused :joy:

Skarlso: Yes. `createServiceAccount` is set to false, i.e.: Disabled, we create an account. If it's true, we DON'T want to create an account because Karpenter will create it. :)

Skarlso:
@Skarlso  Todo to add unit tests for this.

rothgar:
This is mostly working for me with this config

```
kind: ClusterConfig
metadata:
  name: karpenter-3
  region: us-east-2
  version: '1.21'
iam:
  withOIDC: true
karpenter:
  version: '0.5.1'
  createServiceAccount: true

managedNodeGroups:
  - name: ng-1
    minSize: 1
    desiredCapacity: 1
    maxSize: 2
    amiFamily: AmazonLinux2
```

The cluster created and karpenter is running but there was no provisioner (which was a bit of a surprise) and when I created workloads Karpenter didn't add nodes to the cluster. I'm still troubleshooting why it's not working but at least the cluster is created properly.

Skarlso:
@rothgar We don't create a default provisioner. The helm chart used to do that, but it doesn't anymore.

Skarlso:
We still have to figure out what will happen and how we'll deal with this. But that's a completely different story. For now, neither Karpenter, nor `eksctl` creates the default provisioner.

pkg/actions/karpenter/create.go
aclevername: ohh I see. Its not a "do you want eksctl to creates service account", its "do you want to create service account? otherwise eksctl will"

pkg/actions/karpenter/create.go
```diff
@@ -27,18 +27,21 @@ func (i *Installer) Create() error {
 			return i.ClientSet, nil
 		},
 	}
-	karpenterServiceAccountTaskTree := i.StackManager.NewTasksToCreateIAMServiceAccounts([]*api.ClusterIAMServiceAccount{
-		{
-			ClusterIAMMeta: api.ClusterIAMMeta{
-				Name:      karpenter.DefaultServiceAccountName,
-				Namespace: karpenter.DefaultNamespace,
+	i.Config.SetDefaultKarpenterCreateServiceAccount()
```
aclevername, line 30: this feels... a bit odd. Shouldn't this be done before the call to `Install`? Like in `ctl` we load the config, validate it and then call `Create`. This feels like it falls under the validation/defaulting part

Skarlso: Yeah I could call it there I guess. In the validator maybe?

pkg/karpenter/providers/helm.go
```diff
@@ -23,5 +23,10 @@ type HelmInstaller interface {
 	AddRepo(repoURL string, release string) error
 	// InstallChart takes a releaseName's name and a chart name and installs it. If namespace is not empty
 	// it will install into that namespace and create the namespace. Version is required.
-	InstallChart(ctx context.Context, releaseName, chartName, namespace, version string, values map[string]interface{}) error
+	InstallChart(
+		ctx context.Context,
+		releaseName, chartName, namespace, version string,
+		createNamespace bool,
+		values map[string]interface{},
+	) error
```
aclevername, line 31: mhmm perhaps worth chucking into a struct? Or can some of it been on the `helmInstaller` struct? (e.g. ctx)?

pkg/karpenter/providers/helm/helm_test.go
```diff
@@ -196,7 +254,15 @@ var _ = Describe("HelmInstaller", func() {
...
-				Expect(installerUnderTest.InstallChart(context.Background(), "karpenter", "karpenter/karpenter", "karpenter", "0.4.3", values)).To(MatchError(ContainSubstring("failed to install chart: failed to install CRD crds/karpenter.sh_provisioners.yaml: nope")))
+				Expect(installerUnderTest.InstallChart(
+					context.Background(),
+					"karpenter",
+					"karpenter/karpenter",
+					"karpenter",
+					"0.4.3",
+					true,
+					values,
+				)).To(MatchError(ContainSubstring("failed to install chart: failed to install CRD crds/karpenter.sh_provisioners.yaml: nope")))
```
aclevername, line 265: This reads a bit funny `Expect(myFuncCall()).To(MatchError`. Its not as clean as `Expect(err).To(MatchErr` or `Expect(funcCall).To(Succeed())`. worth splitting out the `err := ` outside of the expectation?

Skarlso: I don't mind

pkg/karpenter/providers/helm.go
Skarlso: Yeah thought about it too. 🤔 You just confirmed that it's way too much. :D

pkg/actions/karpenter/create.go
aclevername: yeah. We have a `defaulter` logic somewhere, would fit in there. might be a sub part of validater

pkg/karpenter/providers/helm/helm_test.go
aclevername: plz :smile:

Skarlso: I mean, I don't mind doing it. :D :D :D Sorry.

pkg/karpenter/providers/helm.go
Skarlso: The only problem is that I can't import anything from here, otherwise it's a circular dependency. So I'm creating a models package or something which can define the arguments. Which sounds a bit silly... But I don't really have a choice. Unless I define the interface in the same package as well. Which doesn't work again, because of the fake importing it to check validity which is something again, I can't avoid.

aclevername: where would the circular dependency be? Sounds like a smell if just adding a struct argument causes this :thinking:

Skarlso: Also, I would have to rename the Installer again to something different... :D

Skarlso: The fake, generates a type check. And since it's imported by the test, it can't import the same package.

I guess I can put the test into `helm_test`.

Skarlso: Never mind. I messed something else up.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -325,3 +325,11 @@ func (c *ClusterConfig) SetDefaultFargateProfile() {
...
 	}
 }
+
+// SetDefaultKarpenterCreateServiceAccount disables the create service account option for Karpenter if it's not
+// defined in the configuration file explicitly.
+func (c *ClusterConfig) SetDefaultKarpenterCreateServiceAccount() {
+	if c.Karpenter.CreateServiceAccount == nil {
+		c.Karpenter.CreateServiceAccount = Disabled()
+	}
+}
```
cPu1, line 335: I find the naming a bit verbose. WDYT about calling it just `SetKarpenterDefaults`?

pkg/ctl/cmdutils/configfile.go
```diff
@@ -272,7 +272,9 @@ func NewCreateClusterLoader(cmd *Cmd, ngFilter *filter.NodeGroupFilter, ng *api.
 				}
 			}
 		}
-
+		if clusterConfig.Karpenter != nil {
+			clusterConfig.SetDefaultKarpenterCreateServiceAccount()
+		}
```
cPu1, line 277: I would move this to `SetClusterConfigDefaults` which would allow you to perform defaulting in your tests without having to read a ClusterConfig file.

Skarlso: The default is having `Karpenter` as nil.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -325,3 +325,11 @@ func (c *ClusterConfig) SetDefaultFargateProfile() {
 		},
 	}
 }
+
+// SetDefaultKarpenterCreateServiceAccount disables the create service account option for Karpenter if it's not
+// defined in the configuration file explicitly.
+func (c *ClusterConfig) SetDefaultKarpenterCreateServiceAccount() {
```
cPu1, line 331: I find the naming a bit verbose. WDYT about calling it just `SetKarpenterDefaults`?

pkg/ctl/cmdutils/configfile.go
cPu1: You can still add the same logic in `SetClusterConfigDefaults`.

Skarlso: Alright, works for me

pkg/actions/karpenter/karpenter.go
```diff
@@ -45,6 +45,7 @@ func NewInstaller(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, stackManager
 		CreateServiceAccount: api.IsEnabled(cfg.Karpenter.CreateServiceAccount),
 		ClusterEndpoint:      cfg.Status.Endpoint,
 		Version:              cfg.Karpenter.Version,
+		CreateNamespace:      len(cfg.FargateProfiles) == 0,
```
aclevername, line 48: :thinking: whats this about? also missing a unit test

pkg/karpenter/karpenter_test.go
```diff
@@ -59,5 +59,46 @@ var _ = Describe("Install", func() {
...
+						Namespace:            "karpenter",
+						ClusterName:          "test-cluster",
+						CreateServiceAccount: true,
+						ClusterEndpoint:      "https://endpoint.com",
+						Version:              "0.4.3",
+						CreateNamespace:      true,
+					},
+				}
+			})
+			It("will not create a namespace", func() {
```
aclevername, line 97:
```suggestion
			It("will the namespace", func() {
```

pkg/actions/karpenter/karpenter.go
Skarlso: No it's not. :)

Also, basically, when a fargate profiles is active, it will create the `karpenter` namespace, hence, we should tell helm not to create it itself.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -147,6 +147,23 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
...
+func validateKarpenterConfig(cfg *ClusterConfig) error {
+	if cfg.Karpenter == nil {
+		return nil
+	}
+	if cfg.Karpenter.Version == "" {
+		return errors.New("version field is required if installing Karpenter is enabled")
+	}
+	if len(cfg.FargateProfiles) > 0 && cfg.IAM == nil || (cfg.IAM != nil && IsDisabled(cfg.IAM.WithOIDC)) {
+		return errors.New("OIDC must be set with Karpenter and Fargate Profiles")
+	}
```
cPu1, line 166: Does this condition mean that `withOIDC` must be enabled when both `fargateProfiles` and `karpenter` are set?

pkg/karpenter/karpenter_test.go
Skarlso: Hm? I'm testing that if create namespace is `false` it should **not** create the namespace.

pkg/apis/eksctl.io/v1alpha5/validation.go
Skarlso: Yes.

cPu1: This check will throw an error if `fargateProfiles` is unset and `withOIDC` is disabled, preventing users from using Karpenter if `withOIDC` is not enabled. For example, this standard Karpenter `ClusterConfig` will fail:

```yaml
karpenter:
  version: <>
iam:
  withOIDC: false # or nil
```

cPu1: This should be:

```suggestion
	if len(cfg.FargateProfiles) > 0 && IsDisabled(cfg.IAM.WithOIDC) {
		return errors.New("iam.withOIDC must be set with Karpenter and Fargate Profiles")
	}
```

The nil check for `cfg.IAM` can be omitted as `SetClusterConfigDefaults` ensures that `ClusterConfig.IAM` is always set.

Skarlso: Damn!! Very nice spot Chetan. :) I put the `(` in the wrong place. I meant to put it before the cfg.Iam part. :)

Skarlso: Even better :)

pkg/karpenter/karpenter_test.go
aclevername, line 97: are these supposed to both have the same `It`?

aclevername: opened a new comment

pkg/apis/eksctl.io/v1alpha5/validation.go
cPu1: I was going to suggest adding a unit test for the happy path, but I saw that you just pushed the changes for that.

pkg/actions/karpenter/karpenter.go
aclevername: > No it's not. :)

I can't see a corresponding change in ` pkg/actions/karpenter/create_test.go `

pkg/apis/eksctl.io/v1alpha5/validation.go
Skarlso: Yep. Thanks. :)

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -147,6 +147,23 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
...
+
+func validateKarpenterConfig(cfg *ClusterConfig) error {
+	if cfg.Karpenter == nil {
+		return nil
+	}
+	if cfg.Karpenter.Version == "" {
+		return errors.New("version field is required if installing Karpenter is enabled")
+	}
+	if len(cfg.FargateProfiles) > 0 && IsDisabled(cfg.IAM.WithOIDC) {
+		return errors.New("OIDC must be set with Karpenter and Fargate Profiles")
```
cPu1, line 165: It'd help users if we make the error message more descriptive. 
```suggestion
		return errors.New("iam.withOIDC must be enabled with Karpenter and Fargate profiles")
```
This would tell users exactly what they need to enable if they run into this error.

pkg/karpenter/karpenter_test.go
Skarlso: Uh. What in the. I was looking at the wrong test. This day man... I see now what you changed. :/

pkg/apis/eksctl.io/v1alpha5/validation.go
Skarlso: Sounds good.

pkg/karpenter/karpenter_test.go
Skarlso: No :(

pkg/actions/karpenter/karpenter.go
Skarlso: @aclevername Ah yes. The reason for that is that I'm using a fake Karpenter here and the arguments are just the context. Anything else is set up through `NewKarpenterInstaller` which I can't check here.

Meaning, I can't test that there, is why it's tested in here. 

I can try and add an integration test, but that requires a new cluster which would increase the runtime by another 20-30 minutes.

Skarlso: I guess I can set up a real Karpenter installer with a fake helm installer and veryify that the fake helm installer was called with Namespace off.

Skarlso: Actually, no. :D because I'm setting this up in `NewInstaller` which wires up the correct modules.

Skarlso: Hmmm, maybe I can if I can make helm driver to be `in Memory` for testing. 🤔

Skarlso: Nope, sorry. Can't do it. I thought maybe I can change the helm installer in between, but karpenter installer here is an interface.

What I can do is maybe shift to logic to testable place. Something like, inside Karpenter. But that didn't have access to the config so far, because it didn't needed it. I guess now it does. :)

Skarlso: I've put the choosing logic into Karpenter itself, so now I can just pass in the whole config and test the fargate profiles len logic in there.

Skarlso: @aclevername Okay, sorry for the long rants... How do you find the new changes? I shifted the config processing into Karpenter instead of the `New*` function.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -147,6 +147,23 @@ func ValidateClusterConfig(cfg *ClusterConfig) error {
...
+func validateKarpenterConfig(cfg *ClusterConfig) error {
+	if cfg.Karpenter == nil {
+		return nil
+	}
+	if cfg.Karpenter.Version == "" {
+		return errors.New("version field is required if installing Karpenter is enabled")
+	}
+	if len(cfg.FargateProfiles) > 0 && IsDisabled(cfg.IAM.WithOIDC) {
+		return errors.New("iam.withOIDC must be enabled with Karpenter and Fargate profiles")
+	}
```
aclevername, line 166: isn't oidc just always required? Since we create the IRSA?

aclevername (approved):
lgtm- 1 comment

pkg/apis/eksctl.io/v1alpha5/validation.go
cPu1: Since the default behaviour for Karpenter is to create service accounts for IRSA, we should disallow setting `withOIDC` to false unless `karpenter.CreateServiceAccount` is explicitly set to `true`.

Skarlso: We had a chat about this, and what we'll do is, require oidc always, and in case createServiceAccount is set to true, we pass in the created Role ARN to the helm chart as a service account annotation defined by spec here: https://github.com/aws/karpenter/blob/main/charts/karpenter/values.yaml#L7

Skarlso: It just makes our lives easier.

cPu1: Makes sense, but we don't seem to have validation that disallows disabling `iam.withOIDC` when Karpenter is enabled. 

eksctl will still allow this config:

```yaml
kind: ClusterConfig

metadata:
  name: cluster-with-karpenter
  region: us-west-2
  version: '1.20'

karpenter:
  version: '0.4.3'

iam:
  withOIDC: false
```

Skarlso: It's not yet implemented, correct. :)

Skarlso: that is what I'm doing right now, sorry, that one might have not been clear

pkg/actions/karpenter/karpenter.go
aclevername: lgtm

Skarlso:
```
2021-12-14 18:53:58 [ℹ]  waiting for CloudFormation stack "eksctl-test-karpenter-17-karpenter"
2021-12-14 18:54:01 [ℹ]  kubectl command should work with "/Users/skarlso/.kube/config", try 'kubectl get nodes'
2021-12-14 18:54:01 [✔]  EKS cluster "test-karpenter-17" in "us-west-2" region is ready
➜  karpenter-stack
---
NAMESPACE     NAME                                    READY   STATUS    RESTARTS   AGE
karpenter     karpenter-controller-5dbd7f88dc-fm87d   1/1     Running   0          4m1s
karpenter     karpenter-webhook-788544cbb4-nn49h      1/1     Running   0          4m1s
kube-system   aws-node-pspkx                          1/1     Running   1          18m
kube-system   coredns-86d9946576-lrkm2                1/1     Running   1          28m
kube-system   coredns-86d9946576-slzrt                1/1     Running   1          28m
kube-system   kube-proxy-5rz2l                        1/1     Running   1          18m
```


Successful cluster creation. Services seem to be happy. And I can see the service account with the right role. 

And helm apparently accepted the annotation value nicely
Screenshot 2021-12-14 at 18 57 23

After applying the provisioner though:

```
2021-12-14T17:59:51.797Z    ERROR    controller.controller.provisioning    Reconciler error    {"commit": "6984094", "reconciler group": "karpenter.sh", "reconciler kind": "Provisioner", "name": "default", "namespace": "", "error": "fetching instance types using ec2.DescribeInstanceTypes, WebIdentityErr: failed to retrieve credentials\ncaused by: ValidationError: Request ARN is invalid\n\tstatus code: 400, request id: 5ae5af1e-838e-4c71-ad4e-c1618d99db83"}
```

humm

Skarlso:
Finally, getting somewhere

```
2021-12-15T15:19:56.170Z        ERROR   controller.controller.metrics   Reconciler error        {"commit": "6984094", "reconciler group": "karpenter.sh", "reconciler kind": "Provisioner", "name": "default", "namespace": "", "error": "fetching instance types using ec2.DescribeInstanceTypes, UnauthorizedOperation: You are not authorized to perform this operation.\n\tstatus code: 403, request id: a0a083f1-5847-47b6-b352-d3939d43291b"}

```

The service account is probably missing some access rights.

Skarlso:
Finally! Managed to make it work!

```
2021-12-15T16:36:09.781Z        INFO    controller.provisioning Starting provisioner    {"commit": "6984094", "provisioner": "default"}
2021-12-15T16:36:09.781Z        INFO    controller.provisioning Waiting for unschedulable pods  {"commit": "6984094", "provisioner": "default"}
```

Now, going to update the unit tests.

Skarlso:
Running test with service account disabled. Should also work. I think I might need to update the policy document on that too..

Skarlso:
I found a timing issue where the policy is needed but I haven't created it yet...

pkg/actions/karpenter/create.go
```diff
@@ -21,24 +21,33 @@ func (i *Installer) Create() error {
...
+	if api.IsDisabled(i.Config.Karpenter.CreateServiceAccount) {
+		policyArn := fmt.Sprintf("arn:aws:iam::%s:policy/%s-%s", parsedARN.AccountID, builder.KarpenterManagedPolicy, i.Config.Metadata.Name)
+		iamServiceAccount.AttachPolicyARNs = []string{policyArn}
+	} else {
+		// Create the service account role only.
+		roleName := fmt.Sprintf("eksctl-%s-iamservice-role", i.Config.Metadata.Name)
+		roleARN = fmt.Sprintf("arn:aws:iam::%s:role/%s", parsedARN.AccountID, roleName)
+		iamServiceAccount.RoleOnly = api.Enabled()
+		iamServiceAccount.RoleName = roleName
+		iamServiceAccount.AttachPolicy = makePolicyDocument()
```
aclevername, line 45: why does one scenario use `iamServiceAccount.AttachPolicyARNs = []string{policyArn}` and the other `iamServiceAccount.AttachPolicy = makePolicyDocument()` :thinking: shouldn't it be the same policies everytime?

Skarlso: That's a good point. I guess we could use it since it has all the permissions and I could throw out the makePolicy thing. It was using it before I realised that I'm attaching a role when I should be attaching a policy instead.

But now, I'm sort of in a conundrum, where the policy is required, but it's actually only created after I'm getting to this point. And now I'm in a trap, because I can't create the Karpenter stuff, unless the policies exist. But I'm installing the Karpenter thing via a task separatly because I want a stack and now I have to refactor the whole thing and take out the Helm install part from the cloudformation stack part.

So create the CF stack with the roles and policies, add the mapping, add the accoutn, and run the helm install. I don't remember why I did it like this in the first place, but I do remember getting an error that something didn't exist and it needs to happen before that. I even have a comment about that somewhere. :D

Skarlso: I found the comment, but it's so odd. I must have messed something up. Anyways, I understand things better now, so I'll refactor around it.

Skarlso:
Finally, managed to make it work when we create the service account too and removed a bunch of things that aren't necessary, and restructure the CF stack creation and service account creation order.

Skarlso:
Karpenter creating the service account ALSO WORKS:

```
2021-12-16T15:30:27.660Z        INFO    controller.provisioning Starting provisioner    {"commit": "6984094", "provisioner": "default"}
2021-12-16T15:30:27.660Z        INFO    controller.provisioning Waiting for unschedulable pods  {"commit": "6984094", "provisioner": "default"}
```

aclevername (approved):

PR #4568, opened 2022-01-05
Failing early if missing flux binary during cluster creation

### LINKED ISSUE #4516: Flux binary existence is not checked prior to running `create cluster` with flux
`eksctl create cluster -f cluster.yaml`

**What happened?**

`[!]  exec: "flux": executable file not found in $PATH`

**How to reproduce it?**
See above
It's obvious that flux will fail if it's not in path.

**Versions**

```
$ eksctl info
eksctl version: 0.76.0
kubectl version: v1.21.2-13+d2965f0db10712
OS: linux
```

Skarlso:
### Description

Closes #4516 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/create/cluster.go
+++ b/pkg/ctl/create/cluster.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/aws/amazon-ec2-instance-selector/v2/pkg/selector"
+
 	"github.com/weaveworks/eksctl/pkg/kops"
 	"github.com/weaveworks/eksctl/pkg/utils"
 
@@ -195,6 +197,16 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 	if err := eks.ValidateFeatureCompatibility(cfg, kubeNodeGroups); err != nil {
 		return err
 	}
+
+	// Check if flux binary exists early in the process, so it doesn't fail at the end when the cluster
+	// has already been created with a missing flux binary error which should have been caught earlier.
+	// Note: we aren't running PreFlight here, we just check for the binary.
+	if cfg.HasGitOpsFluxConfigured() {
+		if _, err := exec.LookPath("flux"); err != nil {
+			return fmt.Errorf("flux binary is required when gitops configuration is set: %w", err)
+		}
+	}
+
 	if params.InstallWindowsVPCController {
 		if !eks.SupportsWindowsWorkloads(kubeNodeGroups) {
 			return errors.New("running Windows workloads requires having both Windows and Linux (AmazonLinux2) node groups")


Skarlso:
Not sure how to cover this with a test, since we don't really have anything to mock the os call and local environment does have flux... Maybe I can mess with the environment end clear the `PATH` variable....

cPu1:
>Maybe I can mess with the environment end clear the `PATH` variable....

That might also result in other unintended side effects.

Adding a test for this isn't straightforward if we're using `exec.LookPath`, so I have no objections to not having a test for this change since the validation is quite straightforward.

Skarlso:
@cPu1 What do you think about this... I added a small test in the gitops integration test which looks something like this:

```go
		FWhen("flux binary is not on path", func() {
			It("should fail early", func() {
				cmd := params.EksctlEnableCmd.WithArgs(
					"flux",
					"--config-file", "-",
				).WithStdin(clusterutils.Reader(clusterConfig))
				cmd = cmd.WithCleanEnv()
				session := cmd.Run()
				Expect(session.ExitCode()).To(Equal(1))
				Expect(string(session.Out.Contents())).To(ContainSubstring("asdf"))
			})
		})
```

So I'm using `WithCleanEnv` that should do the trick I believe. :) WDYT?

pkg/ctl/create/cluster.go
```diff
@@ -195,6 +197,17 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 	if err := eks.ValidateFeatureCompatibility(cfg, kubeNodeGroups); err != nil {
 		return err
 	}
+
+	// Check if flux binary exists early in the process, so it doesn't fail at the end when the cluster
+	// has already been created with a missing flux binary error which should have been caught earlier.
+	// Note: we aren't running PreFlight here, we just check for the binary.
+	if cfg.HasGitOpsFluxConfigured() {
+		if _, err := exec.LookPath("flux"); err != nil {
+			logger.Warning(err.Error())
```
cPu1, line 206: This message may not make sense on its own. It'd be more meaningful to users if it's added to the error below.

pkg/ctl/create/cluster.go
```diff
@@ -195,6 +197,17 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
...
+			return errors.New("flux binary is required when gitops configuration is set")
```
cPu1, line 207: Based on the suggestion above:
```suggestion
			return fmt.Errorf("flux binary is required when gitops configuration is set: %w", err)
```

Skarlso: Yeah sounds good

cPu1 (approved):
Quick fix! I left a minor suggestion but otherwise LGTM. :tada:

Issue #4569, opened 2022-01-05
Setting PublicAccessCIDRs with managedNodeGroups timeouts during creation


SamLR:
**What were you trying to accomplish?**

Trying to create a new cluster with:
* Private managed node group
* Public access CIDR set on the VPC

This is a minimal (I think) version of the set up that produces the error.

I think this is due to a race condition or similar in execution order as it only occurs during creation and can be worked around by commenting out the `publicAccessCIDRs` attribute during `create` and then uncommenting and running a subsequent `eksctl utils set-public-access-cidrs --approve -f "ekstctl-test.yaml"`.

If the `managedNodeGroups` attribute is not present then `create` will succeeed as expected. I've not tested this with `nodeGroups`.

**What happened?**

`eksctl create cluster -f eksctl-test.yaml` failed with:
```
2022-01-05 12:30:23 [✖]  waiting for CloudFormation stack "eksctl-public-access-test-nodegroup-ng-1": RequestCanceled: waiter context canceled
caused by: context deadline exceeded
```

**How to reproduce it?**

Using the config file:
```yaml
# eksctl-test.yaml
apiVersion: "eksctl.io/v1alpha5"
metadata:
  name: "public-access-test"
  region: "us-east-2"
kind: "ClusterConfig"
vpc:
  publicAccessCIDRs:
  - "[my ip]/32"
managedNodeGroups:
- desiredCapacity: 2
  instanceType: "m5.xlarge"
  name: "ng-1"
  privateNetworking: true
```

Run:
```
$ eksctl create cluster -f eksctl-test.yaml -v 4
2022-01-05 11:50:27 [ℹ]  eksctl version 0.77.0

...[snip normal output]...

2022-01-05 12:30:18 [ℹ]  waiting for CloudFormation stack "eksctl-public-access-test-nodegroup-ng-1"
2022-01-05 12:30:22 [✖]  unexpected status "CREATE_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-public-access-test-nodegroup-ng-1"
2022-01-05 12:30:22 [ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
2022-01-05 12:30:23 [!]  1 error(s) occurred and cluster hasn't been created properly, you may wish to check CloudFormation console
2022-01-05 12:30:23 [ℹ]  to cleanup resources, run 'eksctl delete cluster --region=us-east-2 --name=public-access-test'
[...]

PR #4576, opened 2022-01-06
Add warning to cluster access about publicAccessCIDRs

SamLR:
### Description

Update the docs to add a warning about a non-obvious interaction between `publicAccessCIDRs` and `nodeGroups`.

Closes https://github.com/weaveworks/eksctl/issues/4569


userdocs/src/usage/vpc-cluster-access.md
```diff
@@ -68,6 +68,10 @@ To update the restrictions using a `ClusterConfig` file, set the new CIDRs in `v
 eksctl utils set-public-access-cidrs -f config.yaml
 ```
 
+!!!warning
+    If setting `publicAccessCIDRs` and creating nodegroups `privateAccess` should be set to `true` to avoid
+    `context deadline exceeded` errors.
```
Skarlso, line 73: This should properly explain why the context deadline exceeds. Also, please link the created issue that you closed previously. We usually don't close issues by hand if they require a PR. We let github do its thing. :) 
https://github.com/weaveworks/eksctl/issues/4569

Skarlso (changes requested):
Hey Sam! Thanks for your Pr. I have a slight suggestion, then this will be okay. :)

cPu1 (approved):

userdocs/src/usage/vpc-cluster-access.md
SamLR: @Skarlso done : )

Also sorry about closing the other PR.

userdocs/src/usage/vpc-cluster-access.md
```diff
@@ -68,6 +68,10 @@ To update the restrictions using a `ClusterConfig` file, set the new CIDRs in `v
 eksctl utils set-public-access-cidrs -f config.yaml
 ```
 
+!!!warning
+    If setting `publicAccessCIDRs` and creating nodegroups `privateAccess` should be set to `true` to avoid
+    `context deadline exceeded` errors due to the nodes being unable to join the cluster.
```
Skarlso, line 73: I was more thinking in the way of WHY it's failing or what it will cause. So something like

```
When setting `publicAccessCIDRs` and creating nodegroups, `privateAccess` must be set to `true` otherwise nodes will be unable to join the cluster due to their ip addresses missing from public access CIRDs.
```

Because when private access is enabled, the nodes will join in a different way. But that's stretching it. :) So the context deadline is irrelevant, an implementation detail. The important part is what happens and why if that makes sense :)

SamLR: The reason I was leaving the context deadline in there is that that's the error people will see (as far as I can tell it's the only consistent error, I've sometimes seen "node unable to join cluster" but not always; I assume depending on where the deadline fires).

I can add in more detail as to what's going on but I think including the error is important as I couldn't find anything about this when I hit it

Skarlso: I understand. In that case, add it to the troubleshooting section? The error is a detail which I think is not something you would care about while reading the documentation and going through it step by step, right? Especially if it's not always context deadling exceeded but node unable to join. :)

SamLR: `deadline exceeded` is always the error I've seen. I've just seen different errors previously to it (e.g. [here](https://gist.github.com/SamLR/a88149789fe46cc658fa8f27b0949f74) it errors with `unexpected status "CREATE_IN_PROGRESS"`; other times I've seen `unable to join`).

I can move it if you think that's a more useful location for it but I'd assume most people will encounter the `publicAccessCIDRs` option here so it makes sense to have the warning here where they'll find it before they trigger the error than providing it to them after.

Skarlso: Meh, it's fine. :D

Skarlso (approved):

Skarlso:
will merge this after we are done with the release. Thanks! :)

Issue #4577, opened 2022-01-06
flux integration test is failing from time to time


Skarlso:
The latest test run failed with this error. We haven't encountered this one yet.

```
2022-01-06T11:05:26.3318719Z [13] error: unable to recognize "/tmp/flux-bootstrap-2411656037/flux-system": no matches for kind "Kustomization" in version "kustomize.toolkit.fluxcd.io/v1beta2"
2022-01-06T11:05:26.3384931Z [13] ‚úó exit status 1
2022-01-06T11:05:26.3427237Z [13] 2022-01-06 11:05:26 [‚Ñπ]  Flux v2 failed to install successfully. check configuration and re-run `eksctl enable flux`
```

Skarlso:
Might be something like https://github.com/weaveworks/eksctl/issues/4395 but there is no deadline exceeded error. But it's very suspicious since there was no change in the code.

Skarlso:
This is most likely due to outdated flux CLI. I'll update it.

Skarlso:
https://github.com/weaveworks/eksctl/blob/main/build/docker/Dockerfile#L44-L47

Yet the integration test is saying:

```
2022-01-06T11:05:16.7568907Z [13] ‚úó flux 0.17.0 <0.24.1 (new version is available, please upgrade)
```

Skarlso:
Ok, so the solution is to switch to using the Flux Github Action to install flux, instead of relying on building the runtime image by hand and pushing it.

Skarlso:
Circleci doesn't need the flux thing any longer. The integration test is now running on Github actions fully which is the only one needing flux to be in the container.

Skarlso:
I will hold off on updating the build image until after the release has been done.

PR #4590, opened 2022-01-10
Removing flux from the build image

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4577

### Checklist
- [ ] Manually tested


build/docker/Dockerfile
```diff
@@ -41,11 +41,6 @@ ENV GITHUB_CLI_VERSION 1.14.0
 RUN curl --silent --location "https://github.com/cli/cli/releases/download/v${GITHUB_CLI_VERSION}/gh_${GITHUB_CLI_VERSION}_linux_amd64.tar.gz" \
     | tar xvz -C /out/usr/local/bin --strip-components=2 gh_${GITHUB_CLI_VERSION}_linux_amd64/bin/gh && gh --version
 
-# Flux release very regularly so we always get and test against the latest
```
Skarlso, line 44: Okay, so... I thought this would trigger the need to update the build image? But when I run:

```
make -f Makefile.docker update-build-image-tag
```

nothing happens... am I missing something?

Skarlso:
Ah. It had to be committed first so git diff works? I see.

Skarlso:
Once this is approved I will run 

```
make -f Makefile.docker push-build-image
```

aclevername (changes requested):
is this because we install flux as part of the test run now?

Does this break https://github.com/weaveworks/eksctl/blob/main/CONTRIBUTING.md#running-the-integration-tests 

```
To run the integration tests in a docker container:

make integration-test-container SSH_KEY_PATH=<path to gitops repo key> \
  AWS_SESSION_TOKEN=<token> \
  AWS_ACCESS_KEY_ID=<access key id> \
  AWS_SECRET_ACCESS_KEY=<secret access key>
```

Skarlso:
@aclevername Yes, and yes it does. We will have to update that to somehow open a volume on a binary or something like that.

Either way, we need to deal with flux version being out of date. :/ I'm not sure how often this Makefile target is used though. What do you recon?

aclevername:
> @aclevername Yes, and yes it does. We will have to update that to somehow open a volume on a binary or something like that.
> 
> Either way, we need to deal with flux version being out of date. :/ I'm not sure how often this Makefile target is used though. What do you recon?

Not sure either, not sure if its used at all. But on the face it of it it would appear to be a good way for new users to easily run the integration tests.

Skarlso:
@aclevername other than a hacky way of going into the container and manually updating flux, I don't have a better solution for this though. :/ So I could, for example, have a FLUX_BIN environment property be mounted in and change integration test to use that if it's set? WDYT?

aclevername:
> @aclevername other than a hacky way of going into the container and manually updating flux, I don't have a better solution for this though. :/ So I could, for example, have a FLUX_BIN environment property be mounted in and change integration test to use that if it's set? WDYT?

mhmm not a massive fan. What happens if you just leave flux on the image, and have the integration tests always use the one it installs instead? Not a massive fan of this easier, but perhaps its the simplest for the moment?

Skarlso:
I can't tell the flux github action where to install flux.

Skarlso:
It will put it under `/usr/local/bin/flux.` which is the same where it would be otherwise too. Hmmm.... So maybe it will just overwrite it and not fail? That would be nice actually.

Skarlso:
eh no, this is `/out/usr/local/bin`? which is weird, so the GHA will put it under `/usr/local/bin`. So whichever happens to come first on `PATH` will win. Which is a bit of a risk of still running the old version.

Skarlso:
Hmm, maybe this will just work. `/usr/local/bin` is definitely before `/out/usr/local/bin` on the path, so it should always call that first. 🤔

aclevername (approved):
lets just scrap support for running the integration tests from within the build image

Skarlso:
@aclevername going to open a new, fresh PR for this

PR #4658, opened 2022-01-19
Added using kubeconfig flag and removed deleting the cluster so we can debug it.

Skarlso:


integration/tests/karpenter/karpenter_test.go
```diff
@@ -45,11 +45,11 @@ var _ = Describe("(Integration) Karpenter", func() {
...
-		cmd := params.EksctlDeleteCmd.WithArgs(
-			"cluster", clusterName,
-			"--verbose", "4",
-		)
-		Expect(cmd).To(RunSuccessfully())
+		// cmd := params.EksctlDeleteCmd.WithArgs(
+		//	"cluster", clusterName,
+		//	"--verbose", "4",
+		// )
+		// Expect(cmd).To(RunSuccessfully())
```
cPu1, line 52: Let's remove this instead of commenting it. We can revert the changes after we're done with debugging.

cPu1 (approved):

Issue #4416, opened 2021-11-04
Cluster deletion failed because of PodDisruptionBudget


AmitBenAmi:
**What were you trying to accomplish?**

Deleting my EKS clusters

**What happened?**

When deleting a cluster, `eksctl` fails to delete the NodeGroups if I have `PodDisruptionBudget` on some of my deployments.
Logs attached under the logs section beneath

On the previous version I was using (`v0.30.0`) it didn't happen.
I saw that the NodeGroups draining process as part of the cluster creation was added in this [commit](https://github.com/weaveworks/eksctl/commit/4b1405856f17f647d201fccd277e585c6eae8a7d)

**How to reproduce it?**

I had 2 clusters:
1. one with 3 nodes
2. one with 14 nodes

When I ran `eksctl delete cluster -f config.yaml`, I got the eviction error, because the PDB is being violated.
After 25 minutes, the task failed.

**Logs**

```bash
2021-11-04 18:51:01 [!]  ignoring DaemonSet-managed Pods: <Daemonsets Pods>                        
2021-11-04 18:51:01 [!]  pod eviction error ("error evicting pod: <Pod of a Deployment with PDB>: Cannot evict pod as it would violate the pod's disruption budget.") on node <Node Name>
```

The PDB:
```yaml
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: pdb
  namespace: default
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app: pdb
status:
  currentHealthy: 1
  desiredHealthy: 2
  disruptionsAllowed: 0
  expectedPods: 3
  observedGeneration: 1

```

**Anything else we need to know?**

**Versions**

```
$ eksctl version: 0.70.0
kubectl version: v1.18.5
OS: Ubuntu
```

tanvp112:
EKS will always respect PodDisruptionBudget. You should remove your PDB before destruct the cluster.

AmitBenAmi:
@tanvp112 there's an option when draining the NodeGroups of `--disable-eviction`, which by default is responsible for this kind of behaviour (only for deletion).
I'm working on a PR to utilize this flag when deleting a cluster and using a deletion flag

AmitBenAmi:
@Skarlso is it possible to re-open this issue? I'll try to finish this work by this weekend.

Skarlso:
@AmitBenAmi sure, no problem

Skarlso:
[...]

PR #4659, opened 2022-01-19
Adding an option for `--disable-nodegroup-eviction` when deleting a cluster

AmitBenAmi:
### Description

When deleting a cluster that has pods with PDBs, I can not delete the whole cluster, since the nodes won't evict all of the workloads, causing the node groups draining step to fail.
Today, there is a flag when draining node groups of: `--disable-eviction`, which ignores PDB errors while draining the node groups.

I've added a flag: `--disable-nodegroup-eviction`, that will exist when deleting a cluster, and achieves the same behavior of the `drain` command with the `--disable-eviction` flag (it literally calls the drain function with the same value).

fixes weaveworks/eksctl#4416

### Checklist
- [X] Manually tested


pkg/actions/cluster/cluster.go
```diff
@@ -15,7 +15,7 @@ import (
 
 type Cluster interface {
 	Upgrade(dryRun bool) error
-	Delete(waitInterval time.Duration, wait, force bool) error
+	Delete(waitInterval time.Duration, wait, force bool, disableNodegroupEviction bool) error
```
aclevername, line 18:
```suggestion
	Delete(waitInterval time.Duration, wait, force, disableNodegroupEviction bool) error
```

pkg/ctl/delete/cluster_test.go
```diff
@@ -0,0 +1,38 @@
...
+				deleteClusterWithRunFunc(cmd, func(cmd *cmdutils.Cmd, force bool, disableNodegroupEviction bool) error {
+					Expect(cmd.ClusterConfig.Metadata.Name).To(Equal(clusterName))
+					Expect(force).To(Equal(forceExpected))
+					Expect(disableNodegroupEviction).To(Equal(disableNodegroupEvictionExpected))
+					count++
+					return nil
+				})
+			})
+			_, err := cmd.execute()
+			Expect(err).To(Not(HaveOccurred()))
```
cPu1, line 30: Nit:
```suggestion
			Expect(err).NotTo(HaveOccurred())
```

pkg/actions/cluster/delete.go
```diff
@@ -159,7 +163,7 @@ func checkForUndeletedStacks(stackManager manager.StackManager) error {
 	return nil
 }
 
-func drainAllNodegroups(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, stackManager manager.StackManager, clientSet kubernetes.Interface, allStacks []manager.NodeGroupStack) error {
+func drainAllNodeGroups(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, clientSet kubernetes.Interface, allStacks []manager.NodeGroupStack, disableEviction *bool, nodeGroupDrainer NodeGroupDrainer, vpcCniDeleter VpcCniDeleter) error {
```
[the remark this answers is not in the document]
AmitBenAmi: Solved

pkg/ctl/delete/cluster_test.go
AmitBenAmi: Fixed

aclevername:
Thanks for opening the PR @AmitBenAmi, descriptions makes sense :ok_hand:  will take a look. +10000 for the lovely tests :smile:

pkg/actions/cluster/delete.go
```diff
@@ -29,6 +28,11 @@ import (
 	"github.com/kris-nova/logger"
 )
 
+type NodeGroupDrainer interface {
+	Drain(nodeGroups []eks.KubeNodeGroup, plan bool, maxGracePeriod, nodeDrainWaitPeriod time.Duration, undo, disableEviction bool) error
+}
+type VpcCniDeleter func(clusterName string, ctl *eks.ClusterProvider, clientSet kubernetes.Interface)
```
cPu1, line 34: We don't need to export this.
```suggestion
type vpcCNIDeleter func(clusterName string, ctl *eks.ClusterProvider, clientSet kubernetes.Interface)
```

AmitBenAmi: Changed

pkg/actions/cluster/delete_test.go
```diff
@@ -0,0 +1,114 @@
+package cluster
```
aclevername, line 1: do we need the tests in this file? Is it not covered by the test in `unowned_test.go` and `owned_test.go`?

AmitBenAmi: I think that the naming of the file is not the best one.
In this file, I only test the `drainAllNodeGroups` function, while in `unowned_test.go` and `owned_test.go` I test the `Delete` function.
The reason this file is called `delete_test.go` is because that `drainAllNodeGroups` function is location in `delete.go` file

pkg/actions/cluster/owned_test.go
```diff
@@ -98,13 +110,127 @@ var _ = Describe("Delete", func() {
...
+
+			c := cluster.NewOwnedCluster(cfg, ctl, nil, fakeStackManager)
+			fakeClientSet = fake.NewSimpleClientset()
+
+			c.SetNewClientSet(func() (kubernetes.Interface, error) {
+				return fakeClientSet, nil
+			})
+
+			mockedDrainer := &drainerMockOwned{}
+			mockedDrainer.On("Drain", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("Mocked error"))
```
aclevername, line 221: rather than `mock.Anything` can we replace it with the arguments we expect (for some at least)? Makes the test more explicit

aclevername: same goes for a few of the other new tests

pkg/actions/cluster/unowned_test.go
```diff
@@ -152,6 +163,180 @@ var _ = Describe("Delete", func() {
 			Expect(fakeStackManager.DeleteStackBySpecCallCount()).To(Equal(1))
 			Expect(*fakeStackManager.DeleteStackBySpecArgsForCall(0).StackName).To(Equal("fargate-role"))
 		})
+
+		It("ignoring nodes draining error when force flag used", func() {
```
aclevername, line 167: I think some more context to each test will make it slightly easier to read:

```suggestion
            When("force is true", func(){
		It("ignores errors when draining the nodes", func() {
```

I think this goes for the rest of the new tests too, its often easier to see the `When` describe whats changed, and the `It` the desired behaviour :smile:

pkg/actions/cluster/delete_test.go
aclevername: gotcha :+1:

pkg/actions/cluster/owned_test.go
AmitBenAmi: Done

pkg/actions/cluster/delete_test.go
```diff
@@ -0,0 +1,127 @@
...
```
aclevername, line 1: mhmm does this test file actually get run when you run `ginkgo -r pkg/action/cluster/`? I would of thought it needs to be in `cluster_test` suite to register. But I get that it needs to have access to the private function `drainAllNodeGroups`.

I'd prefer if we made `drainAllNodeGroups` public, and moved this file into `cluster_test` package.

what do you think :smile: ?

pkg/actions/cluster/delete_test.go
```diff
@@ -0,0 +1,127 @@
...
+type drainerMock struct {
+	mock.Mock
+}
+
+func (drainer *drainerMock) Drain(nodeGroups []eks.KubeNodeGroup, plan bool, maxGracePeriod, nodeDrainWaitPeriod time.Duration, undo, disableEviction bool) error {
+	args := drainer.Called(nodeGroups, plan, maxGracePeriod, nodeDrainWaitPeriod, undo, disableEviction)
+	return args.Error(0)
+}
+
+var _ = Describe("Delete", func() {
```
aclevername, line 28:
```suggestion
var _ = Describe("DrainAllNodeGroups", func() {
```

pkg/actions/cluster/unowned_test.go
AmitBenAmi: Added these contexts for the unit tests

pkg/actions/cluster/delete_test.go
AmitBenAmi: Changed for better context as well, thanks!

pkg/actions/cluster/delete_test.go
AmitBenAmi: This was my initial problem, which caused me to move it to a different package.
I didn't want to make the `drainAllNodeGroups` public so I moved it to be in the same package.
Do you think it is a good thing to make it publicly?

pkg/actions/cluster/owned.go
```diff
@@ -105,8 +110,14 @@ func (c *OwnedCluster) Delete(_ time.Duration, wait, force bool) error {
 		if err != nil {
 			return err
 		}
-		if err := drainAllNodegroups(c.cfg, c.ctl, c.stackManager, clientSet, allStacks); err != nil {
-			return err
+
+		nodeGroupManager := c.newNodeGroupManager(c.cfg, c.ctl, clientSet)
+		if err := drainAllNodeGroups(c.cfg, c.ctl, clientSet, allStacks, disableNodegroupEviction, nodeGroupManager, attemptVpcCniDeletion); err != nil {
+			if force {
+				logger.Warning("error occurred during nodegroups draining")
```
aclevername, line 117:
```suggestion
				logger.Warning("error %q occurred during nodegroups draining, force=true so proceeding with deletion", err.Error())
```

probably find better wording :laughing: but the improvement makes sense :smile:

pkg/actions/cluster/owned_test.go
```diff
@@ -98,13 +112,144 @@ var _ = Describe("Delete", func() {
...
+				disableEviction := false
+
+				mockedDrainer := &drainerMockOwned{}
+				mockedDrainer.On("Drain", kubeNodeGroups, plan, ctl.Provider.WaitTimeout(), nodeDrainWaitPeriod, undo, disableEviction).Return(errors.New("Mocked error"))
+				c.SetNewNodeGroupManager(func(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, clientSet kubernetes.Interface) cluster.NodeGroupDrainer {
+					return mockedDrainer
+				})
+
+				err := c.Delete(time.Microsecond, false, false, false)
+				Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
aclevername: same in some of the other test :smile:

pkg/actions/cluster/owned.go
AmitBenAmi: Done

pkg/actions/cluster/delete.go
```diff
@@ -172,11 +176,12 @@ func drainAllNodegroups(cfg *api.ClusterConfig, ctl *eks.ClusterProvider, stackM
 	}
 
 	logger.Info("will drain %d unmanaged nodegroup(s) in cluster %q", len(cfg.NodeGroups), cfg.Metadata.Name)
-	nodeGroupManager := nodegroup.New(cfg, ctl, clientSet)
-	if err := nodeGroupManager.Drain(cmdutils.ToKubeNodeGroups(cfg), false, ctl.Provider.WaitTimeout(), 0, false, false); err != nil {
+
+	if err := nodeGroupDrainer.Drain(cmdutils.ToKubeNodeGroups(cfg), false, ctl.Provider.WaitTimeout(), 0, false, disableEviction); err != nil {
```
Skarlso, line 180: Not necessary, but Drain has now quite a lot of parameters. Would be nice to have them in a struct rather, where we can clearly read which `0` and which `false`, `false` belongs to which parameter. :)

pkg/actions/cluster/delete_test.go
Skarlso: He means, that he prefers making the function public, and then testing it, rather than having the test code in the same package as the code it self, aka, the test code is in `cluster` package rather than `cluster_test` package which makes it a separate package. The benefit of having `cluster_test` is that you test the public API of the package. If a private function can't be tested through that, it means that it would need a bit of refactoring maybe or it's not working encapsulating, etc.

This isn't a dogma of course. :) It's a suggestion. Try and test private functions through utilising the public function. If it's hard to test, maybe there are certain conditions which are difficult to achieve in the first place?

pkg/actions/cluster/owned.go
```diff
@@ -105,8 +110,14 @@ func (c *OwnedCluster) Delete(_ time.Duration, wait, force bool) error {
...
-		if err := drainAllNodegroups(c.cfg, c.ctl, c.stackManager, clientSet, allStacks); err != nil {
-			return err
+
+		nodeGroupManager := c.newNodeGroupManager(c.cfg, c.ctl, clientSet)
+		if err := drainAllNodeGroups(c.cfg, c.ctl, clientSet, allStacks, disableNodegroupEviction, nodeGroupManager, attemptVpcCniDeletion); err != nil {
+			if force {
+				logger.Warning("an error occurred during nodegroups draining, force=true so proceeding with deletion: %q", err.Error())
+			} else {
+				return err
+			}
```
Skarlso, line 120:
```suggestion
			if !force {
			    return err
			}
			logger.Warning("an error occurred during nodegroups draining, force=true so proceeding with deletion: %q", err.Error())
```

Skarlso: Just a preference. :) I rather skip `else`s to ease reading the code.

pkg/actions/cluster/unowned.go
```diff
@@ -81,8 +86,13 @@ func (c *UnownedCluster) Delete(waitInterval time.Duration, wait, force bool) er
...
 
-		if err := drainAllNodegroups(c.cfg, c.ctl, c.stackManager, clientSet, allStacks); err != nil {
-			return err
+		nodeGroupManager := c.newNodeGroupManager(c.cfg, c.ctl, clientSet)
+		if err := drainAllNodeGroups(c.cfg, c.ctl, clientSet, allStacks, disableNodegroupEviction, nodeGroupManager, attemptVpcCniDeletion); err != nil {
+			if force {
+				logger.Warning("an error occurred during nodegroups draining, force=true so proceeding with deletion: %q", err.Error())
+			} else {
+				return err
+			}
```
Skarlso, line 95:
```suggestion
			if !force {
			    return err
			}
			logger.Warning("an error occurred during nodegroups draining, force=true so proceeding with deletion: %q", err.Error())
```

It's a bummer that this has to be repeated. :/ ¯\\\_(ツ)\_\/¯

Skarlso (approved):
Generally, this is fantastic work. :) I have like two nits, and the fact that it has to be duplicated for owned and unowned. I wished that wouldn't be the case, but nothing you can do about that right now. :)

pkg/actions/cluster/unowned.go
AmitBenAmi: Done

pkg/actions/cluster/owned.go
AmitBenAmi: Done

pkg/actions/cluster/delete.go
AmitBenAmi: Done

pkg/actions/cluster/delete_test.go
aclevername: well summarised @Skarlso . Does that make sense @AmitBenAmi ? Its not a blocker on merging as its a personal preference

aclevername (approved):
lgtm!

AmitBenAmi:
Hey @aclevername ,
What are the next steps in order to merge this PR?

Himangini:
> Hey @aclevername , What are the next steps in order to merge this PR?

@AmitBenAmi we'll merge this as soon as the build is green :) thanks for all the hard work 🎉

AmitBenAmi:
@Himangini Thanks, I pushed another commit to fix the lint errors

Skarlso:
Finally. :D :D Well done! Nice job. Thank you for your contribution @AmitBenAmi.

AmitBenAmi:
@Skarlso Thanks for the help and feedback, it was a pleasure 😄

Issue #4589, opened 2022-01-10
`build-all-distros` job intermittently fails with a timeout error


cPu1:
[This run](https://app.circleci.com/pipelines/github/weaveworks/eksctl/9922/workflows/311e014c-880b-4ad2-adcb-483eb1edf2eb/jobs/13710) failed with a timeout error:

`Timeout exceeded: try increasing it by passing --timeout option`.

This can be fixed by increasing the timeout for `golangci-lint run`.

Himangini:
@Himangini find out why we have this job , add a ticket for this

Skarlso:
@Himangini Why is this reopened?

Himangini:
> @Himangini Why is this reopened?

@Skarlso I am messing with the board. don't worry about it

PR #4684, opened 2022-01-25
Increase lint timeout

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/4589

## Delete the job

We can't. This is the only job testing if eksctl can build on all platforms.

## Why the timeout increase

On my machine, which isn't a bad machine, now-a-days, linter runs in 12 minutes! The CI isn't always faster than that. It times out before that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ endif
 
 .PHONY: lint
 lint: ## Run linter over the codebase
-	golangci-lint run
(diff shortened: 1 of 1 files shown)


aclevername (approved):

cPu1 (commented):
I'm late to review, but it LGTM.

Issue #4628, opened 2022-01-13
Use existing dual-stack VPC for IPv6 clusters


jimmyraywv:
**What help do you need?**
I am trying to use an existing dual-stack VPC for IPv6; I am using the `0.80.0-rc.0` release. I tried the below config. This created a VPC instead of using an existing one.
```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: ipv6
  version: "1.21"
  region: us-east-2

kubernetesNetworkConfig:
  ipFamily: ipv6

...
```
Then I tried the below config.
```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: ipv6
  version: "1.21"
  region: us-east-2

# kubernetesNetworkConfig:
#   ipFamily: ipv6

vpc:
  ipFamily: ipv6
  id: "<VPC_ID>"
  subnets:
    private:
      private-1:
        id: "<SUBNET_ID>"
      private-2:
        id: "<SUBNET_ID>"
      private-3:
        id: "<SUBNET_ID>"
...
```
With the second config, I get the error:
```
Error: loading config file "blog-cluster.yaml": error unmarshaling JSON: while decoding JSON: json: unknown field "ipFamily"
```
- Is the use case of using an existing dual-stack VPC supported for IPv6? I thought it was, based on this [PR](https://github.com/weaveworks/eksctl/pull/4398).
- If it is supported, is there an example config file?
- Are there docs on the new config file schema items that are RC and not yet officially released?

aclevername:
Hey @jimmyraywv. For the initial launch we didn't add support for using a pre-existing VPC for an ipv6 cluster. I'll add this story to our backlog and we will get to it soon :smile:

aclevername:
this is apparently incorrect. we do support using a pre-existing VPC see https://github.com/weaveworks/eksctl/pull/4398 however it sounds like perhaps https://github.com/weaveworks/eksctl/pull/4566 broke it, we are investigating

Himangini:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-PitchBlack-60ed85b4f8399b0017862dbd/issues/weaveworks/eksctl/4628?planning-poker) with ZenHub @aclevername @cPu1 @Skarlso @nikimanoledaki

PR #4686, opened 2022-01-25
fix bug preventing IPv6 clusters using an existing VPC

aclevername:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4628

We were accidentally wiping the VPC config for ipv6 clusters when validating the config file.


integration/tests/existing_vpc/existing_vpc_test.go
```diff
@@ -136,7 +136,7 @@ func createVPC(stackName string, ctl api.ClusterProvider) *api.ClusterVPC {
 	publicSubnets, privateSubnets, vpcID, securityGroup := createVPCStackAndGetOutputs(stackName, ctl)
 
 	By("creating the cluster config from the existing VPC")
-	newVPC := api.NewClusterVPC()
+	newVPC := api.NewClusterVPC(false)
```
Skarlso, line 139: sorry, I'm blind. When is this ever true?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -777,7 +777,7 @@ func NewClusterConfig() *ClusterConfig {
 		KubernetesNetworkConfig: &KubernetesNetworkConfig{
 			IPFamily: DefaultIPFamily,
 		},
-		VPC: NewClusterVPC(),
+		VPC: NewClusterVPC(false),
```
Skarlso, line 780: This... shouldn't be false here, should it?

integration/tests/existing_vpc/existing_vpc_test.go
aclevername: its mainly used for testing. But if you wanted to generate a VPC config for ipv6 its needs to be true

pkg/apis/eksctl.io/v1alpha5/types.go
aclevername: the default new cluster is ipv4, so the cluster should contain an ipv4 vpc

integration/tests/existing_vpc/existing_vpc_test.go
Skarlso: But how or where do you set that?

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -788,15 +788,20 @@ func NewClusterConfig() *ClusterConfig {
 }
 
 // NewClusterVPC creates new VPC config for a cluster
-func NewClusterVPC() *ClusterVPC {
+func NewClusterVPC(ipv6Enabled bool) *ClusterVPC {
```
Skarlso, line 791: Where is this ever called with true?

integration/tests/existing_vpc/existing_vpc_test.go
Skarlso: Ahh! Found it. :D

```go
		if clusterConfig.VPC == nil {
			clusterConfig.VPC = api.NewClusterVPC(ipv6Enabled)
		}
```

pkg/apis/eksctl.io/v1alpha5/types.go
Skarlso: Never mind. :D Found it. I told you I'm blind. :D

pkg/ctl/cmdutils/configfile.go
```diff
@@ -579,7 +577,7 @@ func NewUtilsEnableEndpointAccessLoader(cmd *Cmd, privateAccess, publicAccess bo
 	}
 	l.validateWithConfigFile = func() error {
 		if l.ClusterConfig.VPC == nil {
-			l.ClusterConfig.VPC = api.NewClusterVPC()
+			l.ClusterConfig.VPC = api.NewClusterVPC(false)
```
Skarlso, line 580: This couldn't be called on an IPv6 cluster? Why not do the same check here too? 🤔

aclevername: huh? Could you expand :smile:

Skarlso: In the above this setting is dependent on the ipv6 flag, here it's just false. What's the difference?

aclevername: this func isn't called for `createCluster`, so its not related

Skarlso (approved):

nikimanoledaki (approved):

Issue #4691, opened 2022-01-26
Karpenter test fails intermittently


Skarlso:
Failed in a past couple runs.

aclevername:
With `wait=false` for the helm install the test passes. But the controller starts to fail shortly after:
```
karpenter-controller-777cd6bd6c-vwqsn   0/1     CrashLoopBackOff   1          31s


2022-01-26T14:47:19.983Z        INFO    controller.controller.Termination       Starting Controller     {"commit": "dc47849", "reconciler group": "", "reconciler kind": "Node"}
2022-01-26T14:47:19.983Z        INFO    controller.controller.Termination       Starting workers        {"commit": "dc47849", "reconciler group": "", "reconciler kind": "Node", "worker count": 10}
E0126 14:47:19.984251       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 387 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1dc5800, 0x37a1200})
        k8s.io/apimachinery@v0.20.7/pkg/util/runtime/runtime.go:74 +0x85
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x0})
        k8s.io/apimachinery@v0.20.7/pkg/util/runtime/runtime.go:48 +0x75
panic({0x1dc5800, 0x37a1200})
        runtime/panic.go:1038 +0x215
github.com/awslabs/karpenter/pkg/cloudprovider/aws/apis/v1alpha1.Deserialize(0xc00071ce18)
        github.com/awslabs/karpenter/pkg/cloudprovider/aws/apis/v1alpha1/provider.go:55 +0x96
github.com/awslabs/karpenter/pkg/cloudprovider/aws.(*CloudProvider).GetInstanceTypes(0xc000686fa0, {0x250b3d8, 0xc0007601e0}, 0xa00000000)
        github.com/awslabs/karpenter/pkg/cloudprovider/aws/cloudprovider.go:136 +0x2f
github.com/awslabs/karpenter/pkg/controllers/metrics.(*Controller).updateNodeCounts(0xc000a35ba0, {0x250b3d8, 0xc0007601e0}, 0xc00071cd00)
        github.com/awslabs/karpenter/pkg/controllers/metrics/controller.go:101 +0x65
[...]

PR #4692, opened 2022-01-26
Fix karpenter integration test

aclevername:
### Description

closes https://github.com/weaveworks/eksctl/issues/4691

Bumping the version of karpenter to a newer one seems to resolve the issue

```
[0] JUnit report was created: /home/jake/weave/eksctl/test-results/__home__jake__weave__eksctl__integration__tests__karpenter__karpenter__test__go_1.xml
[0]
[0] Ran 1 of 1 Specs in 1427.681 seconds
[0] SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
[0] PASS
```

### DIFF
--- a/integration/tests/karpenter/karpenter_test.go
+++ b/integration/tests/karpenter/karpenter_test.go
@@ -13,6 +13,7 @@ import (
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	. "github.com/weaveworks/eksctl/integration/runner"
 	"github.com/weaveworks/eksctl/integration/tests"
 	clusterutils "github.com/weaveworks/eksctl/integration/utilities/cluster"
 	"github.com/weaveworks/eksctl/integration/utilities/kube"
@@ -36,6 +37,14 @@ func TestKarpenter(t *testing.T) {
 }
 
 var _ = Describe("(Integration) Karpenter", func() {
+	AfterEach(func() {
+		cmd := params.EksctlDeleteCmd.WithArgs(
+			"cluster", params.ClusterName,
+			"--verbose", "4",
+		)
+		Expect(cmd).To(RunSuccessfully())
+	})
+
 	Context("Creating a cluster with Karpenter", func() {
 		It("should support karpenter", func() {
 			cmd := params.EksctlCreateCmd.
--- a/integration/tests/karpenter/karpenter_test.go
+++ b/integration/tests/karpenter/karpenter_test.go
@@ -1,3 +1,6 @@
+//go:build integration
+// +build integration
+
 package karpenter
 
 import (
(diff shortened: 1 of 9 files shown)


integration/tests/karpenter/karpenter_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build integration
```
Skarlso, line 1: :O

aclevername: i don't know what affect this was having, but it felt wrong

integration/tests/karpenter/karpenter_test.go
```diff
@@ -26,34 +28,25 @@ func init() {
 	if err := api.Register(); err != nil {
 		panic(fmt.Errorf("unexpected error registering API scheme: %w", err))
 	}
-	params = tests.NewParams("")
+	params = tests.NewParams("karp")
```
Skarlso, line 31: I believe I had to do it like this, otherwise it was applying a longer thing. That's why there is the before each thingy that you removed. But it looks like if this works, than it's nice. :)

integration/tests/karpenter/karpenter_test.go
```diff
@@ -26,34 +28,25 @@ func init() {
...
-	})
-
 	Context("Creating a cluster with Karpenter", func() {
 		It("should support karpenter", func() {
 			cmd := params.EksctlCreateCmd.
 				WithArgs(
 					"cluster",
 					"--config-file=-",
 					"--verbose=4",
+					"--kubeconfig", params.KubeconfigPath,
```
Skarlso, line 46: I ran it so many times and this wasn't working. It gave the localhost error thingy. I had to do the write kubeconfig stuff otherwise it failed in CI.

But I hope it will work. :D

aclevername: I'm fairly certain it should be okay!

integration/tests/karpenter/karpenter_test.go
```diff
@@ -26,34 +28,25 @@ func init() {
...
-	var (
-		clusterName string
-	)
-
-	BeforeEach(func() {
-		// the randomly generated name we get usually makes one of the resources have a longer than 64 characters name
-		clusterName = fmt.Sprintf("it-karpenter-%d", time.Now().Unix())
-	})
-
 	Context("Creating a cluster with Karpenter", func() {
```
Skarlso, line 39: Do you want to add back the After to remove the cluster? Otherwise, we'll forget about it, and it will leave some clutter. If it happens to NOT fail, we can always just remove it.

integration/tests/karpenter/karpenter_test.go
Skarlso: I was too! :DDD

integration/tests/karpenter/karpenter_test.go
aclevername: this is defo shorter than other names that also produce long CF stacks. Lets keep it like this for now and see if any failures occur?

integration/tests/karpenter/karpenter_test.go
aclevername: oh shit yeah!

Himangini (approved):
👍🏻

Issue #4530, opened 2021-12-09
Cannot create managed nodegroup in private subnet


aclevername:
```
---
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: jk
  region: ap-northeast-2
  version: "1.21"

addons:
  - name: vpc-cni
    version: latest
  - name: coredns
    version: latest
  - name: kube-proxy
    version: latest

iam:
  withOIDC: true

managedNodeGroups:
  - name: mng-1-private
    desiredCapacity: 1
    subnets:
    - subnet-0ad218f2e9514bc04
```

```
./eksctl create nodegroup -f jk-ipv6.yaml
2021-12-09 10:25:27 [ℹ]  eksctl version 0.70.0-dev+10620385.2021-12-09T10:14:07Z
2021-12-09 10:25:27 [ℹ]  using region ap-northeast-2
2021-12-09 10:25:35 [ℹ]  nodegroup "mng-1-private" will use "" [AmazonLinux2/1.21]
2021-12-09 10:25:39 [✖]  found mis-configured subnets ["subnet-0ad218f2e9514bc04"]. Expected public subnets with property "MapPublicIpOnLaunch" enabled. Without it new nodes won't get an IP assigned
Error: subnets for one or more new nodegroups don't meet requirements. To fix this, please run `eksctl utils update-legacy-subnet-settings --cluster jk`
```

`subnet-0ad218f2e9514bc04` is a private subnet, but the code seems to be treating it as a public subnet.

Himangini:
We need to update the error message here for better UX, perhaps need to improve on the docs for this as well.

Himangini:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-PitchBlack-60ed85b4f8399b0017862dbd/issues/weaveworks/eksctl/4530?planning-poker) with ZenHub @aclevername @cPu1 @nikimanoledaki @Skarlso

cPu1:
This is actually a missing feature in eksctl and not necessarily a bug. When a subnet ID is specified for a nodegroup, eksctl does not try to determine whether that subnet is private or public, and infers this property from the `privateNetworking` field. In your example, eksctl infers that the subnet is public since the `privateNetworking` field is not set. I think we can improve the UX here by returning a more useful error message.

Skarlso:
[...]

PR #4694, opened 2022-01-26
Erroring out if the subnet is not public

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/4530

# test

```
eksctl create nodegroup -f cluster_config.yaml
2022-01-26 17:49:50 [ℹ]  eksctl version 0.82.0-dev+d60d233e.2022-01-26T17:48:15Z
2022-01-26 17:49:50 [ℹ]  using region us-west-2
2022-01-26 17:49:56 [ℹ]  nodegroup "mng-1-private" will use "" [AmazonLinux2/1.21]
2022-01-26 17:49:57 [✖]  subnet "subnet-05110e7292673ede2" not found in the list of public subnets or is not a public subnet
Error: subnets for one or more new nodegroups don't meet requirements. To fix this, please run `eksctl utils update-legacy-subnet-settings --cluster test-private-subnet-4`
```

Might remove that last part or extract this into a separate validation... 🤔  I'll have a think about this. :)

Done, simplified the above error.


pkg/vpc/vpc.go
```diff
@@ -524,16 +522,28 @@ func cleanupSubnets(spec *api.ClusterConfig) {
...
+		return false
+	}
+	var legacySubnets []string
 	for _, sn := range subnets {
 		if sn.MapPublicIpOnLaunch == nil || !*sn.MapPublicIpOnLaunch {
 			legacySubnets = append(legacySubnets, *sn.SubnetId)
 		}
+		if !containsSubnet(*sn.SubnetId) {
+			return fmt.Errorf("subnet %q does not exist or is not a public subnet", *sn.SubnetId)
+		}
```
aclevername, line 541: MEGA nit you could make map a `map[string]interface{}` once containing all the keys, so then each `contains` is just a map lookup instead of another loop over. but this doesn't matter at all :joy:

Skarlso: I ACTUALLY THOUGHT ABOUT THAT! :DDDD 

But it's three or so subnets most of the time, so... :D It's not worth the extra lines + memory to store them in a map. :)

aclevername (approved):

pkg/vpc/vpc.go
```diff
@@ -524,16 +522,28 @@ func cleanupSubnets(spec *api.ClusterConfig) {
...
+		return false
+	}
+	var legacySubnets []string
 	for _, sn := range subnets {
 		if sn.MapPublicIpOnLaunch == nil || !*sn.MapPublicIpOnLaunch {
 			legacySubnets = append(legacySubnets, *sn.SubnetId)
 		}
+		if !containsSubnet(*sn.SubnetId) {
+			return fmt.Errorf("subnet %q does not exist or is not a public subnet", *sn.SubnetId)
```
nikimanoledaki, line 540:
```suggestion
			return fmt.Errorf("subnet %q could not be found in CF template or is not a public subnet", *sn.SubnetId)
```
wdyt?

Skarlso: Sure 👍

cPu1 (approved):

nikimanoledaki (approved):

PR #4702, opened 2022-01-27
Avoid force-pushing to `docs` branch to publish docs

### LINKED ISSUE #4575: [SPIKE] Investigate and fix force pushing to docs branch
Our tag-release action is using a build script which force pushes to the `docs` branch. This should be stopped. Investigate what we can do about it.

```
# Update the site by putting everything from the release into the docs branch
git push --force origin "${release_branch}":docs
```

Timebox: 2 days

cPu1:
### Description

This changelist eliminates the need to force-push to `docs` and to have a dedicated branch for publishing docs, by using Netlify's build hooks feature. Rather than asking Netlify to trigger deployments by monitoring changes to a branch (`docs`), this changelist triggers deployment as part of the release workflow by sending an HTTP request to the build hook URL specifying the release branch. 

There are some GitHub actions for Netlify but they seem complicated for this task and require more configuration than our current setup. A build hook URL seems like the best fit for triggering deployments on a specific branch, while still being able to use the existing `netlify.toml` file.

Closes #4575 

### Checklist
- [x] Manually tested

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,3 +27,7 @@ jobs:
           token: "${{ secrets.WEAVEWORKSBOT_TOKEN }}"
       - name: Open PRs to release branch and default branch
         run: make prepare-release
+      - name: Publish docs to Netlify
+        run: make publish-docs
+        env:
+          NETLIFY_BUILD_HOOK_URL: ${{ secrets.NETLIFY_BUILD_HOOK_URL }}
--- a/build/scripts/publish-docs.sh
+++ b/build/scripts/publish-docs.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -ex
+
+DIR="${BASH_SOURCE%/*}"
+
+if [ -z "${NETLIFY_BUILD_HOOK_URL}" ] ; then
+  echo "NETLIFY_BUILD_HOOK_URL is required"
+  exit 1
+fi
+
+# shellcheck source=tag-common.sh
+. "${DIR}/tag-common.sh"
+
+release_branch=$(release_branch)
+
+curl -X POST -d "trigger_branch=${release_branch}" -d "trigger_title=Triggered+by+Release+action" "${NETLIFY_BUILD_HOOK_URL}"
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -45,7 +45,7 @@ jobs:
         run: make install-site-deps
       - name: Build docs for link check
         run: make build-pages
-        # Using liche action to check generated HTML site
+        # Using link-checker action to check generated HTML site
       - name: Link Checker (generated site)
         id: lc
         uses: peter-evans/link-checker@v1
--- a/Makefile.docs
+++ b/Makefile.docs
@@ -15,3 +15,7 @@ serve-pages: ## Serve the site locally
 .PHONY: build-pages
 build-pages: userdocs/src/usage/schema.json ## Generate the site
 	cd userdocs/ && mkdocs build
+
+.PHONY: publish-docs
+publish-docs:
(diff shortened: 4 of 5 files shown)


build/scripts/publish-docs.sh
```diff
@@ -0,0 +1,15 @@
...
+
+DIR="${BASH_SOURCE%/*}"
+
+if [ -z "${NETLIFY_BUILD_HOOK_URL}" ] ; then
+  echo "NETLIFY_BUILD_HOOK_URL is required"
+  exit 1
+fi
+
+# shellcheck source=tag-common.sh
+. "${DIR}/tag-common.sh"
```
Skarlso, line 11:
```suggestion
source "${DIR}/tag-common.sh"
```

I like it better when the action is explicit. :)

build/scripts/publish-docs.sh
```diff
@@ -0,0 +1,15 @@
...
+
+release_branch=$(release_branch)
```
Skarlso, line 13: This happens to work, right? But this is a shell call to a variable. Do you want to `echo` it? Or what's the purpose here. `release_branch` isn't a command, right?

Skarlso: I see that tag-release is using this as well, so ¯\\\_(ツ)\_\/¯ . But it's still weird. :D

Skarlso (approved):
One nit, otherwise Ok

.github/workflows/release.yaml
```diff
@@ -27,3 +27,7 @@ jobs:
           token: "${{ secrets.WEAVEWORKSBOT_TOKEN }}"
       - name: Open PRs to release branch and default branch
         run: make prepare-release
+      - name: Publish docs to Netlify
+        run: make publish-docs
+        env:
+          NETLIFY_BUILD_HOOK_URL: ${{ secrets.NETLIFY_BUILD_HOOK_URL }}
```
aclevername, line 33: is this the right time to trigger it? It might mean the docs are changed hours before the release is out? or if circleci fails it will never come out

build/scripts/publish-docs.sh
cPu1: `.` is the portable and POSIX-defined way of sourcing a file. `source`, OTOH, is shell-specific and can be missing in other shells like `dash`. Although the interpreter is set to `bash`, it's best to not rely on non-portable features where possible.

build/scripts/publish-docs.sh
cPu1: >This happens to work, right? 

Yes.

>But this is a shell call to a variable.

This is not a shell call to a variable, it's a function call to `release_branch` which is defined in `tag-common.sh`. 

>Do you want to echo it?

No, it's for storing the return value from the `release_branch` function.

.github/workflows/release.yaml
cPu1: > It might mean the docs are changed hours before the release is out?

This is the current behaviour as well. One of the last steps in `make prepare-release` was to force-push to `docs`, which meant that the Netlify deployment was triggered before the release was complete. This is preserving that behaviour without requiring to force-push, although we might want to consider publishing docs after the release is complete.

aclevername: ahh yeah okay. This seems like a good first fix, and then we can improve later to have it happen closer to the final release?

aclevername (approved):

.github/workflows/release.yaml
cPu1: Yup, I will open an issue.

build/scripts/publish-docs.sh
Skarlso: Oh I see it's coming from tag-common 🤦

Issue #4662, opened 2022-01-19
if eksctl create cluster fails, no way to clean up the stack


matti:
```
2022-01-19 17:53:27 [ℹ]  building cluster stack "eksctl-kungen-1-cluster"
2022-01-19 17:53:28 [!]  1 error(s) occurred and cluster hasn't been created properly, you may wish to check CloudFormation console
2022-01-19 17:53:28 [ℹ]  to cleanup resources, run 'eksctl delete cluster --region=eu-north-1 --name=kungen-1'
2022-01-19 17:53:28 [✖]  creating CloudFormation stack "eksctl-kungen-1-cluster": AlreadyExistsException: Stack [eksctl-kungen-1-cluster] already exists
	status code: 400, request id: fce610ce-a478-497f-83b0-867ffa210641
```

but then:
```
eksctl delete cluster --region=eu-north-1 --name=kungen-1
Error: unable to describe cluster control plane: ResourceNotFoundException: No cluster found for name: kungen-1.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "21b14c5a-daca-4ed8-aaa7-43742a2ca3f7"
  },
  Message_: "No cluster found for name: kungen-1."
}
```

also noticed here: https://github.com/weaveworks/eksctl/issues/1386#issuecomment-1011372461

Skarlso:
Hi. 

In that issue, they are talking about the cluster, but what's happening actually is that the cloudformation stack is there while the cluster isn't. This happens if the cluster creation failed on the first part. In which case there is no cluster so it says, can't delete cluster. But the stack was already created. The cloudformation stack that is. In that case, `eksctl` can't do too much since it's looking for the cluster. Maybe we can do something like, look for the cloudformation stack as well (maybe we do that already, I'm not sure, I have to check). But an immediate action you can take is to delete the cloudformation stack and then you can move on. Deleting the CF stack will remove all resources associated with the stack that it might have created so far such as VPCs, Security Groups, etc.

matti:
yes, I already deleted the stack and moved on - but imo eksctl should have atleast some utility function to delete the stack (and possibly some other resources that were created, because as a user I'm not sure what was created for me by eksctl)

jglick:
[...]

PR #4711, opened 2022-01-28
Fix deletion logic in case the cluster does not exist

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/4662

# Testing

Created a cluster and deleted the created cluster. Run delete with `--force`:

```log
➜  stuck-stack eksctl delete cluster -f cluster-managed.yaml --force
2022-01-28 16:21:03 [ℹ]  eksctl version 0.82.0-dev+d3bd6475.2022-01-28T16:19:21Z
2022-01-28 16:21:03 [ℹ]  using region us-west-2
2022-01-28 16:21:03 [ℹ]  deleting EKS cluster "test-stuck-stack-2"
2022-01-28 16:21:06 [✔]  kubeconfig has been updated
2022-01-28 16:21:08 [ℹ]  1 task: { delete cluster control plane "test-stuck-stack-2" [async] }
2022-01-28 16:21:08 [ℹ]  will delete stack "eksctl-test-stuck-stack-2-cluster"
2022-01-28 16:21:09 [✔]  all cluster resources were deleted
```

Screenshot 2022-01-28 at 16 21 42

Second run with a stack which was rollback_complete also worked. Note that we delete everything that is in a COMPLETED state whatever that would be and we don't delete anything in a PROGRESS state.


pkg/ctl/delete/cluster.go
```diff
@@ -65,9 +73,14 @@ func doDeleteCluster(cmd *cmdutils.Cmd, force bool) error {
 		return err
 	}
 
+	// Ignore error here if cluster doesn't exist. We still would like to clean up the stack and other
+	// existing resources which were created even if the cluster failed.
 	cluster, err := cluster.New(cfg, ctl)
 	if err != nil {
-		return err
+		if !force {
```
Skarlso, line 80: This might actually not be needed. I'm running a new test without it, since inside we already ignore if the refresh failed because the cluster didn't exist.

Skarlso: Yep, this wasn't actually needed.

pkg/ctl/delete/cluster.go
```diff
@@ -52,7 +52,15 @@ func doDeleteCluster(cmd *cmdutils.Cmd, force bool) error {
 
 	ctl, err := cmd.NewProviderForExistingCluster()
 	if err != nil {
-		return err
+		if !force {
+			return err
+		}
+		// initialise the controller without refreshing the cluster status.
```
aclevername, line 58: work adding a log about skipping the error?

Skarlso: 👍 will add one

nikimanoledaki (approved):
lgtm!

cPu1 (approved):

Issue #2115, opened 2020-04-30
Allow more granular control over volume configuration on NodeGroup


tomhuang12:
**Why do you want this feature?**
In my use case, I would like to attach a second encrypted EBS volume to the nodes due to security requirements. The second volume is coming from a snapshot of the AMI I am using for the nodes. 

**What feature/behavior/change do you want?**
An example cluster configuration yaml would look something like this:

```
---
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: cluster-1
  region: eu-west-1

nodeGroups:
  - name: ng1
    instanceType: m5.xlarge
    desiredCapacity: 1
    volumeMappings:
    - deviceName: "/dev/xvda"
      volumeSize: 100
      volumeType: gp2
      volumeEncrypted: true
      snapshotId: snap-0e1d1928ad4d81a
      volumeKmsKeyID: 36c0b54e-64ed-4f2d-a1c7-96558764311e
    - deviceName: "/dev/sdb"
      volumeSize: 100
      volumeType: gp2
      volumeEncrypted: true
      snapshotId: snap-012d1928ad4abc1
      volumeKmsKeyID: 36c0b54e-64ed-4f2d-a1c7-96558764311e
```

darth-kermit:
Please implement this. I have been unable to use EKSCtl for this very reason. Cloudformation allows you to configure/add a second volume so it shouldn't be a problem for EKSCtl.

sayboras:
I feel it's making sense to have this. I can spend sometime this week on this if maintainers are ok with this feature. @martina-if @cPu1

scottbisker:
+1 for having this ability.  It will greatly simplify deployments for thing like EdgeFS where a separate volume can be dedicated for the service.

Skarlso:
Interesting, I don't know why we decided to turn a list of block mappings into single items. That's super strange.

Sadly, I can't change that now to be backwards compatible. I can do, however, something like, `additionalBlockMappings` or `additionalVolumeMappings` which then, will be added to the list of mappings. That's the least intrusive and backwards compatible addition I can think of.

Skarlso:
@darth-kermit I know it's a bit late, but this has been implemented now and will be part of the next release this weekend. :)

tomhuang12:
Thank you @Skarlso ! It's been some time 😅

darth-kermit:
Thank you so much! We have been waiting a long time for this and appreciate
[...]

PR #4725, opened 2022-01-31
Allow additional volume mappings

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/2115

Todo: 

Probably some better docs.

## Manual Test

```yaml
kind: ClusterConfig

metadata:
  name: test-additional-volumes-2
  region: us-west-2
  version: '1.20'

managedNodeGroups:
  - name: managed-ng-1
    minSize: 1
    maxSize: 4
    desiredCapacity: 1
    additionalVolumes:
      - volumeName: '/dev/sdb'
```

```
➜  additional-volumes eksctl create cluster -f cluster-managed.yaml
...
2022-02-01 14:39:50 [ℹ]  nodegroup "managed-ng-1" has 1 node(s)
2022-02-01 14:39:50 [ℹ]  node "ip-192-168-76-90.us-west-2.compute.internal" is ready
2022-02-01 14:39:52 [ℹ]  kubectl command should work with "/Users/skarlso/.kube/config", try 'kubectl get nodes'
2022-02-01 14:39:52 [✔]  EKS cluster "test-additional-volumes-2" in "us-west-2" region is ready
```

Screenshot 2022-02-01 at 14 43 09

TODO: 

Update the acceptance test

Currently running acceptance test....

```
using API server https://7D1D98844D284ED6A487D4A8270C7A95.gr7.us-west-2.eks.amazonaws.com

• [SLOW TEST:350.853 seconds]
(Integration) Create Managed Nodegroups
/Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/managed/managed_nodegroup_test.go:57
  cluster with 1 al2 managed nodegroup
  /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/managed/managed_nodegroup_test.go:159
    supports adding bottlerocket and ubuntu nodegroups with additional volumes
    /Users/skarlso/goprojects/weaveworks/eksctl/integration/tests/managed/managed_nodegroup_test.go:160
....
Ran 1 of 21 Specs in 1665.693 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 20 Skipped
PASS | FOCUSED
```

Same for running unmanaged nodegroups.

```
Ran 1 of 43 Specs in 2362.203 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 42 Skipped
PASS | FOCUSED
```


pkg/cfn/builder/block_device_mapping.go
```diff
@@ -1,57 +1,79 @@
...
+		VolumeThroughput: ng.VolumeThroughput,
+	})
+
+	if mapping != nil {
+		mappings = append(mappings, *mapping)
+	}
+
+	for _, volume := range ng.AdditionalVolumes {
+		mappings = append(mappings, *makeBlockDeviceMapping(volume))
+	}
```
aclevername, line 30: we only check if the `mapping` is nil for the default volume, not the additional volumes?

aclevername (dismissed):
worth updating an existing integration test? lgtm

pkg/cfn/builder/block_device_mapping.go
Skarlso: That is a good thought, though I think if someone defined an additional volume they surely would set the size, right? But, maybe that's a problem in any case. Like, if additional volumes is defined, size should be required? WDYT?

Skarlso: In any case I'll add the check and extend the test case to include one  maybe.

aclevername: perhaps we are missing checks in the `validation.go` file that that the minimum required fields are set?

Skarlso: Possibly.

Skarlso: Uh, it's because we just set the mapping field to nil if the size wasn't defined, never really doing anything with it. This is a bit weird. :D

Skarlso: Aaaah, I'm not touching this. :D I'll add the nil check :D

Skarlso: Oh shit, interesting. We have a thing `setVolumeDefaults` that sets default.

Skarlso: OOOOH. Now I see.

Skarlso: We always have these fields set. You can just overwrite them. 🤦 . So I need a different kind of check of the additional thing.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -204,6 +206,28 @@ func setVolumeDefaults(ng *NodeGroupBase, template *LaunchTemplate) {
 	}
 }
 
+func setDefaultsForAdditionalVolumes(ng *NodeGroupBase) {
```
Skarlso, line 209: This is duplicated on the above, but they are using a different struct, so... ¯\\\_(ツ)\_\/¯

integration/tests/managed/managed_nodegroup_test.go
```diff
@@ -157,10 +157,10 @@ var _ = Describe("(Integration) Create Managed Nodegroups", func() {
 	)
 
 	Context("cluster with 1 al2 managed nodegroup", func() {
-		It("supports adding bottlerocket and ubuntu nodegroups", func() {
+		FIt("supports adding bottlerocket and ubuntu nodegroups with additional volumes", func() {
```
Skarlso, line 160: eh

integration/tests/crud/creategetdelete_test.go
```diff
@@ -420,6 +420,11 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 							Name:         "test-containerd",
 							AMIFamily:    api.NodeImageFamilyAmazonLinux2,
 							InstanceType: "p2.xlarge",
+							AdditionalVolumes: []*api.VolumeMapping{
+								{
+									VolumeName: aws.String("/dev/sdb"),
+								},
+							},
```
aclevername, line 427: :eyes: are you adding the check to see it results in the volume being added? *runs*

Skarlso: Nope, because it's expensive. It would fail if it didn't succeed. I would have get the asg, find out which instance is added, do a discribe instance, do a describe volumes and search for the mount. For every nodegroup and every node it would add around 3 more API calls.

userdocs/src/usage/nodegroup-additional-volume-mappings.md
```diff
@@ -0,0 +1,34 @@
+# Additional Volume Mappings
```
aclevername, line 1: does this need to be a new page? or can it be added to the existing page?

integration/tests/crud/creategetdelete_test.go
aclevername: >  because it's expensive.

is 3 api calls expensive? Sorry I know I'm being really annoying, but a lot of our integration tests never check that the desired outcome actuall happens, just that we exit 0

userdocs/src/usage/nodegroup-additional-volume-mappings.md
Skarlso: We have custom DNS and upgrade and custom subnet separately. I thought it would be nice and searchable this way.

I can add it to `manage-nodegroups.md` if you think that's a better place for it. :)

aclevername:
does this work for `unmanaged`?

integration/tests/crud/creategetdelete_test.go
Skarlso: Considering we have a ticket to reduce the number of overwhelming API calls and the number of times it fails because of throttling issue, yes, I think 3 is a lot. :) And that's per nodegroup and per instance. If there are two instances there that will be 6. :)

Skarlso:
> does this work for unmanaged?

The change set includes changes in both of the unit tests and acceptance tests.

Also, see my comment here:

```
Same for running unmanaged nodegroups.

Ran 1 of 43 Specs in 2362.203 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 42 Skipped
PASS | FOCUSED
```

integration/tests/crud/creategetdelete_test.go
aclevername: > Considering we have a ticket to reduce the number of overwhelming API calls and the number of times it fails because of throttling issue, yes, I think 3 is a lot. :)

thats for eksctl itself, not for the tests. We don't care how many API calls the tests make

aclevername:
> > does this work for unmanaged?
> 
> The change set includes changes in both of the unit tests and acceptance tests.
> 
> Also, see my comment here:
> 
> ```
> Same for running unmanaged nodegroups.
> 
> Ran 1 of 43 Specs in 2362.203 seconds
> SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 42 Skipped
> PASS | FOCUSED
> ```

yeah I deleted my comment after I realised haha

integration/tests/crud/creategetdelete_test.go
Skarlso: fine

integration/tests/crud/creategetdelete_test.go
```diff
@@ -434,6 +439,42 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 					WithoutArg("--region", params.Region).
 					WithStdin(clusterutils.Reader(clusterConfig))
 				Expect(cmd).To(RunSuccessfully())
+				config, err := clientcmd.BuildConfigFromFlags("", params.KubeconfigPath)
```
Skarlso, line 442: Will clean this somewhat up tomorrow.

Skarlso: Create a common function in the Matchers. `Check|VerifyNodegroupMount`

aclevername (approved):
lgtm!

userdocs/src/usage/nodegroup-additional-volume-mappings.md
Skarlso: @aclevername I'm actually not sure what the 👍 was for in this case. Move it or leave it as is? :D

integration/tests/assertions.go
```diff
@@ -33,3 +39,42 @@ func ListNodes(clientset kubernetes.Interface, nodeGroupName string) *corev1.Nod
 	Expect(err).NotTo(HaveOccurred())
 	return nodeList
 }
+
+func AssertNodeVolumes(kubeConfig string, region string, nodeGroupName string, volumeName string) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeConfig)
+	Expect(err).NotTo(HaveOccurred())
+	clientSet, err := kubernetes.NewForConfig(config)
+	Expect(err).NotTo(HaveOccurred())
+	nodes, err := clientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
```
cPu1, line 48:
```suggestion
	nodes, err := clientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
		LabelSelector: fmt.Sprintf("%s=%s", api.NodeGroupNameLabel, nodeGroupName),
	})
```

integration/tests/assertions.go
```diff
@@ -33,3 +39,42 @@ func ListNodes(clientset kubernetes.Interface, nodeGroupName string) *corev1.Nod
...
+	Expect(err).NotTo(HaveOccurred())
+	var instanceIDs []string
+	for _, node := range nodes.Items {
+		// only test instances in this nodegroup, any other should be ignored.
+		if node.Labels[api.NodeGroupNameLabel] != nodeGroupName {
+			continue
+		}
```
cPu1, line 55: Based on the above suggestion, this can be removed.
```suggestion
```

integration/tests/assertions.go
```diff
@@ -33,3 +39,42 @@ func ListNodes(clientset kubernetes.Interface, nodeGroupName string) *corev1.Nod
...
+		// aws:///us-west-2c/i-00bb587a7011eb63c
+		split := strings.Split(node.Spec.ProviderID, "/")
+		id := split[len(split)-1]
+		Expect(strings.HasPrefix(id, "i")).To(
+			BeTrue(),
+			fmt.Sprintf("provider ID %q should have instance ID format aws:///us-west-2c/i-00bb587a7011eb63c", node.Spec.ProviderID),
+		)
```
cPu1, line 62:
```suggestion
		Expect(id).To(
			HavePrefix("i"),
			fmt.Sprintf("provider ID %q should have instance ID format aws:///us-west-2c/i-00bb587a7011eb63c", node.Spec.ProviderID),
		)
```

integration/tests/assertions.go
```diff
@@ -33,3 +39,42 @@ func ListNodes(clientset kubernetes.Interface, nodeGroupName string) *corev1.Nod
 	Expect(err).NotTo(HaveOccurred())
 	return nodeList
 }
+
+func AssertNodeVolumes(kubeConfig string, region string, nodeGroupName string, volumeName string) {
```
cPu1, line 43: Minor nit:
```suggestion
func AssertNodeVolumes(kubeConfig, region, nodeGroupName, volumeName string) {
```

cPu1 (approved):
I have suggested some minor improvements in the integration tests but otherwise LGTM! :tada:

integration/tests/assertions.go
Skarlso: Ah that's true, nice. :)

PR #4737, opened 2022-02-03
Show nodegroup type in get nodegroup

### LINKED ISSUE #4695: Indicate whether a cluster’s existing node groups are managed or unmanaged
There should be an easy way to tell whether a cluster's node groups are managed or unmanaged. One solution would be to add a boolean flag `Managed` to the node group obj returned by `eksctl get nodegroup`

See slack thread https://weave-community.slack.com/archives/CAYBZBWGL/p1643180196379949

We have multiple EKS clusters running, some with managed node groups and some with unmanaged. We would like to differentiate between the two types so that we can take advantage of managed node group features. One example of this is upgrading AMIs:
- For managed node groups, we can use `eksctl upgrade nodegroup`
- For unmanaged node groups, we will create a new node group and delete the old one.

cPu1:
### Description

Fixes #4695

TODO:

- [x] Update integration tests

### Checklist
- [x] Manually tested

### DIFF
--- a/integration/tests/crud/creategetdelete_test.go
+++ b/integration/tests/crud/creategetdelete_test.go
@@ -462,9 +462,13 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 					"--name", mngNG1,
 					"-o", "yaml",
 				)
-				Expect(getMngNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MaxSize: 4")))
-				Expect(getMngNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MinSize: 4")))
-				Expect(getMngNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("DesiredCapacity: 4")))
+				Expect(getMngNgCmd).To(RunSuccessfullyWithOutputStringLines(
+					ContainElement(ContainSubstring("Type: managed")),
+					ContainElement(ContainSubstring("MaxSize: 4")),
+					ContainElement(ContainSubstring("MinSize: 4")),
+					ContainElement(ContainSubstring("DesiredCapacity: 4")),
+				),
+				)
 			})
 
 			It("should scale all nodegroups", func() {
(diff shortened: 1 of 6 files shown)


aclevername (approved):
does this new field break any of the integration tests?

cPu1:
> does this new field break any of the integration tests?

It shouldn't, but I'm adding integration tests for this field.

integration/tests/crud/creategetdelete_test.go
```diff
@@ -482,19 +486,27 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
...
 				)
-				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MaxSize: 5")))
-				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("MinSize: 5")))
-				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputString(ContainSubstring("DesiredCapacity: 5")))
+
+				Expect(getUnmNgCmd).To(RunSuccessfullyWithOutputStringLines(
+					ContainElement(HaveSuffix("MaxSize: 5"))),
+					ContainElement(HaveSuffix("MinSize: 5")),
+					ContainElement(HaveSuffix("DesiredCapacity: 5")),
+					ContainElement(HaveSuffix("Type: unmanaged")),
```
nikimanoledaki, line 508: Writing `unmanaged` in the output might be confusing to users since EKS calls them `self-managed`, but on the other hand, our docs call them `unmanaged` everywhere so it's too late anyway, right? 🤷‍♀️

nikimanoledaki (approved):

integration/tests/crud/creategetdelete_test.go
nikimanoledaki: We can ignore this for now since it's not something to be addressed by this PR anyway :)

cPu1: >but on the other hand, our docs call them unmanaged everywhere so it's too late anyway, right? :woman_shrugging:

Yeah, that's true.

cPu1:
I'm running the integration tests on this branch.

Issue #4745, opened 2022-02-04
Karpenter instance profile usage


niallthomson:
**What feature/behavior/change do you want?**

The Karpenter integration creates a node role and associated instance profile that can be used with a provisioner. However, the name of the instance profile created is not explicitly documented.

Additionally, it would be great if this could be optionally or by default set as the default instance profile. This could be activated by a flag and pass in using the helm chart values:

```
aws:
  # -- The default instance profile to use when launching nodes on AWS
  defaultInstanceProfile: ""
```

**Why do you want this feature?**

Users need to figure this out for themselves, which makes getting Karpenter running more difficult.

Skarlso:
Hi!

Thanks for the issue.

For the record, the document says:

>It will create all the necessary prerequisites outlined in Karpenter's [Getting Started](https://karpenter.sh/docs/getting-started/) section including installing Karpenter itself using Helm.

Which includes the created node role that the Getting Started is talking about. :)

But, alas, I do agree that it is useful to expose that setting to `eksctl` config too and is a relatively easy thing to do.

niallthomson:
I agree that the documentation is accurate, but I'd just say its missing some details. For example, the IAM instance profile name (`eksctl-KarpenterNodeInstanceProfile-$CLUSTER_NAME`) is different from the Karpenter Getting Started page (`KarpenterNodeInstanceProfile-${CLUSTER_NAME}`).

These are minor things, overall the integration is fantastic it could just have some little quality of life enhancements :)

Skarlso:
Ah, right. The problem there is, that we prefix everything with `eksctl` and that is definitely missing from the doc. :/ We need the `eksctl` prefix because of the IAM permissions that we have configured for resources that begin wth that prefix.

niallthomson:
I'm happy to submit a PR that introduces a flag for eksctl that sets the default instance profile, for example:

```
karpenter:
   setDefaultInstanceProfile: true
```

Skarlso:
No need for the `set` otherwise LGTM. Please don't forget to update the test which checks if the helm value is passed in correctly. Thanks! :)

Skarlso:
@niallthomson also, thank you that would be very welcomed! :)

afolarin:
>  For example, the IAM instance profile name (eksctl-KarpenterNodeInstanceProfile-$CLUSTER_NAME) 
Came here to post precisely this.

Skarlso:
[...]

PR #4751, opened 2022-02-05
Add support for Karpenter defaultInstanceProfile setting

niallthomson:
### Description

Added support to pass in `defaultInstanceProfile` parameter for Karpenter which configures the default IAM instance profile via the Karpenter Helm chart. When enabled this configures Karpenter to use the node IAM instance profile created by CloudFormation as the default instance profile, which can be overridden in Provisioners if required.

Feature turned on using:

```
karpenter:
  defaultInstanceProfile: true
```

Setting is disabled by default.

Issue #4745 

### Checklist
- [x] Manually tested


Skarlso:
@niallthomson Hi! Thank you for you PR! One thing I don't understand is that `defaultInstanceProfile` is a string in the helm chart. Which means it's configurable to an arbitrary value instead of a set one. But you are using it as a bool value, is there a particular reason for it?

Further, you seem to be passing in the ARN value of the profile, which is not the correct value for this setting. It should just be the name like in the getting started doc here:

```
  --set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME} \
```

pkg/actions/karpenter/create.go
```diff
@@ -73,6 +73,8 @@ func (i *Installer) Create() error {
 		return fmt.Errorf("failed to save the identity config: %w", err)
 	}
 
+	instanceProfileArn := fmt.Sprintf("arn:aws:iam::%s:instance-profile/eksctl-%s-%s", parsedARN.AccountID, builder.KarpenterNodeInstanceProfile, i.Config.Metadata.Name)
+
 	// Install Karpenter
-	return i.KarpenterInstaller.Install(context.Background(), roleARN)
+	return i.KarpenterInstaller.Install(context.Background(), roleARN, instanceProfileArn)
```
Skarlso, line 79: Okay, so this is a cool approach. However, what would be better is to have this as a configuration value in the config file surfaced. Which means users can set it, BUT it needs a default value of the generated one.

So leave this as is, have an optional field of type string and if it IS defined, then just overwrite this value.

But also, pass it through as you do before but instead of a boolean, just always set the value and always provide this field in the helmchart since we never want to NOT provide it because `eksctl` is using a custom value.

niallthomson:
Hey Skarlo,

Thanks for taking a look at the PR!

With regards to the first point, I was limiting the scope of this particular PR to a user that is thinking: "I want `eksctl` to create the IAM instance profile for me AND wire it in as the default instance profile to the Karpenter Helm chart". This means that Karpenter can name the instance profile whatever it wants in CloudFormation and I as a user don't care, I just tell `eksctl` I want it wired in by default in Karpenter as well. This to me seems like a boolean choice,

I think the ability to pass in an arbitrary IAM instance profile in the config file is a very related but separate feature here. I would perhaps even suggest this could be implemented by adding a generic `chartValues` field or similar to the spec since I can imagine users wanting to potentially override anything in the Helm chart eventually (nodeSelectors, tolerations etc).

For the second point I might have screwed up my commit, I tried this initially and it didn't work so I changed it to the name.  I'll need to rework it. Lets figure out the first item and I'll sort this with it.

Skarlso:
>I think the ability to pass in an arbitrary IAM instance profile in the config file is a very related but separate feature here. I would perhaps even suggest this could be implemented by adding a generic chartValues field or similar to the spec since I can imagine users wanting to potentially override anything in the Helm chart eventually (nodeSelectors, tolerations etc).

That is a fair point to make, however, we would like to not be a wrapper around Helm too much. We will probably expose the bare minimum of settings, definitely not all of them.

`eksctl` tries to be a very clean, very short wrapper around Karpetner, so people can get started with it, but not necessarily something that deals with the whole Helm life cycle.

niallthomson:
Perhaps the naming of the configuration parameter is not correct then. I feel like `defaultInstanceProfile` should behave as you outlined, being a name of an IAM instance profile. Perhaps this flag I'm considering should be something like `createDefaultInstanceProfile` that governs both whether its created and whether its set in the Helm chart?

I think you have a point, its just my priority here was the convenience of grabbing the profile `eksctl` already makes and passing it through.

Skarlso:
You will still retain that! :) This PR is on a very good path for that. In fact, you just need a slight adjustment of making it a string, and if set, using that value, if not, use the eksctl value. And always provide it, because `eksctl` also needs to set it. So this:

```go
	defaultInstanceProfileValue := ""
	if *k.ClusterConfig.Karpenter.DefaultInstanceProfile {
		defaultInstanceProfileValue = instanceProfileARN
	}
	values := map[string]interface{}{
		controller: map[string]interface{}{
			clusterName:     k.ClusterConfig.Metadata.Name,
			clusterEndpoint: k.ClusterConfig.Status.Endpoint,
		},
		aws: map[string]interface{}{
			defaultInstanceProfile: defaultInstanceProfileValue,
		},
		serviceAccount: serviceAccountMap,
	}
```

becomes this:

```go
	values := map[string]interface{}{
		controller: map[string]interface{}{
			clusterName:     k.ClusterConfig.Metadata.Name,
			clusterEndpoint: k.ClusterConfig.Status.Endpoint,
		},
		aws: map[string]interface{}{
			defaultInstanceProfile: instanceProfileName,
		},
		serviceAccount: serviceAccountMap,
	}
```

and that is passed in to `Install`.

niallthomson:
Ok awesome. I did consider that path but it obviously changes the behavior thats already been released without a flag. If that is acceptable I'm totally happy with that change.

Skarlso:
Yeah, sounds good to me. Go for it. :) And thanks!

niallthomson:
I've added what I think should be the appropriate changes. Let me know if I messed up how I merged it.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -59,8 +59,10 @@ func SetClusterConfigDefaults(cfg *ClusterConfig) {
 		cfg.VPC.ManageSharedNodeSecurityGroupRules = Enabled()
 	}
 
-	if cfg.Karpenter != nil && cfg.Karpenter.CreateServiceAccount == nil {
-		cfg.Karpenter.CreateServiceAccount = Disabled()
+	if cfg.Karpenter != nil {
```
Skarlso, line 62: What prompted this change?

pkg/karpenter/karpenter.go
```diff
@@ -38,7 +40,7 @@ type Options struct {
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 //counterfeiter:generate -o fakes/fake_chart_installer.go . ChartInstaller
 type ChartInstaller interface {
-	Install(ctx context.Context, serviceAccountRoleARN string) error
+	Install(ctx context.Context, serviceAccountRoleARN string, instanceProfileARN string) error
```
Skarlso, line 43: It's just the name, not the ARN. :)

pkg/actions/karpenter/create.go
```diff
@@ -73,6 +73,11 @@ func (i *Installer) Create() error {
 		return fmt.Errorf("failed to save the identity config: %w", err)
 	}
 
+	instanceProfileName := fmt.Sprintf("eksctl-%s-%s", builder.KarpenterNodeInstanceProfile, i.Config.Metadata.Name)
```
Skarlso, line 76: Almost there. :) But, this won't work, because it's not what it creates. This is what you are looking for:

```go
	instanceProfileName := gfnt.MakeFnSubString(fmt.Sprintf("eksctl-%s-%s", KarpenterNodeInstanceProfile, k.clusterSpec.Metadata.Name))
	instanceProfile := gfniam.InstanceProfile{
		InstanceProfileName: instanceProfileName,
		Path:                gfnt.NewString("/"),
		Roles:               gfnt.NewSlice(roleRef),
	}
	k.newResource(KarpenterNodeInstanceProfile, &instanceProfile)
```

In the file `pkg/cfn/builder/karpenter.go`. This where the role is actually created. Here is where you should use this name or the generated name.

That's this part in this file:

```go
	// Create IAM roles
	taskTree := newTasksToInstallKarpenterIAMRoles(i.Config, i.StackManager, i.CTL.Provider.EC2())
	if err := doTasks(taskTree); err != nil {
		return err
	}
```

The config is passed in there too, so you can use that.

userdocs/src/usage/eksctl-karpenter.md
```diff
@@ -34,8 +34,9 @@ to be set:
 
 ```yaml
 karpenter:
-  version: '0.4.3'
-  createServiceAccount: true # default is false  
+  version: '0.6.0'
+  createServiceAccount: true # default is false
+  defaultInstanceProfile: 'KarpenterNodeInstanceProfile' # default is to use the IAM instance profile created by eksctl
```
Skarlso, line 39: This is great! But, also please add `-cluster-with-karpenter` to signify that the role should be postfixed with the cluster name so other clusters with Karpenter don't get into a naming conflict. :)

Skarlso:
@niallthomson Well done! Almost there. :) Just a few changes needed and we should be golden. :)

pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -744,6 +744,9 @@ type Karpenter struct {
 	// CreateServiceAccount create a service account or not.
 	// +optional
 	CreateServiceAccount *bool `json:"createServiceAccount,omitempty"`
+	// DefaultInstanceProfile override the default IAM instance profile
+	// +optional
+	DefaultInstanceProfile string `json:"defaultInstanceProfile,omitempty"`
```
Skarlso, line 749: Please make this of type `*string` and then get the value as `aws.StringValue()`. The reason for that is that we truly would like to have this as an optional field. Which means that it's either set, is set to empty explicitly or set to nil which will completely remove it from parsed JSON data.

pkg/actions/karpenter/create.go
niallthomson: My intent here was not to rename what eksctl creates but to allow you to use your own IAM instance profile, hence the nature of the changes. If I interpret correctly what you propose would change the name of the IAM instance profile created by eksctl in CFN correct?

pkg/apis/eksctl.io/v1alpha5/types.go
niallthomson: Thanks, I did wonder about this but didn't see any immediate examples of `*string` hence what I went with. Will get this corrected.

pkg/apis/eksctl.io/v1alpha5/defaults.go
niallthomson: I didn't completely revert the change there appropriately. I had originally added additional validation logic there for the initial boolean flag. I've changed this back since I assume there is no defaulting of the value since its now a string.

pkg/apis/eksctl.io/v1alpha5/types.go
Skarlso: Or `ContainerRuntime`.

Skarlso:
@niallthomson Hello. :) Do you have any updates by any chance? This is kind-a needed to make Karpenter work. :D

niallthomson:
> 

@Skarlso I'll update the PR this morning, apologies I got pulled away on other things.

Skarlso:
@niallthomson It's okay, thank you so much for keeping with this! I truly appreciate it! :)

niallthomson:
@Skarlso I'm doing a final round of testing then I'll post here

niallthomson:
@Skarlso I need to update the test in `karpenter_test.go` for the cfn builder part. Right now it just compares a static template using all the expected generated values, but I figure I don't want to just copy the whole string again. Any suggestions what changes I should make here?

```
var expectedTemplate = `{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Description": "Karpenter Stack [created and managed by eksctl]",
  "Mappings": {
    "ServicePrincipalPartitionMap": {
      "aws": {
        "EC2": "ec2.amazonaws.com",
        "EKS": "eks.amazonaws.com",
        "EKSFargatePods": "eks-fargate-pods.amazonaws.com"
      },
      "aws-cn": {
        "EC2": "ec2.amazonaws.com.cn",
        "EKS": "eks.amazonaws.com",
        "EKSFargatePods": "eks-fargate-pods.amazonaws.com"
      },
      "aws-us-gov": {
        "EC2": "ec2.amazonaws.com",
        "EKS": "eks.amazonaws.com",
        "EKSFargatePods": "eks-fargate-pods.amazonaws.com"
      }
    }
  },
[...]
```

Skarlso:
@niallthomson what is it that you want to change? Is it a dynamic value? You can just do `%s` and then `fmt.Stringf()` to replace it with a value.

niallthomson:
@Skarlso I think we are good now!

Skarlso:
Fantastic! Thank you very much!

Skarlso:
@niallthomson ops, there is still an error. :)

Skarlso:
Ignore the link checker failure. :)

Skarlso:
@niallthomson Thank you! Awesome stuff! :)))

Skarlso (approved):

Issue #4696, opened 2022-01-26
eksctl create nodegroup fails with "One or more subnets are duplicate" on us-west-1


sheetaljoshi:
Trying to create a nodegroup using the below yaml, but fails with "One or more subnets are duplicate" on us-west-1.

```yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: my-ipv6-cluster
  version: "1.21"
  region: us-west-1

kubernetesNetworkConfig:
  ipFamily: IPv6

vpc:
  clusterEndpoints:
    publicAccess: true
    privateAccess: true

iam:
  withOIDC: true

addons:
  - name: vpc-cni
    version: v1.10.1-eksbuild.1 # optional
  - name: coredns
    version: v1.8.4-eksbuild.1 # optional
  - name: kube-proxy
    version: v1.21.2-eksbuild.2 # optional

managedNodeGroups:
  - name: x86-al2-on-demand-xl
    amiFamily: AmazonLinux2
    instanceTypes: [ "m6i.xlarge", "m6a.xlarge" ]
    minSize: 1
    desiredCapacity: 2
    maxSize: 3
    volumeSize: 100
    volumeType: gp3
    volumeEncrypted: true
    ssh:
      allow: true
      publicKeyName: ipv6-ssh-key
    updateConfig:
      maxUnavailablePercentage: 33
    labels:
      os-distribution: amazon-linux-2
```

and the error is 

```console
2022-01-26 15:29:10 [ℹ]  waiting for CloudFormation stack "eksctl-my-ipv6-cluster-nodegroup-x86-al2-on-demand-xl"
2022-01-26 15:29:11 [✖]  unexpected status "ROLLBACK_IN_PROGRESS" while waiting for CloudFormation stack "eksctl-my-ipv6-cluster-nodegroup-x86-al2-on-demand-xl"
2022-01-26 15:29:11 [ℹ]  fetching stack events in attempt to troubleshoot the root cause of the failure
2022-01-26 15:29:11 [✖]  AWS::EKS::Nodegroup/ManagedNodeGroup: CREATE_FAILED – "One or more subnets are duplicate (Service: AmazonEKS; Status Code: 400; Error Code: InvalidParameterException; Request ID: e35143d6-cf55-4807-a6a3-26cdc993a841; Proxy: null)"
[...]

PR #4770, opened 2022-02-10
refactor AZ package to be simpler

aclevername:
### Description

It was very complex before, and also required us to do further checks before calling the package such as https://github.com/weaveworks/eksctl/blob/bbebfbb7932f3663196d6a1f9b5dde2dc692ed70/pkg/eks/api.go#L388

closes https://github.com/weaveworks/eksctl/issues/4696


pkg/az/az.go
```diff
@@ -1,168 +1,62 @@
...
 
-// SelectZones returns a list fo az zones to use for the supplied region
-func (a *AvailabilityZoneSelector) SelectZones() ([]string, error) {
-	availableZones, err := a.getZonesForRegion()
-	if err != nil {
-		return nil, err
+func randomSelectionOfZones(region string, availableZones []string) []string {
+	var zones []string
+	desiredNumberOfAZs := api.RecommendedAvailabilityZones
+	if region == api.RegionUSEast1 {
```
Skarlso, line 39: I'm sorry, my memory is failing me, but why is there a restriction for us-east-1 specifically? I remember you said something about this, but don't remember the detail. Which means, maybe it's worth having a note here, since if I don't remember it now, I won't remember it in 2 months when I look at this code segment again for whatever reason and I would need to add or remove a region. :)

pkg/az/az.go
```diff
@@ -1,168 +1,62 @@
...
-}
-
-func (a *AvailabilityZoneSelector) getUsableZones(availableZones []*ec2.AvailabilityZone) []string {
-	usableZones := []string{}
-	for _, zone := range availableZones {
-		zoneUsable := true
-		for _, rule := range a.rules {
-			if !rule.CanUseZone(zone) {
-				zoneUsable = false
+	for len(zones) < desiredNumberOfAZs {
```
Skarlso, line 43: Yeah okay, I see what's going on:

```go
	for len(zones) < desiredNumberOfAZs {
		rand := rand.New(rand.NewSource(time.Now().UnixNano()))
		for _, rn := range rand.Perm(len(availableZones)) {
			zones = append(zones, availableZones[rn])
			if len(zones) == desiredNumberOfAZs {
				break
			}
		}
	}
```

This is like this:

```go
		rand := rand.New(rand.NewSource(time.Now().UnixNano()))
		for _, rn := range rand.Perm(len(availableZones)) {
			zones = append(zones, availableZones[rn])
			if len(zones) == desiredNumberOfAZs {
				break
			}
		}
```

If this ever goes wrong than there is something wrong, right? You don't need the outer for. You either exactly the right number of zones or greater number of zones so this would just quit early.

Skarlso (approved):
Couple suggestions, otherwise, LGTM 👍

pkg/az/az.go
aclevername: image

pkg/az/az.go
```diff
@@ -173,10 +67,31 @@ func (a *AvailabilityZoneSelector) getZonesForRegion() ([]*ec2.AvailabilityZone,
...
+		return nil, fmt.Errorf("error getting availability zones for region %s: %w", region, err)
 	}
 
-	return output.AvailabilityZones, nil
+	return filterZones(region, output.AvailabilityZones), nil
+}
+
+func filterZones(region string, zones []*ec2.AvailabilityZone) []string {
+	var filteredZones []string
+	azsToAvoid := zoneIDsToAvoid[region]
```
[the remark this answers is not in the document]
aclevername:
```
map[string]map[string]struct{
	api.RegionCNNorth1: {
"cnn1-az4": struct{}{},
    }
}
```

While this is probably a more efficient solution, reality is its a bit more complex & messy to read. going to leave as is if its okay :smile:

pkg/az/az.go
Skarlso: No f*cking way. :D Goddamnit, I thought that's new. :D

pkg/az/az.go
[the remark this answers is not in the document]
Skarlso: Is that messy? I don't think that's messy 🤔. Also, you can delete like an entire function! :D That's always nice. :)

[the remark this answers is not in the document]
aclevername:
```
map[string]map[string]struct{
	api.RegionCNNorth1: {
"cnn1-az4": struct{}{},
    }
}
```

hurts my brain. 

> entire function

True! But its a relatively simple to understand func, so it doesn't add much complexity

Issue #4302, opened 2021-10-05
Add MaxInstanceLifetime to ASG/LT


bitva77:
**What feature/behavior/change do you want?**

Add the ability to set the maximum instance lifetime of instances in node groups:

https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-max-instance-lifetime.html

**Why do you want this feature?**

We'd like to ensure that our instances don't hang around for too long and so this would be a nice feature to have.

Skarlso:
@bitva77 If you still would like to take a stab on this, please feel free to do so. :)

Himangini:
We can't support this setting for managed nodegroups, we'll add support for unmanaged nodegroups as part of this ticket.

cPu1:
There's a feature request for this open in containers-roadmap: https://github.com/aws/containers-roadmap/issues/1190

cPu1:
> Can you add it to the backlog?
> 
> I'm not coder but I may know enough to fake it so I might take a stab at a PR as it sounds like there's previous examples to go off of. but don't count on me :)
> 
> Thanks!

We have added this to our backlog and will start work on it soon. Please note that this will feature will be added for unmanaged nodegroups only, as managed nodegroups will require support from the [EKS API](https://github.com/aws/containers-roadmap/issues/1190).

Himangini:
Closing this since we have merged the PR https://github.com/weaveworks/eksctl/pull/4781 for Unmanaged Nodegroups support

PR #4781, opened 2022-02-14
Added MaxInstanceLifetime for unmanaged nodegroups

Skarlso:
Deals with https://github.com/weaveworks/eksctl/issues/4302 for unmanaged nodegroups.

Running manual test... success:

config:

```yaml
kind: ClusterConfig

metadata:
  name: test-max-instance-lifetime-1
  region: us-west-2
  version: '1.20'

nodeGroups:
  - name: ng-1
    minSize: 1
    maxSize: 2
    desiredCapacity: 1
    maxInstanceLifetime: 89000
```

template:
```json
    "NodeGroup": {
      "Type": "AWS::AutoScaling::AutoScalingGroup",
      "Properties": {
        "DesiredCapacity": "1",
        "LaunchTemplate": {
          "LaunchTemplateName": {
            "Fn::Sub": "${AWS::StackName}"
          },
          "Version": {
            "Fn::GetAtt": [
              "NodeGroupLaunchTemplate",
              "LatestVersionNumber"
            ]
          }
        },
        "MaxInstanceLifetime": 89000,
```


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -979,6 +979,11 @@ type NodeGroup struct {
 	// ContainerRuntime defines the runtime (CRI) to use for containers on the node
 	// +optional
 	ContainerRuntime *string `json:"containerRuntime,omitempty"`
+
+	// MaxInstanceLifetime defines the maximum amount of time in seconds an instance stays alive.
+	// To clear a previously set value, this must be set to 0 explicitly.
+	// +optional
+	MaxInstanceLifetime *int `json:"maxInstanceLifetime,omitempty"`
```
aclevername, line 986: is providing this in seconds practical :thinking: I know thats what they want in the stack, we can always convert a more meaningful value into seconds. Perhaps use the `time` type? Although looking behind scenes thats only ` Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`, but providing it this way might be slightly easier?

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -105,6 +105,41 @@ var _ = Describe("ClusterConfig validation", func() {
...
+	Describe("nodeGroups[*].maxInstanceLifetime validation", func() {
+		It("should reject if value is below a day", func() {
+			cfg := api.NewClusterConfig()
+			ng0 := cfg.NewNodeGroup()
+			ng0.Name = "node-group"
+			ng0.MaxInstanceLifetime = aws.Int(5)
+			err := api.ValidateClusterConfig(cfg)
+			Expect(err).NotTo(HaveOccurred())
+			err = api.ValidateNodeGroup(0, ng0)
+			Expect(err).To(HaveOccurred())
```
aclevername, line 117: but what sort of error :thunk:

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -32,6 +32,7 @@ const (
 	MaxIO1Iops    = 64000
 	MinGP3Iops    = DefaultNodeVolumeGP3IOPS
 	MaxGP3Iops    = 16000
+	OneDay        = 86400
```
aclevername, line 35: https://open.spotify.com/track/1OZSVl0JJ1MBzibpuhmmXb?si=7c320cabb5c44e20

pkg/apis/eksctl.io/v1alpha5/types.go
Skarlso: I disagree. I would like to be consistent with what CF provides, because that's well documented. If you change that that's unexpected, which means, you have to explicitly document it so it's apparent which means more burden on the user to discover stuff for the minor addition of using hours instead of secods.

pkg/apis/eksctl.io/v1alpha5/validation_test.go
Skarlso: Ops. :D

pkg/apis/eksctl.io/v1alpha5/types.go
aclevername: > I disagree. I would like to be consistent with what CF provides

folks who use eksctl are necessarily all that familiar with CF 

> because that's well documented. If you change that that's unexpected, which means, you have to explicitly document it so it's apparent which means more burden on the user to discover stuff for the minor addition of using hours instead of secods.

they would have to discover this field in our docs regardless?

I'm open to leaving it as is, but if a day is the minimum value do we expect people to really want the granular control of seconds? Or is setting it in multiples of days/hours a better UX

pkg/apis/eksctl.io/v1alpha5/validation.go
Skarlso: 😄

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -105,6 +105,41 @@ var _ = Describe("ClusterConfig validation", func() {
...
+		})
+		It("should accept 0 as a value as that clears previous settings", func() {
```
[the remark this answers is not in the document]
aclevername: right but how would you do this from an eksctl command :smile:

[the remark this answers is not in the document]
Skarlso: Ummm... that's a good question. I assumed you can update. :D I mean you can change desired capacity in the config too, right? You can call `eksctl nodegroup update`, right?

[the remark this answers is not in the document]
aclevername: i think that onlys for changing scaling config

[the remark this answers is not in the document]
Skarlso: Ah, right. Then we will not allow updating this I guess.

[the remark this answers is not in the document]
Skarlso: I have to look into updating nodegroups via stack. That's the only way to update this.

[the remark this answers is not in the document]
Skarlso: Oh, we don't. Alright then.

aclevername (approved):

cPu1 (approved):

PR #4782, opened 2022-02-14
Move list clusters into cluster pkg and add tests

### LINKED ISSUE #4565: Refactor listing clusters in all regions from ClusterProvider
`ClusterProvider` has a `ListClusters(chunkSize int, listAllRegions bool)` method that creates a new `ClusterProvider` for each supported region when `listAllRegions` is true, thus polluting it with logic to create itself and making its creation from outside using dependency injection useless. Listing clusters in all regions should not be a part of `ClusterProvider`.

Related: https://github.com/weaveworks/eksctl/issues/4213

aclevername:
### Description
Closes #4565 

Refactors it into `pkg/actions/cluster` and adds tests


Skarlso:
And I'm not allowed to comment on existing code, right? :D

aclevername:
> And I'm not allowed to comment on existing code, right? :D

be gentle

pkg/actions/cluster/get.go
```diff
@@ -0,0 +1,139 @@
...
+	eksctlCreatedUnknown api.EKSCTLCreated = "Unknown"
+)
+
+type Description struct {
+	Name   string
+	Region string
+	Owned  api.EKSCTLCreated
+}
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
```
Skarlso, line 27: So the one below doesn't have this, right? I thought this isn't needed. :D or does it only matter once and then it's used for everything? :thunk:

aclevername: If I remember correctly its 1 `go:generate` per package

pkg/actions/cluster/get.go
```diff
@@ -0,0 +1,139 @@
...
+			WaitTimeout: provider.WaitTimeout(),
+		}, nil)
+
+		if err != nil {
+			logger.Critical("error creating provider in %q region: %v", region, err)
+			continue
+		}
+
+		newClusters, err := listClusters(ctl.Provider, int64(chunkSize))
+		if err != nil {
```
Skarlso, line 72: If I see that right, I don't see an error case for this scenario, when there is a failure and the cluster is not added.

Skarlso (approved):
A single comment about a missing test case, otherwise superb refactor, well done.

pkg/actions/cluster/get.go
aclevername: you caught me. I was being lazy

Skarlso: understandable

Issue #3961, opened 2021-07-08
Automatically create ASG tag from nodegroup's labels and taints


nikimanoledaki:
### Background
The Autoscaling group (ASG) uses tags to detect and scale a nodegroup. Such tags are required for the ASG to scale a nodegroup from 0, for example if it was created with `desiredCapacity: 0`.

More context:
- #3793
- https://github.com/aws/containers-roadmap/issues/724#issuecomment-859948729
- #1481 
- #2263

### Problem
Currently, [as per the eksctl docs](https://eksctl.io/usage/autoscaling/#scaling-up-from-0), there is a bit of repetition happening in the config file when we define labels/taints and need them to be added as tags:
```    
nodeGroups:
  - name: ng1-public
    ...
    labels:
      my-cool-label: pizza
    taints:
      feaster: "true:NoSchedule"
    tags:
      k8s.io/cluster-autoscaler/node-template/label/my-cool-label: pizza
      k8s.io/cluster-autoscaler/node-template/taint/feaster: "true:NoSchedule"
```

### Proposed Solution
As suggested by @sammort  in https://github.com/weaveworks/eksctl/issues/3793#issuecomment-875470794, it would be great if the ASG tags were created automatically when labels and/or taints have been defined. 

There is a limit of maximum of 50 tags for an ASG, but there is a [proposal](https://github.com/kubernetes/autoscaler/blob/91bc8e3783aca9b1770675944da18b42ae19d96e/cluster-autoscaler/proposals/circumvent-tag-limit-aws.md) to change this for the Kubernetes cluster autoscaler. For the time being, from our side, we might want to validate that the number of labels, taints, and tags in a config are not greater than 50. Not sure if there is an error from EKS if the count is greater than 50 so this would be good to investigate too. 

Another thing to keep in mind is tag duplicates.

TBBle:
I think #1481 was looking for the same thing, but closed-out due to inactivity. Also some overlay with #2263, which proposed a NodeGroup config option, e.g. `tagForAutoScaler`, to add the core CA tags, and which would make sense to control this tag-copying too.

AndrewFarley:
So, there just is no way I can figure out to scale from 0 reliably without someone implementing this feature afaik?  I can't figure out all the various tag "hints" I need to provide to trick CA into scaling up from 0 reliably, especially when using nodeSelectors.  Can we prioritize this?

TBBle:
[...]

Issue #2263, opened 2020-05-29
Support adding correct Cluster Autoscaler tags to a nodegroup without granting self-scaling permissions


TBBle:
**Why do you want this feature?**

Currently, the correct Cluster Autoscaler tags are only created for a NodeGroup if that NodeGroup also has the Instance Role created to _run_ the Cluster Autoscaler in that group.

This has a few misusages:
* Only the nodegroup where the Cluster Autoscaler is running needs those permissions, but the [documentation](https://eksctl.io/usage/autoscaling/#enable-auto-scaling) implies `--asg-access` (the command-line equivalent of the IAM instance role addon for CA) is the way to mark nodegroups that will _be_ auto-scaled (or rather, it doesn't usefully canvas the alternative of adding those tags yourself).
* When using IRSA, _none_ of the nodegroups will have the IAM instance role, and so none get the tags they need.
* You cannot run the Cluster Autoscaler in a NodeGroup that is _not_ set up for autoscaling, because the same flag does both.

The current workaround is to manually insert the tags into each NodeGroup, while avoiding typos and bitrot, e.g, the pair of tags with a comment at the [IRSA example config](https://github.com/weaveworks/eksctl/blob/9bea46eea14ef31b085528ab4f15dce24b4ab89e/userdocs/src/usage/iamserviceaccounts.md#usage-with-config-files). The presence of the comment suggests this is not a clear UX.

See also https://github.com/weaveworks/eksctl/issues/1481#issuecomment-616792355 which has separately discovered that it's weird to have the "can be autoscaled" setup only be done for "can host autoscaler" nodegroups, since in the scale-up-from-zero case, you'd clearly _never_ try and host the autoscaler in a nodegroup that can reach zero running nodes.

> The scaling targets do not need any additional IAM permissions, and the best practice would be to not grant this permission where it is not needed.

**What feature/behavior/change do you want?**

An extra config flag for `NodeGroup`, something like `enableAutoscaling` which would add the [appropriate tags](https://github.com/weaveworks/eksctl/blob/c84ce3d172c5e4ceef75dbb237d9b5717644e237/pkg/cfn/builder/nodegroup.go#L172). The name needs to make clear that it makes that NodeGroup a candidate for auto-scaling, but does not set up any permissions that might be needed to run the Cluster Autoscaler.

[...]

PR #4785, opened 2022-02-14
Added automatically generating ASG tags from labels and taints

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/3961


userdocs/src/usage/autoscaling.md
```diff
@@ -44,6 +44,21 @@ nodeGroups:
       k8s.io/cluster-autoscaler/node-template/taint/feaster: "true:NoSchedule"
 ```
 
+This is done by `eksctl` automatically if `desiredCapacity` is set to `0`, there is no need to specify these by hand.
```
nikimanoledaki, line 47:
```suggestion
For unmanaged noderoups, this is done by `eksctl` automatically if `desiredCapacity` is set to `0`, there is no need to specify these by hand. This feature is not supported for managed nodegroups.
```

pkg/cfn/builder/nodegroup.go
```diff
@@ -262,12 +265,57 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
...
 	}
 
+	// not using aws.IntValue because it's not a required field
+	// and aws.IntValue would result in 0 if it's nil.
+	if n.spec.DesiredCapacity != nil && *n.spec.DesiredCapacity == 0 {
+		clusterTags, err := generateClusterAutoscalerTags(n.spec)
+		if err != nil {
+			return err
+		}
+		tags = append(tags, clusterTags...)
```
nikimanoledaki, line 275: what if `len(tags)` is greater than 50 here, after they have been appended to the rest of the tags? maybe that validation should be moved here.

nikimanoledaki (approved):
A couple of suggestions, otherwise LGTM!

pkg/cfn/builder/nodegroup.go
Skarlso: Oh snap! You're right!! :O

Skarlso: Frigging love you Niki. Seriously.

aclevername (approved):
lgtm 1 comment

pkg/cfn/builder/nodegroup.go
```diff
@@ -262,12 +265,57 @@ func (n *NodeGroupResourceSet) addResourcesForNodeGroup() error {
...
+	}
+	if len(tags) > MaximumTagNumber {
+		return fmt.Errorf("number of tags is exceeding the configured amount %d, was: %d", MaximumTagNumber, len(tags))
```
aclevername, line 278:
```suggestion
	if len(tags) > MaximumTagNumber {
	    if n.spec.DesiredCapacity != nil && *n.spec.DesiredCapacity == 0 {
	       		return fmt.Errorf("number of tags is exceeding the configured amount %d, was: %d. Due to desiredCapacity==0 we added an additioanl %d tags to ensure the nodegroup is scaled correctly", MaximumTagNumber, len(tags))
	    }
		return fmt.Errorf("number of tags is exceeding the configured amount %d, was: %d", MaximumTagNumber, len(tags))
```

aclevername (approved):
:sparkles: thank you

TBBle:
Just saw this MR... It's late to say, but this probably shouldn't be tied to 'desiredCapacity=0': it's a reasonable use-case for the initial `desiredCapacity` to be, say, 1, to allow, e.g., testing of the newly-deployed nodes, and _then_ scale the group down to 0 (or allow the cluster auto-scaler to scale it down later).

If anything, `minSize=0` might have been a better choice of automatic trigger, since that's what the Cluster Autoscaler would use to decide if it can scale the nodegroup down to 0. (And if unspecified, minSize defaults to desiredCapacity, so the behaviour of the current change is maintained in that case).

Skarlso:
@TBBle I was thinking about that. Also, this fails during `scale` events because it's only inside the node building which is triggered only during the create process. If like you said, this is applied during scale event, the labels won't be transferred correctly.

TBBle:
Yeah, once the nodes are created, I don't think it makes sense to change the definition of the nodes, partly because scaling might be done without the original eksctl config available, and partly because I suspect that might confuse the Cluster Autoscaler (although it'd probably be fine, at worst the wrong nodegroup gets temporarily scaled up, and some pods end up waiting an extra CA cycle or two).

Perhaps it'd actually be better to just expose this behaviour as an explicit boolean config option, rather than trying to derive it from existing config options. I suggested such an option in https://github.com/weaveworks/eksctl/issues/3961#issuecomment-876572534, and that could _also_ resolve #2263 by moving [the tags that activate node-group autoscaling](https://github.com/weaveworks/eksctl/blob/main/pkg/cfn/builder/nodegroup.go#L254-L264) into the newly-added `generateClusterAutoscalerTags` function. (Since where they are added now is actually semantically backwards, see #2263 for the full explanation).

AndrewFarley:
Fully agree this shouldn't be tied to minSize == 0.  What if an user later scaled the ASG to 0, it will break cluster-autoscaler because of the missing annotation.  Please un-tie this to minSize==0 ASAP I'd say.

Skarlso:
@AndrewFarley sorry, I'm not following. It isn't tied to minSize=0.

AndrewFarley:
@Skarlso Sorry, reviewed the code, I mean this...
```
	if n.spec.DesiredCapacity != nil && *n.spec.DesiredCapacity == 0 {
```
That conditional should be removed.

Skarlso:
@AndrewFarley why?

AndrewFarley:
@Skarlso Your logic above will only add tags if someone "from day one" starts with desired instances 0.  This causes it to miss the edge case of if someone scales down their ASG manually to desired instances 0 later.

Skarlso:
Correct. So... why remove this? I agree that there needs to be a followup that does that during scaling, but I don't see why I have to remove this. :) Sorry. :) I might just be tired. :)

Skarlso:
@AndrewFarley ah I see what you mean. You **always** want to add these tags.

I don't see any reason why not right now. But I have to think about this. This might be a problem where we could break people's builds if they have waay to many taints and labels. And suddenly they have to provide an extra setting to disable this. Might not be a problem, but I have to think about it.

AndrewFarley:
@Skarlso Optionally, if I can't always have these tags, can we at least add an argument/option to specify to eksctl at the global level to force these tags to always be enabled?  This way I can remove the manual tags I standardize on all my clients nodeGroups.  Because with your above logic, I'd need to start out all nodeGroups with 0 desired capacity which is quite bad in my experience because then EKSCTL doesn't wait for 1 to get healthy and add to EKS which is a nice sanity check when creating new node groups.  Eg:

```
metadata:
  name: master
  region: us-east-2
  version: "1.21"
iam:
  withOIDC: true
cluster-autoscaler:  # See here
  tagAlways: true      # and here
```

^ Can we do something like this, and make that same conditional above have an "or" in it with tagAlways: true?

Skarlso:
Sure, that's doable. Or something like `autoPropagateLabelsAndTains: true`. Bit expressive though. :)

AndrewFarley:
> Sure, that's doable. Or something like `autoPropagateLabelsAndTains: true`. Bit expressive though. :)

Figure out and maybe ask around the other authors/contributors and find a good standard name for this, and lets do it!  Cheers!!!

TBBle:
[`tagForAutoscaler`](https://github.com/weaveworks/eksctl/issues/3961#issuecomment-876572534) perhaps?

Skarlso:
https://github.com/weaveworks/eksctl/issues/4853 created to follow discussion better.

Issue #2883, opened 2020-12-01
missing --region from nodegroup docs page


todd-dsm:
**What happened?**
Built a first cluster with eksctl and without issue; got what I believe to be expected results.

Asked eksctl to describe the nodegroup from the docs:

```bash
eksctl get nodegroup --cluster=cluster-1 --name=ng-14800d04
Error: getting nodegroup stack summaries: getting nodegroup stacks: no eksctl-managed CloudFormation stacks found for "cluster-1"
```

**What you expected to happen?**
The [managing-nodegroups] page needs to be updated so the first instruction reads:

`eksctl create nodegroup --cluster=<clusterName> --region=<region> [--name=<nodegroupName>]`

**How to reproduce it?**
omit `--region` argument from the command; should fail. cost me a half-hour of head-scratching; please fix for the next person.

The command won't return an expected result without `--region`.

**Versions**

```
$ eksctl version
0.32.0
```

[managing-nodegroups]:https://eksctl.io/usage/managing-nodegroups/

todd-dsm:
Also applies to this page:
https://eksctl.io/usage/iam-identity-mappings/

Without `--region`:
```bash
$ eksctl get iamidentitymapping --cluster=apps-dev
Error: fetching cluster status to determine operability: unable to describe cluster control plane: ResourceNotFoundException: No cluster found for name: apps-dev.
{
  RespMetadata: {
    StatusCode: 404,
    RequestID: "07485935-27ab-4839-a881-5ee6664cdaf3"
  },
  Message_: "No cluster found for name: apps-dev."
}
```

With `--region`:
```bash
$ eksctl get iamidentitymapping --cluster=apps-dev --region=us-west-2
ARN												USERNAME				GROUPS
arn:aws:iam::378220085327:role/eksctl-apps-dev-nodegroup-apps-de-NodeInstanceRole-EJGK73RJR99G	system:node:{{EC2PrivateDNSName}}	system:bootstrappers,system:nodes
```

Callisto13:
Thanks @todd-dsm , I think we could also do with some better error messages in general when resources are not found. https://github.com/weaveworks/eksctl/issues/2980

Himangini:
We need to log what region was being used whenever we run this command [investigate and make sure we always log this for all eksctl commands]. 
We also need to document [if its not mentioned already] that the command will use settings from aws config file if you have any

PR #4791, opened 2022-02-15
log region for every command and improve help message

aclevername:
### Description
Closes https://github.com/weaveworks/eksctl/issues/2883

Moved the logging into `NewCtl` which every command that actually uses the region calls. Added a `Silent` field for the edge-cases where this func gets called multiple times, e.g. `get clusters --all-regions`

Biggest concern was ensuring that when we print json/yaml output via `-o json/yaml` that we don't break the stdout. Please check thet `ctl/get` diffs carefully :smile:


pkg/ctl/get/addon.go
```diff
@@ -47,19 +47,16 @@ func getAddonCmd(cmd *cmdutils.Cmd) {
 }
 
 func getAddon(cmd *cmdutils.Cmd, params *getCmdParams) error {
+	if params.output != printers.TableType {
+		//log warnings and errors to stdout
+		logger.Writer = os.Stderr
+	}
```
aclevername, line 53: this pattern seems to be used a lot in the `get` commands for when we want `json/yaml` output, basically ensures `get addons | jq .` still works, but the user still sees our helpful log messages on `stderr`

pkg/ctl/get/labels.go
```diff
@@ -51,7 +51,6 @@ func getLabels(cmd *cmdutils.Cmd, nodeGroupName string) error {
 	if err != nil {
 		return err
 	}
-	cmdutils.LogRegionAndVersionInfo(cmd.ClusterConfig.Metadata)
```
aclevername, line 54: get labels doesn't have json/yaml support

pkg/ctl/utils/describe_stacks.go
```diff
@@ -66,16 +66,16 @@ func doDescribeStacksCmd(cmd *cmdutils.Cmd, all, events, trail bool, printer pri
 		return err
 	}
 
-	cfg := cmd.ClusterConfig
+	if printer != nil {
+		cmdutils.Silent = true
+	}
```
aclevername, line 71: this command has some funky logic, so just did this. Seeing if I can change it

pkg/ctl/cmdutils/cmd.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/weaveworks/eksctl/pkg/eks"
 )
 
+var Silent = false
```
Skarlso, line 11: FWIW `var Silent bool` does result in false. :) But I understand that you wanted to be explicit. :)

Skarlso: I'm not liking this global variable. :/

aclevername: nor do I, but its the cleanest way to handle the edge-case scenarios where `NewCtl` is called multiple times. I could change the func/have an arg but am I really going to change the 30+ places where this is called for the 1 edge case? I think this is about picking the lesser evils tbh

Skarlso: The part that's happening more than once, can be used with a `sync.Once`. 

```go
    once.Do(func() {
        logRegionAndVersionInfo(c.ClusterConfig.Metadata)
    })
```

Or whatever it is you are guarding against here. This is global, and called only once, no matter how many times you call NewCtl.

pkg/ctl/cmdutils/cmd.go
```diff
@@ -1,13 +1,17 @@
...
+	"sync"
+
 	"github.com/kris-nova/logger"
 	"github.com/spf13/cobra"
 
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
 	"github.com/weaveworks/eksctl/pkg/eks"
 )
 
+var once = &sync.Once{}
```
Skarlso, line 13: hmmm, I believe this should be 
```go
var once sync.Once
```

And not a shared instance 🤔

Skarlso: Yep, because:
> A Once must not be copied after first use.

Skarlso, line 13:
```suggestion
var once sync.Once
```

Skarlso (approved):
One change than LGTM! :)

cPu1 (approved):

Issue #4616, opened 2022-01-12
Remove time-based (TTL) caching of cluster data


cPu1:
When eksctl code needs to read the latest EKS cluster data, it relies on a TTL-based cache that is refreshed when the TTL expires. In most places, this is not required as eksctl itself makes changes to the cluster, so it knows exactly when the cluster data might change. For many commands, the need for this cache can be eliminated by obtaining the cluster data via DescribeCluster in the beginning, when the provider is created. 

The caching code is [pkg/eks/cache.go](https://github.com/weaveworks/eksctl/blob/main/pkg/eks/cache.go)

Related to #4213.

Skarlso:
@cPu1 Interesting. When I was investigating how we could reduce the number of `DescribeStack` calls I was actually finding it nice that it doesn't always call DescribeStack. I found it convenient. Removing the cache means a full restructure of all the Can* calls and RefreshClusterStatus calls. :D

Skarlso:
I know, I know we discussed this already. Just writing it down here too.

PR #4798, opened 2022-02-15
Removed the cache as essentially it is not needed in most cases

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/4616

Notes:

Most of the `Can*` methods where called right after `NewProviderForExistingCluster` which called `RefreshClusterStatus` by force. which means that the cluster info is available by the time Can* is called. In rare cases, this didn't happen, or could error and be ignored the error. In these cases, we just check if the status is set or not. The TTL didn't make actual sense, since the status wouldn't change.

Running integration tests on this branch...

https://github.com/weaveworks/eksctl-ci/actions/runs/1847254636

Outcome: SUCCESS 🟢 🟢


pkg/ctl/delete/cluster.go
```diff
@@ -72,17 +70,13 @@ func doDeleteCluster(cmd *cmdutils.Cmd, force bool, disableNodegroupEviction boo
...
 	}
-	cmdutils.LogRegionAndVersionInfo(meta)
 
+	cmdutils.LogRegionAndVersionInfo(meta)
 	logger.Info("deleting EKS cluster %q", meta.Name)
 	if err := printer.LogObj(logger.Debug, "cfg.json = \\\n%s\n", cfg); err != nil {
 		return err
 	}
 
-	if ok, err := ctl.CanDelete(cfg); !ok {
```
Skarlso, line 82: Can delete notes:

- `CanDelete`
    - Already called NewProvider before it
    - Probably can remove the TTL
    - Interesting `CanDelete` was called only once, but even so, it just errored out if Refresh failed. But that would have happened before it anyways

cPu1 (commented):
One of the points to address in the issue was to ensure that if the local cluster state is being used after an update to the cluster's attributes, the updated attributes are reflected in the state. I'm curious if you found any such occurrences.

Skarlso:
@cPu1 the only one I found that is questionable is `NewOpenIDConnectManager` and `GetCurrentClusterConfigForLogging`. But they seem to be working fine. Other than that, I didn't find a scenario where there was an update and cluster info was needed through `RefreshClusterStatusIfStale`. `SupportsManagedNodes` and `SupportsFargate` happen early in the process.

cPu1 (approved):

Himangini:
The title of this PR doesn't look right, can you fix it ?

Himangini (approved):
LGTM 👍🏻

PR #4809, opened 2022-02-18
Add support for Bottlerocket nodgroups with NVIDIA accelerated hardware

### LINKED ISSUE #4760: Add support for Bottlerocket NVIDIA GPU variant in eksctl
Bottlerocket is releasing an AMI specifically supporting NVIDIA based GPU EC2 instance types in addition to regular AMIs that exist today. For EKS self-managed nodes using Bottlerocket, we expect AMI resolvers in eksctl to auto-detect and resolve to the correct Bottlerocket AMI based on the specified AMI family (Bottlerocket, in this case), instance type and region in the create command, or via the config file. 

For Bottlerocket NVIDIA GPU support, the device plugin is pre-baked into the Bottlerocket GPU enabled AMI and hence we do not expect eksctl to install additional plugins. We want install-nvidia-plugin to be defaulted to false and eksctl to return an error if users pass true. Users do not have to update the K8s device plugin separately since updating their nodes to a new Bottlerocket AMI should have corresponding software including the plugin updated. 

Sample command:
```shell
$ eksctl create cluster 
--name bottlerocket_nvidia_gpu_cluster 
--instance-type g4dn.xlarge
--region us-west-2 
--node-ami-family Bottlerocket 
--install-nvidia-plugin=false
```

SSM params for Bottlerocket: EKS self-managed nodes
X86: /aws/service/bottlerocket/aws-k8s-1.21-nvidia/x86_64/latest/image_id
ARM: /aws/service/bottlerocket/aws-k8s-1.21-nvidia/arm/latest/image_id

Several Bottlerocket users today create and manage EKS clusters with Bottlerocket as the underlying node OS via eksctl. With this feature, Bottlerocket users will have an interface to easily use the NVIDIA GPU based AMIs. For other specialized computing instance types, in the future, Bottlerocket might release new AMIs specific to a given accelerated compute vendor. We want EKS + Bottlerocket users to be able to use eksctl to spin up clusters with self-managed nodes using the new Bottlerocket NVIDIA GPU based AMI.

arnaldo2792:
### Description
Closes #4760

This pull request adds support for Bottlerocket nodegroups with NVIDIA accelerated hardware, the changes are as follows:

```
    Differentiate among accelerated hardware providers

    `IsGPUInstanceType` groups all together NVIDIA and inferentia instances
    which makes it difficult to reuse for OS'es that don't vend one image
    with support for multiple accelerated hardware.

    With this change, the instances with support for NVIDIA GPUs are
    separated in its own function (like inferentia instances), which is
    called by `IsGPUInstanceType`.
```

```
    Autoresolve SSM parameters for Bottlerocket GPU nodegroups

    This ensures that the correct SSM parameter is used to fetch the AMI id
    for Bottlerocket nodegroups with NVIDIA hardware.
```

```
    Skip NVIDIA device plugin daemonset in Bottlerocket node groups

    The Bottlerocket NVIDIA AMI already provides the NVIDIA device plugin,
    so Bottlerocket node groups don't require the NVIDIA device plugin
    daemonset.

    This change makes sure that clusters with only Bottlerocket nodegroups
    don't get the NVIDIA device plugin daemonset. If any of the nodegroups
    in the cluster's configuration uses an OS that requires the device
    plugin, it will be installed.
```

### Manual testing

- With the following configuration file, I verified that the correct AMI was auto resolved based on the `amiFamily` and the `instanceType`, and that the device plugin daemonset wasn't installed:

```yaml
---
kind: ClusterConfig

metadata:
  name: bottlerocket-1-21
  region: us-west-2
  version: '1.21'

nodeGroups:
  - name: al2-without-plugin-ng
    instanceType: m5.large
    desiredCapacity: 1
  - name: br-without-plugin
    instanceType: p3.2xlarge
    desiredCapacity: 1
    amiFamily: Bottlerocket
    iam:
       attachPolicyARNs:
    ssh:
        allow: true
        publicKeyName: <>
    bottlerocket:
      settings:
        motd: "Hello from eksctl!"
        host-containers:
          admin:
            enabled: true
```

```sh
~ ❯ eksctl get nodegroup --cluster bottlerocket-1-21 -o json | jq '.[].Name, .[].Type, .[].ImageID' -r
br-without-plugin
unmanaged
ami-06e0114f5eb33d00c
~ ❯ kubectl get daemonsets -A | grep nvidia
 # returns empty
```

- With the following configuration file, I verified that the correct AMI was auto resolved based on the `amiFamily` and the `instanceType`, and that the device plugin daemonset was installed since I included an AL2 nodegroup:

```yaml
---
kind: ClusterConfig

metadata:
  name: bottlerocket-1-21
  region: us-west-2
  version: '1.21'

nodeGroups:
  - name: br-with-plugin
    instanceType: p3.2xlarge
    desiredCapacity: 1
    amiFamily: Bottlerocket
    iam:
       attachPolicyARNs:
    ssh:
        allow: true
        publicKeyName: <>
    bottlerocket:
      settings:
        motd: "Hello from eksctl!"
        host-containers:
          admin:
            enabled: true
  - name: al2-with-plugin
    instanceType: p3.2xlarge
    desiredCapacity: 1
    iam:
       attachPolicyARNs:
    ssh:
        allow: true
        publicKeyName: <>
```

```sh
❯ eksctl get nodegroup --cluster bottlerocket-1-21 -o json | jq '.[].Name, .[].Type, .[].ImageID' -r
al2-with-plugin
br-with-plugin
br-without-plugin
unmanaged
unmanaged
unmanaged
ami-0a2fbc9dc946cfc58
ami-06e0114f5eb33d00c
ami-06e0114f5eb33d00c

~ ❯ kubectl get daemonsets -A | grep nvidia
kube-system   nvidia-device-plugin-daemonset   3         3         3       3            3           <none>          6m30s
```

### Checklist
- [X] Manually tested

### DIFF
--- a/pkg/ami/ssm_resolver.go
+++ b/pkg/ami/ssm_resolver.go
@@ -60,7 +60,7 @@ func MakeSSMParameterName(version, instanceType, imageFamily string) (string, er
 
 	switch imageFamily {
 	case api.NodeImageFamilyAmazonLinux2:
-		return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/%s/recommended/%s", version, imageType(imageFamily, instanceType), fieldName), nil
+		return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/%s/recommended/%s", version, imageType(imageFamily, instanceType, version), fieldName), nil
 	case api.NodeImageFamilyWindowsServer2019CoreContainer:
 		return fmt.Sprintf("/aws/service/ami-windows-latest/Windows_Server-2019-English-Core-EKS_Optimized-%s/%s", version, fieldName), nil
 	case api.NodeImageFamilyWindowsServer2019FullContainer:
@@ -78,7 +78,7 @@ func MakeSSMParameterName(version, instanceType, imageFamily string) (string, er
 		}
 		return fmt.Sprintf("/aws/service/ami-windows-latest/Windows_Server-20H2-English-Core-EKS_Optimized-%s/%s", version, fieldName), nil
 	case api.NodeImageFamilyBottlerocket:
-		return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/%s", version, instanceEC2ArchName(instanceType), fieldName), nil
+		return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/%s", imageType(imageFamily, instanceType, version), instanceEC2ArchName(instanceType), fieldName), nil
 	case api.NodeImageFamilyUbuntu2004, api.NodeImageFamilyUbuntu1804:
 		return "", &UnsupportedQueryError{msg: fmt.Sprintf("SSM Parameter lookups for %s AMIs is not supported yet", imageFamily)}
 	default:
@@ -108,14 +108,21 @@ func instanceEC2ArchName(instanceType string) string {
 	return "x86_64"
 }
 
-func imageType(imageFamily, instanceType string) string {
+func imageType(imageFamily, instanceType, version string) string {
 	family := utils.ToKebabCase(imageFamily)
-	if instanceutils.IsGPUInstanceType(instanceType) {
-		return family + "-gpu"
+	switch imageFamily {
+	case api.NodeImageFamilyBottlerocket:
+		if instanceutils.IsNvidiaInstanceType(instanceType) {
+			return fmt.Sprintf("%s-%s", version, "nvidia")
+		}
+		return version
+	default:
+		if instanceutils.IsGPUInstanceType(instanceType) {
+			return family + "-gpu"
+		}
+		if instanceutils.IsARMInstanceType(instanceType) {
+			return family + "-arm64"
+		}
+		return family
 	}
-
-	if instanceutils.IsARMInstanceType(instanceType) {
-		return family + "-arm64"
-	}
-	return family
 }
--- a/pkg/ami/ssm_resolver_test.go
+++ b/pkg/ami/ssm_resolver_test.go
@@ -299,8 +299,56 @@ var _ = Describe("AMI Auto Resolution", func() {
 					})
 
 				})
-			})
 
+				Context("and gpu instance", func() {
+					BeforeEach(func() {
+						instanceType = "p3.2xlarge"
+						version = "1.21"
+					})
+
+					Context("and ami is available", func() {
+						BeforeEach(func() {
+							p = mockprovider.NewMockProvider()
+							addMockGetParameter(p, "/aws/service/bottlerocket/aws-k8s-1.21-nvidia/x86_64/latest/image_id", expectedAmi)
+							resolver := NewSSMResolver(p.MockSSM())
+							resolvedAmi, err = resolver.Resolve(region, version, instanceType, imageFamily)
+						})
+
+						It("does not return an error", func() {
+							Expect(err).NotTo(HaveOccurred())
+						})
+						It("calls AWS SSM GetParameter", func() {
+							Expect(p.MockSSM().AssertNumberOfCalls(GinkgoT(), "GetParameter", 1)).To(BeTrue())
+						})
+						It("returns an ami id", func() {
+							Expect(resolvedAmi).To(BeEquivalentTo(expectedAmi))
+						})
+					})
+
+					Context("and ami is NOT available", func() {
+						BeforeEach(func() {
+							p = mockprovider.NewMockProvider()
+							addMockFailedGetParameter(p, "/aws/service/bottlerocket/aws-k8s-1.21-nvidia/x86_64/latest/image_id")
+
+							resolver := NewSSMResolver(p.MockSSM())
+							resolvedAmi, err = resolver.Resolve(region, version, instanceType, imageFamily)
+						})
+
+						It("errors", func() {
+							Expect(err).To(HaveOccurred())
+						})
+
+						It("does NOT return an ami id", func() {
+							Expect(resolvedAmi).To(BeEquivalentTo(""))
+						})
+
+						It("calls AWS SSM GetParameter", func() {
+							Expect(p.MockSSM().AssertNumberOfCalls(GinkgoT(), "GetParameter", 1)).To(BeTrue())
+						})
+
+					})
+				})
+			})
 		})
 	})
 })
--- a/pkg/eks/tasks.go
+++ b/pkg/eks/tasks.go
@@ -334,21 +334,26 @@ func (c *ClusterProvider) ClusterTasksForNodeGroups(cfg *api.ClusterConfig, inst
 		Parallel:  true,
 		IsSubTask: false,
 	}
-	var needsNvidiaButNotNeuron = func(t string) bool {
-		return instanceutils.IsGPUInstanceType(t) && !instanceutils.IsInferentiaInstanceType(t)
-	}
-	var haveNeuronInstanceType, haveNvidiaInstanceType, efaEnabled bool
+	var clusterRequiresNeuronDevicePlugin, clusterRequiresNvidiaDevicePlugin, efaEnabled bool
 	for _, ng := range cfg.NodeGroups {
-		haveNeuronInstanceType = haveNeuronInstanceType || api.HasInstanceType(ng, instanceutils.IsInferentiaInstanceType)
-		haveNvidiaInstanceType = haveNvidiaInstanceType || api.HasInstanceType(ng, needsNvidiaButNotNeuron)
+		clusterRequiresNeuronDevicePlugin = clusterRequiresNeuronDevicePlugin ||
+			api.HasInstanceType(ng, instanceutils.IsInferentiaInstanceType)
+		// Only AL2 requires the NVIDIA device plugin
+		clusterRequiresNvidiaDevicePlugin = clusterRequiresNvidiaDevicePlugin ||
+			(api.HasInstanceType(ng, instanceutils.IsNvidiaInstanceType) &&
+				ng.GetAMIFamily() == api.NodeImageFamilyAmazonLinux2)
 		efaEnabled = efaEnabled || api.IsEnabled(ng.EFAEnabled)
 	}
 	for _, ng := range cfg.ManagedNodeGroups {
-		haveNeuronInstanceType = haveNeuronInstanceType || api.HasInstanceTypeManaged(ng, instanceutils.IsInferentiaInstanceType)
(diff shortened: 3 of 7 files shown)


pkg/eks/tasks.go
```diff
@@ -431,3 +431,23 @@ func (c *ClusterProvider) appendCreateTasksForIAMServiceAccounts(cfg *api.Cluste
...
+		}
+		supported = ng.NodeGroupBase.GetAMIFamily() != api.NodeImageFamilyBottlerocket
+	}
+	for _, mng := range cfg.ManagedNodeGroups {
+		if supported {
+			break
+		}
+		supported = mng.NodeGroupBase.GetAMIFamily() != api.NodeImageFamilyBottlerocket
+	}
+	return supported
```
cPu1, line 452: Only AL2 GPU AMIs require the device plugin, so the condition can be relaxed. Additionally, `NodeGroupBase` is an embedded field, so `GetAMIFamily` (or `AMIFamily`) can be accessed directly. 

```suggestion
	for _, ng := range cfg.NodeGroups {
		if ng.GetAMIFamily() == api.NodeImageFamilyAmazonLinux2 {
			return true
		}
	}
	for _, mng := range cfg.ManagedNodeGroups {
		if mng.GetAMIFamily() == api.NodeImageFamilyAmazonLinux2 {
			return true
		}
	}
	return false
```

pkg/eks/tasks.go
```diff
@@ -431,3 +431,23 @@ func (c *ClusterProvider) appendCreateTasksForIAMServiceAccounts(cfg *api.Cluste
 		spec:            cfg,
 	})
 }
+
+// clusterSupportsDevicePlugin returns true if any of the node/managed node groups
+// uses an AMI family which requires the device plugin to be installed as
+// a daemonset
+func clusterSupportsDevicePlugin(cfg *api.ClusterConfig) bool {
```
cPu1, line 438: What do you think about naming this `clusterRequiresDevicePlugin`?

```suggestion
func clusterRequiresDevicePlugin(cfg *api.ClusterConfig) bool {
```

arnaldo2792:
Forced push addresses the feedback left

cPu1:
> Forced push addresses the feedback left

@arnaldo2792, any thoughts on [this comment](https://github.com/weaveworks/eksctl/pull/4809/files#r809728450)?

pkg/ami/ssm_resolver.go
```diff
@@ -108,14 +108,21 @@ func instanceEC2ArchName(instanceType string) string {
 	return "x86_64"
 }
 
-func imageType(imageFamily, instanceType string) string {
+func imageType(imageFamily, instanceType string, version string) string {
```
cPu1, line 111: Minor nit:
```suggestion
func imageType(imageFamily, instanceType, version string) string {
```

arnaldo2792:
> > Forced push addresses the feedback left
> 
> @arnaldo2792, any thoughts on [this comment](https://github.com/weaveworks/eksctl/pull/4809/files#r809728450)?

I'm sorry I thought I addressed it in my previous version. I fixed it along with your last nit comment :+1:

pkg/ami/ssm_resolver_test.go
```diff
@@ -299,8 +299,56 @@ var _ = Describe("AMI Auto Resolution", func() {
...
+					Context("and ami is NOT available", func() {
+						BeforeEach(func() {
+							p = mockprovider.NewMockProvider()
+							addMockFailedGetParameter(p, "/aws/service/bottlerocket/aws-k8s-1.21-nvidia/x86_64/latest/image_id")
+
+							resolver := NewSSMResolver(p.MockSSM())
+							resolvedAmi, err = resolver.Resolve(region, version, instanceType, imageFamily)
+						})
+
+						It("does not return an error", func() {
```
aclevername, line 337:
```suggestion
						It("errors", func() {
```

arnaldo2792:
Yes @aclevername, AMIs are public, you can pull down the commit and test it yourself :smile: , I'll fix the nit

arnaldo2792:
Forced push addresses the nit pointed out by @aclevername

arnaldo2792:
Forced push is a rebase of `main`

pkg/eks/tasks.go
```diff
@@ -356,7 +356,7 @@ func (c *ClusterProvider) ClusterTasksForNodeGroups(cfg *api.ClusterConfig, inst
 			logger.Info("\t see the following page for instructions: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-deploy/tutorials/tutorial-k8s.html#tutorial-k8s-env-setup-for-neuron")
 		}
 	}
-	if haveNvidiaInstanceType {
+	if haveNvidiaInstanceType && clusterRequiresDevicePlugin(cfg) {
```
aclevername, line 359: Doesn't this mean if my cluster is made up of:
- A regular non-GPU AL2 node
- GPU NVIDIA Bottlerocket node

that it will install the device plugin? Is this the desired behaviour? I just wanna check as I'm not 100% sure I understand when we do/don't need the plugin installed. I would of thought given a non-gpu AL2 only cluster doesn't need it, and a GPU only bottlerocket cluster also doesn't need it, that the combination of the two in the same cluster also wouldn't need it :thinking: 

I validated that it did get installed:
```
cat cluster.yaml
---
kind: ClusterConfig

metadata:
  name: br-gpu-and-al2-nor
  region: us-west-2
  version: '1.21'

nodeGroups:
  - name: br-nvidia
    instanceType: p3.2xlarge
    desiredCapacity: 1
    amiFamily: Bottlerocket
    bottlerocket:
      settings:
        motd: "Hello from eksctl!"
        host-containers:
          admin:
            enabled: true
  - name: al2-normal
    instanceType: m5.large
    desiredCapacity: 1
```

```
k get daemonsets.apps -A
NAMESPACE     NAME                             DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
kube-system   aws-node                         2         2         2       2            2           <none>          50m
kube-system   kube-proxy                       2         2         2       2            2           <none>          50m
kube-system   nvidia-device-plugin-daemonset   2         2         2       2            2           <none>          39m
```

aclevername (commented):
thanks for addressing comments @arnaldo2792 . I spotted one thing playing around this morning, please see the above comment :smile: 

we aren't doing a release until Friday so we have some time to work with :smile:

pkg/eks/tasks.go
arnaldo2792: @aclevername , correct, what you described is what should happen, according to @cPu1 the device plugin should only be installed in AL2 instances, I'll update the code to reflect this. I'd assume that the Neuron device plugin should only be installed in AL2 instances, I'll make the changes to reflect this as well, since I could do a little refactor to remove the function that I added now that I know the conditions required to install the device plugins.

aclevername: > @aclevername , correct, what you described is what should happen, according to @cPu1 the device plugin should only be installed in AL2 instances

I think @cPu1 meant only GPU AL2 Instances? From what I'm reading here https://github.com/weaveworks/eksctl/issues/4760#issuecomment-1040308851

arnaldo2792: Ok, I'll add a different condition somewhere (if it doesn't exist already), so that folks don't try to install Bottlerocket on inferentia hosts because we don't support it just yet. It is a little weird to me that only AL2 hosts require the NVIDIA GPU device plugin, but it is OK to install the Neuron device plugin on all other hosts, that's why I thought the Neuron device plugin would follow the same rule as the NVIDIA GPU device plugin.

aclevername: I might be wrong, I can't find clear documentation on which nodes actually need to have the plugin running, wether it's just the GPU nodes themselves or all nodes. If it's all nodes then the current code you have is correct. If it's only GPU nodes then we only want to install the nvidia plugin if the cluster contains a AL2 GPU node. 

Does that make senses? I'm sorry I don't have more information :/

arnaldo2792: Ideally GPU nodes should have the plugin running only if the AMI family supports them, because the device plugin relies on the driver provided by the AMI. Right, now I don't know for sure if AMIs other than AL2 or Bottlerocket support NVIDIA GPUs or Inferentia hardware. The [unit tests](https://github.com/weaveworks/eksctl/blob/57d0afb297742a579391fd3a8fa6ccafcaad3f10/pkg/apis/eksctl.io/v1alpha5/gpu_validation_test.go#L58) suggest that only AL2 supports both NVIDIA and Inferientia instances (and with my change, Bottlerocket will support NVIDIA GPUs).

I think for now my change should only cover installing the device plugin when there is a nodegroup with AL2 with NVIDIA instances, and then we could revisit what other combinations of AMI families/instance types require device plugins.

Himangini (approved):
Nicely done ⭐  LGTM 👍🏻

pkg/eks/tasks.go
```diff
@@ -334,29 +334,34 @@ func (c *ClusterProvider) ClusterTasksForNodeGroups(cfg *api.ClusterConfig, inst
...
 	}
 	for _, ng := range cfg.ManagedNodeGroups {
-		haveNeuronInstanceType = haveNeuronInstanceType || api.HasInstanceTypeManaged(ng, instanceutils.IsInferentiaInstanceType)
-		haveNvidiaInstanceType = haveNvidiaInstanceType || api.HasInstanceTypeManaged(ng, needsNvidiaButNotNeuron)
+		clusterRequiresNeuronDevicePlugin = clusterRequiresNeuronDevicePlugin ||
+			api.HasInstanceTypeManaged(ng, instanceutils.IsInferentiaInstanceType)
+		// Only AL2 requires the NVIDIA device plugin
+		clusterRequiresNvidiaDevicePlugin = clusterRequiresNvidiaDevicePlugin ||
+			api.HasInstanceTypeManaged(ng, instanceutils.IsNvidiaInstanceType) &&
+				ng.GetAMIFamily() == api.NodeImageFamilyAmazonLinux2
```
aclevername, line 353: my brain gets a bit fuzzy with this chaining, can we add some wrapping brackets?
```suggestion
		clusterRequiresNvidiaDevicePlugin = clusterRequiresNvidiaDevicePlugin || 
			(api.HasInstanceTypeManaged(ng, instanceutils.IsNvidiaInstanceType) &&
				ng.GetAMIFamily() == api.NodeImageFamilyAmazonLinux2)
```

arnaldo2792:
Forced push fixes the nit suggested by @aclevername

arnaldo2792:
Forced push is a rebase of main

aclevername:
correcting my earlier mistake, this will be out int `0.86.0-rc0`, not `85` :smile:

PR #4819, opened 2022-02-18
Add support for more Graviton instance types

otterley:
### Description

Add support for newest Graviton instance families, including im4gn, is4gen, g5g, and x2gd. Also removed some redundant code where longer prefixes were always also matched by shorter prefixes.


cPu1 (approved):

arnaldo2792:
FYI, `g5g` instances will cause inconsistent behavior with AL2, since [this function](https://github.com/weaveworks/eksctl/blob/c65b6911d479fa9517fa6f55d459b787d9c7834e/pkg/ami/ssm_resolver.go#L111) will return the `x86_64` ami version of the EKS Optimized ami.

PR #4825, opened 2022-02-22
remove deprecated field that breaks `upgrade nodegroup` timeout value

### LINKED ISSUE #4787: Default timeout isn't respected in `upgrade nodegroup`
[Failing integration test](https://github.com/weaveworks/eksctl-ci/runs/5209089919?check_suite_focus=true) 
Despite setting a timeout of 45 minutes for `upgrade nodegroup` [here](https://github.com/weaveworks/eksctl/blob/main/pkg/ctl/upgrade/nodegroup.go#L41), the command uses a timeout value of 25 minutes, causing `upgrade nodegroup` to fail when the operation takes longer to complete.

We need to make sure correct default flag is being passed 
`cmdutils.AddTimeoutFlagWithValue(fs, &cmd.ProviderConfig.WaitTimeout, upgradeNodegroupTimeout)`

Need to investigate where the change should go exactly

aclevername:
### Description
Closes #4787

before (print added for debugging):
```
2022-02-22 11:22:40 [ℹ]  eksctl version 0.86.0-dev+551fd9a4.2022-02-22T11:21:57Z
2022-02-22 11:22:40 [ℹ]  using region us-west-2
2022-02-22 11:22:41 [ℹ]  25m0s
```

after
```
2022-02-22 11:25:48 [ℹ]  eksctl version 0.86.0-dev+b7418e27.2022-02-22T11:25:02Z
2022-02-22 11:25:48 [ℹ]  using region us-west-2
2022-02-22 11:25:49 [ℹ]  45m0s
```


cPu1 (approved):
Well spotted! We should also check if there are any commands relying on this behaviour.

nikimanoledaki (approved):

aclevername:
> Well spotted! We should also check if there are any commands relying on this behaviour.

good point. I think this flag was causing ALL of the commands to have this default timeout value, but only some commands had the `AddTimeoutFlag` call done *after* the `AddCommonAWS` which is what made it work. I've changed it so that a new `Cmd` has this set by default, so those who configure the `timeout` flag will still override it, but it should preserve the previous behaviour of `25m` being default

Issue #4439, opened 2021-11-10
Improve UX for `create cluster` and `create nodegroup` when an AMI family value is not capitalised


cPu1:
When the `--node-ami-family` value is incorrect only in the casing, eksctl throws an error saying that the `<ami-family>` is not supported.

```shell
$ eksctl create cluster --node-ami-family=bottlerocket

Error: "bottlerocket" is not supported for managed nodegroups
```

The UX here can be improved by providing a `Did you mean "Bottlerocket"?` error that would aid the user in correcting their config or CLI option, and retry the command.

We do not want to make the `--node-ami-family` option or `nodeGroup.amiFamily` field case-insensitive as eksctl follows Kubernetes-style `UpperCamelCase` (or Pascal case) values for enums and making the value case-insensitive would be inconsistent. For example, kubectl does not allow a lower-case value for `--restart` in `kubectl run --image=<image> --restart=never <name>`. Making it case-insensitive would also allow calls of the form `eksctl create cluster --node-ami-family=bOtTleRoCket`.

### Summary
- After discussing we think it will be nice UX if we make this case insensitive so users don't need to use UpperCamelCase

cPu1:
> This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.

Not stale, this UX improvement can be helpful to new users.

PR #4826, opened 2022-02-22
allow case insensitivity for AMIFamily

aclevername:
### Description
closes https://github.com/weaveworks/eksctl/issues/4439

Cheapest option is to do a 1 time mutation of the `AMIFamily` value to the correct capitalisation 

Before
```
eksctl create nodegroup --name mg-1 --cluster jk --node-ami-family BoTtlerocket
Error: couldn't create cluster provider from options: "BoTtlerocket" is not supported for managed nodegroups
```
After
```
eksctl create nodegroup --name mg-1 --cluster jk --node-ami-family BoTtlerocket
2022-02-22 16:10:23 [ℹ]  eksctl version 0.86.0-dev+ec895808.2022-02-22T16:06:11Z
2022-02-22 16:10:23 [ℹ]  using region us-west-2
2022-02-22 16:10:24 [ℹ]  will use version 1.21 for new nodegroup(s) based on control plane version
...
```


nikimanoledaki (approved):
LGTM!

PR #4829, opened 2022-02-23
set userId prefix when creating clusters

### LINKED ISSUE #4462: Add support for setting a prefix when running integration tests locally
`make integration-test` creates a whole bunch of clusters, e.g. `eksctl-it-<something>`. The problem is on a shared account me and another team member might also be doing this at the same time. This results in confusion of who owns what cluster, since they all look similar. Lets enforce the setting of an env var when running `make integration-test` locally, to add an indentifier. E.g.:

`EKSCTL_INT_NAME=jk make integration-test` would result in all clusters being `eksctl-jk-it-<something>`. We can make `EKSCTL_INT_NAME` required

cc @Skarlso 

### Summary
- Nice thing to implement
- It will be cool if this automatically figures out who you are and adds it to the name [assumed-role?]
- Suggestion to use the `"UserId": "<redacted>:jake@weave.works",` as part of this implementation
- We should use a regex to extract the name from the userId and ignore the rest of the characters to keep the length minimum

aclevername:
### Description
Closes #4462

Example:
` deploying stack "eksctl-ja-it-addons-ferocious-gopher-1645617773-cluster"`

### DIFF
--- a/integration/tests/params.go
+++ b/integration/tests/params.go
@@ -227,7 +227,7 @@ func NewParams(clusterNamePrefix string) *Params {
 	return &params
 }
 
-// attemptSettingUserID will attempt to fetch the first 2 characters of the userID running the
+// attemptSettingUserID will attempt to fetch the first 4 characters of the userID running the
 // integration tests to be used as a prefix for cluster names. It fails silently if an error occurs
 func (p *Params) attemptSettingUserID() {
 	clusterProvider, err := eks.New(&api.ProviderConfig{Region: p.Region}, &api.ClusterConfig{
--- a/integration/tests/params.go
+++ b/integration/tests/params.go
@@ -261,7 +261,7 @@ func (p *Params) attemptSettingUserID() {
 	if len(parts) == 2 {
 		parts = strings.Split(parts[1], "@")
 		if len(parts) == 2 {
-			p.userID = parts[0][0:2]
+			p.userID = parts[0][:4]
 			return
 		}
 	}
--- a/integration/tests/params.go
+++ b/integration/tests/params.go
@@ -251,6 +251,10 @@ func (p *Params) attemptSettingUserID() {
 		return
 	}
 
+	if out.UserId == nil {
+		fmt.Fprintf(GinkgoWriter, msg, fmt.Errorf("UserId value not set in AWS response"))
+	}
+
 	//example value: ABCD1234:my-name@email-provider
 	parts := strings.Split(*out.UserId, ":")
 	if len(parts) == 2 {
--- a/integration/tests/params.go
+++ b/integration/tests/params.go
@@ -253,6 +253,7 @@ func (p *Params) attemptSettingUserID() {
 
 	if out.UserId == nil {
 		fmt.Fprintf(GinkgoWriter, msg, fmt.Errorf("UserId value not set in AWS response"))
+		return
 	}
 
 	//example value: ABCD1234:my-name@email-provider
--- a/integration/tests/params.go
+++ b/integration/tests/params.go
@@ -8,12 +8,15 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/sts"
 	. "github.com/onsi/ginkgo"
 
 	"github.com/weaveworks/eksctl/integration/runner"
 	api "github.com/weaveworks/eksctl/pkg/apis/eksctl.io/v1alpha5"
+	"github.com/weaveworks/eksctl/pkg/eks"
 	"github.com/weaveworks/eksctl/pkg/utils/names"
 )
 
@@ -23,6 +26,7 @@ const (
 
 // Params groups all test parameters.
 type Params struct {
+	userID     string
 	EksctlPath string
 	Region     string
 	Version    string
@@ -158,6 +162,9 @@ func (p *Params) NewClusterName(prefix string) string {
 
 func (p *Params) formatClusterName(prefix string, name string) string {
 	clusterName := fmt.Sprintf("it-%s-%s", prefix, name)
+	if p.userID != "" {
+		clusterName = p.userID + "-" + clusterName
+	}
 	p.addToDeleteList(clusterName)
 	return clusterName
 }
@@ -203,6 +210,8 @@ func NewParams(clusterNamePrefix string) *Params {
 
 	// go1.13+ testing flags regression fix: https://github.com/golang/go/issues/31859
 	flag.Parse()
+
+	params.attemptSettingUserID()
 	if params.ClusterName == "" {
 		params.ClusterName = params.NewClusterName(clusterNamePrefix)
 	} else {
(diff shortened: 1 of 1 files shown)


integration/tests/params.go
```diff
@@ -203,6 +210,8 @@ func NewParams(clusterNamePrefix string) *Params {
 
 	// go1.13+ testing flags regression fix: https://github.com/golang/go/issues/31859
 	flag.Parse()
+
+	params.attemptSettingUserID()
```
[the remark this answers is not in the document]
aclevername: its optional, it attempts to do it, but if it fails just continue as normal without

integration/tests/params.go
```diff
@@ -217,3 +226,39 @@ func NewParams(clusterNamePrefix string) *Params {
...
+	}
+
+	out, err := clusterProvider.Provider.STS().GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		fmt.Fprintf(GinkgoWriter, msg, err)
+		return
+	}
+
+	//example value: ABCD1234:my-name@email-provider
+	parts := strings.Split(*out.UserId, ":")
```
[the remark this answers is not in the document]
aclevername: it isn't required, see the func description :smile: 

Its worked just basing it off my local testing. We don't need to support all the different scenarios, at the end of the day this is for to make our lives easier. If the value is like this for me it should be the same for the rest of the team :smile: 

feel free to give it a run!

integration/tests/params.go
```diff
@@ -217,3 +226,43 @@ func NewParams(clusterNamePrefix string) *Params {
...
+	if out.UserId == nil {
+		fmt.Fprintf(GinkgoWriter, msg, fmt.Errorf("UserId value not set in AWS response"))
+	}
+
+	//example value: ABCD1234:my-name@email-provider
+	parts := strings.Split(*out.UserId, ":")
+	if len(parts) == 2 {
+		parts = strings.Split(parts[1], "@")
+		if len(parts) == 2 {
+			p.userID = parts[0][0:2]
```
Himangini, line 263: can we use the first 4 chars? it'll make it easier for us to identify who's cluster it is :)

integration/tests/params.go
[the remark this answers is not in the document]
nikimanoledaki: My concern was specifically about the integration test account/CI, and I don't have access to it so I can't try it out!

Also, from the go docs about `strings.Split()`: `If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.`

Does this mean that it could log the first 2 chars of the test account? 🤔

integration/tests/params.go
```diff
@@ -217,3 +226,43 @@ func NewParams(clusterNamePrefix string) *Params {
...
+	}
+
+	out, err := clusterProvider.Provider.STS().GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		fmt.Fprintf(GinkgoWriter, msg, err)
+		return
+	}
+
+	if out.UserId == nil {
+		fmt.Fprintf(GinkgoWriter, msg, fmt.Errorf("UserId value not set in AWS response"))
```
Skarlso, line 255: Are you missing a `return` here after this?

Skarlso: Yeah, I think you are missing a return after this :)

aclevername: ooops

integration/tests/params.go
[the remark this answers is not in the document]
aclevername: > Does this mean that it could log the first 2 chars of the test account? thinking Which I guess is not that bad anyway since the CI repo is private but hm

I think it contained some sensitive looking stuff, so I wasn't in favour of this but it is an option

> My concern was specifically about the integration test account/CI, and I don't have access to it so I can't try it out!

I'm less worried about this, either it works which is fine or it doesn't, which doesn't matter since this is mainly being added for us to differentiate our runs in our account

integration/tests/params.go
aclevername: sure, bit concerned that the cluster names are getting even longer but I doubt 2 more will do any harm :smile:

integration/tests/params.go
aclevername, line 255:
```suggestion
		fmt.Fprintf(GinkgoWriter, msg, fmt.Errorf("UserId value not set in AWS response"))
		return
```

integration/tests/params.go
aclevername, line 264:
```suggestion
			p.userID = parts[0][:4]
```

integration/tests/params.go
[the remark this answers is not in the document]
nikimanoledaki: I'm also not in favour of that, I'm saying that that is what this PR does: it prints the first 2 chars of the `userID`, and in the case where the integration tests are run by the integration test account from the CI repo, this might be the AWS Account ID. 

I could be wrong and it doesn't print those 2 chars, and also maybe we don't care about printing those 2 first chars in a private repo (but it's still sensitive info)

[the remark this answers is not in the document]
aclevername: ahh I get what your saying. I think that won't happen in this scenario, it only works if the string matches the format "abc:name@email", and we don't have any email associated that contains the account id (or any sensitive information).

I'll give it a run though to verify :smile:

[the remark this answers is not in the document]
aclevername: yep, since the integration account doesn't use an email its blank `"eksctl-it-addons-adorable-mushroom-1645710242-cluster"`

integration/tests/params.go
```diff
@@ -217,3 +226,44 @@ func NewParams(clusterNamePrefix string) *Params {
...
+		return
+	}
+
+	//example value: ABCD1234:my-name@email-provider
+	parts := strings.Split(*out.UserId, ":")
+	if len(parts) == 2 {
+		parts = strings.Split(parts[1], "@")
+		if len(parts) == 2 {
+			p.userID = parts[0][:4]
```
Skarlso, line 264: For the future, names that are below 4 characters are definitely there. Especially names like Ale, Joy, Hub, Xi, Lo, etc, or even usernames that aren't long enough such as cpu. This will then blow up with an index out of bounds exception. That's why I thought 2 characters is enough for this.

aclevername: true, if/when this happens we can reassess

Skarlso (approved):

integration/tests/params.go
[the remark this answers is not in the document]
nikimanoledaki: thank you for checking :D

integration/tests/params.go
```diff
@@ -217,3 +226,44 @@ func NewParams(clusterNamePrefix string) *Params {
 	params.GenerateCommands()
 	return &params
 }
+
+// attemptSettingUserID will attempt to fetch the first 2 characters of the userID running the
```
Himangini, line 230: small nit after the change 
```suggestion
// attemptSettingUserID will attempt to fetch the first 4 characters of the userID running the
```

Himangini (approved):
LGTM

nikimanoledaki (approved):
🥳

Issue #4779, opened 2022-02-11
ssh-ed25519 keys wrongly assumed as unsupported


tsunamaru:
**What were you trying to accomplish?**
Create cluster with ssh-ed25519 key for ssh access.

**What happened?**
`Error: computing fingerprint for key "~/.ssh/some-ed25519-key.pub": error computing fingerprint for SSH public key: unexpected type of SSH key ("ssh.ed25519PublicKey"); AWS can only import RSA keys`

**How to reproduce it?**
Just try to create any cluster with ed25519 key.

**Anything else we need to know?**
AWS can import ed25519 from EC2 Console:
![image](https://user-images.githubusercontent.com/37580200/153652144-1c2109cb-7a86-40fc-b20d-a724e91c4024.png)

**Versions**
```
$ eksctl info
eksctl version: 0.82.0
kubectl version: v1.21.9
OS: windows
```

#### Summary
- So we need to investigate whether this error is coming from in eksctl
- we need to fix eksctl side so that users can use the ssh key type ssh-ed25519

aclevername:
thanks for opening the issue @tsunamaru . Looks like this is related to the kops package we use to generate fingerprints. Working on a fix :smile:

aclevername:
Fix will be out in 0.85.0-rc.0 later today :smile:

cPu1:
> I can confirm for ssh keys of the type ed25519 this bug still persists:
> 
> ```
> 2022-06-29 13:23:10 [ℹ]  will use version 1.19 for new nodegroup(s) based on control plane version
> 2022-06-29 13:23:10 [ℹ]  nodegroup "<redacted>" will use "" [AmazonLinux2/1.19]
> Error: computing fingerprint for key "<redacted>": error computing fingerprint for SSH public key: unexpected type of SSH key ("ssh.ed25519PublicKey"); AWS can only import RSA keys
> ```

@4thel00z, are you using the latest version of eksctl? The fix was out in 0.86.

4thel00z:
@cPu1 thanks for the quick reply. Yes, I am using a newer version:

```
$ eksctl version
0.103.0
```

cPu1:
> @cPu1 thanks for the quick reply. Yes, I am using a newer version:
> 
> ```
> $ eksctl version
> 0.103.0
> ```

@4thel00z, I tried using an ed25519 SSH key with `create nodegroup` and it worked. Which command is failing for you? Can you share the steps you used to generate the key-pair?

[...]

PR #4835, opened 2022-02-24
add support for ed25519

aclevername:
### Description
Closes https://github.com/weaveworks/eksctl/issues/4779

working on a PR to the `kops` `pki` library we use under the hood, as its the more appropriate place for this IMO. This can be used as a temporary workaround

I wanted to make sure the fingerprint value we produce matches whats displayed in ec2 console. We do this for example with existing rsa keys:
image
```
[ℹ]  using SSH public key "/home/jake/.ssh/id_rsa.pub" as "eksctl-jk-nodegroup-mng-1-9f:f0:b4:09:16:9f:f0:ab:43:4a:79:14:04:71:5b:7b"
```

Now works with the new keys:

image

```
[ℹ]  using SSH public key "/home/jake/weave/eksctl/key.pub" as "eksctl-jk-nodegroup-mng-1-HvE7+gmH78VS53+iPuRDh/gKjVo26OzYU/qOnJWAgyk"

### CODE REVIEW

File: pkg/ssh/client/ssh.go [hunk 1]
```diff
@@ -32,13 +33,11 @@ func LoadKeyFromFile(filePath, clusterName, ngName string, ec2API ec2iface.EC2AP
 	}
 
 	key := string(fileContent)
-
-	fingerprint, err := pki.ComputeAWSKeyFingerprint(key)
+	fingerprint, err := fingerprint(filePath, fileContent)
```
Skarlso, line 36: Interesting that this wasn't picked up by golangci-lint. I guess we have that check disabled.

You are not checking the `err` but using the `fingerprint` result immediately. So first, check the `err`. :)

aclevername: oh yeah. Fixed

Skarlso (approved):

pkg/ssh/client/ssh.go
```diff
@@ -48,6 +48,24 @@ func LoadKeyFromFile(filePath, clusterName, ngName string, ec2API ec2iface.EC2AP
 	return keyName, nil
 }
 
+func fingerprint(filePath string, key []byte) (string, error) {
+	pk, _, _, _, err := ssh.ParseAuthorizedKey(key)
+	if err != nil {
+		return "", fmt.Errorf("parsing key %q: %w", filePath, err)
+	}
+
+	if strings.Contains(pk.Type(), "ed25519") {
```
Skarlso, line 57: Ah hang on. This could just be `pk.Type() == ssh.KeyAlgoED25519`.

aclevername: I'll make it more specific

Issue #4735, opened 2022-02-03
If --zones isn't passed --node-zones are ignored when selecting zones (and then it fails)


pmkc:
**What were you trying to accomplish?**

Create a cluster, pass `--node-zones=ZONE`, but not `--zones`.

This might seem a little contrived, but I only care that my nodes appear in a single AZ. I do not want to tell eksctl which zones to put the control plane and subnets in.

I am writing a programmatic client that does not know that any other AZ exists in the region. I could list zones and select one at random, but I would rather eksctl do that for me.

**What happened?**

eksctl ignored `--node-zones` and selected 3 random zones to create subnets in and then crashed when it didn't create one in the zone that I selected.

```
couldn't find public subnets: mapping doesn't have subnet with AZ us-west-2a: (allSubnets=v1alpha5.AZSubnetMapping{"us-west-2b":v1alpha5.A
ZSubnetSpec{ID:"subnet-030009cd42ded5f17", AZ:"us-west-2b", CIDR:(*ipnet.IPNet)(0xc0008ae030)}, "us-west-2c":v1alpha5.AZSubnetSpec{ID:"subnet-09c7a83017e03a5b9", 
AZ:"us-west-2c", CIDR:(*ipnet.IPNet)(0xc000c5df80)}, "us-west-2d":v1alpha5.AZSubnetSpec{ID:"subnet-0bffd6767a2a50d15", AZ:"us-west-2d", CIDR:(*ipnet.IPNet)(0xc000
```
**How to reproduce it?**

```
eksctl create cluster --managed=True --name=$NAME --node-type=m6i.xlarge --node-zones=us-west-2a --nodes=1 --region=us-west-2
```
This will fail 40% of the time.

**Logs**

```
2022-02-03 00:57:15 []  using region us-west-2
2022-02-03 00:57:15 []  setting availability zones to [us-west-2c us-west-2b us-west-2d]
2022-02-03 00:57:15 []  subnets for us-west-2c - public:192.168.0.0/19 private:192.168.96.0/19
2022-02-03 00:57:15 []  subnets for us-west-2b - public:192.168.32.0/19 private:192.168.128.0/19
2022-02-03 00:57:15 []  subnets for us-west-2d - public:192.168.64.0/19 private:192.168.160.0/19
2022-02-03 00:57:15 []  nodegroup "default" will use "" [AmazonLinux2/1.21]
2022-02-03 00:57:15 []  using Kubernetes version 1.21
[...]

PR #4837, opened 2022-02-24
Added checking zones before creating cluster

Skarlso:
Closes https://github.com/weaveworks/eksctl/issues/4735


Himangini:
Can you use the PR template and add some description, please?

pkg/ctl/create/cluster.go
```diff
@@ -155,6 +156,10 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 		}
 	}
 
+	if err := validateZonesAndNodeZones(cmd, params); err != nil {
+		return err
+	}
```
aclevername, line 161: could this not be with the rest of the validation code & unit tested?

Skarlso: I tried adding it there. But it's not called in time actually and not with the parameters I need.

Skarlso: More precisely this `params.AvailabilityZones` is not available there.

aclevername: ahh sorry yeah validation is the wrong place (even though it probably should be the right place :upside_down_face: ). What about here https://github.com/weaveworks/eksctl/blob/df75054cf0b2696130f465ac7a0af7e4bbe7b851/pkg/ctl/cmdutils/configfile.go#L281-L336

we have a lot of flag validation logic there

Skarlso: Yeah I tried that one too. But decided against it in the end. I don't remember why. It was a week ago :D

Skarlso: But yeah, I put it here first. :)

aclevername: > Yeah I tried that one too. But decided against it in the end. I don't remember why. It was a week ago :D

:thunk: this definitely feels like a suitable place though? It has validation logic for when a user is using command line flags, e.g. https://github.com/weaveworks/eksctl/blob/main/pkg/ctl/cmdutils/configfile.go#L294 

> FYI, what you are showing is nodegroups. This is during cluster creation.

isn't what I linked for clusters? `func NewCreateClusterLoader(cmd *Cmd, ngFilter *filter.NodeGroupFilter, ng *api.NodeGroup, params *CreateClusterCmdParams) ClusterConfigLoader {`

pkg/ctl/cmdutils/configfile.go
```diff
@@ -338,6 +343,23 @@ func NewCreateClusterLoader(cmd *Cmd, ngFilter *filter.NodeGroupFilter, ng *api.
 	return l
 }
 
+func validateZonesAndNodeZones(cmd *cobra.Command) error {
+	if nodeZonesFlag := cmd.Flag("node-zones"); nodeZonesFlag != nil && nodeZonesFlag.Changed {
+		zonesFlag := cmd.Flag("zones")
+		if zonesFlag == nil || !zonesFlag.Changed {
```
nikimanoledaki, line 349:
```suggestion
		if zonesFlag := cmd.Flag("zones"); zonesFlag == nil || !zonesFlag.Changed {
```
to mirror the line above 😌

nikimanoledaki (approved):
LGTM! 👍 (one tiny stylistic nit)

pkg/ctl/cmdutils/configfile.go
Skarlso: fine. :D :D

Skarlso: Oh, but that actually makes this error

```go
			return errors.New("zones must be defined if node-zones is provided and must be a superset of node-zones")
```

Having to put on an `else` which is ugly. :D Sorry :))

nikimanoledaki: I don't see why it would need an `else`, wouldn't it just continue if it doesn't meet the conditions or return the error if it does (what we want)? what am I missing here? 😳

pkg/ctl/cmdutils/configfile_test.go
```diff
@@ -201,6 +202,66 @@ var _ = Describe("cmdutils configfile", func() {
...
+			)
+			BeforeEach(func() {
+				flagSet = pflag.NewFlagSet("test", pflag.ContinueOnError)
+				flagSet.StringSlice("zones", []string{}, "Zones")
+				flagSet.StringSlice("node-zones", []string{}, "Node zones")
+				testCmd = newCmd()
+				testCmd.Flags().AddFlagSet(flagSet)
+			})
+			When("node-zones is a subset of zones", func() {
+				It("does not error", func() {
```
aclevername, line 218:
```suggestion
				It("succeeds", func() {
```

aclevername (approved):

pkg/ctl/cmdutils/configfile.go
Skarlso: How do you mean? If I don't put an else in there, I would have to return nil if there is no error after the inside validation finished. Something like this:

```go
func validateZonesAndNodeZones(cmd *cobra.Command) error {
	if nodeZonesFlag := cmd.Flag("node-zones"); nodeZonesFlag != nil && nodeZonesFlag.Changed {
		if zonesFlag := cmd.Flag("zones"); zonesFlag != nil && zonesFlag.Changed {
			nodeZones := strings.Split(strings.Trim(nodeZonesFlag.Value.String(), "[]"), ",")
			zones := strings.Split(strings.Trim(zonesFlag.Value.String(), "[]"), ",")
			for _, zone := range nodeZones {
				if !utilstrings.Contains(zones, zone) {
					return fmt.Errorf("node-zones %s must be a subset of zones %s; %q was not found in zones", nodeZones, zones, zone)
				}
			}
			return nil
		}
		return errors.New("zones must be defined if node-zones is provided and must be a superset of node-zones")
	}
	return nil
}
```

Otherwise it would always just fall though to the error even if the inside validation finished successfully. Or, I put it on an else like this: 

```go

func validateZonesAndNodeZones(cmd *cobra.Command) error {
	if nodeZonesFlag := cmd.Flag("node-zones"); nodeZonesFlag != nil && nodeZonesFlag.Changed {
		if zonesFlag := cmd.Flag("zones"); zonesFlag != nil && zonesFlag.Changed {
			nodeZones := strings.Split(strings.Trim(nodeZonesFlag.Value.String(), "[]"), ",")
			zones := strings.Split(strings.Trim(zonesFlag.Value.String(), "[]"), ",")
			for _, zone := range nodeZones {
				if !utilstrings.Contains(zones, zone) {
					return fmt.Errorf("node-zones %s must be a subset of zones %s; %q was not found in zones", nodeZones, zones, zone)
				}
			}
		} else if zonesFlag == nil || !zonesFlag.Changed {
			return errors.New("zones must be defined if node-zones is provided and must be a superset of node-zones")
		}
	}
	return nil
}
```

Which is, of course, way more complex to read. :)

So while the first one is fine, I don't really like the double return nil. :D But if you insist I can do that. :D

Skarlso: The reason I like this approach more is that it requires less parsing in the brain. You immediately see the error cases and when it returns and why. :)

PR #4860, opened 2022-02-28
Don't attach the control plane security group to managed nodegroups

tompiscitell:
### Description

When creating Managed Nodegroups from an imported cluster, the control plane security group was getting attached to the worker nodes. This security group is supposed to be attached to the control plane ENIs according to [the EKS docs](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html#control-plane-worker-node-sgs). With this PR, only the cluster security group will get attached to Managed Nodegroups. This is consistent with the behavior of adding Managed Nodegroups to clusters created by eksctl.

### Checklist
- [x] Manually tested


Himangini:
@tompiscitell thanks for the nicely written PR. Core team will review it soon 👍🏻

Skarlso:
Hi @tompiscitell.

That document says earlier than 1.14. What makes you think this is not required for versions above 1.14?

pkg/vpc/importer.go
```diff
@@ -119,19 +119,18 @@ func (si *SpecConfigImporter) SharedNodeSecurityGroup() *gfnt.Value {
 	return si.ClusterSecurityGroup()
 }
 
-// SecurityGroups returns a gftn slice of the ControlPlaneSecurityGroup
-// and the ClusterSecurityGroup
+// SecurityGroups returns a gfnt slice of the ClusterSecurityGroup
 func (si *SpecConfigImporter) SecurityGroups() gfnt.Slice {
-	return gfnt.Slice{si.ControlPlaneSecurityGroup(), si.ClusterSecurityGroup()}
+	return gfnt.Slice{si.ClusterSecurityGroup()}
```
Skarlso, line 124: This change only affects unowned clusters. For Owned clusters we are already doing this. I except this is actually not doing anything, but I have to further check the code.

Skarlso:
@tompiscitell Ok, now that I understand the code a bit better, let me summarise. This change is only applicable to unowned clusters. Meaning, it's a cluster that `eksctl` has not created. For managed clusters we already do the right thing. I'm okay with this change, but please paste in manual tests and steps and outcome so we can see that you tried this code change and an unowned cluster still works in creating managed nodegroups and communication is not disrupted. 

So what you need:
- create an EKS cluster but not by using `eksctl`
- import that cluster
- create a managed nodegroup
- show results

Thanks! :)

Skarlso (changes requested):
Request Manual Test steps then LGTM.

Skarlso (approved):

pkg/vpc/importer.go
aclevername: > This change only affects unowned clusters. For Owned clusters we are already doing this. I except this is actually not doing anything, but I have to further check the code.

why would this change only affect unowned? I'm slightly confused :/

aclevername: ahh makes sense, thank you!

aclevername (approved):

Issue #4858, opened 2022-02-28
releases are tagged sometimes with v prefix and sometimes without


matti:
for example `v0.85.0` vs `0.86.0`

This kind of releasing breaks tools like `asdf` - this old issue: https://github.com/weaveworks/eksctl/issues/4199 is now broken again

Summary
- Have a look at [Chetan's pr](https://github.com/weaveworks/eksctl/pull/4765) 
- we ensure that every official github release uses the v prefixed tag.

Himangini:
@matti Thanks for opening the issue. Can you add more details on how the `asdf` is broken again? As far as I can see from the related issue that they are ignoring `v` from the release tag if exists https://github.com/elementalvoid/asdf-eksctl/pull/1.

Himangini:
After going through the releases page I can see `v0.83` was tagged with a `v` after that all the releases are without `v` 
I am guessing this might be the result of migrating the jobs. Either way, we should have a consistent way of tagging releases.

matti:
I just noticed that suddenly my:
```
asdf plugin add eksctl
asdf install ...
``
was broken and this was the reason. I patched my own version of asdf plugin that tries to download first without `v` prefix and then with `v` prefix - so I'm personally good now, others are not.

Skarlso:
> so I'm personally good now.

Great, we can close this ticket then.

> others are not.

If they exist, I'm sure they'll open their own tickets and then we can deal with those issues. :)

Cheers.

matti:
:D

Skarlso:
@pre we agree. Hence Gini's comment: 
>Either way, we should have a consistent way of tagging releases.

We recently switched from circleci to pure Github actions. I'm guessing something went awry during the process.

aclevername:
Just to summarise whats happened:

- Since `0.37.0` we started tagging every release with a and without the `v` prefix. Example:
https://github.com/weaveworks/eksctl/tree/0.80.0
https://github.com/weaveworks/eksctl/tree/v0.80.0
- We pick 1 of those to be the official github release, in the case of the above example it was the one with the `v` prefix

> Related issues:
> 
> * [New Versioning with v prefix causes problems with installing the latest version #4199](https://github.com/weaveworks/eksctl/issues/4199)
> * [inconsistent tags #4328](https://github.com/weaveworks/eksctl/issues/4328)
[...]

PR #4873, opened 2022-03-02
publish releases on pushes to the v prefixed tag

aclevername:
### Description
Closes https://github.com/weaveworks/eksctl/issues/4858

This fix is broken into two parts:
## part 1- changing the tag the workflows run on
When this workflow runs, it triggers the `do-release.sh` script https://github.com/weaveworks/eksctl/blob/main/build/scripts/do-release.sh#L8

This fetches the tag name, which is equal to the name of the branch/tag the workflow is running on. Example https://github.com/weaveworks/eksctl/runs/5339216464?check_suite_focus=true#step:6:7. This should get changed tot he `v` prefix'd tag now.

## part 2- specifying what the tag should be via the environment variable
its super unclear how `goreleaser` decides what tag to use for the release. You can however use an environment variable to configure this `GORELEASER_CURRENT_TAG` https://goreleaser.com/customization/build/?h=goreleaser_current_tag#define-build-tag.

Combined with the changes in part 1 we can set this environment variable to ensure the version is correct

### DIFF
--- a/.github/workflows/publish-release-candidate.yaml
+++ b/.github/workflows/publish-release-candidate.yaml
@@ -2,7 +2,7 @@ name: publish-release-candidate
 on:
   push:
     tags:
-      - '[0-9]+.[0-9]+.[0-9]+-rc.[0-9]'
+      - 'v[0-9]+.[0-9]+.[0-9]+-rc.[0-9]'
 
 jobs:
   test-and-build:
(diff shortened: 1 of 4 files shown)


aclevername:
just realised this will break how we read the release notes file. Now I am even more confused about how this worked before

Skarlso (approved):

nikimanoledaki (approved):

Issue #4862, opened 2022-02-28
Refactor code to use ValidateClusterConfig


Himangini:
### Description
Follow up from https://github.com/weaveworks/eksctl/issues/4627, https://github.com/weaveworks/eksctl/issues/4627#issuecomment-1044849491
Code areas that should be refactored to use `ValidateClusterConfig`

### 1. `doCreateCluster` in `pkg/ctl/create/cluster.go`
- [ ] [L143-L222](https://github.com/weaveworks/eksctl/blob/5c6580b47340ff41e8f8c3ac42e6c84a1cfdefcb/pkg/ctl/create/cluster.go#L143-L222)

We need to backfill tests here

### 2 `NewCTL` in `pkg/ctl/cmdutils/cmd.go`
- [ ] `NewCTL` [pkg/ctl/cmdutils/cmd.go](https://github.com/weaveworks/eksctl/blob/4388c8d14bde1e0f3cf663b2360d8744bd2e4184/pkg/ctl/cmdutils/cmd.go#L36)

We need to add tests or backfill tests here

### 3. `commonClusterConfigLoader` struct

[commonClusterConfigLoader](https://github.com/weaveworks/eksctl/blob/main/pkg/ctl/cmdutils/configfile.go) is a struct that is quite pervasive in `pkg/ctl`. It might require a larger refactoring solution and effort that is solely focused on that.

Used for validation in multiple `ctl/` packages by a large number of commands, such as:
- [ ] pkg/ctl/cmdutils/kms.go - `utils enable-secrets-encryption`
- [ ] pkg/ctl/cmdutils/scale.go - `scale nodegroup`
- [ ] pkg/ctl/cmdutils/fargate.go - `create fargate`, `delete fargate`
- [ ] pkg/ctl/cmdutils/addon.go - `create addon`, `upgrade addon`, `delete addon`
- [ ] pkg/ctl/cmdutils/gitops.go - `enable flux`
- [ ] pkg/ctl/disassociate/identity_provider.go - `get identityprovider`
- [ ] pkg/ctl/get/identityprovider.go - `disassociate identityprovider`
- [ ] pkg/ctl/associate/identity_provider.go - `associate identityprovider`

### 4. `ConfigLoaderBuilder` struct 

[ConfigLoaderBuilder](https://github.com/weaveworks/eksctl/blob/main/pkg/ctl/cmdutils/builder.go) is used to perform validations by the following commands in `ctl/`:
[...]

PR #4896, opened 2022-03-04
Validate AZ configuration correctly 

### LINKED ISSUE #4891: `availabilityZones` parsing broken after upgrading to eksctl 0.86.0
Create a new cluster with `eksctl 0.86.0`

**What happened?**
Defined `availabilityZones` in ClusterConfig but it randomly generated subnets but not honor the YAML setting.

**How to reproduce it?**
Create a minimal ClusterConfig (see blow) then execute the command as `eksctl create cluster -f ./experimental.yaml --dry-run` and you should see bouncing `availabilityZones` being generated. Tried to search root cause from commit history, seems like it's related to #4735 and #4837.

Additional note, `eksctl create cluster --nodes=us-east-1a,us-east-1b ...` works as expected with no issue.

image

**Logs**

Testing input:
```
$ cat ./experimental.yaml
kind: ClusterConfig

metadata:
  name: experimental
  region: us-east-1

availabilityZones:
- us-east-1a
- us-east-1b
```

Tool version:
```
$ eksctl version
0.86.0
```

Unexpected results:
```
 $ for i in $(seq 1 3); do (echo "[[ Round ${i} ]]";eksctl create cluster -f ./experimental.yaml --dry-run | grep 'availabilityZones' -A 2; echo); done
[[ Round 1 ]]
availabilityZones:
- us-east-1b
- us-east-1f

[[ Round 2 ]]
availabilityZones:
- us-east-1b
- us-east-1d

[[ Round 3 ]]
availabilityZones:
- us-east-1c
- us-east-1e
```

It's was worked as expected before `0.85.0`

Tool version:
```
$ ./eksctl-v0.85.0 version
0.85.0
```

Results:
```
$ for i in $(seq 1 3); do (echo "[[ Round ${i} ]]";./eksctl-v0.85.0 create cluster -f ./experimental.yaml --dry-run | grep 'availabilityZones' -A 2; echo); done
[[ Round 1 ]]
availabilityZones:
- us-east-1a
- us-east-1b

[[ Round 2 ]]
availabilityZones:
- us-east-1a
- us-east-1b

[[ Round 3 ]]
availabilityZones:
- us-east-1a
- us-east-1b
```

**Versions**

```
$ eksctl info
eksctl version: 0.86.0
kubectl version: v1.23.4
OS: darwin
```

nikimanoledaki:
### Description

Closes #4891

During some refactoring, the AZ validation that returned early if AZs were set was accidentally removed, so the AZ configuration was not respected. Reinstating this to fix that bug :) + adding unit-tests so that hopefully it doesn't happen again 😬 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -377,6 +377,13 @@ func (c *ClusterProvider) SetAvailabilityZones(spec *api.ClusterConfig, given []
 		return nil
 	}
 
+	if count := len(spec.AvailabilityZones); count != 0 {
+		if count < api.MinRequiredAvailabilityZones {
+			return api.ErrTooFewAvailabilityZones(spec.AvailabilityZones)
+		}
+		return nil
+	}
+
 	logger.Debug("determining availability zones")
 	zones, err := az.GetAvailabilityZones(c.Provider.EC2(), c.Provider.Region())
 	if err != nil {
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -368,7 +368,7 @@ func ResolveAMI(provider api.ClusterProvider, version string, np api.NodePool) e
 }
 
 // SetAvailabilityZones sets the given (or chooses) the availability zones
-func (c *ClusterProvider) SetAvailabilityZones(spec *api.ClusterConfig, given []string) error {
+func SetAvailabilityZones(spec *api.ClusterConfig, given []string, ec2 ec2iface.EC2API, region string) error {
 	if count := len(given); count != 0 {
 		if count < api.MinRequiredAvailabilityZones {
 			return api.ErrTooFewAvailabilityZones(given)
@@ -385,7 +385,7 @@ func (c *ClusterProvider) SetAvailabilityZones(spec *api.ClusterConfig, given []
 	}
 
 	logger.Debug("determining availability zones")
-	zones, err := az.GetAvailabilityZones(c.Provider.EC2(), c.Provider.Region())
+	zones, err := az.GetAvailabilityZones(ec2, region)
 	if err != nil {
 		return errors.Wrap(err, "getting availability zones")
 	}
--- a/pkg/eks/api_test.go
+++ b/pkg/eks/api_test.go
@@ -179,7 +179,7 @@ var _ = Describe("Setting Availability Zones", func() {
 		When("the given params contain too few AZs", func() {
 			It("returns an error", func() {
 				err := eks.SetAvailabilityZones(cfg, []string{"us-east-2a"}, provider.EC2(), "")
-				Expect(err).To(MatchError(ContainSubstring("only 1 zone(s) specified [us-east-2a], 2 are required (can be non-unique)")))
+				Expect(err).To(MatchError("only 1 zone(s) specified [us-east-2a], 2 are required (can be non-unique)"))
 			})
 		})
 	})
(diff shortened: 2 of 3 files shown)


Skarlso (approved):

cPu1 (approved):

pkg/eks/api.go
```diff
@@ -377,6 +377,13 @@ func (c *ClusterProvider) SetAvailabilityZones(spec *api.ClusterConfig, given []
 		return nil
 	}
 
+	if count := len(spec.AvailabilityZones); count != 0 {
+		if count < api.MinRequiredAvailabilityZones {
+			return api.ErrTooFewAvailabilityZones(spec.AvailabilityZones)
+		}
+		return nil
+	}
+
```
aclevername, line 386: don't we already do this? https://github.com/weaveworks/eksctl/blob/df75054cf0b2696130f465ac7a0af7e4bbe7b851/pkg/apis/eksctl.io/v1alpha5/validation.go#L1296

Skarlso: For given azs, but those don't come from the spec apparently.

Skarlso: They come from `params`:

```go
		if err := ctl.SetAvailabilityZones(cfg, params.AvailabilityZones); err != nil {
			return err
		}
```

Skarlso: Which is set by cli. But the extra check is for when it's coming from the yaml config. Pretty weird, could probably be refactored at some point.

aclevername: > For given azs, but those don't come from the spec apparently.

I don't follow, both codes use `spec/cfg.AvailabilityZones`

nikimanoledaki: Very good point but no it doesn't: this specific function is (only) used by `doCreateCluster`, which does not use `ValidateClusterConfig`. Moved this validation to `ValidateClusterConfig`, but the next step is to make all these `ctl` funcs actually use it. https://github.com/weaveworks/eksctl/issues/4862 addresses this :)

Skarlso: See previous comment. :)

nikimanoledaki:
Added unit tests (which required a slight refactor)

Skarlso (approved):
Very nice changes, well done Niki! :)

pkg/eks/api_test.go
```diff
@@ -153,3 +156,97 @@ func mockDescribeImages(p *mockprovider.MockProvider, amiID string, matcher func
...
+			It("sets them as the AZs to be used", func() {
+				err := eks.SetAvailabilityZones(cfg, []string{"us-east-2a", "us-east-2b"}, provider.EC2(), "")
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+
+		When("the given params contain too few AZs", func() {
+			It("returns an error", func() {
+				err := eks.SetAvailabilityZones(cfg, []string{"us-east-2a"}, provider.EC2(), "")
+				Expect(err).To(MatchError(ContainSubstring("only 1 zone(s) specified [us-east-2a], 2 are required (can be non-unique)")))
```
aclevername, line 182: huge nit, here and a few other places isn't this an exact error match? So we could remove the contain substring part
```suggestion
				Expect(err).To(MatchError("only 1 zone(s) specified [us-east-2a], 2 are required (can be non-unique)"))
```

could totally be wrong but first glace looks like all the errors are internal, so its an exact error match

aclevername (approved):
1 nit

pkg/eks/api_test.go
nikimanoledaki: testing errors and jake - name a more dynamic duo, I'll wait

nikimanoledaki: done!

PR #4897, opened 2022-03-04
set tag env var in rc release script

aclevername:
it was set in the regular release script and not the RC, which is what caught me out in https://github.com/weaveworks/eksctl/pull/4873


cPu1 (approved):
LGTM.

nikimanoledaki (approved):

PR #4905, opened 2022-03-04
Added clarity for overrideBootstrapCommand

rothgar:
### Description

Added a note that you have to specify an AMI. I ran into an error here because I copy pasted the example which didn't work and wasn't clear what I was missing.


cPu1 (approved):
Good catch. We do throw a validation error instructing the user to supply a custom AMI, but this example ClusterConfig was broken.

aclevername (approved):

PR #4911, opened 2022-03-07
Fix setting ownership status when listing clusters

### LINKED ISSUE #4888: 0.83->0.85: eksctl get clusters shows different status for "EKSCTL CREATED"

I wanted to list the current set of clusters, so I ran 
```$ eksctl get clusters```

**What happened?**

Using 0.83, the list of clusters was given, and the correct status of "True" was given in the "EKSCTL CREATED" column

I then updated eksctl to 0.85, and repeated the command. Now the incorrect status of "False" was given for the same clusters.

**How to reproduce it?**

Create a cluster using an old version of eksctl. Verify that "eksctl get clusters" shows that "EKSCTL CREATED" status as True using eksctl 0.83.

Now try "eksctl get clusters" with eksctl 0.85 and see that status turn to False

**Logs**

The verbose command adds no new information.

I upgraded eksctl using brew on OSX.

**Versions**

```
$ eksctl info
```

```
% eksctl info
eksctl version: 0.85.0
kubectl version: v1.22.4
OS: darwin
```

### PR DESCRIPTION
**Pull request by cPu1**
### Description

Instead of using the `clusterName` argument, the method was obtaining the cluster name via the `spec` field.

Fixes #4888 

This changelist correctly sets the `EKSCTL CREATED` column to `True` for clusters created via eksctl:

```
NAME					        REGION		EKSCTL CREATED
test-cluster                              	us-west-2	True
test-cluster2                             	us-west-2	True


```

### Checklist
- [x] Manually tested

### DIFF
--- a/integration/matchers/cluster.go
+++ b/integration/matchers/cluster.go
@@ -1,0 +1,32 @@
+package matchers
+
+import (
+	"strings"
+
+	"github.com/weaveworks/eksctl/integration/runner"
+
+	. "github.com/onsi/gomega"
+)
+
+type GetClusterOutput struct {
+	ClusterName   string
+	Region        string
+	EksctlCreated string
+}
+
+// AssertContainsCluster asserts that the output of the specified command contains the specified cluster
+func AssertContainsCluster(cmd runner.Cmd, out GetClusterOutput) {
+	Expect(cmd).To(runner.RunSuccessfullyWithOutputStringLines(
+		ContainElement(WithTransform(func(line string) GetClusterOutput {
+			fields := strings.Fields(line)
+			const expectedColumns = 3
+			Expect(fields).To(HaveLen(expectedColumns), "Expected `get clusters` to return %d columns", expectedColumns)
+			return GetClusterOutput{
+				ClusterName:   fields[0],
+				Region:        fields[1],
+				EksctlCreated: fields[2],
+			}
+		}, Equal(out))),
+	))
+
+}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by nikimanoledaki** [comment]
Good job getting a fix quickly! :) Could you add unit tests please?

**Review by aclevername — CHANGES_REQUESTED** [review]
can we get some unit tests if possible, and update the unowned cluster integration test?

File: integration/matchers/cluster.go [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+type GetClusterOutput struct {
+	ClusterName   string
+	Region        string
+	EksctlCreated string
+}
+
+// AssertContainsCluster asserts that the output of the specified command contains the specified cluster
+func AssertContainsCluster(cmd runner.Cmd, out GetClusterOutput) {
+	Expect(cmd).To(runner.RunSuccessfullyWithOutputStringLines(
```
cPu1, line 19: Intentionally not using a dot import because runner.Cmd is also being used here.

aclevername (approved):

pkg/cfn/manager/api.go
```diff
@@ -474,36 +477,35 @@ func defaultStackStatusFilter() []*string {
...
-		}
+	input := &cloudformation.DeleteStackInput{
+		StackName: s.StackId,
 	}
 
-	return nil, fmt.Errorf("cannot delete stack %q as it doesn't bear our %q, %q tags", *s.StackName,
-		fmt.Sprintf("%s:%s", api.OldClusterNameTag, c.spec.Metadata.Name),
-		fmt.Sprintf("%s:%s", api.ClusterNameTag, c.spec.Metadata.Name))
-}
+	if cfnRole := c.roleARN; cfnRole != "" {
```
Skarlso, line 490: I feel like creating the `cfnRole` variable is a bit of an overkill here. It's not like it's hard to read / type `c.roleARN` or looks too long or whatever. :)

Skarlso (approved):

pkg/cfn/manager/api.go
cPu1: This line wasn't introduced in this PR, I moved the existing code.

Skarlso: ah

Issue #4925, opened 2022-03-10
Generate interfaces for AWS SDK v2 services from structs


cPu1:
eksctl heavily relies on the AWS interface types for mocking but aws-sdk-go-v2 does not provide interfaces for its API services. As part of the code generation process, interfaces should be generated from aws-sdk-go-v2 structs. [ifacemaker](https://github.com/vburenin/ifacemaker) can help with this but it suffers from [this issue](https://github.com/vburenin/ifacemaker/issues/12), therefore the source generated by `ifacemaker` should be passed through another tool that adds the missing import. 

Part of #4877.

Issue #4914, opened 2022-03-08
Migrate LoggingRetryer to be compatible with aws-sdk-go-v2


cPu1:
The interface and default implementation for the [Retryer](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/retry) has changed, which requires migrating the `LoggingRetryer` to be compatible with [aws.Retryer](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#Retryer).

Part of https://github.com/weaveworks/eksctl/issues/4877

cPu1:
Closed by https://github.com/weaveworks/eksctl/pull/4926.

PR #4926, opened 2022-03-10
Set up AWS SDK v2 and migrate STS service

### LINKED ISSUE #4877: migrate to aws-sdk-go-v2 [1/9]
## Background

As part of #3215 we want to migrate to use the [aws-sdk-go v2](https://github.com/aws/aws-sdk-go-v2). This involves migrating all of the AWS API calls, roughly 500~ to use the new package. A spike was complete and its finding was documented [here](https://github.com/weaveworks/eksctl/issues/3215#issuecomment-1047099306) and that included a demo branch [here](https://github.com/weaveworks/eksctl/pull/4820/files). As the spike documents we can split out the work to migrate to v2 to be done in 9 parts, were we essentially use both v1 and v2 of the API while we migrate.

This ticket is to: add to `ClusterProvider` the setting up of a `v2` aws session (mirroring the configuration we use for `v1`) and add a `STSV2` to the `ClusterProvider`. This should allow us to use both `v1` and `v2` sessions side-by-side while we migrate. Once this step is done we can continue to the next parts and work in parallel

*This ticket needs to be done first before the others*

cPu1:
### Description

This changelist configures `aws-sdk-go-v2` and adds interfaces and types for obtaining AWS services. A code generation script is added that, as part of the build process, generates interfaces from AWS SDK types for mocking. Due to [this issue](https://github.com/vburenin/ifacemaker/issues/12) in ifacemaker, it passes the source generated by ifacemaker through another script `add_import.go` that adds the missing import. 

As part of #4877, the `STS` service is migrated to use `aws-sdk-go-v2`. One usage of STS cannot be migrated to v2 yet as it's blocked on `sigs.k8s.io/aws-iam-authenticator/pkg/token`.

Closes https://github.com/weaveworks/eksctl/issues/4925
Closes #4877

TODO:

- [x] https://github.com/weaveworks/eksctl/issues/4914
- [x] https://github.com/weaveworks/eksctl/issues/4934

### Checklist
- [ ] Manually tested


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -666,6 +668,14 @@ type ClusterProvider interface {
 	WaitTimeout() time.Duration
 	ConfigProvider() client.ConfigProvider
 	Session() *session.Session
+
+	ServicesV2() ServicesV2
```
aclevername, line 672: It might be easier to have the new interfaces live directly on the struct? Like
```suggestion
	STSV2() stsifacev2.STS
```

So that when we come to remove the orginal STS its just a:
- Remove `STS`
- Rename `STSV2` -> `STS`

The current method means this migration would be a bit more fiddly than that. We would have to:
- Remove `STS`
- Move `STS` from `ServicesV2` to the main interface

pkg/awsapi/generate/add_import.go
```diff
@@ -0,0 +1,73 @@
...
+	source, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
+		panic(fmt.Errorf("unexpected error reading from STDIN: %w", err))
+	}
+
+	packageName := os.Args[1]
+	if err = addDotImport(source, packageName); err != nil {
+		panic(err)
+	}
+}
```
aclevername, line 73: is this really easier than just having the sed :thinking: ?

cPu1: It's by no means easier than sed, but it's more robust IMO. We can also modify this program later to avoid using dot imports by qualifying local types with the package name.

aclevername: >  We can also modify this program later to avoid using dot imports by qualifying local types with the package name.

do we need to do that though :shrug: ?

> It's by no means easier than sed, but it's more robust IMO.

I would disagree, the contents of `addImport` is pretty obscure, on a first pass I struggle to understand what its disagree. The sed is just a bit simpler to read, and much easier to test manually

pkg/apis/eksctl.io/v1alpha5/types.go
cPu1: I tried to avoid that approach because it polluted the `ClusterProvider` interface with more methods and required suffixing all service methods with `V2`, but thinking about it more, it requires more changes, so I'll go ahead with adding V2 methods directly on `ClusterProvider`.

build/scripts/generate-aws-interfaces.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -ex
+
+[ "$#" -eq 2 ] || { echo "usage: ${0} <service-name> <interface-name>"; exit 1; }
+
+SERVICE_NAME="${1}"
+INTERFACE_NAME="${2}"
+
+PACKAGE_NAME="github.com/aws/aws-sdk-go-v2/service/${SERVICE_NAME}"
+AWS_SDK_DIR=$(go list -m -f '{{.Dir}}' "${PACKAGE_NAME}")
```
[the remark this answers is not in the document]
cPu1: Yeah, this is something we need to do manually before generating the interfaces because each service is now a Go module. A simple `go get <service-package>` before generating the interfaces will suffice.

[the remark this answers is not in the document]
aclevername: can this be written down somewhere? maybe next to the go generate lines so folks adding new interfaces know what to do

pkg/awsapi/generate/add_import.go
```diff
@@ -0,0 +1,73 @@
+//go:build ignore
```
Skarlso, line 1: @cPu1 Could this be separate from this PR perhaps? It's obscure enough to have it's own improvement. I like using the Go AST, but this requires a bit of testing and documentation on why it exists and what it does.

pkg/awsapi/generate/add_import.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"strconv"
+
+	"github.com/kris-nova/logger"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+)
+
+func addDotImport(code []byte, packageName string) error {
```
aclevername, line 21: can we add a comment describing what this does? Specifically in the context of generating the AWS mocks

pkg/eks/services_v2.go
```diff
@@ -0,0 +1,39 @@
...
+		// the logic used for AWS SDK v1.
+		o.Retryer = retry.NewAdaptiveMode(func(o *retry.AdaptiveModeOptions) {
+			o.StandardOptions = []func(*retry.StandardOptions){
+				func(so *retry.StandardOptions) {
+					so.MaxAttempts = maxRetries
+				},
+			}
+		})
+	})
+}
```
aclevername, line 39: this can be removed right?

pkg/awsapi/generate/add_import.go
Skarlso: Also, why this is needed, and where and how to apply it.

pkg/awsapi/generate/add_import.go
Skarlso: Never mind. I was convinced otherwise.

build/scripts/generate-aws-interfaces.sh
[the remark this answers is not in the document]
Skarlso: I think that can just be part of the script, right? Do a go get ... `asg` beforehand? No need to do that manually.

integration/tests/params.go
```diff
@@ -245,7 +247,7 @@ func (p *Params) attemptSettingUserID() {
 		return
 	}
 
-	out, err := clusterProvider.Provider.STS().GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	out, err := clusterProvider.Provider.STSV2().GetCallerIdentity(context.TODO(), &sts.GetCallerIdentityInput{})
```
Skarlso, line 250: TODO is a marker that we will do something with this later on. I would prefer `Background` which is mire generally used as something that doesn't define an actual context to be used later. The TODO context was defined for the reason that it shows up in TODO searches. :)

pkg/awsapi/generate/add_import.go
cPu1: Sure, I'll add a description.

pkg/connector/connector.go
```diff
@@ -98,7 +102,7 @@ func (c *EKSConnector) RegisterCluster(cluster ExternalCluster) (*ManifestList,
 }
 
 func (c *EKSConnector) createManifests(cluster *eks.Cluster) (*ManifestList, error) {
-	stsOutput, err := c.Provider.STS().GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	stsOutput, err := c.Provider.STSV2().GetCallerIdentity(context.TODO(), &sts.GetCallerIdentityInput{})
```
Skarlso, line 105: Same here for context TODO as above.

pkg/eks/api.go
```diff
@@ -323,9 +336,7 @@ func (c *ClusterProvider) GetCredentialsEnv() ([]string, error) {
 
 // checkAuth checks the AWS authentication
 func (c *ClusterProvider) checkAuth() error {
-
-	input := &sts.GetCallerIdentityInput{}
-	output, err := c.Provider.STS().GetCallerIdentity(input)
+	output, err := c.Provider.STSV2().GetCallerIdentity(context.TODO(), &stsv2.GetCallerIdentityInput{})
```
Skarlso, line 339: *Same here for context TODO

pkg/eks/services_v2.go
cPu1: It's required. `ServicesV2` is an embedded field of `ProviderServies`.

pkg/eks/apiv2.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+
+	if !hasCustomEndpoint {
+		return nil
+	}
+
+	return func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+		envName, ok := serviceIDEnvMap[service]
+		if ok {
```
Skarlso, line 90:
```suggestion
		if envName, ok := serviceIDEnvMap[service]; ok {
```

integration/tests/params.go
cPu1: I'd argue that `context.TODO()` is better suited here and in other places where a context is not available. 

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

build/scripts/generate-aws-interfaces.sh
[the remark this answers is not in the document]
aclevername: the name of the package differs sometimes, e.g. `asg` is `autoscaling` for the go mod. but it could be added

pkg/eks/retryer_v2.go
```diff
@@ -0,0 +1,49 @@
...
+	if !r.Retryer.IsErrorRetryable(err) {
+		return false
+	}
+
+	var oe *smithy.OperationError
+	return errors.As(err, &oe) && oe.Err != nil && isErrorRetryable(oe.Err)
+}
+
+func isErrorRetryable(err error) bool {
+	if aerr, ok := err.(awserr.RequestFailure); ok && aerr.Code() == "EC2MetadataError" {
```
Skarlso, line 42: You can use ApiError here, not? 

```go
	var re *awshttp.ResponseError
	if errors.As(err, &re) {
```

That thing has an http request inside.

integration/tests/params.go
cPu1: I'd argue that `context.TODO()` is better suited here and in other places where a context is not available. 

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

pkg/eks/services_v2.go
aclevername: can we just put everything directly on the struct? it will be easier.

Also all the other PRs will just replace the existing apis with the new ones, STS is one of the few where it will live alongside the V1

build/scripts/generate-aws-interfaces.sh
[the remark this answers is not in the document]
cPu1: Yup, I can add `go get` to the script.

pkg/eks/retryer_v2.go
Skarlso: Never mind, I searched and it's not easy. :D Fuck it, I'm okay with this. :D

integration/tests/params.go
Skarlso: Hmm. Alright, fair point.

aclevername (changes requested):
lgtm, just waiting for comments to be addressed

aclevername:
> I have added #4934 to the TODO checklist, going to work on it before making any more changes.

this PR is already quite big, can it not go in a separate one?

build/scripts/generate-aws-interfaces.sh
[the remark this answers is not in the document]
cPu1: I have added a `go get` to the script.

pkg/eks/services_v2.go
cPu1: I created a new struct because the service methods (STS() etc.) are implemented differently in that the service client is configured in the method itself. Additionally, there's a lot more happening in `pkg/eks/api.go`, so I wanted this to be cleaner and keep the diff minimal. I wouldn't mind adding it to the existing struct though, but this PR has already grown too much.

cPu1:
I'll fix the lint issues shortly.

integration/tests/params.go
aclevername, line 250: can we have the context be passed down from the `ctl/` file? It can still be a `TODO` but having it passed down from the top ensures we re-use it across the whole command

pkg/connector/connector.go
aclevername, line 105: same here

pkg/credentials/filecache.go
```diff
@@ -36,6 +38,16 @@ func (r *RealClock) Now() time.Time {
 	return time.Now()
 }
 
+// Flock provides an interface to handle file locking.
+//counterfeiter:generate -o fakes/fake_flock.go . Flock
+type Flock interface {
+	TryRLockContext(ctx context.Context, retryDelay time.Duration) (bool, error)
+	TryLockContext(ctx context.Context, retryDelay time.Duration) (bool, error)
```
aclevername, line 45: whats the difference?

pkg/credentials/filecache_v2.go
```diff
@@ -0,0 +1,110 @@
+package credentials
```
aclevername, line 1: @Skarlso has the most context on this, going to defer to him but the testing looks good (athough -1 point for using tables :wink: )

pkg/credentials/filecache.go
Skarlso: I specifically didn't want to mock flock because I wanted to make sure that the ability is working with a proper file system. That's why the tests contained a temp folder in which things were happening.

Skarlso: Also if you mock the locking mechanism and the file system means that you aren't testing a large junk of the behaviour.

pkg/credentials/filecache.go
```diff
@@ -173,6 +191,15 @@ func writeCache(filename string, cache cacheFile) error {
 // otherwise fetching the credential from the underlying Provider and caching the results on disk
 // with an expiration time.
 func (f *FileCacheProvider) Retrieve() (credentials.Value, error) {
+	f.once.Do(func() {
```
Skarlso, line 194: Why is this needed?

pkg/credentials/filecache_v2.go
```diff
@@ -0,0 +1,110 @@
...
+		SecretAccessKey: c.Credential.SecretAccessKey,
+		SessionToken:    c.Credential.SessionToken,
+		Source:          c.Credential.ProviderName,
+		CanExpire:       !c.Expiration.IsZero(),
+		Expires:         c.Expiration,
+	}
+}
+
+// Retrieve implements aws.CredentialsProvider.
+func (f *FileCacheV2) Retrieve(ctx context.Context) (aws.Credentials, error) {
```
Skarlso, line 60: So, what exactly changed here in the API that the other cache is no longer working?

cPu1: That's evident in the function signature:

`Retrieve(ctx context.Context) (aws.Credentials, error) {`

vs

`Retrieve() (credentials.Value, error) {`

pkg/credentials/filecache.go
cPu1: When the credentials aren't cached, both caches will try to obtain the MFA token from STDIN, prompting the user twice. This check is for obtaining the credentials from the file cache, only for the first call to Retrieve to ensure it reads the cached token if it was set by the other cache.

pkg/credentials/filecache_v2.go
cPu1: Kubernetes and many other popular projects use table-driven testing (with or without Ginkgo), and it's in fact uncommon for those projects to copy-paste code in tests :slightly_smiling_face:

pkg/credentials/filecache.go
cPu1: >whats the difference?

`TryRLockContext` takes a shared lock, whereas `TryLockContext` takes an exclusive lock.

pkg/eks/services_v2.go
```diff
@@ -0,0 +1,39 @@
...
+
+// STSV2 implements the AWS STS service
+func (s *ServicesV2) STSV2() awsapi.STS {
+	return sts.NewFromConfig(s.config, func(o *sts.Options) {
+		// Disable retryer for STS
+		// (see https://github.com/weaveworks/eksctl/issues/705)
+		o.Retryer = aws.NopRetryer{}
+	})
+}
```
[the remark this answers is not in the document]
cPu1: It does, however, the existing approach in v1 means that all clients will be created even if a command ends up using only some, or none. I was planning to lazily initialise clients and wanted to note it here but it slipped my mind. I have pushed it now.

integration/tests/params.go
cPu1: There's no `ctl` here since this is the integration tests, but I have added it now for eksctl commands.

pkg/credentials/filecache_v2.go
aclevername: > Kubernetes and many other popular projects use table-driven testing

and many projects don't write unit tests at all, doesn't mean its correct :grin:

pkg/eks/services_v2.go
```diff
@@ -0,0 +1,55 @@
...
+				o.StandardOptions = []func(*retry.StandardOptions){
+					func(so *retry.StandardOptions) {
+						so.MaxAttempts = maxRetries
+					},
+				}
+			})
+		})
+	}
+	return s.cloudformation
+}
```
aclevername, line 55: rather than doing the mutex locking and checking if we've already created the client, can we not just construct all the clients as part of `ServicesV2` and then the func's are just getters? Will reduce the complexity a bit

aclevername:
do we need any updates/changes too https://github.com/weaveworks/eksctl/blob/main/integration/tests/caching/credential_caching_test.go

pkg/eks/services_v2.go
cPu1: > Will reduce the complexity a bit

It comes at the cost of initialising all clients even when some or none of them will be used. The approach used for SDK v1 suffers from this problem. The provider stores more than 10 clients and initialises all of them when only a few of them will be used.

aclevername: but instead we have to have this mutex and every getter be is basically a singleton constructor, this seems a lot more complicated to me

cPu1: Each client construction is fairly complex as it requires resolving multiple configurable options and setting defaults. It is wasteful to perform this initialisation for more than 10 clients when only a few of them will be used. I'll also note that the synchronisation code here can be further improved but at the cost of increased complexity.

aclevername: >  It is wasteful to perform this initialisation for more than 10 clients when only a few of them will be used. 

that doesn't concern me at all, 10 in-memory clients isn't breaking the memory bank. I'd rather have much less complex code

cPu1: My concern isn't increased memory usage, it's about wasting CPU resources initialising fields that'll never be used.

aclevername: Is that really a concern we have?

cPu1: I do, not sure about others. It's an improvement over the previous behaviour.

Himangini: > Each client construction is fairly complex as it requires resolving multiple configurable options and setting defaults. It is wasteful to perform this initialisation for more than 10 clients when only a few of them will be used. I'll also note that the synchronisation code here can be further improved but at the cost of increased

+1 on this, i like this approach.

Himangini (commented):
I've gone through all the files, LGTM. Waiting for integration tests to pass, will give a second review again today in an hour or so.

pkg/eks/services_v2.go
aclevername: I really disagree, the code is much more complex this way, all to save the costs of creating a handful more structs? :confused:

Himangini (approved):
LGTM.

pkg/eks/services_v2.go
Himangini: Lets take this offline 👍🏻

pkg/awsapi/generate/add_import.go
```diff
@@ -0,0 +1,76 @@
+//go:build ignore
```
richardcase, line 1: Could we contribute a change to ifacemaker instead of maintaining this?

cPu1: There's an open [PR](https://github.com/vburenin/ifacemaker/pull/47) that claims to fix this issue in ifacemaker, but it has not been reviewed in months. I agree that it'd be best to have it fixed upstream.

pkg/eks/services_v2.go
richardcase: I can see both sides :smile: 

In reality, we don't need to worry too much about optimizing this area and creating all the services would be fine.

On the flip side, its a standard lazy initialization pattern. The one thing i would say is that we are assuming that all new services will use this pattern and that contributors will need to follow it. So, it might be worth documenting it somewhere?

pkg/awsapi/generate/add_import.go
richardcase: Have you tried the change from the PR by any chance? I'm curious if its works

richardcase: Added a comment to the PR :smile:

cPu1: >Have you tried the change from the PR by any chance? I'm curious if its works

That PR's fix does not work. I tried it on both https://github.com/vburenin/ifacemaker/issues/12 and our `generate-aws-interfaces` script. 

(I had actually also tested that potential fix before but it wasn't working then either, otherwise my plan was to use that user's fork).

pkg/eks/client.go
```diff
@@ -40,6 +40,7 @@ func (c *ClusterProvider) NewClient(spec *api.ClusterConfig) (*Client, error) {
 	client := &Client{
 		Config: config,
 	}
+	// TODO: usage of STS v2 is blocked on sigs.k8s.io/aws-iam-authenticator/pkg/token
```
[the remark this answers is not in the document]
cPu1: >And its blocked because sigs.k8s.io/aws-iam-authenticator/pkg/token uses the V1 SDK?

Correct. 

>It would be good to expand on why its blocked and is there a follow-up issue to re-visit this again in the future?

Yeah, we are going to open a follow-up issue: https://github.com/weaveworks/eksctl/pull/4926#pullrequestreview-915916354

pkg/eks/services_v2.go
```diff
@@ -0,0 +1,55 @@
...
+			// Disable retryer for STS
+			// (see https://github.com/weaveworks/eksctl/issues/705)
+			o.Retryer = aws.NopRetryer{}
+		})
+	}
+	return s.sts
+}
+
+// CloudFormationV2 implements the AWS CloudFormation service
+func (s *ServicesV2) CloudFormationV2() awsapi.CloudFormation {
```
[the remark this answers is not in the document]
cPu1: We can't do that in this PR as this one just added this method, but the rest of the code is still using the CloudFormation v1 client. We'll drop the V2 suffix as part of #4882.

pkg/awsapi/generate/add_import.go
richardcase: Now we have write access shall we fix ifacemaker instead of maintaining this in our own repo?

pkg/eks/client.go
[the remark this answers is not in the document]
richardcase: Be good to update the comment and include the issue number in the TODO?

pkg/eks/services_v2.go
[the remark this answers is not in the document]
richardcase: Understood :+1:

pkg/awsapi/generate/add_import.go
cPu1: Hmm, that fix might take some time. The program I wrote can't be used as-is in ifacemaker. The implementation in ifacemaker will also be more complex because a more complete solution would not use a dot import but rather prefix the non-local packages with the remote package's name. I can open an issue in our repo to track this though. WDYT?

richardcase: @Himangini - what are your thought on doing this as a follow-up? My concern if workarounds like this is that they have a habit of becoming the long-term solution.

pkg/eks/client.go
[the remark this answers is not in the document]
cPu1: Agreed, done.

pkg/awsapi/generate/add_import.go
cPu1: >My concern with workarounds like this is that they have a habit of becoming the long-term solution.

Agreed. Best would be to prioritize work on the follow-up issue so it doesn't become a long-term solution.

Himangini: I agree with both 👍🏻  @cPu1 can you create a follow-up issue and add it here as a comment, please. I'll bring this in refinement/planning next week.

richardcase: Awesome thanks @cPu1  and @Himangini

cPu1: I have opened https://github.com/weaveworks/eksctl/issues/4994

pkg/awsapi/generate/add_import.go
nikimanoledaki: >  it's more robust IMO

I'm trying to understand the pros/cons of each `sed` vs `addImport`. @cPu1 could you expand a bit on this, please? `addDotImport` adds quite a lot of complexity compared to `sed`. What do you see as the issues or limitations of `sed`, and how does `addDotImport ` solve this and perform better?

If we do go with this program, I would suggest adding some tests to ensure that it really is more robust and to make it more maintainable.

richardcase (approved):
There are still some difference of opinions on some things and few unresolved comments. From my side i see nothing that is blocking, especially as we have a number of follow-up issues and can revisit things if needed. So with that in mind this /lgtm

pkg/credentials/filecache.go
Himangini: @cPu1 i missed this earlier but just leaving a comment here as a reminder, I think we should add some comments here explaining what these functions do. They look identical and are likely to cause confusion.

cPu1: I have added godoc comments.

pkg/awsapi/generate/add_import.go
cPu1: @nikimanoledaki `sed` is not the right tool to manipulate Go sources. That `sed` snippet can generate programs that have syntax errors.

Here's what's wrong with using that sed snippet:

```bash
$ cat <<EOF | sed "/\"context\".*/a . \"github.com/aws/aws-sdk-go-v2/service/eks\""
// Code generated by ifacemaker; DO NOT EDIT.

package awsapi

import "context"

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}

EOF

// Code generated by ifacemaker; DO NOT EDIT.

package awsapi

import "context"
. "github.com/aws/aws-sdk-go-v2/service/eks"

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}


$ cat <<EOF | sed "/\"context\".*/a . \"github.com/aws/aws-sdk-go-v2/service/eks\""
// Code generated by ifacemaker; DO NOT EDIT.
// Copyright 2020 Weaveworks.
// The awsapi package provides an interface to AWS services. The "context" package is used to carry deadlines and
// cancellation signals.

package awsapi

import (
    "context"
)

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}

EOF

// Code generated by ifacemaker; DO NOT EDIT.
// Copyright 2020 Weaveworks.
// The awsapi package provides an interface to AWS services. The "context" package is used to carry deadlines and
. "github.com/aws/aws-sdk-go-v2/service/eks"
// cancellation signals.

package awsapi

import (
    "context"
. "github.com/aws/aws-sdk-go-v2/service/eks"
)

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}


$ cat <<EOF | sed "/\"context\".*/a . \"github.com/aws/aws-sdk-go-v2/service/eks\""
package awsapi

import (
    "fmt"
)

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}

EOF

package awsapi

import (
    "fmt"
)

// CloudFormation provides an interface to the AWS CloudFormation service.
type CloudFormation interface {
 // ...
}

```



As for the maintenance of the `add_import.go` program, we have an issue for that: https://github.com/weaveworks/eksctl/issues/4994

pkg/eks/services_v2.go
Skarlso: Do we have any kind of benchmarking that supports this "optimisation"? Because without it, this is just going in blind. At first, it actually causes a problem then solve one which might not exist.

Further, this will remove the nice UX of failing early in case the client can't be constructed instead of much later in the process.

And, I think this should have been in a separate PR and not added to a Pr which already huge.

I don't agree with these changes. :/

Skarlso: Oh see, this is just constructing structs? I would argue that this would have impacted performance in any way. I would be very surprised. Which is why a benchmark should have existed first. ¯\_(ツ)_/¯

cPu1: >Do we have any kind of benchmarking that supports this "optimisation"? Because without it, this is just going in blind.

Please go through the entire discussion on this topic. I have already commented this: "this change was not just about optimizing code in terms of CPU cycles, but also to communicate the intent that only a few of these services will be used by most code paths."

This changelist isn't optimizing existing code, it's making sure that no code is unnecessarily run if it's not used.


>Further, this will remove the nice UX of failing early in case the client can't be constructed instead of much later in the process.

Where exactly was it failing early before? The `NewFromConfig` functions do not return an error.

>Oh see, this is just constructing structs? I would argue that this wouldn't have impacted performance in any way. I would be very surprised. Which is why a benchmark should have existed first. ¯\_(ツ)_/¯

"Constructing structs" happens at the end when all the initialization and resolution functions have been run. To "just construct structs" you need to set its fields which requires making several function calls here in the `NewFromConfig` functions. But as I noted above, this is not just trying to improve its performance.

Skarlso: Which would have been more evident if this entire change set would have been a separate PR where all the moving and related parts were visible. I'm not against doing lazy loading, but I was against doing it in an already huge pr with several changes in it, related or not. This should have been in a separate pr for clear visibility.

pkg/credentials/filecache_v2.go
Skarlso: Shouldn't this have implemented rather our own MFA token provider which should have been using a file cache in the background?

Skarlso: But I guess it doesn't matter anymore.

PR #4937, opened 2022-03-11
Migrate ELB and ELBv2 to aws-sdk-go-v2

cPu1:
### Description

This changelist migrates usage of ELB and ELBv2 to aws-sdk-go-v2. Because the usage was fully migrated to v2, the existing methods are changed (instead of adding ELBV2() and ELBV2V2()) to use the v2 interfaces.

This branch is based off https://github.com/weaveworks/eksctl/pull/4926.

### Checklist
- [ ] Manually tested


aclevername:
going to hold off on reviewing until https://github.com/weaveworks/eksctl/pull/4926 is merged

pkg/elb/cleanup.go
```diff
@@ -44,8 +48,18 @@ type loadBalancer struct {
 	ownedSecurityGroupIDs map[string]struct{}
 }
 
+type elb interface {
```
Skarlso, line 51: What is the purpose of these two new unexported interfaces?

cPu1: Both interfaces contain only the set of methods required by `elb.Cleanup`. They don't need to express dependency on the entire ELB and ELBv2 interfaces.

Skarlso: Ah okay, so they aren't used anywhere else. And we don't want to extend these ever?

Skarlso: To be more clear, I'm all for this!! I like removing / minimizing this dependency. I'm asking if we could then, perhaps, do it for the rest of them? :)

Skarlso: Because doing it for only one, will break the consistency in between for what we generate mocks and for what we don't, if that makes sense? :)

cPu1: >Ah okay, so they aren't used anywhere else. 

Correct.

>And we don't want to extend these ever?

If by extend you mean adding more methods then we'd do that if `Cleanup` required more methods from the ELB* interfaces.

>But also, hm. If we do this here, won't that be inconsistent with how the rest of the code deals with the dependency on the other interfaces?

Not every part of the codebase, we do have some places where we/I have used a similar pattern. 

>Because doing it for only one, will break the consistency in between for what we generate mocks and for what we don't, if that makes sense? :)

We don't need to be consistent with following the rest of the code when improvements can be made gradually, like here. This changelist also does not rely on the way mocks are generated.

Skarlso: Sure, but you already have a mock. We already have that dependency. I thought this could be used instead of that whole thing. But you already generated the mock, we already load in the whole package and already have it in our list of dependencies.

And then there is this single function in a single package which uses an unexported interface for two methods out of that generated mock. :)

I'm sure changes can be gradual, sure, but this lone interface here makes little sense. We either want to use our own interface which then can be used by others, so make this exported so it can later be used as well ( gradual process ). Or we use the mock everywhere since it's already there anyways. In which case these two here make little sense. :)

Skarlso: Hang on, let's chat. I think I might be looking at this wrong.

pkg/elb/cleanup.go
```diff
@@ -7,16 +7,21 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
```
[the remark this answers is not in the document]
cPu1: Done.

pkg/eks/api.go
```diff
@@ -253,16 +239,7 @@ func New(spec *api.ProviderConfig, clusterSpec *api.ClusterConfig) (*ClusterProv
...
-	if endpoint, ok := os.LookupEnv("AWS_ELB_ENDPOINT"); ok {
-		logger.Debug("Setting ELB endpoint to %s", endpoint)
-		provider.elb = elb.New(s, s.Config.Copy().WithEndpoint(endpoint))
 
-	}
-	if endpoint, ok := os.LookupEnv("AWS_ELBV2_ENDPOINT"); ok {
-		logger.Debug("Setting ELBV2 endpoint to %s", endpoint)
-		provider.elbv2 = elbv2.New(s, s.Config.Copy().WithEndpoint(endpoint))
-
-	}
```
aclevername, line 265: do we not still need this for the v2 clients?

aclevername (approved):

Skarlso (approved):

pkg/eks/api.go
cPu1: We do, it's handled in the SDK initialization for v2.

PR #4960, opened 2022-03-16
update to go 1.18

aclevername:
### Description
@Skarlso be free child


Skarlso:
OMG OMG OMG OMG OMG OMG OMG OMG OMG

Skarlso (approved):

Skarlso:
Do we have any image builders at some obscure, private location or somewhere which could die because of this? :D

aclevername:
> Do we have any image builders at some obscure, private location or somewhere which could die because of this? :D

nah, think this is all we have to do AFAIK

PR #4963, opened 2022-03-16
Prevent duplication, and add flag to disable shadowing of IAM Identity Mappings

DameonSmith:
### Description
I implemented functionality that will prevent duplication of `aws_auth` entries. The goal of this was to make the `eksctl create iamidentitymapping` command more automation friendly. I had been using this command in some IAC code and ended up having a large number of copies of the exact same record.

This involved changes in 2 areas primarily:
1. The create command itself to exit when a duplicate Identity is found.
2. Creation of a compare function for Identities.
  * This might be more robust if the Identity groups were sorted before being compared. Wasn't sure if it made sense to implement that way. 
  
I also added a flag to the `eksctl create iamidentitymapping` called `--no-shadow` when this flag is set the command will error should there be an Identity using the same ARN already in place. The goal of this change is to be able to provide better guarantees that configurations aren't accidentally changed.

Let me know if there's anything here that doesn't match the projects style or format, or if I'm just contributing to technical debt by working it in the way that I did. 

Also as a side note the tests ran extremely slowly on my machine (1-2 second per package), is that expected behavior?

### Checklist
- [x] Manually tested


Skarlso:
Hi!

Thank you for your contribution! :)

>Also as a side note the tests ran extremely slowly on my machine (1-2 second per package), is that expected behavior?

Unit tests? All tests? Hmm, they can be slow, since they set up a lot of mocks and extra things and a fake http router potentially and a kubernetes client... :D A lot of things.

I'll leave review comments in code.

pkg/ctl/create/iamidentitymapping.go
```diff
@@ -48,6 +51,7 @@ func createIAMIdentityMappingCmd(cmd *cmdutils.Cmd) {
 		fs.StringSliceVar(&options.Groups, "group", []string{}, "Group within Kubernetes to which IAM role is mapped")
 		fs.StringVar(&options.ServiceName, "service-name", "", "Service name; valid value: emr-containers")
 		fs.StringVar(&options.Namespace, "namespace", "", "Namespace in which to create RBAC resources (only valid with --service-name)")
+		fs.BoolVar(&options.NoShadow, "no-shadow", false, "Turn off shadowing of iam identity mappings mappings")
```
Skarlso, line 54: I would rather name this some more descriptive. like, `-no-duplicate-mappings` or `-error-on-duplication` or `no-dulpicates`.

pkg/iam/mapping.go
```diff
@@ -29,6 +29,33 @@ type Identity interface {
 	Account() string
 }
 
+// CompareIdentity takes 2 Identity values and checks to see if they are identitcal
+func CompareIdentity(a, b Identity) bool {
+	sameAccount := (a.ARN() == b.ARN() &&
+		a.Type() == b.Type() &&
+		a.Username() == b.Username() &&
+		a.Account() == b.Account())
```
Skarlso, line 37:
```suggestion
	sameAccount := a.ARN() == b.ARN() &&
		a.Type() == b.Type() &&
		a.Username() == b.Username() &&
		a.Account() == b.Account()
```

pkg/iam/mapping.go
```diff
@@ -29,6 +29,33 @@ type Identity interface {
...
+	}
+
+	aGroups := a.Groups()
+	bGroups := b.Groups()
+	if len(aGroups) != len(bGroups) {
+		return false
+	}
+
+	for i := range aGroups {
+		if aGroups[i] != bGroups[i] {
```
Skarlso, line 50: Hm, in the description you are saying this:

>This might be more robust if the Identity groups were sorted before being compared. Wasn't sure if it made sense to implement that way.



So if this isn't the case, then you are depending here on a thing that doesn't exist, right? And in fact, checking the code, we do not sort groups. Please add a unit test which checks for any kind of order of checking if these are equal. :)

Skarlso (changes requested):
Add tests and code to check unsorted groups are equal.

pkg/iam/mapping.go
DameonSmith: I made it so that it sorts right before the compare.

pkg/ctl/create/iamidentitymapping.go
DameonSmith: Updated to `--no-duplicate-arns`

Skarlso (approved):
LGTM

aclevername (approved):
excellent tests :sparkles: lgtm!

pkg/ctl/create/iamidentitymapping.go
```diff
@@ -133,6 +137,15 @@ func doCreateIAMIdentityMapping(cmd *cmdutils.Cmd, options iamIdentityMappingOpt
 		for _, identity := range identities {
 			arn := identity.ARN()
 
+			if iam.CompareIdentity(id, identity) {
```
Skarlso, line 140: Hey @DameonSmith!

This check here, should have been behind the below if, right? :D :D 

I think we missed that. :)

Skarlso: Or rather, I think this also needs the `&& options.NoDuplicateArns`.

DameonSmith: This check was to just ignore any duplicated identity entirely. Regardless of whether or not the duplicate arn flags was set. 
I was trying to avoid having this kind of pollution:
```
arn: arn1, namespace: default, groups: [group1, group2]
arn: arn1, namespace: default, groups: [group1, group2]
arn: arn1, namespace: default, groups: [group2, group1]
etc...
```
The context was specifically using eskctl in pipelines and IaC to create default roles and users for a cluster. The pipeline was run multiple times and would just input the same config over and over again. My assumption was that if a record was in the aws_auth config map it should just be ignored without an error as there's no functional difference. 

The NoDuplicateArns flag was meant to prevent silent shadowing of existing roles with different permissions. I wanted to ability to have better visibility on something like the "ReadOnlyUsers" role accidentally got the masters group added to it. 

Hopefully that makes sense, and there's not something about my assumptions that were mistaken. Let me know if I need to fix something.

DameonSmith: Oh I didn't realize that was a feature, Is there a reason to allow duplicates? 

Either way I can add that flag as a requirement if that's what you'd like. Should I open another PR or is there some way to open this one?

PR #4995, opened 2022-03-23
Remove usage of v1 api for CloudWatchLogs, CloudTrail and ASG and replace it with v2

aclevername:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4879. Removes the usage of aws-sdk-go v1 for CloudWatchLogs, CloudTrail and ASG and replace it with aws-sdk-go-v2


pkg/actions/nodegroup/scale.go
```diff
@@ -66,19 +67,19 @@ func (m *Manager) scaleUnmanagedNodeGroup(ng *api.NodeGroupBase, stackInfo manag
...
 
 	if ng.DesiredCapacity != nil {
-		input.DesiredCapacity = aws.Int64(int64(*ng.DesiredCapacity))
+		input.DesiredCapacity = aws.Int32(int32(*ng.DesiredCapacity))
 	}
-	out, err := m.ctl.Provider.ASG().UpdateAutoScalingGroup(input)
+
+	_, err := m.ctl.Provider.ASG().UpdateAutoScalingGroup(ctx, input)
 	if err != nil {
-		logger.Debug("ASG update output: %s", out.String())
```
Skarlso, line 81: We don't need this output anymore? 🤔

Skarlso (approved):
LGTM with one question

pkg/actions/nodegroup/scale.go
aclevername: `out.String()` is no longer a thing, so I wasn't really sure if it made sense to just print th estruct. I'm not really sure, but given its a debug and not info output i think its okay to just remove :shrug: wdyt?

Skarlso: nah

PR #5060, opened 2022-04-04
Renamed STS

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/4885.

The final part is to just rename STSV2 to STS.

### Checklist
- [x] Manually tested


pkg/eks/api.go
```diff
@@ -300,7 +299,7 @@ func (c *ClusterProvider) GetCredentialsEnv() ([]string, error) {
 
 // checkAuth checks the AWS authentication
 func (c *ClusterProvider) checkAuth() error {
-	output, err := c.Provider.STSV2().GetCallerIdentity(context.TODO(), &stsv2.GetCallerIdentityInput{})
+	output, err := c.Provider.STS().GetCallerIdentity(context.TODO(), &sts.GetCallerIdentityInput{})
```
cPu1, line 302: We should propagate the context here.

cPu1 (approved):
Just one comment about context propagation but otherwise LGTM.

pkg/eks/api.go
Skarlso: From where?

cPu1: I meant by adding a `Context` parameter to `checkAuth` and passing it from the top-most call.

Skarlso: I don't necessarily agree with that. :) But that's not a hill I'll die on. :D

aclevername: I'd agree with @cPu1 , we've done that for other v2 migrations, not sure why we wouldn't here :thinking:

aclevername (approved):
lgtm- 1 comment

pkg/eks/api.go
Skarlso: Fine, like I said, I don't agree, but I don't care enough. :D

Skarlso: you are the ones who will review it hahahaha

Skarlso: Done :)

pkg/ctl/deregister/cluster.go
```diff
@@ -1,12 +1,14 @@
 package deregister
 
 import (
+	context2 "context"
```
Skarlso, line 4: Huh, weird refactor action Goland... I'll fix this in a sec.

pkg/eks/api.go
Skarlso: Okay, it's not that bad actually. :D

pkg/actions/cluster/get.go
```diff
@@ -58,7 +59,7 @@ func GetClusters(provider api.ClusterProvider, listAllRegions bool, chunkSize in
 			continue
 		}
 		// Reset region and recreate the client.
-		ctl, err := newClusterProvider(&api.ProviderConfig{
+		ctl, err := newClusterProvider(context.TODO(), &api.ProviderConfig{
```
aclevername, line 62: shouldn't this be passed in here too?

Skarlso: It's a serious question. :D Like, is it `GetClusters`? Or is it `getAndPrinterClusters` or is it `doGetCluster` which is added in 

```go
		return doGetCluster(cmd, params, listAllRegions)
	}
```

That's added from `getClusterCmd` that's added in `Command` which is initalised in main. :)

aclevername: I'd stop at ctl

Skarlso: Ok, moved it out until ctl.

Issue #5062, opened 2022-04-04
[Task] Remove support for EKS 1.18


cPu1:
EKS will drop support for creating 1.18 clusters soon. This issue is to [track this change](https://github.com/aws/containers-roadmap/issues/1630) and remove support for creating new EKS 1.18 clusters.

PR #5073, opened 2022-04-05
remove support for 1.18

aclevername:
### Description

Closes https://github.com/weaveworks/eksctl/issues/5062

```
eksctl create cluster --name jk --version 1.18
2022-04-05 16:44:00 [ℹ]  eksctl version 0.93.0-dev+2333ab45f.2022-04-05T16:43:52Z
2022-04-05 16:44:00 [ℹ]  using region us-west-2
Error: invalid version, 1.18 is no longer supported, supported values: 1.19, 1.20, 1.21, 1.22
see also: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
```


Skarlso (changes requested):
Please also update dry-run.md and addons.md. :)

userdocs/src/introduction.md
```diff
@@ -23,7 +23,7 @@ eksctl create cluster --name=cluster-1 --nodes=4
 
 ```
 
-EKS supports versions `1.18`, `1.19`, `1.20`, `1.21` (default) and `1.22`.
+EKS supports versions `1.19`, `1.20`, `1.21` (default) and `1.22`.
 With `eksctl` you can deploy any of the supported versions by passing `--version`.
 
 ```
```
Skarlso, line 29: It doesn't let me comment, but this line below uses 1.18 as an example:
```
eksctl create cluster --version=1.18
```

Skarlso (approved):
thanks!

pkg/actions/addon/addon.go
```diff
@@ -36,10 +34,6 @@ type Manager struct {
 }
 
 func New(clusterConfig *api.ClusterConfig, eksAPI eksiface.EKSAPI, stackManager manager.StackManager, withOIDC bool, oidcManager *iamoidc.OpenIDConnectManager, clientSet kubeclient.Interface, timeout time.Duration) (*Manager, error) {
-	if err := supportedVersion(clusterConfig.Metadata.Version); err != nil {
```
Himangini, line 39: why are we removing this?

pkg/actions/addon/addon.go
```diff
@@ -118,19 +112,6 @@ func (a *Manager) getLatestMatchingVersion(addon *api.Addon) (string, error) {
...
-func supportedVersion(version string) error {
-	supported, err := utils.IsMinVersion(api.Version1_18, version)
-	if err != nil {
-		return err
-	}
-	switch supported {
-	case true:
-		return nil
-	default:
-		return fmt.Errorf("addons not supported on %s. Must be using %s or newer", version, api.Version1_18)
```
Himangini, line 130: why are we removing this check? shouldn't this be 
`return fmt.Errorf("addons not supported on %s. Must be using %s or newer", version, api.Version1_19)`

pkg/actions/addon/addon_test.go
```diff
@@ -17,11 +17,4 @@ var _ = Describe("Addon", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
-
-	When("the version is not supported", func() {
-		It("errors", func() {
-			_, err := addon.New(&api.ClusterConfig{Metadata: &api.ClusterMeta{Version: "1.17"}}, &mocks.EKSAPI{}, nil, false, nil, nil, 5*time.Minute)
-			Expect(err).To(MatchError("addons not supported on 1.17. Must be using 1.18 or newer"))
```
Himangini, line 24: again the same question, shouldn't this be 
`Expect(err).To(MatchError("addons not supported on 1.18. Must be using 1.19 or newer"))`

pkg/addons/default/kube_proxy.go
```diff
@@ -172,13 +155,8 @@ func addArm64NodeSelector(daemonSet *v1.DaemonSet, archLabel string) error {
...
 }
 
-func getLatestKubeProxyImage(input AddonInput, greaterThanOrEqualTo1_18 bool) (string, error) {
+func getLatestKubeProxyImage(input AddonInput) (string, error) {
 	defaultClusterVersion := generateImageVersionFromClusterVersion(input.ControlPlaneVersion)
-	// EKS Addons API only works for 1.18 and above
-	if !greaterThanOrEqualTo1_18 {
-		return defaultClusterVersion, nil
-	}
-
```
Himangini, line 181: don't we need this check for 1.19 and above?

pkg/addons/default/kube_proxy_test.go
```diff
@@ -92,182 +114,178 @@ var _ = Describe("KubeProxy", func() {
...
+								{
+									AddonVersion: aws.String("v1.17.0-eksbuild.1"),
+								},
+								{
+									//latest, unordered list to ensure we sort correctly
+									AddonVersion: aws.String("v1.18.1-eksbuild.2"),
+								},
+								{
+									AddonVersion: aws.String("v1.18.1-eksbuild.1"),
+								},
```
Himangini, line 148: shouldnt this be 
```suggestion
								{
									//latest, unordered list to ensure we sort correctly
									AddonVersion: aws.String("v1.19.1-eksbuild.2"),
								},
								{
									AddonVersion: aws.String("v1.19.1-eksbuild.1"),
								},
```

pkg/addons/default/kube_proxy_test.go
```diff
@@ -29,15 +29,38 @@ var _ = Describe("KubeProxy", func() {
...
+						AddonVersions: []*awseks.AddonVersionInfo{
+							{
+								AddonVersion: aws.String("v1.17.0-eksbuild.1"),
+							},
+							{
+								//latest, unordered list to ensure we sort correctly
+								AddonVersion: aws.String("v1.18.1-eksbuild.2"),
+							},
+							{
+								AddonVersion: aws.String("v1.18.1-eksbuild.1"),
```
Himangini, line 51: shouldn't this be 
```suggestion
								//latest, unordered list to ensure we sort correctly
								AddonVersion: aws.String("v1.19.1-eksbuild.2"),
							},
							{
								AddonVersion: aws.String("v1.19.1-eksbuild.1"),
```

Himangini (commented):
Couple of questions are removal of supported versions checks for addons

pkg/actions/addon/addon.go
Skarlso: Not needed, because addons are supported after 18 and now there is an earlier check which checks if we are even below that. So this check is now superseded by the fact that the minimum version supported is now supported by addons.

Himangini (commented):
Also i noticed some specific checks in tests around 1.18, do we need to update them to 1.19 to make sure everything works fine for 1.19 in the following files?
pkg/actions/addon/create_test.go
pkg/actions/addon/delete_test.go
pkg/actions/addon/get_test.go
pkg/actions/addon/update_test.go
pkg/actions/addon/describe_versions_test.go

pkg/addons/default/kube_proxy.go
Skarlso: Anything below is now no longer supported.

Himangini: > Anything below is now no longer supported.

Yeah, so don't we need this check to return early `if !greaterThanOrEqualTo1_19` ? 🤔

pkg/addons/default/kube_proxy_test.go
Skarlso: Same here as below. :)

pkg/addons/default/kube_proxy.go
aclevername: sorry was out at hospital, thanks for replying @Skarlso ✨ 

Yeah as @Skarlso says since no clusters older than 1.18 can exist, we don't need to check anymore. Its a guarantee that the cluster will be compatible

pkg/actions/addon/addon.go
aclevername: same answer as below

pkg/actions/addon/addon_test.go
aclevername: same answer as below

pkg/addons/default/kube_proxy.go
Himangini: cool 👍🏻

pkg/actions/addon/addon.go
Himangini: noted.

pkg/addons/default/kube_proxy_test.go
Himangini: ok 👍🏻

Himangini:
> I will take a look but I have a feeling it doesn't matte for the tests, any version can be used for unit-testing, since we aren't talking to the real EKS API. We have lots of tests referencing old versions since its just not worth changing. E.g. kube-proxy tests reference 1.15/6 https://github.com/weaveworks/eksctl/blob/main/pkg/addons/default/kube_proxy_test.go

As I mentioned in the earlier comment, the files refer to specific version **1.18** in tests. If we remove the supported version checks and `greaterThanOrEqualTo1_18` checks, then we should remove those tests since it's irrelevant now or update them to use **1.19** otherwise it's confusing. Eg., Why would you test the creation of an addon with a non-supported cluster version?

Himangini:
I didn't know you have a spokesperson for you @aclevername :D

pkg/actions/addon/addon.go
cPu1, line 39: One concern I have with removing this validation is that there might still be existing 1.18 clusters that EKS hasn't upgraded to 1.19 yet. It takes some time for them to gradually upgrade unsupported cluster versions after removing support for creating new clusters on that version. 

A safer approach would be to keep this validation and remove it after a few weeks.

cPu1 (commented):
I left a comment about keeping the existing validation for addons but otherwise LGTM.

pkg/actions/addon/addon.go
aclevername: @cPu1 addons are supported on 1.18, so I don't think there is any harm as there isn't any 1.17 clusters left

cPu1: Ah, so this validation could've been removed some time after removing support for 1.17.

cPu1 (approved):

pkg/actions/addon/addon.go
aclevername: yeah!

Issue #5074, opened 2022-04-05
[Bug] Default `vpc-cni` add-on version not found when creating clusters


etungsten:
### What were you trying to accomplish?

I'm trying to create a new IPv6 EKS 1.22 cluster.

### What happened?

Cluster creation fails after trying to set up vpc-cni add-on.
```
2022-04-05 12:10:20 [ℹ]  waiting for CloudFormation stack "eksctl-etung-1-22-ipv6-cluster"
2022-04-05 12:11:20 [ℹ]  waiting for CloudFormation stack "eksctl-etung-1-22-ipv6-cluster"
2022-04-05 12:12:21 [ℹ]  waiting for CloudFormation stack "eksctl-etung-1-22-ipv6-cluster"
2022-04-05 12:14:24 [ℹ]  daemonset "kube-system/aws-node" restarted
2022-04-05 12:14:25 [!]  1 error(s) occurred and cluster hasn't been created properly, you may wish to check CloudFormation console
2022-04-05 12:14:25 [ℹ]  to cleanup resources, run 'eksctl delete cluster --region=us-west-2 --name=etung-1-22-ipv6'
2022-04-05 12:14:25 [✖]  failed to fetch version  for addon vpc-cni: no version(s) found matching "1.10.0" for "vpc-cni"
```

### How to reproduce it?

Create a cluster with the following cluster configuration:
```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig
metadata:
  name: "etung-1-22-ipv6"
  version: "1.22"
  region: "us-west-2"
kubernetesNetworkConfig:
  ipFamily: IPv6
vpc:
  clusterEndpoints:
    publicAccess: true
    privateAccess: true
iam:
  withOIDC: true
addons:
  - name: vpc-cni
  - name: coredns
  - name: kube-proxy
managedNodeGroups:
  - name: br-mng
    amiFamily: Bottlerocket
    availabilityZones: []
    minSize: 0
    desiredCapacity: 0
    maxSize: 1
```
Then ran `eksctl create cluster -f "${cluster_config_file}" --set-kubeconfig-context=false`

**Logs**

**Anything else we need to know?**

Using eksctl version 0.92.0-rc.0, but this happens with other versions and cluster versions as well.
eksctl documentation says the versions for the add-ons are optional and is set to a default version for a given cluster version.
This could also be because I didn't create a worker node as part of the cluster creation and only created an empty managed nodegroup. But the error message indicates otherwise.

[...]

PR #5078, opened 2022-04-06
Use EKS APIs default vpc-cni addon version for ipv6 clusters

aclevername:
### Description
Closes https://github.com/weaveworks/eksctl/issues/5074

We used to set the version of the `vpc-cni` if no version was set when creating ipv6 clusters. https://github.com/weaveworks/eksctl/commit/c095e3b26c7341276f70fd19f1063579247488e9 this was because the default version available by EKS wasn't supporting ipv6, so we had to mandate a minimum version. Now this minimum version isn't available on the EKS api for `1.22`, so we shouldn't default to it, but we should still check the user provided version is greater than it.

green run https://github.com/weaveworks/eksctl-ci/actions/runs/2103683555


integration/tests/ipv6/ipv6_test.go
```diff
@@ -63,23 +63,20 @@ var _ = PDescribe("(Integration) [EKS IPv6 test]", func() {
...
 				},
 				{
-					Name:    api.KubeProxyAddon,
-					Version: "latest",
+					Name: api.KubeProxyAddon,
 				},
 				{
-					Name:    api.CoreDNSAddon,
-					Version: "latest",
+					Name: api.CoreDNSAddon,
```
aclevername, line 79: we should always check the default version works, encase we have to enforce any versions again in the future

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -270,7 +270,6 @@ func (c *ClusterConfig) unsupportedVPCCNIAddonVersion() (bool, error) {
 	for _, addon := range c.Addons {
 		if addon.Name == VPCCNIAddon {
 			if addon.Version == "" {
-				addon.Version = minimumVPCCNIVersionForIPv6
```
Skarlso, line 273: so.... if it's empty we leave it empty?

aclevername: yep. Default value means the EKS API uses the default value. E.g.:
```
 eksctl utils describe-addon-versions --name vpc-cni --kubernetes-version 1.22
2022-04-06 15:22:57 [ℹ]  eksctl version 0.88.0
2022-04-06 15:22:57 [ℹ]  using region us-west-2
2022-04-06 15:22:57 [ℹ]  describing addon versions for addon: vpc-cni
{
  Addons: [{
      AddonName: "vpc-cni",
      AddonVersions: [
        {
          AddonVersion: "v1.10.2-eksbuild.1",
          Architecture: ["amd64","arm64"],
          Compatibilities: [{
              ClusterVersion: "1.22",
              DefaultVersion: false,
              PlatformVersions: ["*"]
            }]
        },
        {
          AddonVersion: "v1.10.1-eksbuild.1",
          Architecture: ["amd64","arm64"],
          Compatibilities: [{
              ClusterVersion: "1.22",
              DefaultVersion: true,
              PlatformVersions: ["*"]
            }]
        },
        {
          AddonVersion: "v1.9.3-eksbuild.1",
          Architecture: ["amd64","arm64"],
          Compatibilities: [{
              ClusterVersion: "1.22",
              DefaultVersion: false,
              PlatformVersions: ["*"]
            }]
        },
        {
```

`v1.10.1-eksbuild.1` would be used

Skarlso (approved):

nikimanoledaki (approved):

PR #5088, opened 2022-04-07
Remove sorting in case the switch is off

Skarlso:
### Description

Don't sort in case the duplication suppression is not enabled.

Closes https://github.com/weaveworks/eksctl/issues/5089

Incoming local crud test run result...

### Checklist
- [ ] Manually tested


aclevername (approved):
is it worth changing it so the sorting logic just mutates a local copy of the identities?

Skarlso:
>is it worth changing it so the sorting logic just mutates a local copy of the identities?

Thought about that. For now, it's fine, since, we don't really integration test that flag, and I don't really think we ever will because it's not worth the resource. So I decided to just leave it. It's a quicker and smaller fix, easier to review too instead of me adding a `copy(dst, src)` shenanigans. :)

cPu1 (approved):

PR #5116, opened 2022-04-12
Handle pod eviction errors correctly

hintofbasil:
### Description

Currently any eviction error causes the draining of a node to stop and a new
node to start draining.  Eviction errors are common, expected occurrences
especially when PDBs are used in the cluster.
By having any error abort the draining of a node we slow down the entire node
draining process as many of the pods further in the list could happily be
drained.

This change separates recoverable and irrecoverable eviction errors and retries
only the recoverable.  Unrecoverable errors fail the entire command.

An important aspect of this is that the `evictPods` function becomes blocking
until a node is drained or the process times out.  This is required as the
current implementation begins draining another node on the first eviction
error.  We would rather keep trying and eventually time out than make a bad
situation worse by draining a new node.

### Checklist
- [x] Manually tested


pkg/drain/evictor/pod.go
```diff
@@ -204,7 +203,7 @@ func (d *Evictor) daemonSetFilter(pod corev1.Pod) PodDeleteStatus {
 		if controllerRef.Name == ignoreDaemonSet.Name {
 			switch ignoreDaemonSet.Namespace {
 			case pod.Namespace, metav1.NamespaceAll:
-				return makePodDeleteStatusWithWarning(false, daemonSetWarning)
+				return makePodDeleteStatusSkip()
```
hintofbasil, line 206: This error is super noisy and adds next to zero value.

There may be a better way to disable it / filter it out but this was the best I found

pkg/drain/nodegroup.go
```diff
@@ -27,6 +29,13 @@ import (
 // retryDelay is how long is slept before retry after an error occurs during drainage
 const retryDelay = 5 * time.Second
 
+var recoverablePodEvictionErrors = [...]string{
```
hintofbasil, line 32: This list may not be exhaustive.  These are the ones I found through manual testing.

hintofbasil:
Logs from the manual testing

```
2022-04-12 19:16:54 [ℹ]  eksctl version 0.93.0-dev+920c2f91.2022-04-12T18:32:04Z
2022-04-12 19:16:54 [ℹ]  using region eu-west-1
2022-04-12 19:16:58 [ℹ]  comparing 2 nodegroups defined in the given config ("/tmp/cellsdev-1-eu-west-1a-1.yaml") against remote state
2022-04-12 19:17:04 [ℹ]  6 nodegroup(s) present in the config file (mixed-instances-12xl-a0adaeb2,mixed-instances-4xl-002a6f2a,mixed-instances-8xl-722a76be,nodes-default-gateway-ingress-5a731ab7,nodes-ingress-3aa9f35a,nodes-public-ingress-24e77938) will be excluded
2022-04-12 19:17:04 [ℹ]  1 nodegroup (mixed-instances-4xl-002a6f2a-test) was included (based on the include/exclude rules)
2022-04-12 19:17:04 [ℹ]  will drain 1 nodegroup(s) in cluster "cellsdev-1-eu-west-1a-1"
2022-04-12 19:17:07 [ℹ]  starting parallel draining, max in-flight of 5
2022-04-12 19:17:08 [ℹ]  cordon node "ip-172-16-1-163.eu-west-1.compute.internal"
2022-04-12 19:17:09 [ℹ]  cordon node "ip-172-16-10-58.eu-west-1.compute.internal"
2022-04-12 19:17:09 [ℹ]  cordon node "ip-172-16-15-44.eu-west-1.compute.internal"
2022-04-12 19:17:09 [ℹ]  cordon node "ip-172-16-21-109.eu-west-1.compute.internal"
2022-04-12 19:17:10 [ℹ]  cordon node "ip-172-16-29-0.eu-west-1.compute.internal"
2022-04-12 19:17:10 [ℹ]  cordon node "ip-172-16-29-49.eu-west-1.compute.internal"
2022-04-12 19:17:10 [ℹ]  cordon node "ip-172-16-30-253.eu-west-1.compute.internal"
2022-04-12 19:17:10 [ℹ]  cordon node "ip-172-16-4-65.eu-west-1.compute.internal"
2022-04-12 19:18:11 [!]  2 pods are unevictable from node ip-172-16-4-65.eu-west-1.compute.internal
2022-04-12 19:19:10 [!]  1 pods are unevictable from node ip-172-16-29-49.eu-west-1.compute.internal
2022-04-12 19:20:16 [!]  1 pods are unevictable from node ip-172-16-29-49.eu-west-1.compute.internal
2022-04-12 19:20:26 [✔]  drained all nodes: [ip-172-16-15-44.eu-west-1.compute.internal ip-172-16-21-109.eu-west-1.compute.internal ip-172-16-30-253.eu-west-1.compute.internal ip-172-16-10-58.eu-west-1.compute.internal ip-172-16-29-0.eu-west-1.compute.internal ip-172-16-29-49.eu-west-1.compute.internal ip-172-16-1-163.eu-west-1.compute.internal ip-172-16-4-65.eu-west-1.compute.internal]
2022-04-12 19:20:26 [ℹ]  will delete 1 nodegroups from cluster "cellsdev-1-eu-west-1a-1"
2022-04-12 19:20:28 [ℹ]  1 task: { 1 task: { delete nodegroup "mixed-instances-4xl-002a6f2a-test" [async] } }
2022-04-12 19:20:28 [ℹ]  will delete stack "eksctl-cellsdev-1-eu-west-1a-1-nodegroup-mixed-instances-4xl-002a6f2a-test"
2022-04-12 19:20:28 [✔]  deleted 1 nodegroup(s) from cluster "cellsdev-1-eu-west-1a-1"
```

pkg/drain/nodegroup.go
```diff
@@ -204,29 +224,53 @@ func (n *NodeGroupDrainer) toggleCordon(cordon bool, nodes *corev1.NodeList) {
...
+			if len(pods) == 0 {
+				return nil
+			}
+			if w := list.Warnings(); w != "" {
+				logger.Warning(w)
+			}
+			// This log message is important but can be noisy.  It's useful to only
+			// update on a node every minute.
+			if time.Now().Sub(previousReportTime) > time.Minute*1 && len(pods) > 0 {
+				logger.Warning("%d pods are unevictable from node %s", len(pods), node)
```
hintofbasil, line 255: This count also includes currently terminating pods.
I think there is value in that - but maybe the wording could be improved.

Skarlso:
I actually noticed this as well! Thank you for doing this. I will check the code out. :)

pkg/drain/nodegroup.go
Skarlso: Hmm, I wonder if we could check more reliably than this, using some Kubernetes `const` for the error message. I'm a bit reluctant to rely on error messages that are subject to change, trimming, extra spaces, etc...

pkg/drain/nodegroup.go
```diff
@@ -235,3 +279,12 @@ func cordonStatus(desired bool) string {
 	}
 	return "uncordon"
 }
+
+func isEvictionErrorRecoverable(err error) bool {
```
Skarlso, line 283: We have a string contains in here somewhere if you want something like this.

But I think we might be able to detect the error a bit better perhaps than relying on the text.

pkg/drain/nodegroup.go
```diff
@@ -143,32 +158,37 @@ func (n *NodeGroupDrainer) Drain() error {
 
 			logger.Debug("already drained: %v", mapToList(drainedNodes.Items()))
 			logger.Debug("will drain: %v", newPendingNodes.List())
-			for i, node := range newPendingNodes.List() {
-				if err := sem.Acquire(ctx, 1); err != nil {
-					logger.Critical("failed to acquire semaphore: %w", err)
-				}
 
-				go func(i int, node string) {
+			errorGroup, _ := errgroup.WithContext(ctx)
```
Skarlso, line 162:
```suggestion
			g, _ := errgroup.WithContext(ctx)
```

pkg/drain/nodegroup.go
```diff
@@ -143,32 +158,37 @@ func (n *NodeGroupDrainer) Drain() error {
...
+			for _, node := range newPendingNodes.List() {
+				node := node
+				errorGroup.Go(func() error {
```
Skarlso, line 165:
```suggestion
				g.Go(func() error {
```

pkg/drain/nodegroup.go
```diff
@@ -143,32 +158,37 @@ func (n *NodeGroupDrainer) Drain() error {
...
+					if err := sem.Acquire(ctx, 1); err != nil {
+						return errors.Wrapf(err, "failed to acquire semaphore")
+					}
 					defer sem.Release(1)
+
+					drainedNodes.Set(node, nil)
 					logger.Debug("starting drain of node %s", node)
-					pending, err := n.evictPods(node)
+					err := n.evictPods(ctx, node)
 					if err != nil {
```
Skarlso, line 174:
```suggestion
					if err := n.evictPods(ctx, node); err != nil {
```

pkg/drain/nodegroup.go
```diff
@@ -143,32 +158,37 @@ func (n *NodeGroupDrainer) Drain() error {
...
 						logger.Debug("waiting for %.0f seconds before draining next node", n.nodeDrainWaitPeriod.Seconds())
 						time.Sleep(n.nodeDrainWaitPeriod)
 					}
-				}(i, node)
+					return nil
+				})
 			}
+			// We need to loop even if this is an error to check whether the error was a
+			// context timeout or something else.  This lets us log timout errors consistently
+			evictErr = errorGroup.Wait()
```
Skarlso, line 191:
```suggestion
			evictErr = g.Wait()
```

pkg/drain/nodegroup.go
```diff
@@ -204,29 +224,53 @@ func (n *NodeGroupDrainer) toggleCordon(cordon bool, nodes *corev1.NodeList) {
...
+				previousReportTime = time.Now()
+			}
+			logger.Debug("%d pods to be evicted from %s", pods, node)
+			failedEvictions := false
+			for _, pod := range pods {
+				if err := n.evictor.EvictOrDeletePod(pod); err != nil {
+					if !isEvictionErrorRecoverable(err) {
+						return errors.Wrapf(err, "unrecoverable error evicting pod: %s/%s", pod.Namespace, pod.Name)
+					}
+					logger.Debug("pod eviction failed recoverably: %q", err)
```
Skarlso, line 265:
```suggestion
					logger.Debug("recoverable pod eviction failure: %q", err)
```

pkg/drain/nodegroup.go
```diff
@@ -204,29 +224,53 @@ func (n *NodeGroupDrainer) toggleCordon(cordon bool, nodes *corev1.NodeList) {
...
+			if len(errs) > 0 {
+				return fmt.Errorf("errs: %v", errs) // TODO: improve formatting
+			}
+			if list == nil {
+				return nil
+			}
+			pods := list.Pods()
+			if len(pods) == 0 {
+				return nil
+			}
```
Skarlso, line 248:
```suggestion
			if list == nil || len(list.Pods()) == 0 {
				return nil
			}
```

pkg/drain/nodegroup.go
Skarlso: Shouldn't this happen only in case of a failure, though? Why would there be a warning if this is still working through some pods?

pkg/drain/nodegroup.go
Skarlso:
```
	"TooManyRequests",
	"NotFound",
```

These two are definitely meta error `	StatusReasonNotFound StatusReason = "NotFound"` and `StatusReasonTooManyRequests StatusReason = "TooManyRequests"`.

The other two appear to be some aws specific things. How did you encounter them? Can you print the whole error so we can see what interface it implements at that point?

pkg/drain/nodegroup.go
Skarlso: We should also use the context here so `Wait` can cancel it when it's done so there are no stragglers.

pkg/drain/nodegroup.go
hintofbasil: I've updated this to use the checker functions built into apimachinery.

I've re-ran a manual test and it works perfectly.

pkg/drain/nodegroup.go
hintofbasil: `GetPodsForEviction` returns terminating pods.  This causes the for loop to keep looping allowing this message to log.

pkg/drain/nodegroup.go
hintofbasil: This has been replaced with more intelligent error checking logic.

pkg/drain/nodegroup.go
Skarlso: Ah okay.

pkg/drain/nodegroup.go
```diff
@@ -36,16 +45,17 @@ type Evictor interface {
...
+	ng                    eks.KubeNodeGroup
+	waitTimeout           time.Duration
+	nodeDrainWaitPeriod   time.Duration
+	podEvictionWaitPeriod time.Duration
+	undo                  bool
+	parallel              int
 }
 
-func NewNodeGroupDrainer(clientSet kubernetes.Interface, ng eks.KubeNodeGroup, waitTimeout, maxGracePeriod, nodeDrainWaitPeriod time.Duration, undo, disableEviction bool, parallel int) NodeGroupDrainer {
+func NewNodeGroupDrainer(clientSet kubernetes.Interface, ng eks.KubeNodeGroup, waitTimeout, maxGracePeriod, nodeDrainWaitPeriod time.Duration, podEvictionWaitPeriod time.Duration, undo, disableEviction bool, parallel int) NodeGroupDrainer {
```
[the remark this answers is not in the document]
hintofbasil: Good catch.  Added this in as a command line argument.

hintofbasil:
I can see that all the CI checks have passed.  Let me know if you think this PR is ready and I can rebase it into a single commit.

Skarlso:
Hi @hintofbasil. Will do. Please bear with us for a while as the whole team is off for Easter. :)

Himangini (approved):
LGTM 👍🏻 nicely done

pkg/actions/cluster/delete_test.go
```diff
@@ -68,7 +70,7 @@ var _ = Describe("DrainAllNodeGroups", func() {
 					vpcCniDeleterCalled++
 				}
 
-				err := cluster.DrainAllNodeGroups(cfg, ctl, fakeClientSet, nodeGroupStacks, false, 1, mockedDrainer, vpcCniDeleter)
+				err := cluster.DrainAllNodeGroups(cfg, ctl, fakeClientSet, nodeGroupStacks, false, 1, mockedDrainer, vpcCniDeleter, time.Second*0)
```
Skarlso, line 73: It's enough to pass in 0. :)

pkg/drain/nodegroup.go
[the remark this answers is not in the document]
Skarlso: Oh, never mind. :D I see you also changed the code to pass it in now. Good stuff. :)

pkg/drain/evictor/pod.go
Skarlso: Yeah, that error came from some copied over eviction code from 3 years ago. I don't think we need it... :D

Skarlso:
@hintofbasil Thank you for your contribution! :)

hintofbasil:
Thank you for the reviews!

Hopefully there will be more contributions in the future.

PR #5118, opened 2022-04-13
Revert "Adding containerd as a runtime parameter option to windows bootstrapping"

Skarlso:
Reverts weaveworks/eksctl#5083

It looks like `containerd` support is not working yet with Windows. There are several problems and errors when running a windows workload on a `containerd` enabled cluster.

Some of these are:

```
Warning  Failed             9s    kubelet                  Failed to pull image "mcr.microsoft.com/windows/servercore:20H2": rpc error: code = Unknown desc = failed to pull and unpack image "mcr.microsoft.com/windows/servercore:20H2": failed to extract layer sha25 │
│ 6:d00f903a379eeaefcb3d589e8fefd8c9b1fe5f7caad260511d6ebb99411b0289: hcsshim::ImportLayer failed in Win32: The process cannot access the file because it is being used by another process. (0x20): unknown
```

and

```
Warning  FailedCreatePodSandBox  0s                     kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "6d1190746ed84cfb679d2fbf0d4cb9565fc66b467f71f304ab2f8c738d86c4e5": plugin type="vpc │
│ -shared-eni" name="vpc" failed (add): hnsCall failed in Win32: An address provided is invalid or reserved. (0x803b002f) 
```

The error occurs during the image pull phase. We will try to escalate to AWS for further analysis on their side. `eksctl` is only passing in a parameter, so I'm not sure what the problem could be if it were on our side.


cPu1 (approved):

Himangini (approved):
👍🏻

PR #5187, opened 2022-04-29
Support AWS Local Zones

### LINKED ISSUE #4734: Cluster Creation in a Region with Local Zones
Story comment below https://github.com/weaveworks/eksctl/issues/4734#issuecomment-1068961392

---

When you try to create a cluster in an AWS Region that has one or more local zones enabled, and you are also creating the VPC as part of the cluster creation process, then eksctl passes the subnets in a local zone to the cluster. Unlike regional subnets, EKS can't place network interfaces into Local Zone subnets. As a result, the cluster creation process fails.

AWS customers are trying to use local zones in EKS clusters and are trying to create EKS clusters that include worker nodes in local zones. Currently, they have to create the VPC with regional and local zone subnets as a separate step and then explicitly specify only regional subnets in their configuration file while creating an EKS cluster. They would like to do this with a one step process.

cPu1:
### Description

Adds support for AWS Local Zones based on [this proposal](https://github.com/weaveworks/eksctl/issues/4734#issuecomment-1103809860). 

Closes #4734 

### Checklist
- [x] Manually tested


Himangini (approved):
LGTM 👍🏻

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -718,6 +733,10 @@ func ValidateNodeGroup(i int, ng *NodeGroup) error {
 		}
 	}
 
+	if len(ng.LocalZones) > 0 && len(ng.AvailabilityZones) > 0 {
+		return errors.New("cannot specify both localZones and availabilityZones")
```
Skarlso, line 737: How come? Can't you have a hybrid cluster with specific workloads running in a localzone vs some other nodegroup which defines availability zones that will take core a different workload?

Skarlso: Oh wait, never mind. This is the same nodegroup.

cPu1: >Can't you have a hybrid cluster with specific workloads running in a localzone vs some other nodegroup which defines availability zones that will take core a different workload?

You can. As you noticed, this validation is to disallow nodegroups having both local zones and AZs defined.

Skarlso (approved):
Well done! Nice work!

Issue #5015, opened 2022-03-29
Backfill unit tests for `eksctl create cluster`


nikimanoledaki:
## Context
The test coverage for `eksctl create cluster`, one of the most important commands of eksctl, is quite low. This blocks any refactoring efforts around this command because it can lead to regressions and bugs. This ties back to both the testing spike as well as the validation spike: we need solid unit tests before refactoring things such as how we do the validation for this command.

## Scope
- The aim here is to backfill unit tests for it, with a focus on the following: https://github.com/weaveworks/eksctl/blob/main/pkg/ctl/create/cluster.go
- If possible, move any non-CLI-related logic from `pkg/ctl/cluster/cluster.go` to `pkg/actions/cluster/create.go` to follow the pattern of other commands. It might be easiest to add the unit tests and do the code migration at the same time. 
- This might require refactoring commands that are untested due to the Provider being too powerful.

Skarlso:
This will be much more than a 3.

Skarlso:
Hmmm... orrrr..... we could use a test environment flag which would be a URL that overrides all of the services. So we could test with a `httptest` server and still test create cluster to check if the right calls are made.

Skarlso:
The httptest server works nicely. Sadly, not all of the endpoints go through v2 yet. So this is blocked until that is completely done.

Skarlso:
Last blocking piece https://github.com/weaveworks/eksctl/issues/4884

PR #5259, opened 2022-05-16
Refactoring the provider to unit test create cluster

Skarlso:
### Description

Ground work for https://github.com/weaveworks/eksctl/issues/5015.

#### What happens in this PR?

1. Renamed `Provider` to `AWSProvider` since it provides AWS functionality
2. ClusterProvider is no longer a KubeProvider. It has been extracted into a separate part so it can be mocked more easily
3. doClusterCreate now gets the ctl passed in from the outside so we can mock it
4. Several small refactors and test changes for mock calls

Integration test run https://github.com/weaveworks/eksctl-ci/actions/runs/2330469236

_edit_: New test run https://github.com/weaveworks/eksctl-ci/actions/runs/2332032290

### Checklist
- [ ] Manually tested


pkg/ctl/create/cluster.go
```diff
@@ -135,11 +140,6 @@ func doCreateCluster(cmd *cmdutils.Cmd, ngFilter *filter.NodeGroupFilter, params
 		}()
 	}
 
-	ctl, err := cmd.NewCtl()
```
Skarlso, line 138: The above logger initialisation breaks now that newCtl happens before this. NewCtl contains a `once` for the logger. I'm working on the fix.

Skarlso: Fixed.

Skarlso:
Integration tests passed except for a single test which timed out on TLS. That appears to be a fluke from AWS though.

pkg/eks/api.go
```diff
@@ -48,18 +46,26 @@ import (
 // ClusterProvider stores information about the cluster
 type ClusterProvider struct {
 	// core fields used for config and AWS APIs
-	Provider api.ClusterProvider
+	AWSProvider api.ClusterProvider
+	// KubernetesProvider offers helper methods to handle Kubernetes operations
+	KubernetesProvider KubeProvider
```
cPu1, line 51: If we embed this field, we wouldn't have to reference this field via `ctl` everywhere and it would also make the diff smaller. 
```suggestion
	KubeProvider
```

cPu1 (approved):
Just one comment about considering embedding `KubeProvider`, but otherwise LGTM.

My [deadline-based Context PR](https://github.com/weaveworks/eksctl/pull/5282) will result in conflicts, so I'll wait for this to be merged before merging mine.

pkg/eks/api.go
Skarlso: Hmmm, not a bad idea. I guess it would result in a smaller diff since I don't have to call it explicitly everywhere. 🤔

Skarlso: Hm, not that much :) The AWSProvider rename touched more. :) But I've done it anyways.

Himangini (approved):
LGTM 👍🏻

PR #5303, opened 2022-05-23
Identify watch.Error event types received from ResultChan

### LINKED ISSUE #5108: [Bug] `event = watch.Event{Type:"", Object:runtime.Object(nil)}` when waiting for nodegroup
From time to time, it appears that the waiter loses the node context:

https://github.com/weaveworks/eksctl-ci/runs/5982018686?check_suite_focus=true

It might be that the node disappears or the waiting times out because of an AWS issue. In that case, we shouldn't keep logging the empty event but fail immediately. Otherwise, we'll fill up the log to 1.800.000+ lines!

The bug is that we don't check if the Watcher closed the channel or not due to an internal error. So we read on a closed channel which returns the default empty value of the event immediately. This is why we get so many log messages. :)

cPu1:
### Description

This changelist identifies `watch.Error` event types and returns this as an error. This would help debug issues with the node watcher code failing. 

Closes #5108

### Checklist
- [ ] Manually tested


Himangini (approved):
👍🏻

pkg/eks/client.go
```diff
@@ -169,6 +170,15 @@ func WaitForNodes(ctx context.Context, clientSet kubernetes.Interface, ng KubeNo
 				logger.Debug("the watcher channel was closed... stop processing events from it")
 				return fmt.Errorf("the watcher channel for the nodes was closed by Kubernetes due to an unknown error")
 			}
+			if event.Type == watch.Error {
```
Skarlso, line 173: Yep I tried this. Sadly since we debug log the event before this in every case it actually had a nil event so this didn't do anything. But let's hope that there will be a time when this changes. 😊

Skarlso (approved):

pkg/eks/client.go
Skarlso: Nope, it wasn't. I think there is a log about that at some point in one of the runs if you can see that. ¯\_(ツ)_/¯ 

But that doesn't matter, because eventually it might and THEN we will have what went wrong with this. :)

PR #5314, opened 2022-05-24
Backfill create cluster unittests

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/5015

### Checklist
- [x] Manually tested


Skarlso:
test run: https://github.com/weaveworks/eksctl-ci/runs/6572778661?check_suite_focus=true

Only ipv6 failed. It doesn't look related, but I'll investigate and run it locally to be sure.

Skarlso:
Yay, all integration tests passed! :) https://github.com/weaveworks/eksctl-ci/actions/runs/2377602031

🟢 🟢

pkg/cfn/manager/api.go
```diff
@@ -178,8 +178,23 @@ func (c *StackCollection) CreateStack(ctx context.Context, stackName string, res
...
+		StackName: &stackName,
+	})
+	if err != nil && !IsStackDoesNotExistError(err) {
+		return err
+	}
+	if stack != nil {
+		switch stack.StackStatus {
+		case types.StackStatusCreateComplete, types.StackStatusDeleteComplete:
+			close(errCh) // no need to wait on this task anymore
+			return nil
```
nikimanoledaki, line 192: instead of `return nil`, should this not be a descriptive error about why it's returning early, which would be passed to the `errCh`? maybe there's something that I'm not seeing 🤔

pkg/eks/client.go
```diff
@@ -126,6 +128,38 @@ func (c *KubernetesProvider) ServerVersion(rawClient *kubewrapper.RawClient) (st
 	return rawClient.ServerVersion()
 }
 
+// WaitForControlPlane waits till the control plane is ready
+func (c *KubernetesProvider) WaitForControlPlane(meta *api.ClusterMeta, clientSet *kubewrapper.RawClient, waitTimeout time.Duration) error {
+	successCount := 0
+	operation := func() (bool, error) {
+		_, err := c.ServerVersion(clientSet)
+		if err == nil {
+			if successCount >= 5 {
```
nikimanoledaki, line 137: (I know this code was just moved but) 5 retries seems a bit arbitrary when the current default timeout is `25m` 🤔

pkg/cfn/manager/api.go
Skarlso: There is no error here. We got what we were looking for.

pkg/eks/client.go
Skarlso: ¯\\\_(ツ)\_/¯ :D I don't make the retry rules hahaha

pkg/cfn/manager/api.go
Skarlso: To be more clear. This check if the stack has the desired status, which is `CREATE_COMPLETE` or `DELETE_COMPLETE` which the below check will be looking for as well. This is to avoid having to wait a minute until the next check if the stack already has the state we want anyways.

nikimanoledaki: I saw that, but at this point the stack has not been created yet. If a stack is found with `CREATE_COMPLETE`, couldn't it be a totally different one from a different operation?

The check below checks that the stack that was just created by `c.createStackRequest` was properly created, which is different.

pkg/cfn/manager/api.go
```diff
@@ -178,8 +178,23 @@ func (c *StackCollection) CreateStack(ctx context.Context, stackName string, res
...
+		}
```
cPu1, line 193: This will result in a bug. If a cluster with the same name already exists, the current behaviour is to throw an error saying that it already exists. But this will skip that check and attempt to create a Kubernetes client without having the values for the endpoint and CA data, failing with a cryptic error. 

(I've had this comment typed out for hours but forgot to hit submit).

pkg/cfn/manager/api.go
```diff
@@ -178,8 +178,23 @@ func (c *StackCollection) CreateStack(ctx context.Context, stackName string, res
...
+	}
```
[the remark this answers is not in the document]
Skarlso: That would happen at the createStackCall a few calls back. This is in the Wait part of the chain.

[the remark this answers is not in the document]
Skarlso: ah, yes, it appears so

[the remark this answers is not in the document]
Skarlso: I removed this call, but for the record, sadly, because of how the wait works:
```go
		stack, err := waiter.WaitForStack(ctx, c.cloudformationAPI, *stack.StackId, *stack.StackName, func(attempts int) time.Duration {
			// Wait 30s for the first two requests, and 1m for subsequent requests.
			if attempts <= 2 {
				return 30 * time.Second
			}
			return 1 * time.Minute
		})
```

This will make the unit test run for at least 30 seconds. ¯\\\_(ツ)\_/¯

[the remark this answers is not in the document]
nikimanoledaki: Could skip the initial 30s wait and add to the retries.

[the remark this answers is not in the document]
Skarlso: Technically it doesn't make sense since we are waiting on an API call, which will not even be registered in a second. :D
Also, if you mean attempts, that's the parameter coming from WaitForStack. :) Or did you mean to track the retries externally?

cPu1 (approved):

Issue #5277, opened 2022-05-17
[Feature] nvidia device plugin not scheduled on GPU nodes if they have additional taints


yuvipanda:
### What were you trying to accomplish?

I'm trying to setup an unmanaged GPU nodegroup with some additional taints, and have eksctl automatically install the nvidia device plugin daemonset correctly.

### What happened?

Because I've additional taints on my nodegroup and the nvidia-device-plugin daemonset doesn't have tolerations for those, it's never scheduled on the GPU nodegroup.

### How to reproduce it?

Here's the appropriate nodegroup definition:

```
      {
         "availabilityZones": [
            "us-west-2b"
         ],
         "desiredCapacity": 0,
         "iam": {
            "withAddonPolicies": {
               "autoScaler": true
            }
         },
         "instanceType": "p2.xlarge",
         "labels": {
            "hub.jupyter.org/node-purpose": "user",
            "k8s.dask.org/node-purpose": "scheduler",
            "node.kubernetes.io/instance-type": "p2.xlarge"
         },
         "maxSize": 500,
         "minSize": 0,
         "name": "nb-p2-xlarge",
         "ssh": {
            "publicKeyPath": "ssh-keys/uwhackweeks.key.pub"
         },
         "tags": {
            "k8s.io/cluster-autoscaler/node-template/label/hub.jupyter.org/node-purpose": "user",
            "k8s.io/cluster-autoscaler/node-template/label/k8s.dask.org/node-purpose": "scheduler",
            "k8s.io/cluster-autoscaler/node-template/label/node.kubernetes.io/instance-type": "p2.xlarge",
            "k8s.io/cluster-autoscaler/node-template/taint/hub.jupyter.org/dedicated": "user:NoSchedule",
            "k8s.io/cluster-autoscaler/node-template/taint/hub.jupyter.org_dedicated": "user:NoSchedule"
         },
         "taints": {
            "hub.jupyter.org/dedicated": "user:NoSchedule",
            "hub.jupyter.org_dedicated": "user:NoSchedule"
         },
         "volumeSize": 80
      },
```
The created nvidia-device-plugin daemonset only has the following tolerations:

```yaml
      - key: CriticalAddonsOnly
        operator: Exists
      - key: nvidia.com/gpu
        operator: Exists
        effect: NoSchedule
```

So it never gets scheduled on my GPU nodes.

[...]

PR #5345, opened 2022-05-30
Add tolerations for taints on NVIDIA specific node groups

Skarlso:
### Description

Closes https://github.com/weaveworks/eksctl/issues/5277

I believe this will do the trick, but I have to manually test it first. Is this right, @cPu1? I'm a bit unfamiliar with this part of the code. :)

## TODO:

- [x] Manual run

### Checklist
- [x] Manually tested


Skarlso:
With the following config file:

```yaml
kind: ClusterConfig

metadata:
  name: gb-test-cluster-1
  region: us-west-2
  version: '1.22'

nodeGroups:
  - name: ng-1
    minSize: 1
    maxSize: 2
    desiredCapacity: 1
    instanceType: p2.xlarge
    taints:
      feaster: "true:NoSchedule" 

```

Achieved this on the nvidia pod:

```
Tolerations:                 CriticalAddonsOnly op=Exists
                             feaster=true
                             node.kubernetes.io/disk-pressure:NoSchedule op=Exists
                             node.kubernetes.io/memory-pressure:NoSchedule op=Exists
                             node.kubernetes.io/not-ready:NoExecute op=Exists
                             node.kubernetes.io/pid-pressure:NoSchedule op=Exists
                             node.kubernetes.io/unreachable:NoExecute op=Exists
                             node.kubernetes.io/unschedulable:NoSchedule op=Exists
                             nvidia.com/gpu:NoSchedule op=Exists

```

pkg/addons/device_plugin.go
```diff
@@ -197,11 +209,59 @@ func (n *NvidiaDevicePlugin) Deploy() error {
...
+	}
+	return nil
+}
+
 // A EFADevicePlugin deploys the EFA Device Plugin to a cluster
 type EFADevicePlugin struct {
 	rawClient kubernetes.RawClientInterface
 	region    string
 	planMode  bool
+	spec      *api.ClusterConfig
```
cPu1, line 264: We don't need to add this field since it isn't used.

pkg/addons/device_plugin.go
```diff
@@ -216,17 +276,22 @@ func (n *EFADevicePlugin) Manifest() []byte {
...
-func NewEFADevicePlugin(rawClient kubernetes.RawClientInterface, region string, planMode bool) DevicePlugin {
+func NewEFADevicePlugin(rawClient kubernetes.RawClientInterface, region string, planMode bool, spec *api.ClusterConfig) DevicePlugin {
 	return &EFADevicePlugin{
-		rawClient,
-		region,
-		planMode,
+		rawClient: rawClient,
+		region:    region,
+		planMode:  planMode,
+		spec:      spec,
```
cPu1, line 294: Same as above, no need to add this field.

pkg/addons/device_plugin.go
Skarlso: So I was thinking about this. I only added it because of consistency, but do you think we should actually do this for other plugins as well? Won't they eventually face the same issue?

cPu1: Right, other plugins will also face the same issue, but it might be out of scope for this PR. If we're adding this to other plugins, I'd say we either add the full implementation or leave them as-is.

Skarlso: True that.

cPu1 (approved):

Himangini (commented):
>  Extend unit tests

I can't see this in the files diff, where is this added? 🤔

Skarlso:
Right, that didn't happen because there aren't any. :D

Himangini (approved):
👍🏻 
At some point, we should add some tickets to improve testing around this and taints in general 💡

PR #5566, opened 2022-08-02
Changed albIngress to awsLoadBalancerController in usage iam-policies example

akhal3d96:
Changed albIngress to awsLoadBalancerController in usage iam-policies example because it's deprecated


Himangini (approved):
thanks @akhal3d96 Nice spot 🎉

cPu1 (approved):

PR #5680, opened 2022-09-07
Stop using 's' as indicator that a method returns a list

### LINKED ISSUE #5219: [Task] Stop using `s` as an indicator that we are returning a list
For example:

```go
	DescribeNodeGroupStack(ctx context.Context, nodeGroupName string) (*Stack, error)
	DescribeNodeGroupStacks(ctx context.Context) ([]*Stack, error)
```

This is super annoying and actually caused a bug to happen: https://github.com/weaveworks/eksctl/issues/5217

It's because our unit tests were mocking the wrong function since the author missed the `s`:

```go
fakeStackManager.DescribeNodeGroupStackReturns(nil, fmt.Errorf("error describing cloudformation stack"))
```

This should have been

```
fakeStackManager.DescribeNodeGroupStacksReturns(nil, fmt.Errorf("error describing cloudformation stack"))
```

It would have been more visible if the extension would say `List` instead of an additional `s`.

The test is passing because the mock was happily providing a canned `nil, nil` as a return.

This also suggests that we should more often verify if the mock was called with the right parameters.

TiberiuGC:
### Description

We are having a couple of methods that return either some instance or a list of those instances, which are very similar in name. In fact, the one that returns a list ends in an additional `s` (e.g. `DescribeNodeGroupStack` and `DescribeNodeGroupStacks`). This means that using and testing these functions can get error prone in particular scenarios (closes #5219). We've decided to replace the `s` with `List` for such cases. 

### Checklist
- [ ] Manually tested


pkg/cfn/manager/nodegroup.go
```diff
@@ -103,8 +103,8 @@ func (c *StackCollection) propagateManagedNodeGroupTagsToASGTask(ctx context.Con
 }
 
 // DescribeNodeGroupStacks calls DescribeStacks and filters out nodegroups
```
Himangini, line 105:
```suggestion
// DescribeNodeGroupStackList calls DescribeStackList and filters out nodegroups
```
💡

pkg/cfn/manager/cluster.go
```diff
@@ -43,7 +43,7 @@ func (c *StackCollection) createClusterTask(ctx context.Context, errs chan error
 
 // DescribeClusterStack calls DescribeStacks and filters out cluster stack
```
Himangini, line 44:
```suggestion
// DescribeClusterStack calls DescribeStackList and filters out cluster stack
```

pkg/cfn/manager/api.go
```diff
@@ -619,7 +619,7 @@ func fmtStacksRegexForCluster(name string) string {
 }
 
 // DescribeStacks describes the existing stacks
```
Himangini, line 621:
```suggestion
// DescribeStackList describes the existing stacks
```

Himangini (commented):
small nits :D

Himangini (approved):
LGTM 👍🏻

Himangini (approved):
👍🏻

cPu1 (changes requested):
`DescribeStackList` does not read well and implies that `StackList` is the resource that needs to be described, but there is no such resource. `ListStacks` is more descriptive IMO, but there's already a method by that name so we should rename the existing `ListStacks` method to something else, e.g., `ListStacksWithStatuses`.

pkg/ctl/cmdutils/filter/nodegroup_filter.go
```diff
@@ -93,7 +94,8 @@ func (f *NodeGroupFilter) SetOnlyLocal(ctx context.Context, eksAPI awsapi.EKS, l
 
 // SetOnlyRemote uses StackLister to list existing nodegroup stacks and configures
 // the filter to exclude nodegroups already defined in the config file. It will include the
-//  nodegroups that exist in the cluster but not in the config
+//
```
cPu1, line 97: Redundant line.

cPu1 (approved):

PR #5797, opened 2022-10-18
Bump nvidia-device-plugin to 0.12.3

yuvipanda:
### Description
0.9 is a bit over 18 months old at this point. Matching versions of drivers, CUDA and packages 
can be a bit difficult with older versions (see, for 
example https://github.com/pangeo-data/pangeo-docker-images/issues/390).

Tries to match https://github.com/weaveworks/eksctl/pull/3696.

### Checklist
- [x] Manually tested


yuvipanda:
I've tested that a device-plugin with this manifest works. Haven't tested running it via `eksctl` itself though.

Himangini:
> I've tested that a device-plugin with this manifest works. Haven't tested running it via `eksctl` itself though.

Would you be able to test this with eksctl? I just glanced over the linked old PR and looks like this needs a bit more changes in `addons/assets.go`?

PR #6023, opened 2022-12-05
Port coredns lameduck fix to EKS 1.24

pnmatich:
### Description

In this MR, I added the a lameduck fix to the coredns 1.24 file. In [MR #5695](https://github.com/weaveworks/eksctl/pull/5695), this same fix that was added to the coredns 1.20, 1.21, 1.22, and 1.23 files. 

This lameduck fix is important, because it allows coredns pods to shutdown gracefully. This is particularly important when coredns is being auto-scaled.

### Checklist
- [ ] Manually tested


pnmatich:
Can a maintainer approve tests for this :pray: 

Anything else I need to do here?

pnmatich:
I'm not sure how to add a label for this. I don't think I have permissions to do that. In my opinion, it should be `kind/bug`.

Himangini (approved):
LGTM 👍🏻

wind0r:
Sorry for asking but when can we expect this merged?
This currently is blocking our Kubernetes update.

gohmc:
Nice patch, but remind that the changes made by update-coredns does not consider add-on configuration values, e.g nodeSelector will be lost, this is no good for cluster that runs coredns in specific node group. The current image tag _must_ match with the [tag in the patch](https://github.com/weaveworks/eksctl/blob/main/pkg/addons/default/assets/coredns-1.24.json#L173)? What [happen](https://github.com/weaveworks/eksctl/blob/main/pkg/addons/default/assets/coredns-1.24.json#L110) if the addon was scheduled to run on Fargate? Note that when EKS update the add-on, the add-on spec can be overwritten, e.g change to readinessProbe will be reverted to /health.

Possibly, a more stable fix is update-coredns will also change the add-on from EKS managed to self-managed, so that this operation will not conflict with EKS managed add-on and server side apply.

Issue #5698, opened 2022-09-14
[Feature] Clean up `write-kubeconfig` lock file after command finishes


jpmcb:
### What feature/behavior/change do you want?

Currently, when a kubeconfig is written using the `write-kubeconfig` util, the lock file is left over. Here's an example where I write a kubeconfig to my local directory tracked by git:

```
❯ eksctl utils write-kubeconfig --cluster beautiful-creature-1663175112 --kubeconfig ./my-kubeconfig.yaml

❯ git status

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        my-kubeconfig.yaml
        my-kubeconfig.yaml.eksctl.lock
```

Looking over the code, looks like this is needed so `eksctl` doesn't attempt multiple writes at the same time to the default kubeconfig. But it'd be great if `eksctl` also cleaned up this file so it's not left laying around.

### Why do you want this feature?
Generally, it makes sense to me that the "behind the scenes" things are cleaned up and that the user doesn't have unexpected files from some operation.

Himangini:
thanks for reporting this @jpmcb We'll review and attempt to resolve this soon. 👍🏻

PR #6030, opened 2022-12-09
Write kubeconfig lock file to tmp so that it gets automatically cleaned up

TiberiuGC:
### Description

Closes https://github.com/weaveworks/eksctl/issues/5698

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/utils/kubeconfig/kubeconfig.go
+++ b/pkg/utils/kubeconfig/kubeconfig.go
@@ -308,20 +308,23 @@ func getAWSIAMAuthenticatorVersion() (string, error) {
 	return parsedVersion.Version, nil
 }
 
-/* KubectlVersionFormat is the format used by kubectl version --format=json, example output:
-{
-  "clientVersion": {
-    "major": "1",
-    "minor": "23",
-    "gitVersion": "v1.23.6",
-    "gitCommit": "ad3338546da947756e8a88aa6822e9c11e7eac22",
-    "gitTreeState": "archive",
-    "buildDate": "2022-04-29T06:39:16Z",
-    "goVersion": "go1.18.1",
-    "compiler": "gc",
-    "platform": "linux/amd64"
-  }
-} */
+/*
+KubectlVersionFormat is the format used by kubectl version --format=json, example output:
+
+	{
+	  "clientVersion": {
+	    "major": "1",
+	    "minor": "23",
+	    "gitVersion": "v1.23.6",
+	    "gitCommit": "ad3338546da947756e8a88aa6822e9c11e7eac22",
+	    "gitTreeState": "archive",
+	    "buildDate": "2022-04-29T06:39:16Z",
+	    "goVersion": "go1.18.1",
+	    "compiler": "gc",
+	    "platform": "linux/amd64"
+	  }
+	}
+*/
 type KubectlVersionData struct {
 	Version string `json:"gitVersion"`
 }
@@ -344,7 +347,7 @@ func getKubectlVersion() string {
 }
 
 func lockFileName(filePath string) string {
-	return filePath + ".eksctl.lock"
+	return filepath.Join(os.TempDir(), fmt.Sprintf("%s.eksctl.lock", filepath.Base(filePath)))
 }
 
 // ensureDirectory should probably be handled in flock
--- a/pkg/utils/kubeconfig/kubeconfig.go
+++ b/pkg/utils/kubeconfig/kubeconfig.go
@@ -347,7 +347,7 @@ func getKubectlVersion() string {
 }
 
 func lockFileName(filePath string) string {
-	return filepath.Join(os.TempDir(), fmt.Sprintf("%s.eksctl.lock", filepath.Base(filePath)))
+	return filepath.Join(os.TempDir(), fmt.Sprintf("%x.eksctl.lock", utils.FnvHash(filePath)))
 }
 
 // ensureDirectory should probably be handled in flock
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"hash/fnv"
 	"regexp"
 	"strings"
 
@@ -9,7 +10,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-var matchFirstCap = regexp.MustCompile("([0-9]+|[A-Z])")
+var (
+	matchFirstCap = regexp.MustCompile("([0-9]+|[A-Z])")
+	fnvHash       = fnv.New32a()
+)
 
 // ToKebabCase turns a CamelCase string into a kebab-case string
 func ToKebabCase(str string) string {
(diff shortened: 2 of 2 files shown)


stmcginnis (commented):
Looks like some of the `Expect` calls can cause an issue by being in goroutines:

```
/home/runner/work/eksctl/eksctl/pkg/utils/kubeconfig/kubeconfig_test.go:353
  When you, or your assertion library, calls Ginkgo's Fail(),
  Ginkgo panics to prevent subsequent assertions from running.

  Normally Ginkgo rescues this panic so you shouldn't see it.

  However, if you make an assertion in a goroutine, Ginkgo can't capture the
  panic.
  To circumvent this, you should call

  	defer GinkgoRecover()

  at the top of the goroutine that caused this panic.
```

Due to the panic it is not printing out what `err` is. I wonder if there is a concurrency issue with this. I've pulled this down and run locally, and I didn't hit the error (data sample of 1).

It might be a good idea to make the lock file deletion best effort. In other words, just have something like:

```go
_ = os.Remove(fl.Path())
```

And just ignore if there were errors.

stmcginnis:
OK, sample size of 2 did the trick. On my second run it did hit this error.

After adding the recommended defer for Ginko and running a few more times, I was able to hit this again and get some more details:

```
• [FAILED] [0.145 seconds]
Kubeconfig [It] safely handles concurrent read-modify-write operations
/home/stmcg/src/weaveworks/eksctl/pkg/utils/kubeconfig/kubeconfig_test.go:323

  [FAILED] Expected
      <*errors.withStack | 0x4000cb4648>: {
          error: <*errors.withMessage | 0x4001251cc0>{
              cause: <*fs.PathError | 0x4000e8fb60>{
                  Op: "open",
                  Path: "/tmp/3479146172.lock",
                  Err: <syscall.Errno>0x11,
              },
              msg: "unable to modify kubeconfig /tmp/3479146172",
          },
          stack: [0xbec924, 0xfb43bc, 0x77f04],
      }
  to be nil
  In [It] at: /home/stmcg/src/weaveworks/eksctl/pkg/utils/kubeconfig/kubeconfig_test.go:361
```

I wonder if it would be better to just do the cleanup [in the unlock](https://github.com/weaveworks/eksctl/blob/f9792c596a974259aef08123e2c913eb11dc79a9/pkg/utils/kubeconfig/kubeconfig.go#L376). The lock file really isn't needed after that point, and it serves no purpose. Also, nice symmetry since [`lockConfigFile` creates the file](https://github.com/weaveworks/eksctl/blob/f9792c596a974259aef08123e2c913eb11dc79a9/pkg/utils/kubeconfig/kubeconfig.go#L367).

This failure doesn't look related to that though. It actually looks to me like there is an existing bug in concurrent kubeconfig updates. The code creates two separate `Flock` instances for the same file path. Internally to the Flock packages implementation, they lock on the individual instance's mutex, which wouldn't prevent another instance from running the same thing, then [ignore if grabbing the file lock on the actual lock file fails](https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock_unix.go#L57).  So I think the failure is an existing issue, and this change must just change the timing such that it's a little more likely to hit than it was before.

But this brings up another issue too. With this change to delete the lock file after it is unlocked... when there are two concurrent updates going for the same lock file, the second `Write` call should theoretically block waiting for the lock file to be unlocked. In the meantime, the first process holds the lock and does its thing. Then when it's done, it unlocks and deletes the file. Then that second goroutine that was waiting on the lock proceeds and tries to use the lock file but it is now gone.

So maybe it's better to just put the lock file in `/tmp` or something and not try to explicitly delete it? I don't like leaving files around, but at least in that case it would be out of the way and eventually get cleaned up, rather than being left wherever the kubeconfig file is written.

jpmcb:
I think Sean is right: the [`Flock` struct](https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock.go#L29-L35) is a simple wrapper around a read write mutex:
```go
type Flock struct {
	path string
	m    sync.RWMutex
	fh   *os.File
	l    bool
	r    bool
}
```

And each subsequent call to `lockConfigFile` creates a new `flock`: https://github.com/weaveworks/eksctl/blob/f9792c596a974259aef08123e2c913eb11dc79a9/pkg/utils/kubeconfig/kubeconfig.go#L367

Which then is going to create a new RW mutex. From reading the `flock` code, it doesn't seem that those mutexes are mutually exclusive based on the _file path name_. A.k.a, creating a new `flock` with a the same filepath as an existing `flock` _will_ create a new RW mutex and won't prevent concurrent operations if one `flock` is used in one thread and the other flock is used in another.

I wonder if the expectation from `flock` is that a single `flock` be used across many goroutines to gain/prevent access to read and/or write a file. In that case, we _shouldn't_ be creating new locks inside threads, we should hoist that logic to then pass those mutexes down into the individual goroutines.

cPu1:
>So maybe it's better to just put the lock file in /tmp or something and not try to explicitly delete it? I don't like leaving files around, but at least in that case it would be out of the way and eventually get cleaned up, rather than being left wherever the kubeconfig file is written.

I'm on board with this. Writing the lock file to `/tmp` and not trying to removing it later is the cleanest solution IMO.

pkg/utils/kubeconfig/kubeconfig.go
```diff
@@ -344,7 +347,7 @@ func getKubectlVersion() string {
 }
 
 func lockFileName(filePath string) string {
-	return filePath + ".eksctl.lock"
+	return os.TempDir() + filepath.Base(filePath) + ".eksctl.lock"
```
stmcginnis, line 350: This isn't quite right. This will try to place the lock file in the root (`/`) rather than the tmp dir. It might be better to do `fmt.Println(filepath.Join(os.TempDir(), fmt.Sprintf("%s.eksctl.lock", filepath.Base(filePath))))`:

https://go.dev/play/p/6bDaitdArcR

TiberiuGC: Absolutely, a silly mistake 😞 . Thanks for the spot ✨!

stmcginnis: Easy mistake - I only spotted it because I've done the same. ;)

stmcginnis (approved):
Looks great to me!

jpmcb (approved):
👏🏼👏🏼👏🏼👏🏼 

```
❯ ./eksctl utils write-kubeconfig --cluster test --region us-west-2
2022-12-12 21:26:57 [!]  failed to determine authenticator version, leaving API version as default v1alpha1: failed to parse versions: unable to parse first version "unversioned": Invalid character(s) found in major number "unversioned"
2022-12-12 21:26:57 [✔]  saved kubeconfig as "/home/ubuntu/.kube/config"

❯ ls -la /tmp/config.eksctl.lock
-rw------- 1 ubuntu ubuntu 0 Dec 12 21:26 /tmp/config.eksctl.lock
```

cPu1 (changes requested):
As noted in my comment, we need to create an identifier for the lockfile name using the full path of the kubeconfig file rather than just the filename, otherwise multiple eksctl processes using a different kubeconfig path but the same filename will contend for the same lock.

pkg/utils/utils.go
```diff
@@ -47,3 +51,11 @@ func CompareVersions(a, b string) (int, error) {
 	}
 	return aVersion.Compare(bVersion), nil
 }
+
+// FnvHash computes the hash of a string using the Fowler–Noll–Vo hash function
+func FnvHash(s string) []byte {
+	fnvHash.Write([]byte(s))
+	defer fnvHash.Reset()
+
+	return fnvHash.Sum(nil)
```
cPu1, line 60: We should initialise the hash here and remove the call to `Reset`.

cPu1 (approved):

PR #6176, opened 2023-01-20
Fix updating kube-proxy to a newer version

### LINKED ISSUE #6027: [Bug] kube-proxy image version after eksctl upgrade is incorrect causing the kube-proxy to fail

I was trying to upgrade the eks cluster using the command
`eksctl upgrade cluster --name <cluster-name> --region us-east-1 --version 1.23 --approve`

### What happened?
The cluster was upgraded successfully without any other errors on the console. But the `kube-proxy` image updated in the `kube-proxy` `DaemonSet` was incorrect.

As per the AWS EKS docs here - https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html, the image to be used must have the version `v1.23.8-eksbuild.2` OR `v1.23.13-minimal-eksbuild.2`; whereas the image that was updated in my cluster was `v1.23.13-eksbuild.2`. Seems like the default-type & minimal-type versioning was mix-matched here which caused the container startup to fail due to `ImagePullBackOff` error.

As per the below image extracted from the docs referred above, the issue is only specific to k8s 1.23 because the minor version is different for default & minimal image types. Also, even though not applicable for `1.23`, the eksbuild version is different for default & minimal in few other kubernetes versions. It must also be ensured that, this specific build version must not mis-match.
image

Although, I see from https://eksctl.io/usage/addon-upgrade/ that the 3 add-ons are added automatically & are available inside the cluster, they are not shown on the eks cluster console or through `eksctl get addons --cluster <cluster-name>` command. As a result, I cannot use the `eksctl utils update-kube-proxy --cluster=<clusterName>` command as well to get this fixed. I had to manually update the image version using kubectl to make the upgrade work.

### How to reproduce it?
Upgrade cluster to version 1.23 using `eksctl upgrade cluster` command

**Logs**
Nothing much relevant that I could see

N/A

**Versions**

```
eksctl version: 0.118.0
kubectl version: v1.20.5
OS: linux
```

cPu1:
### Description

The version returned by `DescribeAddonVersions` cannot be used as-is for kube-proxy as it does not always map to the container image version. But for now, it does map to the minimal image tag.

This changelist adds a `minimal` string to the version obtained from `DescribeAddonVersions` to map it to a valid container image.

Fixes #6027

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/addons/default/kube_proxy.go
+++ b/pkg/addons/default/kube_proxy.go
@@ -224,7 +224,23 @@ func getLatestImageVersionFromEKS(ctx context.Context, eksAPI awsapi.EKS, contro
 	sort.SliceStable(versions, func(i, j int) bool {
 		return versions[j].LessThan(versions[i])
 	})
-	return versions[0].Original(), nil
+
+	return toMinimalVersion(versions[0]), nil
+}
+
+func toMinimalVersion(v *version.Version) string {
+	preRelease := v.Prerelease()
+	if preRelease == "" {
+		return v.Original()
+	}
+	const versionPrefix = "v"
+	var tagPrefix string
+	if strings.HasPrefix(v.Original(), versionPrefix) {
+		tagPrefix = versionPrefix
+	}
+
+	minimalBuildTag := fmt.Sprintf("%s%s-minimal-%s", tagPrefix, v.Core(), preRelease)
+	return minimalBuildTag
 }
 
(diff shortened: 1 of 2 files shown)


TiberiuGC (approved):

johnpetersjr:
I tested with this release candidate, and I can now properly create node groups with 1.21 version of EKS with proper kube-proxy image.  Thank you!

PR #6332, opened 2023-02-21
Correct cluster and nodegroup name validation

### LINKED ISSUE #6303: [Bug] Cannot get details about my nodegroup, fails validation regex
image
I have a cluster, probably created by Terraform.
I decided to get details about cluster - success.
Then I decided to get all nodegroups in da cluster - success
Then I wanted to get one exactly nodegroup, and it fails with a validation of the name. It seems, it doesn't allow glyphes and underscores in the namings of nodegroup, thou the AWS is totally fine to have such names.
Either I am doing something wrong or the validation RegEx is too strict and doesn't allow real, existing names, being asked.
Version of the eksctl 0.130.0

yuxiang-zhang:
### Description
Stems from #6303. 

Corrected the name validation regex to prevent names starting or ending with `-`. 
Added some test cases for name validation. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -713,13 +713,13 @@ func (ue *unsupportedFieldError) Error() string {
 
 // IsInvalidNameArg checks whether the name contains invalid characters
 func IsInvalidNameArg(name string) bool {
-	re := regexp.MustCompile(`[^a-zA-Z0-9\-]+`)
-	return re.MatchString(name)
+	re := regexp.MustCompile(`^([A-Za-z0-9][-A-Za-z0-9]*)?[A-Za-z0-9]$`)
+	return !re.MatchString(name)
 }
 
 // errInvalidName error when invalid characters for a name is provided
 func ErrInvalidName(name string) error {
-	return fmt.Errorf("validation for %s failed, name must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]*", name)
+	return fmt.Errorf("validation for %s failed, name must satisfy regular expression pattern: ([A-Za-z0-9][-A-Za-z0-9]*)?[A-Za-z0-9]", name)
 }
 
 func validateNodeGroupName(name string) error {
--- a/pkg/apis/eksctl.io/v1alpha5/validation_test.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation_test.go
@@ -53,31 +53,30 @@ var _ = Describe("ClusterConfig validation", func() {
 		})
 	})
 
+
 	Describe("nodeGroups[*].name validation", func() {
 		var (
 			cfg *api.ClusterConfig
+			ng *api.NodeGroup
 			err error
 		)
 
 		BeforeEach(func() {
 			cfg = api.NewClusterConfig()
-			ng0 := cfg.NewNodeGroup()
-			ng0.Name = "ng_invalid-name-10"
-			ng1 := cfg.NewNodeGroup()
-			ng1.Name = "ng100_invalid_name"
-			ng2 := cfg.NewNodeGroup()
-			ng2.Name = "ng100@invalid-name"
+			ng = cfg.NewNodeGroup()
 		})
 
-		It("should reject invalid nodegroup names", func() {
-			err = api.ValidateClusterConfig(cfg)
-			Expect(err).NotTo(HaveOccurred())
+		DescribeTable("rejecting invalid names", 
+			func(ngName string) {
+				ng.Name = ngName
 
-			for i, ng := range cfg.NodeGroups {
-				err = api.ValidateNodeGroup(i, ng, cfg)
+				err = api.ValidateNodeGroup(0, ng, cfg)
 				Expect(err).To(HaveOccurred())
-			}
-		})
+			},
+			Entry("starts with [^A-Za-z0-9]", "-ng-invalid-name-10"),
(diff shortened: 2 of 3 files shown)


pkg/utils/names/names.go
```diff
@@ -72,3 +73,7 @@ func RandomName(length int, chars string) string {
 	}
 	return string(randomName)
 }
+
+func ForStack(name string) string {
+	return strings.ReplaceAll(name, "_", "-")
+}
```
Himangini, line 79: I am not a fan of this. We don't manipulate user input for cluster or nodegroup names. This might have undesired effects on the user.

pkg/actions/fargate/tasks.go
```diff
@@ -20,7 +21,7 @@ type createFargateStackTask struct {
 func (t *createFargateStackTask) Describe() string { return "create fargate IAM stacK" }
 
 func makeClusterStackName(clusterName string) string {
-	return "eksctl-" + clusterName + "-fargate"
+	return names.ForStack("eksctl-" + clusterName + "-fargate")
```
Himangini, line 24: this is not required, pls remove this.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -713,13 +713,15 @@ func (ue *unsupportedFieldError) Error() string {
 
 // IsInvalidNameArg checks whether the name contains invalid characters
 func IsInvalidNameArg(name string) bool {
-	re := regexp.MustCompile(`[^a-zA-Z0-9\-]+`)
-	return re.MatchString(name)
+	startsWithNonAlphanumericRe := regexp.MustCompile(`^[^A-Za-z0-9]`)
+	containsUnsupportedCharacterRe := regexp.MustCompile(`[^A-Za-z0-9\-_]+`)
+
+	return startsWithNonAlphanumericRe.MatchString(name) || containsUnsupportedCharacterRe.MatchString(name)
```
Himangini, line 719: why do we need two regexes? Can we not use the original one 
```re := regexp.MustCompile(`[^a-zA-Z0-9\-]+`)```
to check and allow `underscores` ?
For ref see here https://github.com/weaveworks/eksctl/pull/3942

```suggestion
re := regexp.MustCompile(`[^a-zA-Z0-9\-_]+`)
	return re.MatchString(name)
```

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -713,13 +713,15 @@ func (ue *unsupportedFieldError) Error() string {
...
 }
 
 // errInvalidName error when invalid characters for a name is provided
 func ErrInvalidName(name string) error {
-	return fmt.Errorf("validation for %s failed, name must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]*", name)
+	return fmt.Errorf("validation for %s failed, name must satisfy regular expression pattern: [A-Za-z0-9][A-Za-z0-9\\-_]*", name)
```
Himangini, line 724:
```suggestion
	return fmt.Errorf("validation for %s failed, name must satisfy regular expression pattern: [a-zA-Z][a-zA-Z0-9\\-_]*", name)
```

Himangini (changes requested):

pkg/utils/names/names.go
yuxiang-zhang: Yeah, I am also not a fan. I'm trying to open the discussion for a solution on this problem. Not sure if this is the reason why you didn't support underscore in the first place?

pkg/apis/eksctl.io/v1alpha5/validation.go
yuxiang-zhang: This regex doesn't reject names that start with `-` or `_`, but yeah I'll combine the two.

pkg/utils/names/names.go
Himangini: We didn't support underscore because it wasn't supported in CFN before. This validation was added a long time ago. See here for more details: https://github.com/weaveworks/eksctl/issues/2943
https://github.com/weaveworks/eksctl/issues/2943#issuecomment-859390934

yuxiang-zhang: I suspected that was the case. Thanks for the link!

pkg/apis/eksctl.io/v1alpha5/validation.go
Himangini: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html

yuxiang-zhang: As it stands, eksctl has been accepting names starting with a number because of the regex problem we mentioned

pkg/apis/eksctl.io/v1alpha5/validation_test.go
```diff
@@ -56,28 +56,37 @@ var _ = Describe("ClusterConfig validation", func() {
...
+		)
+
+		DescribeTable("accepting valid names",
+			func(ngName string) {
+				ng.Name = ngName
+
+				err = api.ValidateNodeGroup(0, ng, cfg)
+				Expect(err).NotTo(HaveOccurred())
+			},
+			Entry("starts with [0-9]", "123--"),
```
A-Hilaly, line 86: Interesting, does EKS allow double hyphens in cluster names?

A-Hilaly (approved):
LGTM!
Left one question

Issue #6334, opened 2023-02-22
[Bug] Upgrading using API ignores specified AMI release version


tomr-jet:
### What were you trying to accomplish?

Upgrade managed node groups to a specific AMI version.

### What happened?

The AMI version specified in the `eksctl upgrade nodegroup` command was ignored and it defaults to the latest available AMI version.

### How to reproduce it?

Running the command:
`eksctl upgrade nodegroup --cluster xxxx --region eu-west-1 --name xxxxx --kubernetes-version 1.22 --release-version 1.22.17-20230203`

In the AWS console I can see the AMI version `1.22.17-20230217` is being used:
![Screenshot 2023-02-22 at 12 07 32](https://user-images.githubusercontent.com/125892521/220615725-d6a9a6e2-6333-4997-a26b-7354af3ee9b5.png)

**Logs**

**Anything else we need to know?**

- Cluster and node groups initially provisioned with Terraform.

- It looks like `ReleaseVersion` is missing from the input to the `UpdateNodegroupVersion` API call:
https://github.com/weaveworks/eksctl/blob/5746832673792da6dba9ab892d9cda9ce6a0efad/pkg/actions/nodegroup/upgrade.go#L98

**Versions**

```
$ eksctl info
eksctl version: 0.130.0
kubectl version: v1.26.1
OS: darwin
```

PR #6340, opened 2023-02-23
[Upgrade nodegroup via EKS API] Make AMI release version customisable

tomr-jet:
### Description

Fixes https://github.com/weaveworks/eksctl/issues/6334

Where the command line option `--release-version` is being ignored when upgrading node groups. 

### Checklist
- [X] Manually tested


TiberiuGC:
Hi @tomr-jet . Thanks for opening a PR for this issue. If you notice [the workflow](https://github.com/weaveworks/eksctl/blob/858e189060c07fb4e7f15d506903d36ba0a82e96/pkg/actions/nodegroup/upgrade.go#L174-L289) for nodegroups that are eksctl created (i.e. have a cloudformation stack) you can see that there are various validations that come together with setting `releaseVersion` e.g.

```
	if options.KubernetesVersion != "" && options.ReleaseVersion != "" {
		return errors.New("only one of kubernetes-version or release-version can be specified")
	}
```

```
	if usesCustomAMI && (options.KubernetesVersion != "" || options.ReleaseVersion != "") {
		return errors.New("cannot specify kubernetes-version or release-version when using a custom AMI")
	}
```

I'll be running some manual tests and see which of these are needed here (i.e. upgrading via EKS API) as-well.

TiberiuGC:
@tomr-jet coming back to you on the above. I think we can skip the first validation I referenced in the above message as we want to be able to specify kubernetes version and release version at the same time. Otherwise we always force the latest release version when upgrading kubernetes version, which is exactly the behaviour we want to change with this PR.

However, as per the EKS docs, we need to make sure that we don't specify release version when using a launch template with custom AMI. So we need to add the second validation for this workflow as-well. Just modifying [these lines](https://github.com/weaveworks/eksctl/blob/6580a4d75a4bbc2a02c06691bb24183e84a12a8b/pkg/actions/nodegroup/upgrade.go#L120-L122) and updating unit tests accordingly should suffice to get this PR merged.

image

TiberiuGC:
> 	if options.KubernetesVersion != "" && options.ReleaseVersion != "" {
> 		return errors.New("only one of kubernetes-version or release-version can be specified")
> 	}

Just leaving a note here for future reference.  The first validation remains required for nodegroups that are `eksctl` created and unfortunately the same behaviour cannot be achieved for them. That's a CloudFormation specific restriction not allowing `kubernetesVersion` and `releaseVersion` to be updated together or at the same time with any other stack parameter. 

Suggestions for this PR remain the same.

TiberiuGC (approved):

PR #6554, opened 2023-04-25
Force kubeconfig get-token output format to json

### LINKED ISSUE #6547: [Bug] When the output format of AWS CLI is set to 'table', kubectl will encounter an error when connecting to the EKS created by eksctl
`eksctl create cluster -f test.yml`

### What happened?
At the end of the EKS creation, the following error is displayed:
```bash
2023-04-20 09:16:38 [✖️]  unable to use kubectl with the EKS cluster (check 'kubectl version'): WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Unable to connect to the server: getting credentials: decoding stdout: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
```
kubectl cannot connect to eks.

### How to reproduce it?

set the output to `table` when `aws configure`, and create an eks using eksctl

eksctl should set the output format to `json` in the aws command line when generating `~/.kube/config` instead of relying on the user's AWS profile configuration, which would be a better solution.

yuxiang-zhang:
### Description
Fixes #6547 

AWS CLI introduced a change in https://github.com/aws/aws-cli/pull/7726 as a result of supporting `--output` flag in the `eks get-token` API. We now have to hard-code `--output json` which takes precedence, otherwise the output format will default to [other options](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html#json-output:~:text=output%20format%20in-,three%20ways,-%3A). Note that the AWS CLI version must be newer than `1.27.86` or `2.11.1`. 

### Checklist
- [x] Manually tested


Himangini:
If the AWS CLI version is lower than `1.27.86` or `2.11.1` does it error when `--output json` is used?

yuxiang-zhang:
> If the AWS CLI version is lower than `1.27.86` or `2.11.1` does it error when `--output json` is used?

No, if it's lower then the `--output` flag will be ignored, but this bug will still be present. Though to me, this isn't an eksctl bug. 

On a related note, have you considered using `aws eks update-kubeconfig` for this? not that it would solve this specific problem, but it does the same thing as `write-kubeconfig`

Himangini (approved):
👍🏻

Himangini:
> On a related note, have you considered using `aws eks update-kubeconfig` for this? not that it would solve this specific problem, but it does the same thing as `write-kubeconfig`

I haven't. I'll add a note to look into this further at some point. 👍🏻

PR #6662, opened 2023-06-01
OIDC IDP creation through CF template

vulkoingim:
### Description

**⚠️ This PR is nowhere near ready to be merged and it was a quick PoC to verify that a solution like this would even work. I'm opening it in order to start a discussion on whether such a change is even desirable. As it stands, it works for a particular use-case, where a cluster has to be created from an air-gapped environment and some APIs are unreachable**

Creating a cluster with `oidc` enabled from an air-gapped environment (private subnet, no internet gateway), would fail due to the `eksctl` trying to reach endpoints which are inherently public and for which PrivateLink endpoints don't exist. 

There are a couple of changes in this PR:

##### OIDC Thumbprint

The first endpoint is `oidc.eks.{region}.amazonaws.com`, and it's required to [obtain](https://github.com/weaveworks/eksctl/blob/a4637775c48850be77380289a56bb928bf4129d8/pkg/iam/oidc/api.go#L145-L171C2) the [thumbprint for the identity provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html). 

Since the cluster's oidc provider is always the regional `oidc.eks.{region}`, and the required thumbprint is of the top-level CA - then they are always the same per region and can be obtained beforehand. If we provide them beforehand, we avoid the need to talk to that endpoint.

Looking at the certs, they are generally valid from 2009 until ~2034, so I don't expect that they change often (unless they get rotated, but at that point there would be bigger issues), so supplying the value manually should be safe.

```bash
openssl s_client -servername oidc.eks.eu-central-1.amazonaws.com -showcerts -connect oidc.eks.eu-central-1.amazonaws.com:443  </dev/null

 3 s:C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", CN = Starfield Services Root Certificate Authority - G2
   i:C = US, O = "Starfield Technologies, Inc.", OU = Starfield Class 2 Certification Authority
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Sep  2 00:00:00 2009 GMT; NotAfter: Jun 28 17:39:16 2034 GMT
```

The first change adds `iam.oidcThumbrint` field, and if passed would skip the call to get the thumbprint.

##### Creating the IAM IDP

The second problem is that the actual creation of the IAM IDP requires talking to the IAM API. Since there is no PrivateLink endpoint for it, and it's always public (I imagine that's a side effect of it being in the `us-east-1` region), the IAM IDP can't be created, and also its ARN can't be retrieved.

This can be worked by creating the resources through CF, waiting until they are complete and subsequently getting the ARN from the stack. Right now this change is done by directly calling the CF API.

The problem I see here would be if an IDP already exists, then the CF template would fail. And we wouldn't be able to get the ARN. As a workaround we could have the option to pass the ARN of the IDP in the configuration file, then we can skip template creation. Such a feature would be useful only for clusters that already exist - new clusters shouldn't require that.

##### Creating ServiceAccounts

ServiceAccount creation requires the IDP ARN. I've added a very hacky flag to prompt the manager to look it up from the CF template.

##### Creating NodeGroups

Determining if IRSAs are used requires a call to IAM. [Right now it just returns true](https://github.com/vulkoingim/eksctl/pull/2/files#diff-0109290a87c411fc455c9ac85657661bd794463cacfc112ffa271bfb22facbd3R149). But that probably can be inferred by a check for certain cf templates.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/ctl/create/iamserviceaccount.go
+++ b/pkg/ctl/create/iamserviceaccount.go
@@ -22,6 +22,10 @@ func createIAMServiceAccountCmd(cmd *cmdutils.Cmd) {
 	})
 }
 
+var (
+	discoverIDPFromCF bool
+)
+
 func createIAMServiceAccountCmdWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *cmdutils.Cmd, overrideExistingServiceAccounts, roleOnly bool) error) {
 	cfg := api.NewClusterConfig()
 	cmd.ClusterConfig = cfg
@@ -52,6 +56,7 @@ func createIAMServiceAccountCmdWithRunFunc(cmd *cmdutils.Cmd, runFunc func(cmd *
 		fs.StringVar(&serviceAccount.AttachRoleARN, "attach-role-arn", "", "ARN of the role to attach to the iamserviceaccount")
 		fs.StringVar(&serviceAccount.RoleName, "role-name", "", "Set a custom name for the created role")
 		fs.BoolVar(roleOnly, "role-only", false, "disable service account creation, only the role will be created")
+		fs.BoolVar(&discoverIDPFromCF, "discover-cf-idp", false, "discover the idp from cf template")
 
 		cmdutils.AddStringToStringVarPFlag(fs, &serviceAccount.Tags, "tags", "", map[string]string{}, "Used to tag the IAM role")
 
@@ -76,6 +81,14 @@ func doCreateIAMServiceAccount(cmd *cmdutils.Cmd, overrideExistingServiceAccount
 
 	cfg := cmd.ClusterConfig
 	meta := cmd.ClusterConfig.Metadata
+
+	if discoverIDPFromCF {
+		// just set this to a dummy value, which would prompt the oidc manager to discover
+		// the role ARN from the cf template
+		// doesnt' seem to be used anywhere else
+		dummy := "blbalb"
+		cfg.IAM.OIDCThumbprint = &dummy
+	}
 
 	printer := printers.NewJSONPrinter()
 
--- a/pkg/iam/oidc/api.go
+++ b/pkg/iam/oidc/api.go
@@ -5,20 +5,24 @@ import (
 	"crypto/sha1"
 	"crypto/tls"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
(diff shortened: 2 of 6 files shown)


pkg/iam/oidc/api.go
```diff
@@ -110,6 +181,90 @@ func (m *OpenIDConnectManager) CreateProvider(ctx context.Context) error {
...
+				Key:   gfnt.NewString(k),
+				Value: gfnt.NewString(v),
+			})
+		}
+
+		hostWithoutPort := strings.Split(m.issuerURL.Host, ":")[0]
+		newURL := *m.issuerURL
+		newURL.Host = hostWithoutPort
+
+		fmt.Println(m.issuerURL.String())
```
yue9944882, line 197: use logger rather than print

pkg/iam/oidc/api.go
```diff
@@ -5,20 +5,24 @@ import (
 	"crypto/sha1"
 	"crypto/tls"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
+	"github.com/aws/smithy-go"
+	cft "github.com/weaveworks/eksctl/pkg/cfn/template"
+	cf "github.com/weaveworks/goformation/v4/cloudformation/cloudformation"
```
yue9944882, line 12: format imports

pkg/ctl/create/iamserviceaccount.go
```diff
@@ -77,6 +82,14 @@ func doCreateIAMServiceAccount(cmd *cmdutils.Cmd, overrideExistingServiceAccount
 	cfg := cmd.ClusterConfig
 	meta := cmd.ClusterConfig.Metadata
 
+	if discoverIDPFromCF {
+		// just set this to a dummy value, which would prompt the oidc manager to discover
+		// the role ARN from the cf template
+		// doesnt' seem to be used anywhere else
+		dummy := "blbalb"
```
yue9944882, line 89: isnt empty value better here?

vulkoingim:
@yue9944882 thanks for the comments. As I mentioned, the PR was hacked together quickly just to verify that it would work. I don't expect any of those changes to actually persist in a final version.

What I'm expecting is an answer on whether such a change would be desired and if I should spend time on getting it in order.

PR #6756, opened 2023-07-03
fix index out of range error when aws-node, kube-proxy or coredns are missing

wind0r:
### Description
After https://github.com/weaveworks/eksctl/pull/6695 eksctl stops working on clusters without aws-node (e.g cilium as cni) 

This happens because `clientSet.AppsV1().DaemonSets(metav1.NamespaceSystem).Get(ctx, AWSNode, metav1.GetOptions{})` always returns a resource. This resource shouldn't be used if also an error is returned.
See: https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/apps/v1/daemonset.go#L74

But the code is checking for nil. (Also true for CoreDNS and kube-proxy)
```
	if clusterDaemonSet == nil {
		return false, nil
	}
```

I think this got unnoticed because the client mock incorrectly returns a nil resource on error.

```
2023-07-03 09:14:45 [!]  "aws-node" was not found
panic: runtime error: index out of range [0] with length 0
goroutine 1 [running]:
github.com/weaveworks/eksctl/pkg/addons/default.DoesAWSNodeSupportMultiArch({0x5d02c50?, 0xc00013c000?}, {0x5d1c790?, 0xc000789e00?})
	github.com/weaveworks/eksctl/pkg/addons/default/aws_node.go:48 +0x4c5
github.com/weaveworks/eksctl/pkg/addons/default.DoAddonsSupportMultiArch({0x5d02c50, 0xc00013c000}, {0x5d1c790, 0xc000789e00})
	github.com/weaveworks/eksctl/pkg/addons/default/addons.go:39 +0x113
github.com/weaveworks/eksctl/pkg/actions/nodegroup.(*Manager).checkARMSupport(0xc000c1b200?, {0x5d02c50, 0xc00013c000}, 0xc001185e30, 0xc000c54e60?, 0x1)
	github.com/weaveworks/eksctl/pkg/actions/nodegroup/create.go:292 +0x5b
github.com/weaveworks/eksctl/pkg/actions/nodegroup.(*Manager).Create(0xc000c54e60, {0x5d02c50, 0xc00013c000}, {0x1, 0x1, 0x1, {0x0, {0x5cb5200, 0xc000134008}}, 0x1, ...}, ...)
	github.com/weaveworks/eksctl/pkg/actions/nodegroup/create.go:92 +0x6b6
github.com/weaveworks/eksctl/pkg/ctl/create.createNodeGroupCmd.func1(0xc0007c9440, 0xc0007c9500, {{0x1, 0x1, 0x0}, {0x1, 0x0, {0x0, 0x0, 0x0}}, ...})
	github.com/weaveworks/eksctl/pkg/ctl/create/nodegroup.go:71 +0x5e7
github.com/weaveworks/eksctl/pkg/ctl/create.createNodeGroupCmdWithRunFunc.func1(0xc00032d800?, {0xc000a406f0?, 0x0?, 0x3?})
	github.com/weaveworks/eksctl/pkg/ctl/create/nodegroup.go:100 +0xb8
github.com/spf13/cobra.(*Command).execute(0xc00032d800, {0xc000a406c0, 0x3, 0x3})
	github.com/spf13/cobra@v1.7.0/command.go:940 +0x862
github.com/spf13/cobra.(*Command).ExecuteC(0xc0005e3500)
	github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.7.0/command.go:992
main.main()
	github.com/weaveworks/eksctl/cmd/eksctl/main.go:102 +0x5db
```

### Checklist
- [x] Manually tested


TiberiuGC:
> 
	if clusterDaemonSet == nil {
		return false, nil
	}
	
Right above the condition you highlighted there is a check for non nil error. If I understand correctly, shouldn't this be enough to avoid the scenario which you believe produces the panic?

https://github.com/weaveworks/eksctl/blob/cccd7322f9ddfb873b18573ed6d91d3a3d176a24/pkg/addons/default/aws_node.go#L34-L40

wind0r:
No the `err != nill` check isnt enough since the makeGetError function is "silencing" the NotFoundError:

Also in a error case we want to return `false, err` and if the Daemonset doesn't exist we want to return `true, nil`.

https://github.com/weaveworks/eksctl/blob/cccd7322f9ddfb873b18573ed6d91d3a3d176a24/pkg/addons/default/addons.go#L74-L83

https://github.com/weaveworks/eksctl/blob/cccd7322f9ddfb873b18573ed6d91d3a3d176a24/pkg/addons/default/aws_node.go#L173-L175

cPu1 (approved):
Good catch. LGTM.

TiberiuGC:
> No the `err != nill` check isnt enough since the makeGetError function is "silencing" the NotFoundError:
> 
> Also in a error case we want to return `false, err` and if the Daemonset doesn't exist we want to return `true, nil`.
> 
> https://github.com/weaveworks/eksctl/blob/cccd7322f9ddfb873b18573ed6d91d3a3d176a24/pkg/addons/default/addons.go#L74-L83
> 
> https://github.com/weaveworks/eksctl/blob/cccd7322f9ddfb873b18573ed6d91d3a3d176a24/pkg/addons/default/aws_node.go#L173-L175

Makes sense, thanks! 👍

TiberiuGC (approved):

PR #6814, opened 2023-07-14
Select one subnet for AZs where multiple are present and no VPC config provided

TiberiuGC:
### Description

Internally, we store subnets as a map, where the key can be either the subnet name of the subnet AZ. As we populate this map when importing a VPC config, if there are multiple subnets in the same availability zone, they will conflict with each other. This happens because, when we find the second subnet, we already have the first one stored by AZ in the map, and the existing validation will see this as a conflict between local state provided in the config file, and the remote state we are trying to import. However, if the user defined this mapping in the config file and chooses to use subnet name as key, the 2 subnets in same AZ won't generate a conflict.

With this PR, we aim to correctly validate config file against remote config, if one is provided. Moreover, if no config file is provided, only one subnet from each AZ will be picked, as this is enforced by the [AZSubnetMapping](https://github.com/eksctl-io/eksctl/blob/dc102c8db64741400ada83bfb98b5c65b03ae310/pkg/apis/eksctl.io/v1alpha5/vpc.go#L32-L36) that's being populated using AZs as keys.

Closes https://github.com/eksctl-io/eksctl/issues/6702

### Checklist
- [x] Manually tested


pkg/apis/eksctl.io/v1alpha5/vpc.go
```diff
@@ -268,75 +268,95 @@ func DefaultCIDR() ipnet.IPNet {
...
-			return fmt.Errorf("subnet ID %q is not the same as %q", network.ID, subnetID)
+		return err
+	}
+
+	// if a VPC config was provided as part of the config file,
+	// we need to validate it against the remote config
+	// and return an error in case of missmatch
+	subnetKey, err := validateLocalConfigAgainstRemote(localSubnetsConfig, remoteSubnet, makeSubnetAlias(subnet))
+	if err != nil {
+		return fmt.Errorf("missmatch found between local and remote VPC config: %w", err)
```
cPu1, line 286: Minor nit here and in tests:

```suggestion
		return fmt.Errorf("mismatch found between local and remote VPC config: %w", err)
```

cPu1 (approved):
I have left a comment about a minor typo but otherwise LGTM.

Issue #6570, opened 2023-05-04
[Spike] Investigate how to improve unit test coverage for 'ctl' package


TiberiuGC:
An older [document](https://docs.google.com/document/d/1ZVkEogWNZzOebJ0QMN6eYrVpBFs1W7EfDqqtpZ4ozRo/edit#heading=h.u9tibp2h2kr2) suggests the following

```
I'm relatively sure that part of the reason for the low unit test coverage in the ctl/ package is due to the fact that we're hardcoding the Provider in each ctl with the following function:

ctl, err := cmd.NewProviderForExistingCluster()
if err != nil {
return err
}

This means that we can't mock calls to EKS that are made by ctl commands.
```

We should investigate if there's a reasonable way to mock the provider and then increase test coverage, separately, for each relevant command.

Timebox: 1-2 days

TiberiuGC:
Posting findings here:

## What is the main deterrent?

Most of the commands in the CTL package hardcode the cluster provider by using the function call below.     
```
ctl, err := cmd.NewProviderForExistingCluster(ctx)
if err != nil {
    return err
}
```

However, very little of the provider’s functionality is actually being used in the CTL package. See a list below:

```
ctl.CanOperate
ctl.NewOpenIDConnectManager
ctl.NewStdClientSet
ctl.NewRawClient
ctl.AWSProvider.EKS
ctl.AWSProvider.AWSConfig
ctl.AWSProvider.EC2
```
Moreover, some of the functions (first two) are not even part of any interfaces that the provider consists of, making it mode difficult to mock them in a clean way. 


## How to overcome it?

### Option 1 
Simply pass the cluster provider as a function argument in the following template

```
eksctlCommand() {
	…
	cmd.CobraCommand.RunE = func(_ *cobra.Command, args []string) error {
        	…
		return doEksctlCommand(…, clusterProvider)
    }
} 
```
Afterwards we can use existing mocks as in - `FakeKubeProvider`, `MockProvider` and/or configure `ProviderStatus` for `ctl.CanOperate` calls. Additionally we will probably need another interface as part of the `clusterProvider` just for `ctl.NewOpenIDConnectManager` calls.

### Option 2 

Similar to option 1 in terms of template, but additionally create an interface that amasses the functions calls referenced above, and mock them using counterfeiter.

In addition to the above work, each command usually relies on a manager struct inside actions package for doing the actual work. We will need interfaces and mocks for these managers as-well.


## What functionality should be tested?

[...]

PR #6851, opened 2023-07-26
Add option to print specific events when running `eksctl describe-stacks`

### LINKED ISSUE #6571: [Feature] Add option to print specific events when running eksctl describe-stacks
When running `eksctl utils describe-stacks` command, the user has the options to provide `--events` flag in order to see a full event log as they would in the AWS Console. However, this could be considered noisy, depending on use case, and they may want to only filter for failed events. We could potentially provide this options via a new CLI flag e.g. `--failed-only` and filter by `ResourceStatus` only for the following types:

```
	ResourceStatusCreateFailed             ResourceStatus = "CREATE_FAILED"
	ResourceStatusDeleteFailed             ResourceStatus = "DELETE_FAILED"
	ResourceStatusUpdateFailed             ResourceStatus = "UPDATE_FAILED"
	ResourceStatusUpdateRollbackFailed     ResourceStatus = "UPDATE_ROLLBACK_FAILED"
	ResourceStatusRollbackFailed           ResourceStatus = "ROLLBACK_FAILED"
```

Note that we also run `eksctl utils describe-stacks --events` whenever an integration test spec fails, so adding this new flag there would probably be useful as-well.

TiberiuGC:
### Checklist
- [x] Manually tested

Adding unit tests for these changes is blocked by #6570.

Closes #6571

When running `eksctl utils describe-stacks` command, the user has the options to provide `--events` flag in order to see a full event log as they would in the AWS Console. However, this could be considered noisy, depending on use case, and they may want to only filter for e.g. failed events. This PR adds a CLI flag called `--resource-status` that accepts one of the [following values](https://github.com/aws/aws-sdk-go-v2/blob/806208b12d2f51bf0a7eb79e56cb48fb65c39fab/service/cloudformation/types/enums.go#L763-L789) (case insensitive) and filters the results accordingly.


cPu1 (approved):

pkg/ctl/utils/describe_stacks.go
```diff
@@ -48,14 +52,15 @@ func describeStacksCmd(cmd *cmdutils.Cmd) {
...
-		return doDescribeStacksCmd(cmd, all, events, trail, printer)
+		return doDescribeStacksCmd(cmd, all, events, trail, resourceStatus, printer)
 	}
 
 	cmd.FlagSetGroup.InFlagSet("General", func(fs *pflag.FlagSet) {
 		cmdutils.AddClusterFlagWithDeprecated(fs, cfg.Metadata)
 		cmdutils.AddRegionFlag(fs, &cmd.ProviderConfig)
 		fs.BoolVar(&all, "all", false, "include deleted stacks")
 		fs.BoolVar(&events, "events", false, "include stack events")
+		fs.StringSliceVar(&resourceStatus, "resource-status", nil, "resource statuses to filter events by, e.g. `CREATE_FAILED`, `UPDATE_FAILED`")
```
a-hilaly, line 63: How about naming this flag `--events-filter|type`? do we already use `--resource-status` in other places?

a-hilaly (commented):
LGTM, one tiny comment below

pkg/ctl/utils/describe_stacks.go
TiberiuGC: I don't think we use `--resource-status` in any other places. We could use your suggested approach if we find other fields of interest to filter by, I guess. I'll merge the PR as is for now, but I'm open to changes for future improvements.

PR #6903, opened 2023-08-03
Fix PR Rebase workflow

### LINKED ISSUE #6900: [Bug] PR Rebase workflow job is always skipped
`/rebase` properly triggers the workflow. However the job itself is being skipped.

https://github.com/eksctl-io/eksctl/actions/workflows/rebase.yaml

TiberiuGC:
### Checklist
- [x] Manually tested

Closes #6900 

The PR rebase workflow points to an incorrect job i.e. `Rebase`, which results in the workflow always being skipped. The actual name of the job should be `Automatic Rebase`.


.github/workflows/rebase.yaml
```diff
@@ -7,14 +7,13 @@ jobs:
     name: Rebase
     if: |
       (github.event.issue.pull_request != '' && contains(github.event.comment.body, '/rebase')) &&
-      (github.event.comment.author_association == 'MEMBER' || github.event.comment.user.login == github.event.pull_request.user.login)
```
cPu1, line 10: What were the values for these two fields when you triggered the workflow?

TiberiuGC: github.event.comment.author_association = "Owner"
github.event.comment.user.login = "TiberiuGC"
github.event.pull_request.user.login = "" 😕

TiberiuGC: > Don't we need to remove the `&&` on L9?

🙈 🙈 🙈

TiberiuGC: So I did some digging and if the workflow is triggered by an `issue_comment` event, the PR opener should be retrieved using `github.event.issue.user.login`, instead of `github.event.pull_request.user.login` (which is used if the worflow is triggered by a `pull_request` event).

I propose to keep the condition as below:

`github.event.comment.author_association` == `OWNER` || 
`github.event.comment.author_association` == `MEMBER` || 
(`github.event.issue.user.login` == `github.event.comment.user.login`)

This way, owners and members will be able to rebase any PRs, while other type of contributors will only be able to rebase their PRs.

TiberiuGC:
Updated the PR as per the suggestion in the comment above ☝️ 
https://github.com/eksctl-io/eksctl/pull/6903#discussion_r1283043231

.github/workflows/rebase.yaml
Himangini: tbh, I'd just remove this condition altogether but I am fine with the above suggestion 👍🏻

.github/workflows/rebase.yaml
```diff
@@ -7,14 +7,16 @@ jobs:
     name: Rebase
     if: |
```
Himangini, line 8:
```suggestion
    if: >-
```
noticed on https://github.com/cirrus-actions/rebase/blob/master/README.md#installation

Himangini (approved):
LGTM

cPu1 (approved):

Issue #7028, opened 2023-08-30
[Bug] eksctl fail to delete cluster if the Application load balancer name created by Ingress is 31 characters


arestuzi:
### What were you trying to accomplish?

#  eksctl delete cluster --name eks-workshop 

### What happened?

1. eksctl version - latest version
After execute eksctl delete cluster --name eks-workshop , it returns following error.
2023-08-30 16:45:42 [ℹ]  deleting EKS cluster "eks-workshop"
2023-08-30 16:45:42 [ℹ]  will drain 0 unmanaged nodegroup(s) in cluster "eks-workshop"
2023-08-30 16:45:42 [ℹ]  starting parallel draining, max in-flight of 1
2023-08-30 16:45:43 [ℹ]  deleted 0 Fargate profile(s)
2023-08-30 16:45:43 [✔]  kubeconfig has been updated
2023-08-30 16:45:43 [ℹ]  cleaning up AWS load balancers created by Kubernetes objects of Kind Service or Ingress
Error: cannot obtain information for ALB from Ingress nginx/nginx-ingress: cannot obtain security groups for ALB abc5ca85-default-terryares-5555-: cannot describe ELB: operation error Elastic Load Balancing v2: DescribeLoadBalancers, https response error StatusCode: 400, RequestID: xxxxx-8a5d-xxxxx-bed9-e47c9f844965, api error ValidationError: The load balancer name '0abc5ca85-default-terryares-5555-' cannot end with a hyphen(-)

related code: 
https://github.com/eksctl-io/eksctl/blob/0c2ced9b2f3ec5613d6ef3a7be5327dae1f62b67/pkg/elb/cleanup.go#L220

### How to reproduce it?
Include ALL the steps to reproduce the bug.
1. deploy a ingress with annotation alb.ingress.kubernetes.io/load-balancer-name: "abc5ca85-default-terryares-5555"
2. eksctl delete cluster --name eks-workshop  delete the cluster
If using a config file, include it here, removing any sensitive information!
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
  namespace: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  namespace: nginx
[...]

PR #7029, opened 2023-08-31
Fix ELB cleanup failure due to resource name end with a hyphen

0xlen:
Initiate the bug fix thread to mitigate the ELB cleanup issue due to incorrect resource name. (See #7028)

### Description

The current ELB cleanup logic doesn't handle the case when the trimmed ELB hostname string is ending with hyphen (`-`). This can cause the eksctl throw the error when making DescribeLoadBalancers API calls due to incorrect resource format. The issue can be reproduced by using the testing file [1] with the following error:

```
$ eksctl version
0.154.0
$ eksctl delete cluster --region us-west-2 --name eks
2023-08-31 00:44:39 [ℹ]  deleting EKS cluster "eks"
2023-08-31 00:44:40 [ℹ]  will drain 0 unmanaged nodegroup(s) in cluster "eks"
2023-08-31 00:44:40 [ℹ]  starting parallel draining, max in-flight of 1
2023-08-31 00:44:41 [ℹ]  deleted 0 Fargate profile(s)
2023-08-31 00:44:42 [ℹ]  cleaning up AWS load balancers created by Kubernetes objects of Kind Service or Ingress
Error: cannot obtain information for ALB from Ingress default/testcase-0: cannot obtain security groups for ALB k8s-default-testcase-93d8419948-: cannot describe ELB: operation error Elastic Load Balancing v2: DescribeLoadBalancers, https response error StatusCode: 400, RequestID: ......, api error ValidationError: The load balancer name 'k8s-default-testcase-93d8419948-' cannot end with a hyphen(-)
```

By fixing the resource name issue, the expected deletion can be successfully executed [2].

[1] https://gist.github.com/0xlen/d3bab9e46a8e797760e4731817543b28
[2] https://gist.github.com/0xlen/15cd7f43fca4cc8483e70ed7fc85bcb5

### Checklist
- [X] Manually tested

- [X] Refactored something and made the world a better place :star2:


TiberiuGC:
Hi @0xlen , thanks for opening a PR for this issue! 🚀 

Before approving I wanted to confirm that trimming the ending `-` now properly covers all possible naming formats. Are you positive about that?

pkg/elb/cleanup_test.go
```diff
@@ -0,0 +1,49 @@
...
+				"abcdefghijklmnopqrstuvw000",
+				"abcdefghijklmnopqrstuvw001",
+				"myloadbalancer",
+				"my-loadbalancer",
+			}
+
+			for i, elbDNSName := range testELBDNSnames {
+				name := getIngressELBName(context.TODO(), "", elbDNSName)
+				Expect(name).To(Equal(expectELBName[i]))
+			}
```
a-hilaly, line 46: Tiny suggestion:
```go
			testCases := [][]struct{hostname, metadataName, expected string}{
				{
				    "bf647c9e-default-appingres-350b-1622159649.eu-central-1.elb.amazonaws.com", 
				    "",
				    "bf647c9e-default-appingres-350b",
				  },
				{
				    "internal-k8s-default-testcase-93d8419948-541000771.us-west-2.elb.amazonaws.com",
				    "",
				    "k8s-default-testcase-93d8419948",
				},
                // ...
			}

			for i, tc := range testCases {
				name := getIngressELBName(context.TODO(), tc.metadatName, []string{tc.hostname})
				Expect(name).To(Equal(tc.expected))
			}
```

pkg/elb/cleanup.go
```diff
@@ -199,6 +199,26 @@ func getServiceLoadBalancer(ctx context.Context, ec2API awsapi.EC2, elbAPI Descr
...
+		return ""
+	}
+	name := strings.TrimPrefix(hostNameParts[0], "internal-") // Trim 'internal-' prefix for ALB DNS name which is not a part of name.
+	idIdx := strings.LastIndex(name, "-")
+	if (idIdx) != -1 {
+		name = name[:idIdx] // Remove the ELB ID and last hyphen at the end of the hostname (ELB name cannot end with a hyphen)
+	}
+	if len(name) > 32 {
+		name = name[:32]
+	}
```
a-hilaly, line 217: Do we need this if we already remove the ELB ID?

a-hilaly (commented):
Thank you @0xlen ! Left few questions/suggestions below.

pkg/elb/cleanup.go
```diff
@@ -199,6 +199,26 @@ func getServiceLoadBalancer(ctx context.Context, ec2API awsapi.EC2, elbAPI Descr
...
 }
 
+func getIngressELBName(ctx context.Context, metadataName string, hosts []string) string {
+	// Expected e.g. bf647c9e-default-appingres-350b-1622159649.eu-central-1.elb.amazonaws.com where AWS ALB name is
+	// bf647c9e-default-appingres-350b (cannot be longer than 32 characters).
+	hostNameParts := strings.Split(hosts[0], ".")
+	if len(hostNameParts[0]) == 0 {
+		logger.Debug("%s is ALB Ingress, but probably not provisioned or something other unexpected, skip", metadataName)
+		return ""
+	}
```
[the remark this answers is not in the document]
0xlen: Thanks @a-hilaly, that's a great point! I tried to revise this part and let `getIngressLoadBalancer` handle it. To have no interruption deletion of `eksctl delete` command, I think it would be better to keep the current business logic to skip the ALB resource clean up [1] if there's unexpected error when getting ELB name.

https://github.com/eksctl-io/eksctl/blob/6e8e73c13cee21b3a833f111e63ade8c03536d04/pkg/elb/cleanup.go#L220-L224

Update: https://github.com/eksctl-io/eksctl/pull/7029/commits/b638eacf9e0b8d75ae174a7a729b59d4cf2a4550#diff-bf8761431813288f0320030b76870e9d4a40dd0e80e1ca725ef262184c498b90R237-R241

pkg/elb/cleanup_test.go
0xlen: Thanks for the advice, this has been revised here: https://github.com/eksctl-io/eksctl/pull/7029/commits/b638eacf9e0b8d75ae174a7a729b59d4cf2a4550#diff-581a28d52a0e2589ad7027966d2b578dfe8dce701181ce35e53c2d4b06f5b302

pkg/elb/cleanup.go
0xlen: That's a good question. From my research ELB document [1] explicitly mentioned the rule when giving the name of load balancer, I think we still don't expect there's an ELB name exceeds 32 characters.

It might be better revise this part as exception case and throw the error directly, and let `getIngressLoadBalancer ` handle proper action (like skip ALB deletion right now, or interrupt the clean up process in the future).

Update: https://github.com/eksctl-io/eksctl/pull/7029/commits/4989bf86801145276785f5691c310e808f33cf12#diff-bf8761431813288f0320030b76870e9d4a40dd0e80e1ca725ef262184c498b90R214-R215

[1] https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_CreateLoadBalancer.html#API_CreateLoadBalancer_RequestParameters

pkg/elb/cleanup.go
```diff
@@ -199,6 +199,26 @@ func getServiceLoadBalancer(ctx context.Context, ec2API awsapi.EC2, elbAPI Descr
...
+	name := strings.TrimPrefix(hostNameParts[0], "internal-") // Trim 'internal-' prefix for ALB DNS name which is not a part of name.
+	idIdx := strings.LastIndex(name, "-")
+	if (idIdx) != -1 {
+		name = name[:idIdx] // Remove the ELB ID and last hyphen at the end of the hostname (ELB name cannot end with a hyphen)
+	}
+	if len(name) > 32 {
+		name = name[:32]
+	}
+	logger.Debug("ALB resource name: %s", name)
```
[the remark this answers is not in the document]
0xlen: Thanks for the suggestion! But I am not sure if this update is what you are looking for? Or does it mean the debug message of showing ALB resource name should be moved to `getIngressLoadBalancer`?

- https://github.com/eksctl-io/eksctl/pull/7029/commits/4989bf86801145276785f5691c310e808f33cf12#diff-bf8761431813288f0320030b76870e9d4a40dd0e80e1ca725ef262184c498b90R217

I did not fully understand what could be the difference as it seems like not all logger in the package is applying the defer directive, I assume it might be some performance improvement, or, it helps the readability (which I think would be what you are mentioning). Any details would be appreciated!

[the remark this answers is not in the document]
0xlen: Moved the logger to `getIngressLoadBalancer`: https://github.com/eksctl-io/eksctl/compare/4989bf86801145276785f5691c310e808f33cf12..b73c890febc5feec8b4ae554067260aecfe73a91

pkg/elb/cleanup.go
```diff
@@ -199,6 +199,24 @@ func getServiceLoadBalancer(ctx context.Context, ec2API awsapi.EC2, elbAPI Descr
 	return &lb, nil
 }
 
+func getIngressELBName(ctx context.Context, hosts []string) (string, error) {
```
a-hilaly, line 202: We can remove `ctx context.Context` from the params

pkg/elb/cleanup.go
```diff
@@ -215,17 +233,12 @@ func getIngressLoadBalancer(ctx context.Context, ec2API awsapi.EC2, elbAPI Descr
...
-	// Expected e.g. bf647c9e-default-appingres-350b-1622159649.eu-central-1.elb.amazonaws.com where AWS ALB name is
-	// bf647c9e-default-appingres-350b (cannot be longer than 32 characters).
-	hostNameParts := strings.Split(hosts[0], ".")
-	if len(hostNameParts[0]) == 0 {
-		logger.Debug("%s is ALB Ingress, but probably not provisioned or something other unexpected, skip", metadata.Name)
+	name, err := getIngressELBName(ctx, hosts)
+	if err != nil {
+		logger.Debug("%s is ALB Ingress, but probably not provisioned or something other unexpected when getting ALB resource name, skip: %s", metadata.Name, err)
 		return nil, nil
 	}
```
a-hilaly, line 240: Maybe we should error if the `parsed name exceeds maximum of 32 characters` is returned? thoughts @cPu1 ?

a-hilaly (approved):
Great stuff @0xlen , thank you! I have one nit/question below

pkg/elb/cleanup.go
0xlen: Agree, I removed the unused `ctx` parameter.

pkg/elb/cleanup.go
0xlen: I was referring original general text as considering there can have `cannot get the hostname` or `parsed name exceeds maximum of 32 characters`. Would be interested to know what could be the proper error message and I am happy to revise that.

a-hilaly:
@0xlen Thank you for iterating on this! LGTM. I'll let the eksctl experts have a say on this before we merge it.

TiberiuGC (approved):
LGTM 👍

PR #7065, opened 2023-09-13
Allow managed suite specs to run in parallel

TiberiuGC:
### Description

Closes https://github.com/eksctl-io/eksctl-ci/issues/101

Managed suite creates, validates and deletes a bunch of completely unrelated nodegroups. This PR refactors the specs such that they can run in parallel. Additionally, this PR removes duplicate specs (which are also present in CRUD suite) and moves the nodegroup upgrade spec to upgrade suite. With these improvements, runtime went down to less than half, i.e. 45m. 

N.B. Moving the upgrade spec will obviously increase the run time of upgrade suite (by around 10 mins), however, this allows tests from managed suite to be logically independent and hence run in parallel, which still brings great improvements in runtime.

image

### Checklist
- [x] Manually tested


integration/tests/managed/managed_nodegroup_test.go
```diff
@@ -142,504 +140,387 @@ var _ = Describe("(Integration) Create Managed Nodegroups", func() {
...
+				"nodegroup",
+				"--verbose", "4",
+				"--cluster", params.ClusterName,
+				publicNodeGroup,
+			)
+			Expect(cmd).To(RunSuccessfully())
+		})
+
+		It("supports a private nodegroup", func() {
+			By("creating it")
```
cPu1, line 233: Here and elsewhere, we should make all `By` messages more specific to the test specs.

integration/utilities/kube/kube.go
```diff
@@ -26,6 +28,7 @@ func NewTest(kubeconfigPath string) (*harness.Test, error) {
 		return nil, err
 	}
 	test := h.NewTest(t)
+	test.Namespace += fmt.Sprintf("-%d", t.ParallelProcess())
```
cPu1, line 31: This will change the namespace in other tests, which should not cause any failures, but for now, we should do this only if the test is being run with `-p`.

cPu1 (approved):

PR #7116, opened 2023-09-26
Set AWS credential expiry window to 30 minutes

### LINKED ISSUE #7095: [Bug] Need to refresh token before each CloudFormation stack operation
`eksctl` currently reuses the same AWS credentials, including session token, for every AWS operation it performs during an operation such as creating a cluster. This can be a problem for complex configurations that take a long time to process, such as one including a cluster, managed node groups, multiple IRSA entries, and a Karpenter setup, because the token is passed to CloudFormation (which operates on the principal's behalf) as-is and cannot be refreshed by it. If the total completion time of all tasks exceeds the STS token expiration time -- currently 15 minutes -- a stack creation can fail with an invalid token error. This typically won't happen with simple configurations, such as just a VPC and cluster, but can happen with more complicated ones. 

This isn't an issue for many users, but it is a problem in CodeBuild environments where the credential source is the ECS credential provider. 

This issue is essentially identical to #4221.

I would recommend `eksctl` configure the AWS auth provider to maximize the validity time of the temporary token it requests (I believe 1 hour is the maximum) and configure the auth provider to refresh the token if it will expire within 1 hour. This will ensure CloudFormation has all the time it needs to perform the work it needs to do without the token expiring.

otterley:
### Description

Set the AWS credential expiry window to 30 minutes. This will force the AWS client to renew any endpoint-provided credentials (e.g. EC2 instance credentials, ECS-provided credentials) if they would expire in less than 30 minutes, and ensure the passed credentials don't expire during long CloudFormation stack operations.

This PR also includes an update to address a broken reference to `pathlib` retracted version v1.0.0, replacing it with v0.15.0. (The retracted version was available via the public Go module proxy, but not available directly to those of us who aren't allowed to use the proxy.)

Closes #7095

### Checklist
- [x] Manually tested


a-hilaly (approved):

TiberiuGC:
Hi @otterley ! Thanks for opening a PR for this improvement.

I realise none of the load options set are being unit tested atm, but I think we can add a test for this one in particular, and improve the overall coverage afterwards. I'll leave a code snipped I came up with rather quickly, so feel free to think of a better way of testing if you wish. (this should be added under `pkg/eks/api_test.go`)

```
		// check that load config options were setup properly
		_, options := fakeConfigurationLoader.LoadDefaultConfigArgsForCall(0)
		lo := &config.LoadOptions{}
		for _, loadOptionsFunc := range options {
			Expect(loadOptionsFunc(lo)).NotTo(HaveOccurred())
		}

		// check that credentials cache was setup properly
		cco := &aws.CredentialsCacheOptions{}
		lo.CredentialsCacheOptions(cco)
		Expect(cco.ExpiryWindow).To(Equal(30 * time.Minute))
		Expect(cco.ExpiryWindowJitterFrac).To(Equal(float64(0)))
```

otterley:
Hi @TiberiuGC - I'd be happy to do this, but the testing code is a bit difficult for me to make sense of, so I don't know where this would go. Since you already have a sense of where it should go and how it should be composed, I'd welcome your patch - you can submit a PR against this branch and I'll gladly merge it.

TiberiuGC:
Sure, no worries, I'll update the tests myself.

otterley:
Thanks @TiberiuGC ! Are we OK to merge?

TiberiuGC:
@otterley sorry for keeping you waiting, just merged the PR so it will get caught in today's release candidate.

PR #7451, opened 2024-01-04
Update well-known policy for ebsCSIController

teraflik:
### Description

The IAM condition key `StringLike` was used incorrectly in the policy and it doesn't work with wildcard (*) in the key itself. Wildcards are only supported in the value of the key.

This fixes an issue in cases where a volume dynamically provisioned via the older in-tree CSI plugin is being deleted by the new EBS CSI driver, because such volumes don't have the tags used in the policy. 

>rpc error: code = Internal desc = Could not delete volume ID "vol-0894ac8afbxxxxx": DeleteDisk could not delete volume: UnauthorizedOperation: You are not authorized to perform this operation. User: arn:aws:sts::xxxxxxx:assumed-role/ebs-csi-controller-preprod-eu/170435858907323xxxx is not authorized to perform: ec2:DeleteVolume on resource: arn:aws:ec2:eu-west-1:xxxxxxx:volume/vol-0894ac8afbxxxxx because no identity-based policy allows the ec2:DeleteVolume action.

The changes made are inspired from the AWS managed `AmazonEBSCSIDriverPolicy`.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/iam_test.go
+++ b/pkg/cfn/builder/iam_test.go
@@ -547,18 +547,6 @@ const expectedEbsPolicyDocument = `{
 	},
 	{
 	  "Action": [
-		"ec2:CreateVolume"
-	  ],
-	  "Condition": {
-		"StringLike": {
-		  "aws:RequestTag/kubernetes.io/cluster/*": "owned"
-		}
-	  },
-	  "Effect": "Allow",
-	  "Resource": "*"
-	},
-	{
-	  "Action": [
 		"ec2:DeleteVolume"
 	  ],
 	  "Condition": {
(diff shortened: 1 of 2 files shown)


teraflik:
Sorry to ping you @cPu1 / @TiberiuGC but do you think this can be moved forward?

yuxiang-zhang (approved):
Found the corresponding update: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/b1d476c49a5c84bc4bb1ae09d1c2dc67e7e03f18

teraflik:
Thanks @yuxiang-zhang for finding that. Now approval to run the workflow is needed to proceed.

teraflik:
Updated the test case.

teraflik:
Missed a comma in json, fixed now.

PR #7483, opened 2024-01-16
Handle unordered public endpoint CIDRs from EKS in endpoint updates

Emberwalker:
### Description

For some clusters, EKS can return the list of public endpoint CIDRs out of order, and won't allow updates where the incoming and current sets have set equality (i.e. regardless of order of CIDR entries). This change restores the set equality check that was removed in commit 72605fbfb1d5b7a8f72dd58d184e7fd52a54cc88 and adds an additional test case to cover this case.

In our clusters, EKS always appears to return the CIDRs in a specific order, but not sorted - possibly due to the accumulation and removal of entries over time. Without this change, eksctl 0.167.0 will always attempt an update and get a 400 InvalidParameterException ("Cluster is already at the desired configuration with [...]") - with this applied, it correctly identifies no update is needed.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/ctl/utils/vpc_helper.go
+++ b/pkg/ctl/utils/vpc_helper.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/kris-nova/logger"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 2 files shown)


pkg/ctl/utils/vpc_helper.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/kris-nova/logger"
+	"k8s.io/apimachinery/pkg/util/sets"
```
a-hilaly, line 9: Nice, TIL about apimachinery string sets!

a-hilaly (approved):

PR #4205, opened 2021-09-09
Drain nodegroups during cluster deletion

nikimanoledaki:
### Description

We drain nodegroups prior to deleting them individually but not prior to deleting them when we delete the whole cluster. Draining nodegroups before deleting them is important so that pods can be evicted (and not scheduled on those pods again) and resources, such as IP addresses, can be released. Not draining nodes before deleting them may block the deletion of some resources and cause leftover resources such as IP addresses.

This PR focuses on unmanaged nodegroups because managed nodegroups get drained by EKS anyway.

I tried to do a bit of refactoring to backfill tests. I also extracted the CF call that gets `allStacks` since it was used elsewhere in the code where we delete `unowned` clusters. 

Related issues: #1849, #536, https://github.com/weaveworks/eksctl/pull/523, #3726

### Demo
The following demo is for a eksctl-created cluster with unmanaged nodegroups.

```
./eksctl delete cluster --name drain-unmanaged
2021-09-13 11:19:04 [ℹ]  eksctl version 0.68.0-dev+fbe5484e.2021-09-13T10:40:28Z
2021-09-13 11:19:04 [ℹ]  using region us-west-2
2021-09-13 11:19:04 [ℹ]  deleting EKS cluster "drain-unmanaged"
2021-09-13 11:19:07 [ℹ]  will drain 1 nodegroup(s) in cluster "drain-unmanaged"
2021-09-13 11:19:11 [ℹ]  cordon node "ip-192-168-25-121.us-west-2.compute.internal"
2021-09-13 11:19:11 [ℹ]  cordon node "ip-192-168-94-224.us-west-2.compute.internal"
2021-09-13 11:19:12 [!]  ignoring DaemonSet-managed Pods: kube-system/aws-node-shq67, kube-system/kube-proxy-jl5g2
2021-09-13 11:19:13 [!]  ignoring DaemonSet-managed Pods: kube-system/aws-node-s2nm4, kube-system/kube-proxy-2jnjc
2021-09-13 11:19:15 [!]  ignoring DaemonSet-managed Pods: kube-system/aws-node-shq67, kube-system/kube-proxy-jl5g2
2021-09-13 11:19:16 [!]  ignoring DaemonSet-managed Pods: kube-system/aws-node-s2nm4, kube-system/kube-proxy-2jnjc
[...]

PR #7515, opened 2024-01-30
Skip draining managed nodegroups when deleting cluster

yuxiang-zhang:
### Description

As pointed out in #4205, the code intended to only drain unmanaged nodegroups since EKS should take care of managed nodegroups. However, `cmdutils.ToKubeNodeGroups` function has been populating the `DrainInput` with both unmanaged and managed nodegroups. This causes an inconsistency where when using `eksctl delete cluster --config-file` with a config that specify the managed nodegroup config, that managed nodegroup get drained; but `eksctl delete cluster --name` deleting the same cluster will skip draining the same managed nodegroup, because https://github.com/yuxiang-zhang/eksctl/blob/1d9cd7dc8370716d711e45bf17e2967434eb88f4/pkg/actions/cluster/delete.go#L174-L179 only populates the unmanaged nodegroups. 

Regardless of using `--config-file`, the behaviour for deleting a cluster should be consistent: either drain managed nodegroups, or skip draining managed nodegroups. This change sticks with the original intent from #4205, that is to skip draining managed nodegroups. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/cluster/delete.go
+++ b/pkg/actions/cluster/delete.go
@@ -178,10 +178,11 @@ func drainAllNodeGroups(ctx context.Context, cfg *api.ClusterConfig, ctl *eks.Cl
 		}
 	}
 
+	// EKS automatically drains managed nodegroups
 	logger.Info("will drain %d unmanaged nodegroup(s) in cluster %q", len(cfg.NodeGroups), cfg.Metadata.Name)
 
 	drainInput := &nodegroup.DrainInput{
-		NodeGroups:            cmdutils.ToKubeNodeGroups(cfg),
+		NodeGroups:            cmdutils.ToKubeNodeGroups(cfg.NodeGroups, []*api.ManagedNodeGroup{}),
 		MaxGracePeriod:        ctl.AWSProvider.WaitTimeout(),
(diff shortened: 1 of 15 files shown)


a-hilaly (approved):

pkg/actions/cluster/delete.go
```diff
@@ -178,10 +178,11 @@ func drainAllNodeGroups(ctx context.Context, cfg *api.ClusterConfig, ctl *eks.Cl
 		}
 	}
 
+	// EKS automatically drains managed nodegroups
 	logger.Info("will drain %d unmanaged nodegroup(s) in cluster %q", len(cfg.NodeGroups), cfg.Metadata.Name)
 
 	drainInput := &nodegroup.DrainInput{
-		NodeGroups:            cmdutils.ToKubeNodeGroups(cfg),
+		NodeGroups:            cmdutils.ToKubeNodeGroups(cfg.NodeGroups, []*api.ManagedNodeGroup{}),
```
yuxiang-zhang, line 185: Changed the function signature for allowing specifying empty list of either unmanaged/managed nodegroups

PR #7522, opened 2024-02-01
Add support for Access Entry type

### LINKED ISSUE #7446: [Bug] Access Entries for Self-Managed Nodes not created by eksctl not working
Create an access entry for an EC2_LINUX IAM role for Karpenter to use.

### What happened?

When attempting to create an Access Entry for an EC2 Linux self-managed node group used by Karpenter via the config file, eksctl is expecting a policy ARN to be passed. The documentation for creating these types of access entries seems to be lacking, or perhaps eksctl currently only supports creation of these entries for self-managed node groups that eksctl creates (as opposed to compatibility for external self-managed node groups, such as Karpenter nodes).

### How to reproduce it?

First tried to create the entry without specifying the type. This won't work because it will try to create as a standard entry, which doesn't allow `system:*` to be passed in the `kubernetesGroups`.

```
accessConfig:
  authenticationMode: API_AND_CONFIG_MAP
  accessEntries:
    - principalARN: arn:aws:iam::11111111111:role/KarpenterNodeInstanceProfile
      kubernetesUsername: system:node:{{EC2PrivateDNSName}}
      kubernetesGroups:
        - system:bootstrappers
        - system:nodes
```

`The kubernetes group name system:nodes is invalid, it cannot start with system`

Then, tried passing the type as EC2_LINUX, but eksctl is expecting a policyARN to be passed. 

```
accessConfig:
  authenticationMode: API_AND_CONFIG_MAP
  accessEntries:
    - principalARN: arn:aws:iam::11111111111:role/KarpenterNodeInstanceProfile
      kubernetesUsername: system:node:{{EC2PrivateDNSName}}
      kubernetesGroups:
        - system:bootstrappers
        - system:nodes
      accessPolicies:
        - accessScope:
            type: EC2_LINUX
```

`Error: accessEntries[0].policyARN must be set to a valid AWS ARN`

I believe my second attempt is the correct approach, and that eksctl should not be requiring a policyARN to be passed for this type. 

**Logs**
`Error: accessEntries[0].policyARN must be set to a valid AWS ARN`

**Versions**

```
eksctl version: 0.167.0-dev+4280da7cd.2023-12-22T09:42:31Z
kubectl version: v1.29.0
OS: darwin
```

yuxiang-zhang:
### Description

Resolves #7446

### Checklist
- [x] Manually tested

### DIFF
--- a/examples/40-access-entries.yaml
+++ b/examples/40-access-entries.yaml
@@ -6,7 +6,7 @@ kind: ClusterConfig
 metadata:
   name: access-entries-cluster
   region: us-west-2
-  version: '1.25'
+  version: '1.29'
 
 nodeGroups:
   - name: ng
@@ -17,6 +17,7 @@ accessConfig:
   authenticationMode: API
   accessEntries:
     - principalARN: arn:aws:iam::111122223333:user/my-user-name
+      type: STANDARD # optional Type
       kubernetesGroups: # optional Kubernetes groups
         - group1 # groups can used to give permissions via RBAC
         - group2
@@ -34,3 +35,5 @@ accessConfig:
         - policyARN: arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy
           accessScope:
             type: cluster
+    - principalARN:  arn:aws:iam::111122223333:role/role-name-2
+      type: EC2_LINUX
--- a/pkg/apis/eksctl.io/v1alpha5/access_entry.go
+++ b/pkg/apis/eksctl.io/v1alpha5/access_entry.go
@@ -6,15 +6,23 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
 )
 
 // AccessEntry represents an access entry for managing access to a cluster.
 type AccessEntry struct {
+	// existing IAM principal ARN to associate with an access entry
 	PrincipalARN ARN `json:"principalARN"`
+	// `EC2_LINUX`, `EC2_WINDOWS`, `FARGATE_LINUX` or `STANDARD`
+	// +optional
+	Type string `json:"type,omitempty"`
+	// set of Kubernetes groups to map to the principal ARN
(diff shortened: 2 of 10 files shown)


examples/40-access-entries.yaml
```diff
@@ -6,7 +6,7 @@ kind: ClusterConfig
 metadata:
   name: access-entries-cluster
   region: us-west-2
-  version: '1.25'
```
a-hilaly, line 9: Did this example work before? or it's not being tested at all?

pkg/apis/eksctl.io/v1alpha5/access_entry.go
```diff
@@ -6,16 +6,20 @@ import (
...
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
 )
 
 // AccessEntry represents an access entry for managing access to a cluster.
 type AccessEntry struct {
+	// existing IAM principal ARN to associate with an access entry
 	PrincipalARN ARN `json:"principalARN"`
-	// +optional
```
a-hilaly, line 14: Not very sure here but i think `// +optional` are supposed to be a one line comment. They are parsed by controller-gen if i'm not mistaken..

a-hilaly: See https://github.com/kubernetes-sigs/controller-tools/pull/211/files#diff-3b88366c37358db48ac32a74f079eb2d1c89727c3cd99ddb552908a05794c94fR66-R70

a-hilaly: Do we use `controller-gen` to generate CRDs for eksctl?

examples/40-access-entries.yaml
yuxiang-zhang: Yeah it worked with non `STANDARD` types. I just updated to show the new fields in use

a-hilaly (dismissed):
Thank you @yuxiang-zhang ! 👍

a-hilaly (approved):

PR #7523, opened 2024-02-02
Check for empty region before invoking API in AWS SDK

a2ush:
### Description

Currently, you will receive the following error if you **forget to set the AWS region code**.

$ eksctl create cluster
Error: checking AWS STS access – cannot get role ARN for current session: operation error STS: GetCallerIdentity, https response error StatusCode: 0, RequestID: , request send failed, Post "https://sts..amazonaws.com/": Bad Request

When you look closely, you can notice that there is something wrong with this endpoint "https://sts..amazonaws.com/".
However, it is difficult to immediately resolve the problem from this error message.

**This PR provides a more understandable error message if you run into the same situation**.
I think it helps to navigate the next action to solve the problem.
```
(demo)
$ ./eksctl create cluster
Error: no region code: please set the AWS region to ~/.aws/config file or environment variables.
```

note:
The Installation document already has "prerequisite" description about aws configs.
https://eksctl.io/installation/#prerequisite
 > You can use ~/.aws/credentials file or environment variables. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/eks/api.go
+++ b/pkg/eks/api.go
@@ -186,6 +186,10 @@ func newAWSProvider(spec *api.ProviderConfig, configurationLoader AWSConfigurati
 		return nil, err
 	}
 
+	if cfg.Region == "" {
+		return nil, fmt.Errorf("AWS Region must be set, please set the AWS Region in AWS config file or as environment variable")
+	}
+
 	if spec.Region == "" {
 		spec.Region = cfg.Region
 	}
--- a/pkg/eks/api_test.go
+++ b/pkg/eks/api_test.go
@@ -118,6 +118,14 @@ var _ = Describe("eksctl API", func() {
 			},
 			err: fmt.Sprintf("cache file %s is not private", cacheFilePath),
 		}),
+		Entry("region code is not set", newAWSProviderEntry{
+			updateFakes: func(fal *fakes.FakeAWSConfigurationLoader) {
+				fal.LoadDefaultConfigReturns(aws.Config{
+					Region: "",
+				}, nil)
+			},
+			err: "AWS Region must be set, please set the AWS Region in AWS config file or as environment variable",
+		}),
 		Entry("creates the AWS provider successfully", newAWSProviderEntry{}),
 	)
 


a2ush:
kind/improvement

pkg/eks/api.go
```diff
@@ -187,6 +187,9 @@ func newAWSProvider(spec *api.ProviderConfig, configurationLoader AWSConfigurati
 	}
 
 	if spec.Region == "" {
+		if cfg.Region == "" {
+			return nil, fmt.Errorf("no region code: please set the AWS region to ~/.aws/config file or environment variables")
+		}
 		spec.Region = cfg.Region
```
yuxiang-zhang, line 193: https://github.com/eksctl-io/eksctl/blob/4e3d0e45dd37db58c2718c130d6bb5b9f4f7b334/pkg/eks/apiv2.go#L48-L50

Looks like `cfg.Region` is set from `spec.Region` in `newV2Config()` on L184. 

We should just move this check on `spec.Region` to before L184 and error if it's empty then.

pkg/eks/api.go
```diff
@@ -187,6 +187,9 @@ func newAWSProvider(spec *api.ProviderConfig, configurationLoader AWSConfigurati
 	}
 
 	if spec.Region == "" {
+		if cfg.Region == "" {
+			return nil, fmt.Errorf("no region code: please set the AWS region to ~/.aws/config file or environment variables")
```
yuxiang-zhang, line 191: also I'd prefer this error message: `"AWS Region must be set, please set the AWS Region in AWS config file or as AWS_REGION environment variable"`

a2ush: I'll change this message : ) Thank you

pkg/eks/api.go
a2ush: Thank you for your suggestion.

I understood the current logic as the following;
 * if `Spec.Region` is not empty, `cfg.Region` is set from `spec.Region` in `newV2Config()`
 * if `Spec.Region` is empty, `cfg.Region` is set from `~/.aws/config` or `AWS_REGION` in `newV2Config()`

So, I think we should check the value of `cfg.Region` after L184.
If `cfg.Region` is empty, we can determine AWS region is not defined.

//note
```
	cfg, err := newV2Config(spec, credentialsCacheFilePath, configurationLoader)
	if err != nil {
		return nil, err
	}

	fmt.Printf("spec.Region : %v\n", spec.Region)
	fmt.Printf("cfg.Region : %v\n", cfg.Region)

	if cfg.Region == "" {
		return nil, fmt.Errorf("AWS Region must be set, please set the AWS Region in AWS config file or as AWS_REGION environment variable")
	}
```

[result]
```
$ ./eksctl create cluster --dry-run
spec.Region : 
cfg.Region : ap-northeast-1

$ rm ~/.aws/config
$ ./eksctl create cluster --dry-run 
spec.Region : 
cfg.Region : 
Error: AWS Region must be set, please set the AWS Region in AWS config file or as AWS_REGION environment variable

$ export AWS_REGION=us-east-1      
$ ./eksctl create cluster --dry-run
spec.Region : 
cfg.Region : us-east-1

$ ./eksctl create cluster -f sample.yaml --dry-run
spec.Region : ap-northeast-1
cfg.Region : ap-northeast-1
```

yuxiang-zhang (approved):

PR #7661, opened 2024-03-14
Include MixedInstancesPolicy LaunchTemplate for validation

### LINKED ISSUE #7644: [Bug] warning on missing launch template when using mixed instance type node group
Scaling up an (unmanaged) nodegroup which has mixed instances (using instancesDistribution)

### What happened?
The scale up throws a warning that no launch template can be found (even if one exists). The actual scale up succeeds, but the warning is wrong.

### How to reproduce it?
`eksctl scale ng --cluster <cluster_name> --nodes=2 --nodes-min=0 --nodes-max=4 <ng_name>`

**Logs**
2024-03-13 10:12:46 [!]  nodegroup with Auto Scaling group "ng_name" does not have a launch template

I think it is related to the output of the `aws autoscaling describe-auto-scaling-groups`, which returns a different structure in case mixed instance policy is applied. eksctl expects LaunchTemplate as a top level property in the ASG struct, but for mixed instance node groups the property is nested one level deeper under the MixedInstancesPolicy property (so launch template is in **_MixedInstancesPolicy.LaunchTemplate_**)

**Versions**
eksctl version: 0.162.0
kubectl version: v1.26.9
OS: linux

yuxiang-zhang:
### Description

Fixes #7644

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/nodegroup/scale.go
+++ b/pkg/actions/nodegroup/scale.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/autoscaling"
+	autoscalingtypes "github.com/aws/aws-sdk-go-v2/service/autoscaling/types"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	awseks "github.com/aws/aws-sdk-go-v2/service/eks"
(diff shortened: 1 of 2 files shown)


yuxiang-zhang:
> Can we add an unit test for this?

👍 will do, I'm trying to confirm whether this solution is correct

PR #7702, opened 2024-04-05
docs: add CreateTags for Easy Setup of CNI Addon

hofq:
### Description
For easy setup of the CNI Addon as in the Example, the CreateTags Permission is missing.

### Checklist
- [ ] Manually tested

### DIFF
--- a/userdocs/src/usage/addons.md
+++ b/userdocs/src/usage/addons.md
@@ -40,6 +40,7 @@ addons:
       - ec2:DeleteNetworkInterface
       - ec2:DescribeInstances
       - ec2:DescribeTags
+      - ec2:CreateTags
       - ec2:DescribeNetworkInterfaces
       - ec2:DescribeInstanceTypes
       - ec2:DetachNetworkInterface


cPu1 (changes requested):
Hi @hofq, VPC CNI needs the `ec2:CreateTags` permission only on network interfaces, but the proposed change will allow it access to all resource types due to the `*` wildcard. 

This list is also obsolete as it's missing permissions to other actions like `ec2:DescribeSubnets` which is required for the relatively new prefix mode feature. I would recommend removing `attachPolicy` from this example and instead link to this [official page](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEKS_CNI_Policy.html#AmazonEKS_CNI_Policy-json). 

Also, unless you're scoping down permissions for your specific use case, it's recommended to use the leave `attachPolicy` and `attachPolicyARNs` unset for `vpc-cni` as eksctl will default to the recommended `AmazonEKS_CNI_Policy` managed policy.

PR #7714, opened 2024-04-17
Allow nodegroup creation after a cluster subnet is deleted

TiberiuGC:
### Description

With this PR, eksctl now accounts for the possibility of a user deleting a subnet from their cluster, operation which is achieved outside eksctl. This scenario became much more likely with introduction of #7218. Deleting a cluster subnet creates a configuration drift from cloudformation stack. This means, eksctl now needs to be aware that the subnets that it imports from the cluster stack may have been deleted and thus can't be used anymore.

Additionally, this PR introduces a new validation, in which it fails early if users try to create private nodegroups (i.e. set `privateNetworking:true`) in a cluster where all private subnets were deleted. Vice versa holds as-well, eksctl failing early if users try to create public nodegroups (i.e. don't set `privateNetworking:true`) in a cluster where all public subnets were deleted.

In terms of integration tests, I've tried adding a test on top of [the spec](https://github.com/eksctl-io/eksctl/blob/ce836e8774089fde63551ff69be7bec698bf5b47/integration/tests/managed/managed_nodegroup_test.go#L491) that covers `eksctl utils update-cluster-vpc-config`, but deleting subnets there proved tricky due to dependencies created earlier in the test.
  
Fixes https://github.com/eksctl-io/eksctl/issues/7538 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/vpc/vpc.go
+++ b/pkg/vpc/vpc.go
@@ -329,12 +329,17 @@ func UseFromClusterStack(ctx context.Context, provider api.ClusterProvider, stac
 		// if a subnet is present on the stack outputs, but actually missing from VPC
 		// e.g. it was manually deleted by the user using AWS CLI/Console
 		// than log a warning and don't import it into cluster spec
+		stackDriftFound := false
 		for _, ssID := range stackSubnets {
 			if !slices.Contains(vpcSubnets, ssID) {
-				logger.Warning("%s was found on cluster cloudformation stack, but has been removed from %s outside of eksctl", ssID, spec.VPC.ID)
+				stackDriftFound = true
+				logger.Warning("%s was found on cluster cloudformation stack outputs, but has been removed from VPC %s outside of eksctl", ssID, spec.VPC.ID)
 				continue
 			}
 			toBeImported = append(toBeImported, ssID)
+		}
+		if stackDriftFound {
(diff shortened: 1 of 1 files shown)


pkg/vpc/vpc.go
```diff
@@ -300,7 +301,42 @@ func UseFromClusterStack(ctx context.Context, provider api.ClusterProvider, stac
...
+		for _, o := range output.Subnets {
+			vpcSubnets = append(vpcSubnets, *o.SubnetId)
+		}
+
+		// if a subnet is present on the stack outputs, but actually missing from VPC
+		// e.g. it was manually deleted by the user using AWS CLI/Console
+		// than log a warning and don't import it into cluster spec
+		for _, ssID := range stackSubnets {
+			if !slices.Contains(vpcSubnets, ssID) {
+				logger.Warning("%s was found on cluster cloudformation stack, but has been removed from %s outside of eksctl", ssID, spec.VPC.ID)
```
cPu1, line 334: Minor nit:

```suggestion
				logger.Warning("%s was found on cluster's CloudFormation stack outputs, but has been removed from VPC %s outside of eksctl", ssID, spec.VPC.ID)
```

cPu1: Also, when this condition holds true, we should also log the set of subnets that will be used for the nodegroup.

cPu1 (approved):

PR #7756, opened 2024-05-15
Update aws-node from 1.12.6 to 1.18.1

consideRatio:
1.18.1 is recommended for EKS clusters, where its documented that "For
all Kubernetes releases, we recommend installing the latest VPC CNI
release." as read at https://github.com/aws/amazon-vpc-cni-k8s?tab=readme-ov-file#recommended-version.

The latest available addon for various k8s minor versions are listed at
https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html#updating-vpc-cni-add-on,
and it currently sais 1.18.1 for k8s 1.23 to 1.29.

### DIFF
--- a/pkg/addons/default/assets/aws-node.yaml
+++ b/pkg/addons/default/assets/aws-node.yaml
@@ -1,5 +1,5 @@
 ---
-# Source: crds/customresourcedefinition.yaml
+# Source: aws-vpc-cni/crds/customresourcedefinition.yaml
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


consideRatio:
A key confusion point is why `aws-node.yaml` is hardcoded inside `eksctl` codebase, and why we may want to use that over having `addons` list the addon named `vpc-cni` etc.

Maybe the way forward isn't to bump this, but to help users transition to using `addons` as declared in the eksctl config file?

consideRatio:
I pushed an update meant to fix the unit tests -- we now have two containers instead of just one in the aws-node pod. The added container relates to NetworkPolicy enforcement - which is disabled by default in the daemonset, and doesn't seem to get functional after enabling it either.

It would be great if networkpolicy enforcement could be configured and enabled successfully, but it could be considered separate from this PR as well of course.

consideRatio:
Thank you @TiberiuGC!

For reference about the network policy stuff, the steps to enable it are listed here [under "self-managed add-on"](https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html) I think - I've followed them and failed to see a result so far though

PR #7813, opened 2024-06-03
Disallow IRSA config if addon has existing pod identity associations

cPu1:
### Description

Disallows setting IRSA config if addon has existing pod identity associations.

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/actions/addon/update.go
+++ b/pkg/actions/addon/update.go
@@ -77,7 +77,7 @@ func (a *Manager) Update(ctx context.Context, addon *api.Addon, podIdentityIAMUp
 	}
 
 	var deleteServiceAccountIAMResources []string
-	if len(summary.PodIdentityAssociations) > 0 && !addon.UseDefaultPodIdentityAssociations {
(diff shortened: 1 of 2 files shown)


pkg/actions/addon/update.go
```diff
@@ -77,7 +77,7 @@ func (a *Manager) Update(ctx context.Context, addon *api.Addon, podIdentityIAMUp
 	}
 
 	var deleteServiceAccountIAMResources []string
-	if len(summary.PodIdentityAssociations) > 0 && !addon.UseDefaultPodIdentityAssociations {
+	if len(summary.PodIdentityAssociations) > 0 && !addon.UseDefaultPodIdentityAssociations && !a.clusterConfig.AddonsConfig.AutoApplyPodIdentityAssociations {
 		if addon.PodIdentityAssociations == nil {
```
a-hilaly, line 81: qq/non-blocking: What's the reason behind using an array pointer here? (`PodIdentityAssociations *[]PodIdentityAssociation`

a-hilaly (approved):
👍

pkg/actions/addon/update.go
cPu1: To remove pod identity associations from an addon, a user needs to explicitly set this field to the empty slice. A pointer is used to distinguish between an unspecified field and an empty slice.

PR #7831, opened 2024-06-12
Fail early when installing ADOT if cert-manager is not present on cluster

TiberiuGC:
### Description

Related to https://github.com/eksctl-io/eksctl/issues/6171

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/addon/create.go
+++ b/pkg/actions/addon/create.go
@@ -3,9 +3,11 @@ package addon
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
+	v1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/actions/addon/create.go
```diff
@@ -474,6 +480,35 @@ func (a *Manager) createStack(ctx context.Context, resourceSet builder.ResourceS
...
+			return fmt.Errorf("creating K8s client set: %w", err)
+		}
+		listDeploymentsOut, err := clientSet.AppsV1().Deployments("").List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return fmt.Errorf("listing k8s deployments: %w", err)
+		}
+		for _, name := range certManagerDeployments {
+			if !slices.ContainsFunc(listDeploymentsOut.Items, func(d v1.Deployment) bool {
+				return d.Name == name
+			}) {
```
a-hilaly, line 502: Maybe we could use label selectors instead?

pkg/actions/addon/create.go
```diff
@@ -474,6 +480,35 @@ func (a *Manager) createStack(ctx context.Context, resourceSet builder.ResourceS
...
 
+func (a *Manager) checkAddonPrereqs(ctx context.Context, addonName string) error {
+	switch addonName {
+	case api.AdotAddon:
+		certManagerDeployments := []string{
+			"cert-manager",
+			"cert-manager-cainjector",
+			"cert-manager-webhook",
+		}
+		clientSet, err := a.createClientSet()
```
a-hilaly, line 491: QQ: do we re-create clientSets for each seperate "component" in eksctl?

a-hilaly (commented):
My 2c: I understand that this is a highly requested feature but I'm not sure if it's right for us to try to solve this problem. Managing components like the cert manager sounds like a very tricky/risky task. Helm charts could evolve over time, potentially breaking things, and CRDs might change. This raises questions like whether we want to force CRD upgrades. What about version compatibility matrices? Additionally, what if users are utilizing cert-manager for other purposes? (and probably there are some other risks..)

Perhaps users could just manage cert-manager using `helm` CLi and take ownership of the cert-manager deployment themselves?

PR #7838, opened 2024-06-15
use string in logging instead of wrapping error

moreandres:
Using %w outside of fmt.Errorf is not appropriate because %w is intended specifically for error wrapping.


cPu1 (approved):
This is a known issue that affects some other logging calls as well.

PR #8081, opened 2024-12-11
Migrate dependent goformation module to a local fork

cheeseandcereal:
### Description

This PR migrates away from using the goformation fork at https://github.com/weaveworks/goformation/tree/master.next to a 'local' fork within this repo under the `goformation/` subdirectory.

The fork at `weaveworks/goformation` is really only used for this project, so maintaining a separate repo for a separate fork which requires manual updates adds complexity to the update process when releasing new features. Given that the upstream repository `awslabs/goformation` is archived, we are on our own for the maintenance of this fork, and shouldn't encourage its use elsewhere anyways.

Longer term, we may want to investigate alternatives to this dependent module/package, but for now, this just moves the dependency into the same repository here so we can more easily update it.

This PR also updates some of the golangci-lint config since it was using deprecated options; we are excluding goformation from being linted with the rest of the project.

I was able to build/run the tests/confirm everything is still working.

### Checklist
- [x] Manually tested


a-hilaly (approved):
Great call, thanks Adam!

PR #8125, opened 2025-01-10
Upgrade to go 1.22

bryantbiggs:
### Description

1.21 reached end of support in August of 2024. 
1.22 was chosen because there were errors thrown on 1.23 - tabled for later date. 

This PR is a bit larger than usual given the way the prior Docker images were configured in order to update the go version. Previously, it appears that you would need to perform the following to update the go version:
1. Create your git branch
2. Update the go version references throughout the codebase to the desired version
3. Build and push the [`eksctl` "build" image](https://github.com/eksctl-io/eksctl/tree/main/build/docker) to ECR. This means that only maintainers with write access to ECR can update go versions, and this is done locally! Not great
4. Grab the sha digest of the "build" image pushed to ECR, update the main [Dockerfile](https://github.com/eksctl-io/eksctl/blob/389f0717a0b8d374e57bd83a75209e59578229ab/Dockerfile#L1) with the new digest
5. Proceed to open pull request with changes

This is interesting for two reasons:
1. It requires a human in the loop which opens up various issues in terms of supply chain, access, and does not allow OSS contributions for this update
2. The "build" image includes a lot of dev type tooling (helm, kubectl, aws-iam-authenticator, etc.) but the `eksctl` binary is copied out of this dev build image into the resulting final image. If this psuedo development environment image was of value, there are probably better ways to achieve this using multi-stage builds and removing the need to have a person push the intermediate image to ECR from their local machine

To simplify, this PR removes the dev build image entirely which removes the majority of the complexity in this process. The resulting image that is pushed to ECR is based on other K8s/EKS images. I did not get a chance to dig into the code generation and whether any of that is used in the resulting final executable or if that is only for testing, but there could be further simplification - tabled for a later date.

With this changes, we therefore can remove the following:
- The `.requirements` file is removed which avoids version skew issues; any binaries needed are instead installed via `go install ...` in the Makefile [before their use](https://github.com/eksctl-io/eksctl/compare/main...bryantbiggs:feat/go-1-22?expand=1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R140)
  - This removes the need for the `install-build-deps` Make target
- With building the binary in the final image, the `eksctl-build` image, scripts, and manifests are removed
  - This removes the `.github/workflows/ecr-publish-build.yaml` workflow
  - This removes the need for `Makefile.docker`
- `.github/workflows/build-all-distros-nightly.yaml` workflow is removed since its redundant of the CI checks, and has been failing for some time
- `.github/workflows/cache-dependencies.yaml` workflow is removed since there are already a number of cache points in the workflows and we are over-caching which is rendering this moot 
  image

With this change, an outside OSS contributor can now update the version go and build, test, and validate that change before creating a PR without any maintainer involvement. 

### Checklist
- [x] Manually tested


bryantbiggs:
Link checker failing because it looks like macports.org is offline - will wait to see if it comes back online before adding to the ignored links

Makefile
```diff
@@ -139,7 +138,10 @@ delete-integration-test-dev-cluster: build ## Delete the test cluster for use wh
 ##@ Code Generation
 
 .PHONY: generate-always
-generate-always: pkg/addons/default/assets/aws-node.yaml ## Generate code (required for every build)
+generate-always: ## Generate code (required for every build)
+	go install github.com/maxbrunsfeld/counterfeiter/v6
+	go install github.com/vektra/mockery/v2
+	go install github.com/vburenin/ifacemaker
```
a-hilaly, line 144: super nit: we can move the go installs under a differ make rule and just call it here:
```suggestion
generate-always: install-gen-tools ## Generate code (required for every build)
```

Makefile
```diff
@@ -91,18 +88,20 @@ unit-test: check-all-generated-files-up-to-date unit-test-no-generate
 
 .PHONY: unit-test-no-generate ## Run unit test only
 unit-test-no-generate:
+	go install github.com/cloudflare/cfssl/cmd/...@latest
 	CGO_ENABLED=0 go test  -tags=release ./pkg/... ./cmd/... $(UNIT_TEST_ARGS)
 
 .PHONY: unit-test-race
 unit-test-race: ## Run unit test with race detection
+	go install github.com/cloudflare/cfssl/cmd/...@latest
 	CGO_ENABLED=1 go test -race ./pkg/... ./cmd/... $(UNIT_TEST_ARGS)
```
a-hilaly, line 97: is `cfssl` binary being called inside the unit tests?

a-hilaly (commented):
Awesome! Thank you @bryantbiggs for working on this! :) i left a couple comments in line

Makefile
bryantbiggs: unfortunately, yes https://github.com/eksctl-io/eksctl/blob/83e5cccfb245d3f281fc031605f8082af3e64ced/pkg/iam/oidc/testdata/Makefile#L15 

the unit tests are not "pure" unit tests - for example, they reach out to ECR to pull images

Makefile
bryantbiggs: good call - added a new target for installing tools https://github.com/eksctl-io/eksctl/pull/8125/commits/3331498639878cdaccec4cc26d4dba4c33120cbd

should this target run automatically where tools are required, or should that be an explicit step that CI/users take on their own

a-hilaly (approved):
👍

PR #8134, opened 2025-01-10
Remove `xtables.lock` which should be managed by the VPC CNI

bryantbiggs:
### Description

- Remove `xtables.lock` which should be managed by the VPC CNI
	- This is fixed in v1.18.1 of the VPC CNI https://github.com/aws/amazon-vpc-cni-k8s/pull/2841 which is the current default, but we should bump https://github.com/eksctl-io/eksctl/pull/7899 
- Update self-managed `aws-node.yaml` to latest from [here](https://github.com/aws/eks-charts/tree/master/stable/aws-vpc-cni) to ensure changes are in sync with respect to managed version of add-on (and that the xtables mount is still relevant)

### Checklist
- [ ] Manually tested

### DIFF
--- a/pkg/addons/default/assets/aws-node.yaml
+++ b/pkg/addons/default/assets/aws-node.yaml
@@ -405,7 +405,8 @@ spec:
     type: RollingUpdate
   selector:
     matchLabels:
-      k8s-app: aws-node
+      app.kubernetes.io/name: aws-node
+      app.kubernetes.io/instance: aws-vpc-cni
   template:
     metadata:
       labels:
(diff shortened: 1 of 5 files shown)


bryantbiggs:
Looks like this depends on https://github.com/eksctl-io/eksctl/pull/7899/files#diff-00df199028ae76ae35f1678fdb2184bf611cfb868abc937e86828e2aa2ee0150R597

a-hilaly:
@bryantbiggs not very sure, but do we need to remove the xtables volume mount with this change?

bryantbiggs:
> @bryantbiggs not very sure, but do we need to remove the xtables volume mount with this change?

@a-hilaly doesn't look like it - I just did `helm template . --include-crds ...` using the latest [here](https://github.com/aws/eks-charts/tree/master/stable/aws-vpc-cni) and updated the `aws-node.yaml`, but the xtables mount is still present.

PR #8140, opened 2025-01-13
Enable default addon creation for auto mode clusters

gustavodiaz7722:
### Description

This change enables the default addon creation workflow for auto mode clusters. Importantly though this change does not install networking addons by default. To start with only metrics-server addon will be installed by default. This will align EKSCTL experience with the default AWS EKS console experience, where by default auto mode clusters come with metrics-server enabled. 

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/actions/addon/tasks.go
+++ b/pkg/actions/addon/tasks.go
@@ -29,8 +29,14 @@ func CreateAddonTasks(ctx context.Context, cfg *api.ClusterConfig, clusterProvid
 			if addonInfo.IsDefault && !slices.ContainsFunc(cfg.Addons, func(a *api.Addon) bool {
 				return strings.EqualFold(a.Name, addonName)
 			}) {
-				addons = append(addons, &api.Addon{Name: addonName})
-				autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
+				if !cfg.IsAutoModeEnabled() {
+					addons = append(addons, &api.Addon{Name: addonName})
+					autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
+				} else if addonInfo.IsDefaultAutoMode {
+					addons = append(addons, &api.Addon{Name: addonName})
+					autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
+				}
+
 			}
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


pkg/actions/addon/tasks.go
```diff
@@ -29,8 +29,14 @@ func CreateAddonTasks(ctx context.Context, cfg *api.ClusterConfig, clusterProvid
...
 				return strings.EqualFold(a.Name, addonName)
 			}) {
-				addons = append(addons, &api.Addon{Name: addonName})
-				autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
+				if !cfg.IsAutoModeEnabled() {
+					addons = append(addons, &api.Addon{Name: addonName})
+					autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
+				} else if addonInfo.IsDefaultAutoMode {
+					addons = append(addons, &api.Addon{Name: addonName})
+					autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
```
bryantbiggs, line 37: hard to tell without the context of the PR (please update the PR body explaining the motivation behind the changes and intent), but when Auto Mode is enabled, I believe we do NOT want the add-ons deployed since they are provided automatically as part of Auto Mode. The only time we really want add-ons to be deployed is when there is a node group involved (either managed or self-managed)

pkg/actions/addon/tasks.go
```diff
@@ -29,8 +29,14 @@ func CreateAddonTasks(ctx context.Context, cfg *api.ClusterConfig, clusterProvid
...
+				}
```
a-hilaly, line 38:
```suggestion
				if !cfg.IsAutoModeEnabled() || addonInfo.IsDefaultAutoMode {
					addons = append(addons, &api.Addon{Name: addonName})
					autoDefaultAddonNames = append(autoDefaultAddonNames, addonName)
				}
```

a-hilaly (commented):
Thank you @gustavodiaz7722 ! I left one tiny suggestion below

pkg/actions/addon/tasks.go
gustavodiaz7722: My apologies for not including description in the PR body, I had originally included one but it seemed to have not published. I have updated the description

gustavodiaz7722: Yes you are correct, we do not want to install core networking addons on auto mode clusters, since these addons are provisioned as part of auto mode. This change does not enable installing vpc-cni, coredns, or kube-proxy, it only enables installing metrics-server by default. 

Metrics-server was chosen to be installed by default based on default AWS EKS Console experience.  Console will call EKS Addons create-addon api to provision metrics-server for the customer on cluster creation.

a-hilaly (approved):
👍

artem-nefedov:
This introduced a problem making it impossible to disable metrics-server in auto mode, see #8164

PR #8199, opened 2025-02-06
Enable NVIDIA Plugin automatically for AL2023 GPU AMIs

dims:
We now have AL2023 GPU based AMI(s):
- https://aws.amazon.com/blogs/containers/amazon-eks-optimized-amazon-linux-2023-accelerated-amis-now-available/
- https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html

Just like we enable nvidia device plugin for AL2 based ones, we should do the same for AL2023 as well!


bryantbiggs (approved):
thank you!

a-hilaly (approved):
👍

PR #8206, opened 2025-02-10
pre-Generate EC2 instance type details out of band

dims:
Use `DescribeInstanceTypes` API from a helper `ec2geninfo` command to pre-generate information we need instead of hardcoding similar to VPC CNI or Karpenter. This avoid making expensive calls at runtime to get the information needed.

There's also prior art here as well:
- https://github.com/clowdhaus/eksnode/blob/main/eksnode/files/ec2-instances.yaml
- https://github.com/clowdhaus/eksnode/tree/main/eksnode-gen

Once we land this, the follow up would be to write a GH action that runs `ec2geninfo` and files a PR automatically when there are changes. This takes toil way from maintainers.


pkg/utils/instance/instance_types.go
```diff
@@ -0,0 +1,8221 @@
...
+}
+
+type InstanceInfo struct {
+	InstanceType             string
+	InstanceStorageSupported bool
+	EFASupported             bool
+	NvidiaGPUSupported       bool
+	NvidiaGPUType            string
+	NeuronSupported          bool
+	NeuronGPUType            string
```
bryantbiggs, line 20:
```suggestion
	NeuronDeviceType            string
```

Ref https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/arch/neuron-hardware/neuron-devices.html#neurondevices-arch

pkg/utils/instance/instance.go
```diff
@@ -1,72 +1,50 @@
...
-		strings.HasPrefix(instanceType, "p5") ||
-		strings.HasPrefix(instanceType, "g3") ||
-		strings.HasPrefix(instanceType, "g4") ||
-		strings.HasPrefix(instanceType, "g5") ||
-		strings.HasPrefix(instanceType, "g6")
+	return InstanceTypesMap[instanceType].NvidiaGPUSupported
 }
 
 // IsInferentiaInstanceType returns true if the instance type requires AWS Neuron
 func IsInferentiaInstanceType(instanceType string) bool {
```
bryantbiggs, line 39: we shouldn't need separate checks for Inferentia vs Trainium - it should just be a check for `NeuronSupported`, similar to `NvidiaGPUSupported`


We could proceed with this current change but we'll have to exhaustively check the Neuron device types (`Inferentia`, Inferentia2`, etc. Same for `Trainium`, `Trainium2`) or do a prefix/startswith check instead

dims: added explicit checks for Inferentia/Inferentia2 and Trainium/Trainium2.

pkg/utils/instance/instance_types.go
dims: Done

pkg/utils/instance/instance.go
dims: we can revisit the places we are making the calls and see if we want to consolidate. for this PR i'd like to get the checks right (as-is! but killing things like "g3.xlarge" "p2.large" that are no longer there)

bryantbiggs: agreed - I was about to comment about p2 but thats an issue we can tackle separately; this just helps to remove the hardcoding of instance type name prefixes which is awesome

I think once the checks are green, we are good to merge

bryantbiggs (approved):

pkg/utils/instance/instance.go
dims: thanks @bryantbiggs

dims:
xref: https://github.com/eksctl-io/eksctl/issues/8000

PR #8217, opened 2025-02-13
Update GPU ec2 instance types in integration tests

cheeseandcereal:
### Description

This PR updates the instance types used in the integration tests to stop using P3 or G4ad instances which are much older. P3 instances are scheduled for retiring at some point, so we should stop using them.

Replaced with more modern equivalents.

### Checklist
- [ ] Manually tested

### DIFF
--- a/integration/tests/crud/creategetdelete_test.go
+++ b/integration/tests/crud/creategetdelete_test.go
@@ -924,7 +924,7 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 				"--timeout=45m",
 				"--cluster", params.ClusterName,
 				"--nodes", "1",
-				"--instance-types", "p3.2xlarge,p3.8xlarge,g3s.xlarge,g4ad.xlarge,g4ad.2xlarge",
+				"--instance-types", "g6.xlarge,g6.2xlarge",
 				"--node-private-networking",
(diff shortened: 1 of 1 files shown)


integration/tests/crud/creategetdelete_test.go
```diff
@@ -924,7 +924,7 @@ var _ = Describe("(Integration) Create, Get, Scale & Delete", func() {
 				"--timeout=45m",
 				"--cluster", params.ClusterName,
 				"--nodes", "1",
-				"--instance-types", "p3.2xlarge,p3.8xlarge,g3s.xlarge,g4ad.xlarge,g4ad.2xlarge",
+				"--instance-types", "g6.xlarge,g6.2xlarge",
```
bryantbiggs, line 927: should work but might run into insufficient capacity with this type FYI 

if we go a bit larger to 4x, 8x, 16x - the price won't be too drastic given its run a short duration but there will be more available capacity (customers avoid these because they really need the GPUs, not the added vCPU/memory - so they all aim for the x, 2x size) 
image

cheeseandcereal: Yeah. I wanted to try using another instance type entirely, but they're either also old or gigantic (like the P5 instances).

We can try this for now, and then if there are issues adjust

PR #8223, opened 2025-02-18
Documenting the use of kubernetesUsername while creating an access entry

jolivataws:
### Description

The intention of this change is to document that it is possible to specify the kubernetesUsername field while creating an access entry. 

Currently there are no reference to the parameter which currently is leading to the users to believe that this capability is not available for the eksctl.

I have modified the example provided that it is given to the users to showcase that the use of the field is possible.

### Checklist
- [X] Manually tested
- [] (Core team) Added labels for change area (e.g. `area/nodegroup`) and kind (e.g. `kind/improvement`)

- [] Backfilled missing tests for code in same general area :tada:
- [] Refactored something and made the world a better place :star2:

### DIFF
--- a/examples/40-access-entries.yaml
+++ b/examples/40-access-entries.yaml
@@ -21,6 +21,7 @@ accessConfig:
       kubernetesGroups: # optional Kubernetes groups
         - group1 # groups can used to give permissions via RBAC
         - group2
+      kubernetesUsername: username # optional Kubernetes username
     - principalARN: arn:aws:iam::111122223333:role/role-name-1
       accessPolicies: # optional access polices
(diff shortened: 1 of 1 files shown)


examples/40-access-entries.yaml
```diff
@@ -21,6 +21,7 @@ accessConfig:
       kubernetesGroups: # optional Kubernetes groups
         - group1 # groups can used to give permissions via RBAC
         - group2
+      kubernetesUsername: username # optional Kubernetes username
```
bryantbiggs, line 24: these are just examples, but the config schema does show/document this option https://eksctl.io/usage/schema/#accessConfig-accessEntries-kubernetesUsername

PR #8225, opened 2025-02-19
Fix issue installing the nvidia device plugin unconditionally on al2023

dims:
fix for https://github.com/eksctl-io/eksctl/issues/8224

inadvertently slipped into https://github.com/eksctl-io/eksctl/commit/987f978bc855e5947f7e192466e5e37a3725f68b

Added test cases to cover various scenarios, we did not have enough.


bryantbiggs (approved):
thank you!

PR #8230, opened 2025-02-20
Remove use of intermediate build image

bryantbiggs:
### Description

- Round two of #8125; removing the intermediate build image
- The `.requirements` file has been removed; binaries/libs required for building/testing have been placed under the `install-tools` make command. Future: investigate using `go tools ...` which was added in 1.24
- The `build/docker/*` directory and associated files are removed, removing the intermediate build image artifacts
  - This removes the need for the `.github/workflows/ecr-publish-build.yaml` workflow
  - This removes the need for `Makefile.docker`
- `.github/workflows/build-all-distros-nightly.yaml` workflow is removed since its redundant of the CI checks, and has been failing for some time

### Checklist
- [ ] Manually tested


.github/workflows/release-candidate.yaml
```diff
@@ -7,22 +7,14 @@ jobs:
   rc:
     name: Push release candidate tag
     runs-on: ubuntu-latest
-    container: public.ecr.aws/eksctl/eksctl-build:26e3c36557da8ae2db5d995cea673e05cc60cfce
```
bryantbiggs, line 10: removing the build image was previously reverted because these `container` references were not cleaned up from the CI scripts

Dockerfile
```diff
@@ -1,18 +1,19 @@
...
+RUN <<EOT
+    make install-tools
+    make build
+    chown 65532 eksctl
+EOT
 
-FROM scratch
-COPY --from=build /out /
-ENTRYPOINT ["eksctl"]
+FROM public.ecr.aws/eks-distro/kubernetes/go-runner:v0.16.4-eks-1-32-6 AS go-runner
```
cheeseandcereal, line 17: Why switch to using a base image instead of scratch?

Makefile
```diff
@@ -33,11 +32,15 @@ endif
 
 ##@ Dependencies
 .PHONY: install-all-deps
-install-all-deps: install-build-deps install-site-deps ## Install all dependencies for building both binary and user docs)
+install-all-deps: install-site-deps ## Install all dependencies for building both binary and user docs)
```
cheeseandcereal, line 35: why is install-all-deps no longer installing the other dependencies?

.github/workflows/test-and-build.yaml
```diff
@@ -17,8 +17,8 @@ jobs:
         uses: ./.github/actions/setup-build
       - name: Unit test
         run: |
-          PATH=$PATH:$(go env GOPATH)/bin make build
-          PATH=$PATH:$(go env GOPATH)/bin make unit-test-no-generate
+          make build
+          make unit-test-no-generate
```
cheeseandcereal, line 21: Why is setting the go bin path no longer necessary? Is this just coming with the container we use now?

Makefile
```diff
@@ -33,11 +32,15 @@ endif
...
 
-.PHONY: install-build-deps
-install-build-deps: ## Install dependencies (packages and tools)
-	build/scripts/install-build-deps.sh
+.PHONY: install-tools
+install-tools: ## Install dependencies for code generation and test execution
```
cheeseandcereal, line 38: Is there a reason for renaming this target? There are other references to the old target name `install-build-deps` in various other places that also need to be updated.

cheeseandcereal (changes requested):
Changes also need to be made in the other repos in this org to support these changes

Makefile
bryantbiggs: its a bit mis-leading, the aren't just build deps but also deps required for testing (cert generation), and this matched the `tools.go` a bit better. I can go either way though - let me know which way we want go and I'll make sure any references are updated

.github/workflows/test-and-build.yaml
bryantbiggs: we don't use a container, its setup by https://github.com/eksctl-io/eksctl/blob/ec903e328ae1f341cce2e0815d89c8b4a1e867db/.github/actions/setup-build/action.yaml#L8 which automatically adds it to the path

Makefile
bryantbiggs: this looks like a mistake - will re-add

Dockerfile
bryantbiggs: scratch images are not ideal - typically you want something with at least ca-certificates, timezone data, etc.

however, as noted above, this was just following the convention of other EKS projects 
- [Ref](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/c486a6fdbe1e1982f0852fcd7360c91de6367d4f/Dockerfile#L26)
- [Ref](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/dba6d7090f33e5ccd1452214691e0d1433d7881a/Dockerfile#L31-L44)

Makefile
bryantbiggs: corrected

Makefile
bryantbiggs: reverted

PR #8231, opened 2025-02-20
Re-add support for env var keys from AWS SDK

dims:
We dropped support for `AWS_ENDPOINT_URL` and `AWS_ENDPOINT_URL_*` environment vars that are documented here in commit e64db43bd455518d0712cff213df738a501d80ac:
- https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-sdk-compat
- https://docs.aws.amazon.com/sdkref/latest/guide/ss-endpoints-table.html

We are not going back to honoring what is in the shared config files, but `AWS_ENDPOINT_URL` itself is a handy short-cut for sure. 

In this PR, we will explicitly list all the env vars that we support, loop through them (in a specific order) and use whichever is defined.

xref: https://github.com/eksctl-io/eksctl/pull/8231


bryantbiggs (approved):

dims:
xref: https://github.com/eksctl-io/eksctl/issues/7968

PR #8233, opened 2025-02-21
Add missing coredns-1.32 json file

dims:
Short term fix for https://github.com/eksctl-io/eksctl/issues/8222

copy over pkg/addons/default/assets/coredns-1.31.json in pkg/addons/default/assets/coredns-1.32.json

### DIFF
--- a/pkg/addons/default/assets/coredns-1.32.json
+++ b/pkg/addons/default/assets/coredns-1.32.json
@@ -1,0 +1,379 @@
+{
+  "apiVersion": "v1",
+  "items": [
+    {
+      "apiVersion": "v1",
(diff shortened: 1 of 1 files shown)


bryantbiggs:
I'm not familiar with this but I thought it was automatic - looks like our IAM role is missing permissions https://github.com/eksctl-io/eksctl/actions/runs/13428257927/job/37515115269 and its been failing for a few weeks (since 1.32 was released)

bryantbiggs:
looks like we need the file to be created first before it can be updated - https://github.com/eksctl-io/eksctl/actions/runs/13448101739/job/37577563227

merging

PR #8239, opened 2025-02-21
Enable auto-update of ec2 info

dims:
- Add a new target for make `update-ec2-info`
- fix path of the file to update to match the fact that we are running from root of the repository
- updating the github action to call this new make target in addition to the other things it already does
- EC2 API data for `"p5en.48xlarge"` was wrong before when we ran this automation, when we run it now, they corrected the Nvidia GPU type.

TODO:
- need to setup a secret `UPDATE_EC2_INFO_ROLE_ARN`, the role should allow using `DescribeInstanceTypes` API

### DIFF
--- a/.github/workflows/update-generated.yaml
+++ b/.github/workflows/update-generated.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        resource: ["coredns", "aws-node", "nvidia-device-plugin"]
+        resource: ["coredns", "aws-node", "nvidia-device-plugin", "ec2-info"]
     name: Update ${{ matrix.resource }} and open PR
     runs-on: ubuntu-latest
     container: public.ecr.aws/eksctl/eksctl-build:26e3c36557da8ae2db5d995cea673e05cc60cfce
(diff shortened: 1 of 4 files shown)


dims:
cc @bryantbiggs

.github/workflows/update-generated.yaml
```diff
@@ -35,6 +35,14 @@ jobs:
...
         role-session-name: eksctl-update-coredns-assets
         role-to-assume: ${{ secrets.UPDATE_COREDNS_ROLE_ARN }}
+    - name: Configure AWS credentials for ec2 info update
+      if: ${{ matrix.resource == 'ec2-info' }}
+      uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722 # v4.1.0
+      with:
+        aws-region: us-west-2
+        role-duration-seconds: 900
+        role-session-name: eksctl-update-ec2-info-assets
+        role-to-assume: ${{ secrets.UPDATE_EC2_INFO_ROLE_ARN }}
```
bryantbiggs, line 45: could probably just use one role that has list/describe permissions - if we need to scope it, we have session tags that can be used

bryantbiggs (approved):

.github/workflows/update-generated.yaml
bryantbiggs: we also need to ensure this doesn't run in an account that has preview access to new instance types that are not yet released

cheeseandcereal:
Added a `UPDATE_EC2_INFO_ROLE_ARN` secret to the repo with an IAM role that should have `ec2:DescribeInstanceTypes` permissions.

cheeseandcereal (approved):

PR #8246, opened 2025-02-23
Do not run ci jobs in forks

dims:
Just adds noise in a fork when these actions run unnecessarily.


dims:
cc @bryantbiggs

bryantbiggs (approved):
makes sense!

PR #8249, opened 2025-02-25
Fix failure when VPC CNI is configured to use both iam.withOIDC and useDefaultPodIdentityAssociations

dims:
xref: https://github.com/eksctl-io/eksctl/issues/8147

This explicit check comes at a time when the OIDC endpoint is getting created (and may not have been yet!), so we end up getting `but since OIDC is disabled on the cluster` message which is not really true, because we did ask the cluster to be created with `cfg.IAM.WithOIDC` set to true!

xref: https://github.com/eksctl-io/eksctl/issues/8141

During create, we were passing `podIdentityIAMUpdater` to be nil which caused the panic.

### DIFF
--- a/pkg/actions/addon/tasks.go
+++ b/pkg/actions/addon/tasks.go
@@ -18,7 +18,7 @@ import (
 	"github.com/weaveworks/eksctl/pkg/utils/tasks"
 )
 
-func CreateAddonTasks(ctx context.Context, cfg *api.ClusterConfig, clusterProvider *eks.ClusterProvider, iamRoleCreator IAMRoleCreator, forceAll bool, timeout time.Duration, region string) (*tasks.TaskTree, *tasks.TaskTree, *tasks.GenericTask, []string) {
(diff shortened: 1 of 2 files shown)


pkg/actions/addon/tasks.go
```diff
@@ -178,15 +178,11 @@ func createAddonManager(ctx context.Context, clusterProvider *eks.ClusterProvide
...
-		}
-		oidcProviderExists, err = oidc.CheckProviderExists(ctx)
-		if err != nil {
-			return nil, err
-		}
+		// We are here because we are creating a cluster with OIDC enabled.
+		// To avoid a race, assume that it is getting created properly
+		// and do not check explicitly if it is already ready. If we really
+		// want to be sure then we would have to loop/wait/timeout here.
+		oidcProviderExists = true
```
bryantbiggs, line 185: I'm not great with go - but is this just simply:

```go
  oidcProviderExists := api.IsEnabled(cfg.IAM.WithOIDC)
```

removing the variable initialization above

dims: Sorry should have been a draft :)

bryantbiggs (approved):

PR #8262, opened 2025-02-28
Run unit tests with codegen diff check

bryantbiggs:
### Description

- Update codegen to use [new script/format](https://github.com/eksctl-io/eksctl/blob/5dd8d4c2ac7d01e5ebca6f6923e4d8b5c25b3627/build/scripts/update-codegen.sh#L30) since we've [updated to a version](https://github.com/eksctl-io/eksctl/pull/8210/files#diff-35f539189f171374fbd4326909476220b3c99a6320e72d7dce95bbb23ba3844dL9-R13) that has since changed the script/format
- Run unit tests with codegen diff check on every PR to catch this in the future and ensure we update the generated files

This would have caught issues with #8242 and #8210

### Checklist
- [ ] Manually tested


bryantbiggs:
this is expected to fail on the first go - making sure it catches this failure

cheeseandcereal:
When we run code generation, new versions of the aws golang v2 sdk are automatically fetched, and used in codegen for `./pkg/awsapi/...`

Won't this make it so that, if any of the aws services we use in that package are updated, this will just start failing all unit tests until someone updates and merges a fix, and everyone else rebases? Either that, or every PR is going to have to include autogen updates from AWS service updates in each of their PRs, and keep them up to date as AWS keeps their service up to date, making stale PRs more difficult to deal with?

I'm not convinced this is the right approach here.

PR #8267, opened 2025-02-28
Update zz_generated.deepcopy.go, also add new make target for the same

dims:
- Copied the script from @bryantbiggs 's PR - https://github.com/eksctl-io/eksctl/pull/8262/files
- Added a handy make target `update-deepcopy`
- Ran the target to generate the new zz_generated.deepcopy.go

For now, let's leave this as a manual step we have to remember to do.


cheeseandcereal (approved):
Looks good. We can figure out how to better integrate this into our workflows (aside from just manually running) later. New target lgtm though.

PR #8268, opened 2025-02-28
Fix failure when VPC CNI is configured to use both iam.withOIDC and useDefaultPodIdentityAssociations

dims:
xref: https://github.com/eksctl-io/eksctl/issues/8141

During create, we were passing podIdentityIAMUpdater to be nil which caused the panic.

Going to try half of https://github.com/eksctl-io/eksctl/pull/8249


cheeseandcereal (approved):
LGTM!

PR #8285, opened 2025-03-01
Standardize on k8s 1.32.2 for go.mod

dims:
Fix for https://github.com/eksctl-io/eksctl/issues/8284

use a specific version consistently in both the require and replace for kubernetes related dependencies.


cheeseandcereal (approved):
Looks good. Since we have removed use of the interim build image yet, we'll need to build and push it after merging this.

PR #8289, opened 2025-03-05
Update goformation from latest schema

dims:
We have TON of code we generate but do not use. So let's clean it up.

- Add an Allow list in generate.go with just the services that we need/use in this repo.
- Run `go:generate` in goformation.go to generate from latest schema
- Adjust some of the existing code as the struct(s) have changed a bit.

Only the first commit is useful to review. The second commit is mechanical output from `go:generate`.


cheeseandcereal:
Because of the # of files, changed, I can't view the whole thing on github. I will download it and check.

Can you help point me to the non-generated/interesting/important files to review for changes in this CR @dims ?

dims:
@cheeseandcereal skip d8c49c2f55c4ff83ace505d1a3737be63eca0198 commit, the rest of the 3 commits are easily reviewable in GH or locally.

cheeseandcereal (approved):
Looks great to me! Thanks for the code reaping!

PR #8303, opened 2025-03-20
Skip instance type support validation for custom AMI ID

ndbaker1:
### Description

closes: https://github.com/eksctl-io/eksctl/issues/8304

When using a nodegroup with an a custom AMI based on a supported AMIFamily (ex. for an arm64 gpu AL2023 derivative) you will/could be blocked in the validation stage with the error
```
Error: could not create cluster provider from options: ARM GPU instance types are not supported for unmanaged nodegroups with AMIFamily AmazonLinux2023
```

This should be supported for any `AMIFamily` as long as a custom AMI is provided, as we can't make an assumption about the details of the AMI.

### Reproduction

```bash
eksctl create ng -f config.yaml --dry-run
```

using the following config

```yaml
# config.yaml
kind: ClusterConfig

metadata:
  name: testbed
  region: us-west-2

nodeGroups:
  - name: test-nvidia-arm
    instanceType: g5g.2xlarge
    desiredCapacity: 1
    amiFamily: AmazonLinux2023
    ami: <ami-XXX>
    iam:
       attachPolicyARNs:
```

### Testing

building locally `make build` and running `./eksctl create ng -f config.yaml --dry-run` output valid data. also ran without `--dry-run` to confirm test node launched and joined cluster.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -973,8 +973,8 @@ func ValidateNodeGroup(i int, ng *NodeGroup, cfg *ClusterConfig) error {
 		return err
 	}
 
-	if instanceutils.IsARMGPUInstanceType(SelectInstanceType(ng)) && ng.AMIFamily != NodeImageFamilyBottlerocket {
-		return fmt.Errorf("ARM GPU instance types are not supported for unmanaged nodegroups with AMIFamily %s", ng.AMIFamily)
+	if err := validateInstanceTypeSupport(ng); err != nil {
+		return err
 	}
 
 	if err := validateInstancesDistribution(ng); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -1045,6 +1045,23 @@ func ValidateNodeGroup(i int, ng *NodeGroup, cfg *ClusterConfig) error {
...
+	if IsAMI(ng.AMI) {
+		return nil
+	}
+	if instanceutils.IsARMGPUInstanceType(SelectInstanceType(ng)) {
+		switch ng.AMIFamily {
+		case NodeImageFamilyBottlerocket:
+		default:
+			return fmt.Errorf("ARM GPU instance types are not supported for unmanaged nodegroups with AMIFamily %s", ng.AMIFamily)
+		}
+	}
```
cheeseandcereal, line 1061: Is there a reason to make this a switch-case statement over a simple `&& !=` in the existing `if` statement since the switch is only really checking one thing?

cheeseandcereal (changes requested):
It looks like the unit tests are failing after this change. Could you take a look?

pkg/apis/eksctl.io/v1alpha5/validation.go
ndbaker1: since more families that support `ARMGPU` besides `Bottlerocket` will eventually be added, i decided to refactor this whitelisting block. If you'd like to keep the original until there are required changes I'm fine with that too

ndbaker1:
> It looks like the unit tests are failing after this change. Could you take a look?

should be fixed now, accidentally left some broken test lines behind the first time 🫠

cheeseandcereal:
Seems like the unit tests are still failing after the changes btw

pkg/apis/eksctl.io/v1alpha5/validation.go
cheeseandcereal: Sure, that seems reasonable. we can leave it as-is

cheeseandcereal (approved):

PR #8312, opened 2025-03-23
Fix auto mode creation when no permissions boundary is set

cheeseandcereal:
### Description

After merging #8307, creation of clusters with auto mode without an explicit permissions boundary broke because they were getting set with the 'zero' arn value of `arn:::::`, this fixes that bug.

Tested by manually creating clusters before and after.

### Checklist
- [x] Manually tested


dims (approved):
LGTM

PR #8322, opened 2025-03-31
Enable creating IPV6 clusters with pod identities in addition to IRSA

simonmarty:
- **Fix typo on "associations"**
- **Add VS Code folder to gitignore**
- **Enable creating IPV6 clusters with pod identities**

### Description

Currently, creating an IPV6 cluster fails validation if `withOIDC` is not set to `true`. Since the advent of EKS pod identities, OIDC is no longer necessary as long as the pod identity agent is added, and the VPC CNI addon is configured with a pod identity association. Updated the validation logic to allow either pod identities or IRSA for IPV6 clusters.

Unrelated: caught two typos and added the VS Code folder to the gitignore since that's my IDE of choice (and a pretty common one at that).

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -609,7 +609,6 @@ func (c *ClusterConfig) validateKubernetesNetworkConfig() error {
 				return fmt.Errorf("the default core addons must be defined for IPv6; missing addon(s): %s; either define them or use EKS Auto Mode", strings.Join(missing, ", "))
 			}
 			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
-
 				return fmt.Errorf("either pod identity or oidc needs to be enabled if IPv6 is set; set either one or use EKS Auto Mode")
 			}
 
--- a/pkg/apis/eksctl.io/v1alpha5/validation.go
+++ b/pkg/apis/eksctl.io/v1alpha5/validation.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -607,8 +608,19 @@ func (c *ClusterConfig) validateKubernetesNetworkConfig() error {
 			if missing := c.addonContainsManagedAddons([]string{VPCCNIAddon, CoreDNSAddon, KubeProxyAddon}); len(missing) != 0 {
 				return fmt.Errorf("the default core addons must be defined for IPv6; missing addon(s): %s; either define them or use EKS Auto Mode", strings.Join(missing, ", "))
 			}
-			if c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC) {
-				return fmt.Errorf("oidc needs to be enabled if IPv6 is set; either set it or use EKS Auto Mode")
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
+
+				return fmt.Errorf("either pod identity or oidc needs to be enabled if IPv6 is set; set either one or use EKS Auto Mode")
+			}
+
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) == 0 && !c.AddonsConfig.AutoApplyPodIdentityAssociations {
+				// Assuming user intends to use pod identities if the pod identity agent addon is added.
+				vpcCNIAddonEntry := c.Addons[slices.IndexFunc(c.Addons, func(a *Addon) bool { return a.Name == VPCCNIAddon })]
+
+				if !vpcCNIAddonEntry.UseDefaultPodIdentityAssociations &&
+					(vpcCNIAddonEntry.PodIdentityAssociations == nil || (vpcCNIAddonEntry.PodIdentityAssociations != nil && len(*vpcCNIAddonEntry.PodIdentityAssociations) == 0)) {
+					return fmt.Errorf("Set one of: addonsConfig.autoApplyPodIdentityAssociations, useDefaultPodIdentityAssociations on the vpc-cni addon, apply a custom pod identity on the vpc-cni addon")
+				}
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -607,8 +608,19 @@ func (c *ClusterConfig) validateKubernetesNetworkConfig() error {
...
-			if c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC) {
-				return fmt.Errorf("oidc needs to be enabled if IPv6 is set; either set it or use EKS Auto Mode")
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
+
+				return fmt.Errorf("either pod identity or oidc needs to be enabled if IPv6 is set; set either one or use EKS Auto Mode")
+			}
+
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) == 0 && !c.AddonsConfig.AutoApplyPodIdentityAssociations {
+				// Assuming user intends to use pod identities if the pod identity agent addon is added.
+				vpcCNIAddonEntry := c.Addons[slices.IndexFunc(c.Addons, func(a *Addon) bool { return a.Name == VPCCNIAddon })]
```
simonmarty, line 618: I don't like the linear search here but I don't have any better ideas.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -607,8 +608,19 @@ func (c *ClusterConfig) validateKubernetesNetworkConfig() error {
...
 				return fmt.Errorf("the default core addons must be defined for IPv6; missing addon(s): %s; either define them or use EKS Auto Mode", strings.Join(missing, ", "))
 			}
-			if c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC) {
-				return fmt.Errorf("oidc needs to be enabled if IPv6 is set; either set it or use EKS Auto Mode")
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
+
+				return fmt.Errorf("either pod identity or oidc needs to be enabled if IPv6 is set; set either one or use EKS Auto Mode")
+			}
+
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) == 0 && !c.AddonsConfig.AutoApplyPodIdentityAssociations {
```
cheeseandcereal: I'm a little confused by this check. So if the pod identity addon _does_ exist, then we check for OIDC, but if it doesn't, then we check for pod identity settings on the addon itself? Am I reading that correctly? If so, I'm not really sure I understand. Shouldn't it be if there is no pod identity addon, then OIDC needs to be enabled, otherwise either OIDC _OR_ pod identity can be used?

simonmarty: The first if block
```
if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
```

Checks that there is at least one way to provide credentials to the VPC CNI addon. If there is not at least one provider (either Pod identity Agent or IRSA), then we exit early.

The second block

```
if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) == 0 && !c.AddonsConfig.AutoApplyPodIdentityAssociations {
```

Performs extra validations needed _if_ we are using pod identities (namely, that the VPC CNI addon has a pod identity configured in one way or another.

pkg/apis/eksctl.io/v1alpha5/validation.go
simonmarty: I can do that if that's more idiomatic. I'm not a golang expert.

pkg/apis/eksctl.io/v1alpha5/validation.go
simonmarty:
```
&& !c.AddonsConfig.AutoApplyPodIdentityAssociations
```
Could be moved to a separate if block inside the first, before the `getAddon` call. I can add comments too if that helps.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -607,8 +616,18 @@ func (c *ClusterConfig) validateKubernetesNetworkConfig() error {
...
-				return fmt.Errorf("oidc needs to be enabled if IPv6 is set; either set it or use EKS Auto Mode")
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) != 0 && (c.IAM == nil || c.IAM != nil && IsDisabled(c.IAM.WithOIDC)) {
+				return fmt.Errorf("either pod identity or oidc needs to be enabled if IPv6 is set; set either one or use EKS Auto Mode")
+			}
+
+			if len(c.addonContainsManagedAddons([]string{PodIdentityAgentAddon})) == 0 && !c.AddonsConfig.AutoApplyPodIdentityAssociations {
+				// Assuming user intends to use pod identities if the pod identity agent addon is added.
+				vpcCNIAddonEntry := c.getAddon(VPCCNIAddon)
+
+				if !vpcCNIAddonEntry.UseDefaultPodIdentityAssociations &&
```
cheeseandcereal, line 627: now that `getAddon` is a function that can return nil, you need to do a nil check first here.

pkg/apis/eksctl.io/v1alpha5/validation.go
cheeseandcereal: Ah ok, that makes sense. Yeah I think given the amount of conditionals here, a comment would help :)

pkg/apis/eksctl.io/v1alpha5/validation.go
simonmarty: `getAddon(VPCCNIAddon)` should never return `nil` because
```go
missing := c.addonContainsManagedAddons([]string{VPCCNIAddon, CoreDNSAddon, KubeProxyAddon}); len(missing) != 0
```

is `false` on L616. Do you still want me to add a `nil` check given that?

cheeseandcereal: Yeah I think it would be good, otherwise if we modify this code in the future it may be easy to miss.

cheeseandcereal (approved):
Thanks for the contribution!

PR #8349, opened 2025-04-11
Enable enclave options in EC2 Launch Template

tarikdem:
### Description

Unmanaged nodegroups schema does not currently support passing `EnclaveOptions` for the EC2 Launch Template which we rely on for provisioning nitro-enabled nodegroups.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/apis/eksctl.io/v1alpha5/types.go
+++ b/pkg/apis/eksctl.io/v1alpha5/types.go
@@ -1335,7 +1335,7 @@ type NodeGroup struct {
 
 	// EnclaveEnabled determines if the EC2 instance will be Nitro enclave enabled
 	// +optional
-	EnclaveEnabled bool `json:"enclaveEnabled,omitempty"`
+	EnclaveEnabled *bool `json:"enclaveEnabled,omitempty"`
 }
 
 // GetContainerRuntime returns the container runtime.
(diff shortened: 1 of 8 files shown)


pkg/apis/eksctl.io/v1alpha5/types.go
```diff
@@ -1332,6 +1332,10 @@ type NodeGroup struct {
 	// The cluster should have been created with all of the local zones specified in this field.
 	// +optional
 	LocalZones []string `json:"localZones,omitempty"`
+
+	// EnclaveEnabled determines if the EC2 instance will be Nitro enclave enabled
+	// +optional
+	EnclaveEnabled bool `json:"enclaveEnabled,omitempty"`
```
cheeseandcereal, line 1338: This should be a pointer so we can differentiate between not provided (nil pointer) and default boolean value of false.

pkg/cfn/builder/nodegroup.go
```diff
@@ -459,6 +459,9 @@ func newLaunchTemplateData(ctx context.Context, n *NodeGroupResourceSet) (*gfnec
 		UserData:          gfnt.NewString(userData),
 		MetadataOptions:   makeMetadataOptions(ng.NodeGroupBase),
 		TagSpecifications: makeTags(ng.NodeGroupBase, n.options.ClusterConfig.Metadata),
+		EnclaveOptions: &gfnec2.LaunchTemplate_EnclaveOptions{
+			Enabled: gfnt.NewBoolean(ng.EnclaveEnabled),
+		},
```
cheeseandcereal, line 464: We should only add this `EnclaveOptions` part if the EnclaveEnabled boolean was explicitly provided. Otherwise, this will default to always including this block in the cloudformation which could have implications for unsupported regions, etc.

cheeseandcereal (changes requested):

cheeseandcereal (approved):
Looks good. Thanks for the contribution!

PR #8442, opened 2025-07-14
Add announcement for docs on AWS, remove old announcements

geoffcline:
* add new announcement that the site will be moving
* remove old announcements


a-hilaly (approved):
👍

PR #8586, opened 2025-11-13
Enable termination protection by default for eksctl CFN stacks

abhay-krishna:
### Description

Enable termination protection by default for the CloudFormation stacks that eksctl spins up during cluster creation.

### Checklist
- [x] Manually tested


pkg/cfn/manager/api.go
```diff
@@ -120,8 +120,9 @@ func NewStackCollection(provider api.ClusterProvider, spec *api.ClusterConfig) S
 // DoCreateStackRequest requests the creation of a CloudFormation stack
```
abhay-krishna: The actual change is in this file, rest are mostly generated changes.

.github/workflows/greetings.yml
```diff
@@ -11,6 +11,6 @@ jobs:
     steps:
     - uses: actions/first-interaction@1c4688942c71f71d4f5502a26ea67c331730fa4d #v3.1.0
       with:
-        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
NicholasBlaskey: thanks for updating this

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -840,18 +840,18 @@ func validateNodeGroupBase(np NodePool, path string, controlPlaneOnOutposts bool
...
 	if ng.VolumeType != nil {
 		volumeType := *ng.VolumeType
-		if ng.VolumeIOPS != nil && !(volumeType == NodeVolumeTypeIO1 || volumeType == NodeVolumeTypeIO2 || volumeType == NodeVolumeTypeGP3) {
+		if ng.VolumeIOPS != nil && volumeType != NodeVolumeTypeIO1 && volumeType != NodeVolumeTypeIO2 && volumeType != NodeVolumeTypeGP3 {
 			return fmt.Errorf("%s.volumeIOPS is only supported for %s, %s and %s volume types", path, NodeVolumeTypeIO1, NodeVolumeTypeIO2, NodeVolumeTypeGP3)
 		}
 
 		if volumeType == NodeVolumeTypeIO1 {
-			if ng.VolumeIOPS != nil && !(*ng.VolumeIOPS >= MinIO1Iops && *ng.VolumeIOPS <= MaxIO1Iops) {
+			if ng.VolumeIOPS != nil && (*ng.VolumeIOPS < MinIO1Iops || *ng.VolumeIOPS > MaxIO1Iops) {
```
NicholasBlaskey: do you know what linter is requiring this? Is it something we could disable? Is it included by default in the v2 golangci? 

I think in this case it makes this code less clear. Also it seems flipping a bunch of if conditions could lead to a regression though I think all the changes look correct to me though.

abhay-krishna: This is enforced by the `staticcheck` linter's check [QF1001](https://staticcheck.dev/docs/checks/#QF1001). The check just recommends applying De Morgan's identities to resolve the negation through distribution. I would argue that this makes the current code more readable in this case as it expresses the intent clearly (outside the range) and eliminates the need to mentally parse a negation (NOT inside the range). Also I think the "lesser-than-min, greater-than-max" pattern is widely used in validation code across codebases for out-of-range checks. I agree with you in that the application of De Morgan's laws could lead to individual negated variables, like `(!X || !Y)` which could lead to more confusion than having one negation, but here since we're dealing with numeric inequalities, we are just able to flip the inequality to eliminate any negation.

NicholasBlaskey: I think in some of these cases it is debatable which is better. I think some linters can often be a tradeoff between freedom to express code how you want it and preventing potential issues. 

I'm okay with leaving the current updates in, but would you be okay with disabling the linter so developers have the option to choose which is clearer for the specific case in the future?

abhay-krishna: Yes that works, I have included the changes but disabled the specific check QF1001.

PR #8692, opened 2026-03-06
Fix EFA interface count for instances where MaximumNetworkCards exceeds EFA limit

kprahulraj:
### Description

Fixes https://github.com/eksctl-io/eksctl/issues/8685

We tried to set EFA's by using MaximumNetworkCards. This worked for all previous EFA-capable instances in cases where `MaximumEfaInterfaces = MaximumNetworkCards`, but starts to break in the latest instance types.

To fix this using `NetworkInfo.EfaInfo.MaximumEfaInterfaces` to determine the EFA interface count, with fallback to `MaximumNetworkCards` when `EfaInfo` is not populated.

### Checklist
- [x] Manually tested

### DIFF
--- a/pkg/cfn/builder/network_interfaces.go
+++ b/pkg/cfn/builder/network_interfaces.go
@@ -49,19 +49,31 @@ func buildNetworkInterfaces(
 		}
 
 		var numEFAs = math.MaxFloat64
+		var efaStartCard int
 		for _, it := range info.InstanceTypes {
 			networkInfo := it.NetworkInfo
-			numEFAs = math.Min(float64(aws.ToInt32(networkInfo.MaximumNetworkCards)), numEFAs)
 			if !aws.ToBool(networkInfo.EfaSupported) {
 				return fmt.Errorf("instance type %s does not support EFA", it.InstanceType)
 			}
+			if networkInfo.EfaInfo != nil && networkInfo.EfaInfo.MaximumEfaInterfaces != nil {
+				maxEfa := aws.ToInt32(networkInfo.EfaInfo.MaximumEfaInterfaces)
+				numEFAs = math.Min(float64(maxEfa), numEFAs)
+				// If EFA interfaces < network cards, network card 0 is ENA-only
+				if maxEfa < aws.ToInt32(networkInfo.MaximumNetworkCards) {
+					efaStartCard = 1
+				}
+			} else {
+				numEFAs = math.Min(float64(aws.ToInt32(networkInfo.MaximumNetworkCards)), numEFAs)
+			}
 		}
 
-		firstNI.InterfaceType = gfnt.NewString("efa")
+		if efaStartCard == 0 {
+			firstNI.InterfaceType = gfnt.NewString("efa")
+		}
 		nis := []gfnec2.LaunchTemplate_NetworkInterface{firstNI}
 		// The primary network interface (device index 0) must be assigned to network card index 0
 		// device index 1 for additional cards
-		for i := 1; i < int(numEFAs); i++ {
+		for i := 1; i < int(numEFAs)+efaStartCard; i++ {
 			ni := defaultNetworkInterface(securityGroups, 1, i)
 			ni.InterfaceType = gfnt.NewString("efa")
 			nis = append(nis, ni)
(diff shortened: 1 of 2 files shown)


mselim00 (approved):
this PR lgtm in terms of the fix intended for `p6-b200`, but it's worth pointing out:

most EFA-enabled instance types work with an `efa` interface on device index 0 for each network card. as far as I know, there's 4 special cases we have to handle, and those are enumerated in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html

considering each one:
1. `p5[e]?.48xlarge`
Supports 32 network cards, and 32 EFA interfaces. Assigning an `efa` interface to the 0th index of each card works as we currently do, but is not the recommendation. we should instead assign `efa-only` interfaces, and the one on network card index 0 should use device index 1

2. `p6-b200.48xlarge`
Supports 8 network cards and 8 EFA interfaces.
Same recommendation as p5's

3. `p6e-gb200.36xlarge`
Supports 17 network cards but only 16 EFA interfaces.
Even though there are 16 slots where an EFA interface is supported, the recommendation is to only allocate either 8 or 4 (200 gbps vs 400 gbps / interface bandwith). the example recommendation  for 4 allocates device index 0 of network cards indexed 1, 5, 9, 13. These interfaces would need to be `efa-only`, ENA would not be supported for those

4. `p6-b300.48xlarge`
Supports 17 network cards but only 16 EFA interfaces.
Similar to `p6e-gb200`, except in this case the layout does support utilizing each EFA interface without competing for the same bandwidth, it's just that the first network card does not support any. This is the usecase that the PR currently supports.

***

the takeaway from this is (separate from this PR):
1. we should find a way to switch to using `efa-only` instead of `efa` interface types - this is a breaking change though, as some users do require multiple interfaces available on host networking
2. we should add in an override strategy for instance types like `p6e-gb200.36xlarge`, where the optimal allocation is not immediately self-evident from the API response.

erezzarum:
We should have support for two scenarios:
1. Maximize EFA, conserve IP addresses - use ENA only (interfaceType=interface) for Primary network interface (NCI0,DI0) rest of interfaces required for EFA are interfaceType=efa-only, if going this path, we can safely use DeviceIndex=1 for all efa-only interfaces, e.g: for p5.48xlarge we can safely use NCI0,DI0=interface, NCI0-31,DI1=efa-only, this is because p5 and p5e supports up to max 2 network interfaces per card, so this is safe.
2. Maximize EFA, Maximize IP bandwidth (ENA) - use ENA only (interfaceType=interface) for Primary network interface (NCI0,DI0), if NCI0 supports EFA , use efa-only for NCI0,DI1. then for next unusued NCI (e.g 1), use DI0 for efa-only and if using ENA on the same NCI use DI1. we conserve IP addresses and use the maximum IP bandwidth available.

For 2, we could safely do the following for p5en and later generations (excluding p5 and p5e), primary network interface NCI0,DI0, NCI<N>,DI1 for every network card index that support EFA as efa-only, NCI<N>,DI2 for every network interface card index that we  want to use as ENA as interface type "interface" (ena only), this will work on p5en+ instances because they support >2 network interfaces per card.

Example per use case
p5.48xlarge,p5e.48xlarge
1. 1 IP address: NCI0,DI0 interface, NCI0-31,DI1 efa-only
2. 8 IP addresses: NCI0,DI0 interface, NCI0,DI1 efa-only, NCI1-31,DI0 efa-only, NCI4,8,12,16,20,24,28,DI1 interface 

p5en.48xlarge
1. 1 IP address: NCI0,DI0 interface, NCI0-15,DI1 efa-only
2. 8 IP addresses: NCI0,DI0 interface, NCI2,4,6,8,10,12,14,DI1 interface, NCI0-15,DI2 efa-only

p6-b200.48xlarge
1. 1 IP address: NCI0,DI0 interface, NCI0-7,DI1 efa-only
2. 8 IP addresses: NCI0,DI0 interface, NCI1-7,DI1 interface, NCI0-7,DI2 efa-only

p6-b300.48xlarge
1. 1 IP address: NCI0,DI0 interface, NCI1-16,DI1 efa-only
2. 17 IP addresses: NCI0,DI0 interface, NCI1-16,DI1 inteface, NCI1-16,DI2 efa-only

p6e-gb200.36xlarge
1. 1 IP address: NCI0,DI0 interface, NCI1,5,9,13,DI1 efa-only
2. 5 IP addresses: NCI0,DI0 interface, NCI2,6,10,14,DI1 interface, NCI1,5,9,13,DI2 efa-only

By doing so, we can also have only EFA SG on the efa-only interfaces and the normal SG on the ENA-only (interface) interfaces

PR #8774, opened 2026-06-22
Add support for EKS on Outposts with EC2 instance store

DimitriPL5:
- Add etcdInstanceType, etcdPlacement, and controlPlanePlacement fields
- Add remoteNetworkConfig validation (max 1 network/pod list, pod networks
  required for outpost clusters)
- Add outpost config cross-field validations
- Relax subnet-on-outpost check for instance store clusters
- Reject CONFIG_MAP auth mode for instance store clusters

### Description
- Updated eksctl support to allow customers to create clusters on their Outposts which use instance store. 

### References
- https://aws.amazon.com/about-aws/whats-new/2026/06/amazon-eks-aws-outposts-ec2-instance-store/

### Checklist
- [x] Manually tested


pkg/vpc/vpc.go
```diff
@@ -541,7 +541,7 @@ func importSubnetsForTopology(ctx context.Context, ec2API awsapi.EC2, spec *api.
 		return err
 	}
 
-	if spec.IsControlPlaneOnOutposts() {
+	if spec.IsControlPlaneOnOutposts() && spec.Outpost.EtcdInstanceType == "" {
```
kushcha-amzn, line 544: Why do we need this change?

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -104,13 +108,27 @@ func (c *ClusterConfig) validateRemoteNetworkingConfig() error {
...
 		// For creates, this is a no-op since HasRemoteNetworkingConfigured() gates the CFN builder.
 		if rnc.RemotePodNetworks != nil && len(rnc.RemotePodNetworks) == 0 {
 			logger.Warning("remoteNetworkConfig has empty remoteNodeNetworks and remotePodNetworks; this will remove all remote networks on upgrade, or be ignored on create")
 			return nil
 		}
+		if c.IsControlPlaneOnOutposts() {
+			if len(rnc.RemotePodNetworks) == 0 {
+				return setNonEmpty("remoteNetworkConfig.remotePodNetworks")
+			}
+			return nil
```
kushcha-amzn, line 130: Nit - Should it also print a similar warning to what's on L123?

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -104,13 +108,27 @@ func (c *ClusterConfig) validateRemoteNetworkingConfig() error {
 		return fmt.Errorf("remoteNetworkConfig requires authenticationMode to be either %q or %q", ekstypes.AuthenticationModeApiAndConfigMap, ekstypes.AuthenticationModeApi)
 	}
 
+	if len(rnc.RemoteNodeNetworks) > 1 {
+		return fmt.Errorf("only one remoteNodeNetwork is allowed in remoteNetworkConfig.remoteNodeNetworks")
```
kushcha-amzn, line 112: This isnt something we added for voyager right? Its always been there?

kushcha-amzn: Why does the validation need to come now?

pkg/vpc/vpc.go
DimitriPL5: EKS on Outposts with EC2 instance store doesn't validate that the provided subnets are hosted on an outpost, so we relaxed this check. Didn't remove it completely for backwards compatibility

pkg/apis/eksctl.io/v1alpha5/validation.go
DimitriPL5: Good callout. The difference here is that empty remoteNodeNetworks is an expected path for outpost clusters (only remotePodNetworks is required). The warning on L123 covers the case where BOTH are empty, which is an unusual/ambiguous state. Here, having only pod networks is the normal outpost use case, so a warning would generate noise for every valid outpost cluster create

pkg/apis/eksctl.io/v1alpha5/validation.go
DimitriPL5: Correct, this isn't Voyager-specific. The EKS API has always enforced max 1 remoteNodeNetwork and max 1 remotePodNetwork entry since hybrid nodes launched. eksctl just never had client-side validation for it. We're adding it now as part of this PR since we were already touching the remoteNetworkConfig validation path for the outpost-specific checks. It gives customers a clear local error instead of a server-side rejection.

pkg/apis/eksctl.io/v1alpha5/defaults.go
```diff
@@ -57,10 +57,10 @@ func SetClusterConfigDefaults(cfg *ClusterConfig) {
 
 	if cfg.AccessConfig == nil {
 		cfg.AccessConfig = &AccessConfig{
-			AuthenticationMode: getDefaultAuthenticationMode(cfg.IsControlPlaneOnOutposts()),
+			AuthenticationMode: getDefaultAuthenticationMode(cfg.IsControlPlaneOnOutposts(), cfg.Outpost != nil && cfg.Outpost.EtcdInstanceType != ""),
```
michaelhtm, line 60: nit: can we have a helper for this one as well?
```suggestion
			AuthenticationMode: getDefaultAuthenticationMode(cfg.IsControlPlaneOnOutposts(), cfg.IsOutpostV2()),
```

michaelhtm (dismissed):

michaelhtm (approved):

PR #8780, opened 2026-06-30
Add cluster version rollback support with rollbackConfig

michaelhtm:
### Description
Allow rolling back (downgrading) an EKS cluster's Kubernetes version by
one minor version via `eksctl upgrade cluster --version <lower>`.

- ResolveUpgradeVersion now resolves both the next and previous supported
  versions, permits a single-version downgrade (N->N-1), and rejects
  multi-version downgrades and downgrading the oldest supported version,
  replacing the old "cannot upgrade to a lower version" error.
- Add metadata.rollbackConfig.timeoutMinutes (api.RollbackConfig) to
  configure how long EKS waits before automatically cancelling an
  in-progress version update. Validated to the API's 120-10080 range
  and plumbed into UpdateClusterVersion.
- Bump aws-sdk-go-v2/service/eks to v1.88.0, which ships RollbackConfig
  on UpdateClusterVersion (and the CancelUpdate operation); regenerate
  the awsapi EKS interface, EKS mock, deepcopy helpers, and schema.

### Checklist
- [x] Manually tested


pkg/eks/versions_manager.go
```diff
@@ -179,15 +178,37 @@ func (cvm *ClusterVersionsManager) ResolveUpgradeVersion(desiredVersion string,
...
 	}
 
-	if desiredVersion == nextVersion {
+	c, err := utils.CompareVersions(desiredVersion, currentVersion)
+	if err != nil {
+		return "", fmt.Errorf("couldn't compare versions: %w", err)
+	}
+
+	// Handle upgrade
+	if c > 0 {
```
kprahulraj: what about c==0 case? If the current and desired version are same I believe this is going to downgrade path, which may be incorrect

pkg/eks/update.go
```diff
@@ -236,6 +236,12 @@ func (c *ClusterProvider) UpdateClusterVersion(ctx context.Context, cfg *api.Clu
 	if cfg.Metadata.ForceUpdateVersion != nil {
 		input.Force = *cfg.Metadata.ForceUpdateVersion
 	}
+	if rollbackConfig := cfg.Metadata.RollbackConfig; rollbackConfig != nil {
+		input.RollbackConfig = &ekstypes.RollbackConfig{}
+		if rollbackConfig.TimeoutMinutes != nil {
```
kprahulraj: when TimeoutMinutes is nil, are we sending an empty SDK RollbackConfig struct to the api? Will this result in any problem?

michaelhtm: you're right. added a check for timeoutMinutes before constructing rollbackConfig

pkg/eks/versions_manager.go
michaelhtm: line 177 handles the early return case for `desired == currentVersion`

pkg/apis/eksctl.io/v1alpha5/validation.go
sapphirew, line 1: nit: could consider adding validations around the rest of pre-requisites for version rollback https://docs.aws.amazon.com/eks/latest/userguide/rollback-cluster.html

sapphirew (approved):
Thanks @michaelhtm! lgtm, despite a non-blocking comment.

PR #8821, opened 2026-08-12
Add control plane component config support

chithreshazad:
### Description

Add support for configuring EKS control plane components via ClusterConfig: kubeAPIServerConfig, kubeSchedulerConfig, and kubeControllerManagerConfig.

Changes span the layers this feature touches:

- pkg/apis/eksctl.io/v1alpha5/types.go: ClusterConfig members and nested config structs.
- pkg/cfn/builder/cluster.go: map the config onto the ControlPlane resource. The makeKube* helpers return nil when unset, so the emitted template is byte-identical to today's when the feature is unused.
- pkg/cfn/builder/beta.go: add the three properties to the addBetaResources copy-list. That list copies by name, so a property missing from it is silently dropped on the beta/gamma custom-resource path.
- pkg/goformation/cloudformation/eks: vendored CFN structs (eight new files plus three properties on Cluster).
- pkg/ctl/utils and pkg/ctl/cmdutils: the `eksctl utils update-control-plane-component-config` command and its loader, following the pattern of update-zonal-shift-config. The makeKube*Request helpers mirror their makeKube* counterparts in pkg/cfn/builder/cluster.go, differing only where the SDK's types force it: ServiceNodePortRange.MinPort/MaxPort are non-pointer int32, and ScoringStrategy.Type is a named string type. Each returns nil when nothing is set, so a component the user omits is left unchanged rather than sent as an empty object, and the command fails before calling AWS when no values are set at all.
- zz_generated.deepcopy.go and assets/schema.json: regenerated.
- pkg/printers/testdata: add the three new Cluster fields to the golden files. They are exhaustive snapshots of ekstypes.Cluster, so the aws-sdk-go-v2/service/eks bump to v1.91.0 changes their output.

Testing:

- pkg/ctl/utils: unit tests for the request-building helpers, and for the command rejecting a config file whose components are all empty. The latter needs a real config file, since the loader only requires that one of the three components is present and returns before the command body runs otherwise.
- integration/tests/cluster_config: assert all three components round-trip through DescribeCluster on create, and cover the update command. The update spec changes only eventTTL and then asserts the other two components are unchanged, pinning the merge semantics: a component absent from the request is preserved, not cleared. The command blocks until the update reaches a terminal state, which takes roughly seven minutes, so the spec overrides EksctlUtilsCmd's default five-minute timeout. Also initialise the EKS client in BeforeSuite even when create is skipped. It was left nil under -eksctl.skip.create, which panicked every spec in the suite, including the pre-existing upgradePolicy one.

Verified end-to-end against a live cluster: every value set in ClusterConfig round-trips unchanged through the generated template and back out of DescribeCluster. The update path was exercised separately against the same cluster, including a rejected out-of-range value. The integration suite above passes in full against a freshly created cluster.

### Checklist
- [X] Manually tested


naclonts (dismissed):

naclonts (dismissed):
Thanks for the contribution @chithreshazad!

naclonts (approved):

PR #8872, opened 2026-09-17
Add connectionTracking support to nodegroup config

### LINKED ISSUE #8871: [Feature] Support EC2 connection tracking timeouts in nodegroup config
### What feature/behavior/change do you want?

Expose EC2's connection tracking timeouts as a first-class field in the nodegroup schema, for both `nodeGroups` and `managedNodeGroups`:

```yaml
managedNodeGroups:
  - name: ng-1
    instanceType: c8in.8xlarge
    connectionTracking:
      tcpEstablishedTimeout: 432000   # 60-432000
      udpStreamTimeout: 180           # 60-180
      udpTimeout: 60                  # 30-60
```

`eksctl` would pass these through to `ConnectionTrackingSpecification` on the network interfaces of the launch template it generates for the nodegroup.

### Why do you want this feature?

Security groups track connections to and from a node and drop traffic on connections that have been idle longer than EC2's timeout. The default idle timeout for established TCP connections depends on the instance's Nitro generation: Nitro v6 instance types (the `*8i` families, excluding P6e-GB200) use **350 seconds**, where earlier generations used **432,000 seconds**. See [Nitro instance types](https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-nitro-instances.html) and [connection tracking](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/security-group-connection-tracking.html).

Workloads that hold long-lived but mostly idle TCP connections — WebSocket streams, gRPC channels, database connection pools — can therefore start silently losing connections after moving a nodegroup to a current-generation instance type. The fix EC2 documents is to set `TcpEstablishedTimeout` on the instance's network interface, which for a nodegroup means setting it in the launch template.

Today `eksctl` has no way to do that. The only option is to supply a custom launch template via `launchTemplate.id`, which turns off most of the native nodegroup schema — `instanceType`, `ami`, `securityGroups`, `preBootstrapCommands`, `overrideBootstrapCommand`, `volumeSize`, `disableIMDSv1` and about nine others are all rejected once a launch template is supplied. Reimplementing a whole nodegroup automation setup inside hand-authored launch templates just to set one timeout is a steep cost for what is a single launch template property.

This is not specific to one workload. Nitro v6 is the platform for the current generation of x86 instance families, so every user adopting them inherits the new default, and the exposure grows as adoption does.

A few things make this a small change:

- The generated CloudFormation binding is already vendored and completely unused: `pkg/goformation/cloudformation/ec2/aws-ec2-launchtemplate_connectiontrackingspecification.go`, and `LaunchTemplate_NetworkInterface` already carries a `ConnectionTrackingSpecification` field. No goformation regeneration is needed.
- `pkg/cfn/builder/network_interfaces.go` is the single place where nodegroup network interfaces are built.
- Karpenter already exposes the same control as `spec.connectionTracking` on `EC2NodeClass` ([docs](https://karpenter.sh/docs/concepts/nodeclasses/)), so there is a reference implementation and an existing, validated shape for the field. `eksctl` is the remaining gap for people managing nodegroups.

One wrinkle worth flagging for whoever picks this up: a non-EFA **managed** nodegroup currently emits no `NetworkInterfaces` at all — `pkg/cfn/builder/managed_launch_template.go` only calls `buildNetworkInterfaces` when EFA is enabled, and otherwise sets instance-level `SecurityGroupIds`. So populating the network interface alone is a no-op for the common managed nodegroup; that path also needs to emit a single interface when the new field is set.

I have a change ready and will open a PR against this issue.

gustavodiaz7722:
Closes #8871

### Description

Adds an optional `connectionTracking` block to the nodegroup schema, so the EC2 connection tracking timeouts can be set without abandoning the native nodegroup config for a hand-written launch template.

```yaml
managedNodeGroups:
  - name: ng-1
    instanceType: c8in.8xlarge
    connectionTracking:
      tcpEstablishedTimeout: 432000

nodeGroups:
  - name: ng-2
    instanceType: m8i.4xlarge
    connectionTracking:
      tcpEstablishedTimeout: 3600
      udpStreamTimeout: 180
      udpTimeout: 60
```

The field sits on `NodeGroupBase`, so one addition covers `nodeGroups` and `managedNodeGroups`. Timeouts left unset are omitted from the template, so EC2 keeps its own default for the instance type; at least one must be set.

**Where the values land.** `buildNetworkInterfaces` now populates `ConnectionTrackingSpecification` on every network interface it builds — the primary interface, and for an EFA-enabled nodegroup the EFA interfaces too. That matches what Karpenter does for `EC2NodeClass`. The VPC CNI copies the primary interface's settings onto the interfaces it creates for pods (added in `v1.21.2` / `v1.22.1`), so pod traffic inherits them.

**The one non-obvious change.** A non-EFA managed nodegroup previously emitted no `NetworkInterfaces` at all — `makeLaunchTemplateData` only called `buildNetworkInterfaces` when EFA was enabled, and otherwise set instance-level `SecurityGroupIds`. Populating the interface alone would therefore have been a silent no-op for the common managed nodegroup. Such a nodegroup now emits a single network interface, carrying the security groups in `Groups`, **when and only when `connectionTracking` is set**. `AssociatePublicIpAddress` is left unset, so the subnet's auto-assign public IPv4 setting still applies — [EC2 documents](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/working-with-ip-addresses.html) that auto-assign is only lost with more than one interface, or with an existing interface at device index 0. I verified this on a real instance rather than relying on the docs; see below. This is also the same shape self-managed nodegroups have always used, since `nodegroup.go` calls `buildNetworkInterfaces` unconditionally.

A nodegroup that does not set `connectionTracking` generates exactly the template it did before. All existing launch template golden files are unchanged.

**Validation.** The ranges are checked at config load time, against the values in the EC2 API model: `tcpEstablishedTimeout` 60–432000, `udpStreamTimeout` 60–180, `udpTimeout` 30–60. This is worth doing client-side because EC2 does *not* reject out-of-range values at `CreateLaunchTemplate` time — I confirmed it stores them and defers enforcement, so without this check a user would get a launch template that only fails later, when nodes launch.

`connectionTracking` is also added to the set of fields rejected alongside a user-supplied `launchTemplate.id`, since `eksctl` does not build the launch template in that case and the value would otherwise be silently dropped.

### Manual testing

**Through the CLI.** Built the binary and drove the real thing. Valid config, resolved through `create cluster --dry-run` (round-trips through defaulting and validation for both nodegroup types):

```
$ eksctl create cluster -f valid.yaml --dry-run
...
managedNodeGroups:
- amiFamily: AmazonLinux2023
  connectionTracking:
    tcpEstablishedTimeout: 432000
...
nodeGroups:
- amiFamily: AmazonLinux2023
  connectionTracking:
    tcpEstablishedTimeout: 3600
    udpStreamTimeout: 180
    udpTimeout: 60
```

Rejections:

```
$ eksctl create cluster -f bad-range.yaml --dry-run
Error: value for managedNodeGroups[0].connectionTracking.tcpEstablishedTimeout must be within range 60-432000

$ eksctl create cluster -f bad-udp.yaml --dry-run
Error: value for nodeGroups[0].connectionTracking.udpTimeout must be within range 30-60

$ eksctl create cluster -f bad-empty.yaml --dry-run
Error: at least one of managedNodeGroups[0].connectionTracking.tcpEstablishedTimeout, managedNodeGroups[0].connectionTracking.udpStreamTimeout or managedNodeGroups[0].connectionTracking.udpTimeout must be set

$ eksctl create cluster -f bad-lt.yaml --dry-run
Error: cannot set instanceType, ami, ..., placement, connectionTracking in managedNodeGroup when a launch template is supplied
```

**The generated CloudFormation.** Rendered the templates both nodegroup types produce and checked them against AWS. The managed nodegroup's launch template comes out as:

```json
"NetworkInterfaces": [
    {
        "ConnectionTrackingSpecification": {
            "TcpEstablishedTimeout": 432000,
            "UdpStreamTimeout": 180,
            "UdpTimeout": 60
        },
        "DeviceIndex": 0,
        "Groups": ["sg-..."],
        "NetworkCardIndex": 0
    }
]
```

with no instance-level `SecurityGroupIds`, as EKS requires when a network interface is specified. `aws cloudformation validate-template` accepts both the managed and the self-managed stack.

**End to end on a real instance, including the public-subnet case.** I built a launch template with exactly that shape and launched an instance from it into a subnet with `MapPublicIpOnLaunch=true`. The instance got its public IPv4 address, so the subnet's auto-assign setting is still honoured with the interface block present:

```
$ aws ec2 describe-instances --instance-ids i-... \
    --query 'Reservations[0].Instances[0].{State:State.Name,PublicIp:PublicIpAddress,PrivateIp:PrivateIpAddress}'
{
    "State": "running",
    "PublicIp": "54.218.87.182",
    "PrivateIp": "172.31.40.51"
}
```

And the timeouts reached the live ENI, not just the launch template:

```
$ aws ec2 describe-network-interfaces --network-interface-ids eni-... \
    --query 'NetworkInterfaces[0].{ConnectionTracking:ConnectionTrackingConfiguration,PublicIp:Association.PublicIp}'
{
    "ConnectionTracking": {
        "TcpEstablishedTimeout": 432000,
        "UdpStreamTimeout": 180,
        "UdpTimeout": 60
    },
    "PublicIp": "54.218.87.182"
}
```

So the chain holds end to end: config → CloudFormation → launch template → attached ENI. Test resources were torn down afterwards.

One note from that exercise, in case anyone reproduces it by hand: with a `NetworkInterfaces` block in the launch template, the subnet has to be supplied on the interface — `RunInstances` with an instance-level `--subnet-id` is refused with `InvalidParameterCombination`. That doesn't affect this change, since `eksctl` never puts a subnet in the launch template and EKS/the ASG supplies them, and the existing EFA path already launches managed nodegroups with an interface block.

**What I have not done** is a full `eksctl create cluster` against a real cluster, so the managed-nodegroup node-join path is covered by the EC2-level evidence above rather than by a cluster run. I'd suggest an integration test for a public-subnet managed nodegroup so this doesn't rest on a one-off manual run — happy to add it here if you'd like it in scope, or leave it as a follow-up.

### Checklist
- [x] Manually tested


gustavodiaz7722:
I've now verified the public-subnet case on live EC2, so the caveat in the description can come off.

I built a launch template with exactly the shape this change emits for a managed nodegroup — a single network interface at device index 0, security groups in `Groups`, no instance-level `SecurityGroupIds`, `AssociatePublicIpAddress` left unset — and launched an instance from it into a subnet with `MapPublicIpOnLaunch=true`:

```
"NetworkInterfaces": [
    {
        "ConnectionTrackingSpecification": {
            "TcpEstablishedTimeout": 432000,
            "UdpStreamTimeout": 180,
            "UdpTimeout": 60
        },
        "DeviceIndex": 0,
        "Groups": ["sg-..."],
        "NetworkCardIndex": 0
    }
]
```

**The instance got its public IPv4 address**, so the subnet's auto-assign setting is still honoured with the interface block present:

```
$ aws ec2 describe-instances --instance-ids i-... \
    --query 'Reservations[0].Instances[0].{State:State.Name,PublicIp:PublicIpAddress,PrivateIp:PrivateIpAddress}'
{
    "State": "running",
    "PublicIp": "54.218.87.182",
    "PrivateIp": "172.31.40.51"
}
```

And the timeouts reached the live ENI, not just the launch template:

```
$ aws ec2 describe-network-interfaces --network-interface-ids eni-... \
    --query 'NetworkInterfaces[0].{ConnectionTracking:ConnectionTrackingConfiguration,PublicIp:Association.PublicIp}'
{
    "ConnectionTracking": {
        "TcpEstablishedTimeout": 432000,
        "UdpStreamTimeout": 180,
        "UdpTimeout": 60
    },
    "PublicIp": "54.218.87.182"
}
```

So the chain holds end to end: config → CloudFormation → launch template → attached ENI. Test resources were torn down afterwards.

Two smaller notes from the same exercise:

- With a `NetworkInterfaces` block in the launch template, the subnet has to be supplied on the interface — `RunInstances` with an instance-level `--subnet-id` is refused with `InvalidParameterCombination`. That doesn't affect this change, since `eksctl` never puts a subnet in the launch template and EKS/the ASG supplies them, and the existing EFA path already launches managed nodegroups with an interface block. Flagging it only because it's the kind of thing that's easy to trip over when reproducing this by hand.
- I re-ran `golangci-lint` with the repo's own config and it caught a `revive` `if-return` violation in my validation hook, fixed in the second commit. Lint is clean on the changed packages now; the only remaining finding repo-wide is a pre-existing `govet` one in `pkg/ctl/cmdutils/configfile.go`, which this PR doesn't touch.

I'd still suggest an integration test covering the public-subnet managed nodegroup so this doesn't rest on a one-off manual run — happy to add it here if you'd like it in scope, or leave it as a follow-up.

One thing I can't move: the GitHub Actions runs on this PR are sitting in `queued` pending first-time-contributor approval, so `Lint` and `Unit tests` haven't reported yet. Locally `go build ./...`, `go test ./pkg/...`, `gofmt`, `golangci-lint`, and the regenerated `schema.json` / `zz_generated.deepcopy.go` are all clean, and the docs deploy preview passed.

pkg/apis/eksctl.io/v1alpha5/validation.go
```diff
@@ -969,7 +979,29 @@ func validateNodeGroupBase(np NodePool, path string, controlPlaneOnOutposts bool
 		}
 	}
 
-	return nil
+	return validateConnectionTracking(ng.ConnectionTracking, path)
+}
+
+func validateConnectionTracking(connectionTracking *ConnectionTracking, path string) error {
```
sapphirew: Configurable connection-tracking timeouts are supported only on Nitro instances. This currently validates presence and numeric ranges, so a non-Nitro instanceType, a mixed instance list containing one, or a type produced by instanceSelector can pass validation and then fail when EC2 launches the node. Could we validate the final resolved instance types (for example, via DescribeInstanceTypes and requiring Hypervisor == nitro) and add a non-Nitro test? Karpenter filters non-Nitro instance types for the analogous field.

userdocs/src/usage/nodegroup-connection-tracking.md
```diff
@@ -0,0 +1,69 @@
...
+interfaces it creates for pods, so pod traffic inherits them too. That copying was added in
+VPC CNI `v1.21.2` and `v1.22.1`; on older versions the timeouts apply to node traffic only.
+
+Because `connectionTracking` is applied through the launch template that `eksctl` generates,
+it cannot be combined with a nodegroup that supplies its own launch template through
+`launchTemplate.id`. Set the timeouts in your own launch template in that case.
+
+!!! note
+    Changing `connectionTracking` on an existing nodegroup does not update the nodes that
+    are already running. Create a new nodegroup, or replace the existing nodes, to pick up
```
sapphirew: This guidance looks misleading for the normal eksctl workflow. eksctl update nodegroup does not apply NodeGroupBase fields, and the generated managed-nodegroup launch-template reference does not select the newly created template version. Replacing instances after editing connectionTracking can therefore keep using the old timeouts. Could we describe this as creation-only and recommend creating a new nodegroup, or document the explicit launch-template version update required before replacing nodes?

